PR #3, opened 2015-08-04
EchoDeviceTypes

larsfinander:
-Heartbeat attribute added
-Hearbeat event generated for each poll/checkin


larsfinander:
Made by mistake.

PR #15, opened 2015-08-07
Adding Nest and TED5000 Device type

tomforti:


tylerlange:
Thanks for sharing!

We unfortunately won't be able to accept this pull request. We have an official Nest integration that we've been working on and we are unable to publish this community integration due to the Nest partnership agreement.

The other piece you submitted, TED5000, looks really cool! In order for us to publish an integration to our customers it'd need to be wrapped in a Service Manager SmartApp so that a user can walk through the entire setup flow from the SmartThings Mobile app. More information about that is available at http://docs.smartthings.com. Check out the portions on web services SmartApps and service manager SmartApps.

Thanks,
Tyler
SmartThings

PR #43, opened 2015-08-25
Update smartsense-moisture.groovy

workingmonk:
The fix was put it months ago, but accidentally ended up in the Private Repo. Deleting the DTH in the private repo and updating the fix here.


workingmonk:
@tylerlange can you check this? this fix already exists in prod, just moving the change over to the Public repo

workingmonk:
@mrnohr you refrenced this for CREX-2640. Ok for me to merge this?

mrnohr:
Yes #45 changed the same file, but it should be smooth to merge

PR #45, opened 2015-08-25
CREX-2640: Adding labels to moisture sensor attribute states.

whfern:
We need to have labels present so that the states can be properly
labeled in the new versions of client applications. This causes a slight
experience degradation in older versions of the app, but the impact is
understood.


mrobinet:
:+1:

mrnohr:
See #43 which also updates this file. Should be fine to merge

mrnohr:
:+1:

whfern:
Merging this after reviewing #43

PR #47, opened 2015-08-26
fixing the issue with custom icon not working

workingmonk:


workingmonk:
@mrnohr can you verify this?

PR #54, opened 2015-08-26
REVISED: SmartWeather Station Tile

bflorian:
I edited the SmartWeather Station Tile to include values for windchill (I live in Minneapolis, windchill is an unfortunate fact of life here), wind direction, wind gust, heat index, visibility, solar radiation, uv, pressure, & dew point.


remasse:

PR #58, opened 2015-08-26
BeaconThings

bflorian:
BeaconThings is an iOS app that detects iBeacons, allows users to register them with SmartThings, and then reports presence to SmartThings. This allows users to enable micro-location within their home/office/etc. The app connects to SmartThings via OAuth, then creates a virtual presence sensor for each beacon added. See beaconthingsapp.com for some more information and please feel free to contact me for more info.


remasse:

PR #59, opened 2015-08-26
Added Water Sensor Option

bflorian:
added water condition.  allows me to turn off outside cameras so they do not sense motion during these times.  also added polling condition to schedule (i.e. every 15 minutes).


remasse:

PR #65, opened 2015-08-26
DJ Roomba!

bflorian:
Integrate a roomba equipped with a RooWifi component. RooWifi is required to give the roomba an IP address and rest services to control.


remasse:

PR #66, opened 2015-08-26
z-wave garage door opener (works better with mobile app)

bflorian:
Changed from original ST implementation to add "momentary" and "relay switch" capabilities (along with implementation.)  This allows this device type to be used as a garage door in the ST mobile app "doors & locks" stuff


remasse:

PR #67, opened 2015-08-26
Aeon HEMv2+ Device

bflorian:
Update/replacement for original Aeon HEMv2 submission request. 

Aeon HEMv2+ now supports alternate displays:
- Energy, Power, Current & Voltage
  Display 1: Low & High values since reset() for Power, Current, Voltage
                  Calculated cost (preference defines $/kWh)
  Display 2: L1 & C1 values for Energy, Power, Current & Voltage

Toggle between views with a button, All stat collection continues while in the "background" (displaying other view). Preferences define delay between reports for kWh (default 2 minutes) and the other 3 values (default 30 seconds).

See also https://github.com/SANdood/Aeon-HEM-v2


remasse:

PR #69, opened 2015-08-26
First release device type for Fibaro FGS-221 Double Relay Switch

bflorian:
Working device type supporting both switches of the Fibaro FGS-221 Double Relay Switch.

Does not support all available configuration options but has the advantage over the default switch device type in that both switches work. Also provides a way to configure the switches as momentary – the device ships with both set as toggle switches.

TODOs: Support for simulation in the IDE. Support for the second switch in SmartApps.


remasse:

PR #70, opened 2015-08-26
Smart device control with Ring

bflorian:
A wearable device, Ring, can be a trigger to control various devices. You can control smart devices with your gesture of your index finger in the air. It looks like magic!


remasse:

PR #72, opened 2015-08-26
Aeon Micro Motor Controller

bflorian:
Allows control of Aeon Micro Motor Controller, a Z-Wave device, both from the mobile app tiles, and SmartApps.  The device type maps the three hardware states into ST:  Up, Down, Stop, and provides commands with the same names.  When used with motor shades, generally the motor has an adjustable automatic stop at the full height or extension of the shade, so the physical switch would just select Up, and that was that.  The motor stopped despite the switch.  The physical switch would typically also have an Off position.  The micro motor controller does as well, hence the stop() command.  Below is the ReadMe for this device type:

To use this device type in SmartApps, use "capability.doorControl" in Preferences to select the controller.  
Then you can reference the state of the device with device.currentValue("motor") or device.currentMotor.  
The device can be in one of three states: up, down, stop, e.g. if(device.currentMotor == "up") {...}  
There are three commands available: up(), down(), stop(), e.g. device.up()

After you include the Aeon device into ST, edit its device type in the IDE to be "Aeon Motor Controller"

Down turns on motor M1, and Up turns on motor M2.

### DIFF
--- a/devicetypes/bravenel/aeon-motor-controller.src/aeon-motor-controller.src/aeon-motor-controller.src/aeon-motor-controller.groovy
+++ b/devicetypes/bravenel/aeon-motor-controller.src/aeon-motor-controller.src/aeon-motor-controller.src/aeon-motor-controller.groovy
@@ -1,0 +1,128 @@
+/**
+ *  Aeon Motor Controller
+ *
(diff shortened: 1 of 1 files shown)


remasse:

PR #74, opened 2015-08-26
Smart Water Heater

bflorian:
Using a SmartenNet high voltage Zigbee switch, I created this program to turn off and on the water heater based on a complex schedule, including time-based and presence based decision making. This app, just by itself, has lowered my water bill by 10-20% without any loss of hot water.

There was an error when a fresh install was performed and certain parameters were null. I have gracefully trapped those so that a new user can successfully install this app. There is a 'quirk' that if a day schedule is not set and the user attempts to save, then they will receive an 'unexpected error' when attempting to exit the app. This is mitigated within the HREF text field marking certain parameters 'required'. If this is not per standard there is a possible work around I can implement. However, it would require the changing of the interface slightly.

Made additional modifications to the app to allow for handling if the night schedule is not chosen. Currently, the application does not allow for graceful handling of 'required' HREF pages. As such, I have put in text showing what is required. If this still doesn't meet the criteria I will change it again to have the main schedule on the setup page and eliminate the Href page.


remasse:

PR #75, opened 2015-08-26
GE Bulb setLevel fix

bflorian:
Convert value from String to integer


workingmonk:
the path for this file is incorrect:
smartthings/ge-link-bulb.src/ge-link-bulb.src/ge-link-bulb.src/ge-link-bulb.groovy

remasse:

PR #76, opened 2015-08-26
Turn On Only It If I Arrive After Sunset for 30 Minutes

bflorian:
Turns On Selected Devices Only It If I Arrive After Sunset for 30 Minutes


remasse:

PR #79, opened 2015-08-27
Fix for main tile issue

mrnohr:
In our attempt to roll out an experimental new feature early, we prevented users from being able to choose which tile they want to set as the "main" tile.

This temporary fix will help users so they can continue to change the main tile. This is temporary because there is a better way to do it by changing back-end code, which we will get in soon. Once that happens we can rip out these weirdly named tiles.


mrnohr:
@workingmonk OK to merge this?

workingmonk:
:+1:

PR #83, opened 2015-08-30
Fixing the enrollment issue for the remaining IAS devices. 

workingmonk:
Also cleared the long standing typo "Confuguring"

### DIFF
--- a/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
+++ b/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
@@ -143,8 +143,7 @@ def setLevel(value) {
 
 def configure() {
 
-	String zigbeeId = swapEndianHex(device.hub.zigbeeId)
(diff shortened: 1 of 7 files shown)


workingmonk:
@lbredeso can you take a look at this?

lbredeso:
Thank goodness "Confuguring" is gone now.  :-)

Does the CIE need to be written first for these devices in the case where the enroll request is sent by the device before CIE is written in `configure()` (this would be in `enrollResponse()`)?  It isn't happening in any of these DeviceTypes, but is for CentraLite devices, so I wasn't sure if that was intentional or not.

workingmonk:
For now, I am not modifying the behavior of these device types. This will be handled during the zigbee library update for these device types. Will do extensive testing then when the code modifications are huge

lbredeso:
Sure, that makes sense.

:+1:

susie78:
Thanks

PR #85, opened 2015-08-31
Rollback changes for active tile fix

mrnohr:
This is to rollback the changes from #79. That was a temporary fix to help users that were having problems selecting active tiles. There is a change that will be deployed on the platform to fix the problem. Once that is deployed, we can merge this to rollback the temporary fix.

cc @workingmonk


workingmonk:
@mrnohr If the issue has been resolved, can you fix the conflicts and merge this?

workingmonk:
:+1:

PR #86, opened 2015-08-31
[DVCSMP-811] Garage door preference based DTH with icon support

workingmonk:


lbredeso:
:+1: 

I just finished testing this out in AppEngine and it all looks good to me.

PR #87, opened 2015-09-01
Small fix to the wording

workingmonk:


workingmonk:
Changing the wording based on input from Tyler Hall

PR #89, opened 2015-09-01
[PROB-516] as Integer causes the temp to be off by 1

workingmonk:
Updating the as Integer to Math.round
Math.round throws an error when the value is null whereas the 'as Integer' operation doesn't
Handling the above case with NPE checks

Need someone with groovy expertise to look into this to confirm


kris-schaller:
:+1:

PR #93, opened 2015-09-02
MSA-62: Routine Director

tslagle13:
Submitting 'Routine Director' for publication

### DIFF
--- a/smartapps/tslagle13/routine-director.src/routine-director.groovy
+++ b/smartapps/tslagle13/routine-director.src/routine-director.groovy
@@ -1,0 +1,338 @@
+/**
+ *  Rotuine Director
+ *
+ *
+ *  Changelog
+ *
+ *  2015-09-01
+ *     --Added Contact Book
+ *     --Removed references to phrases and replaced with routines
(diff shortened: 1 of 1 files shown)


workingmonk:
@tslagle13 are we following a convention to not have the new smartapps under smartthings folder? it is currently under smartapps/tslage13 folder. That is probably the namespace causing that to happen.

cc @kris-schaller

kris-schaller:
What namespace issue? @workingmonk @tslagle13

kris-schaller:
I was told that we can't update apps from the superuser account if the file name doesn't match the app name with existing GitHub integration. @workingmonk is this true?

workingmonk:
correct, the file name has to match the app-name

workingmonk:
the namespace issue is the location of the app. it is currently under smartapps/tslage13 rather than smartapps/smartthings

kris-schaller:
ah, technically @tslagle13 developed this outside of ST employement, which is why it's residing under his account.

workingmonk:
i thought when we get the apps in, we get them into the smartthings folder and keep the author name as SmartThings if it was developed internally other wise change it, in this case to "Tim Slagle" because he developed it outside

PR #105, opened 2015-09-08
Re-enabling garage door option in the preference after the 1.0.17 AE release

workingmonk:
cc @lbredeso @joneilloine


lbredeso:
:+1:  I think this should be fine, but we should probably wait until RevD incremental has been rolled out, unless we're ok with it not functioning as expected for those folks.

workingmonk:
@lbredeso Won't merge it to prod until Rev D has the update. For now, will push it to staging for SQE to test this

PR #106, opened 2015-09-08
Hue update to match the new UI

juano2310:


workingmonk:
cc @mrnohr

devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
```diff
@@ -32,24 +32,51 @@ metadata {
...
+		standardTile("refresh", "device.switch", height: 2, width: 2, inactiveLabel: false, decoration: "flat") {
 			state "default", label:"", action:"refresh.refresh", icon:"st.secondary.refresh"
 		}
-	}
-
-	main(["switch"])
-	details(["switch", "levelSliderControl", "rgbSelector", "refresh", "reset"])
+		controlTile("levelSliderControl", "device.level", "slider", height: 1, width: 2, inactiveLabel: false, range:"(0..100)") {
+			state "level", action:"switch level.setLevel"
+		}
```
mrnohr: Not sure why levelSliderControl, level, satuartionSliderControl, saturation, hueSliderControl, and hue were added here. It doesn't look like they are ever used.

devicetypes/smartthings/hue-bridge.src/hue-bridge.groovy
```diff
@@ -15,19 +15,27 @@ metadata {
 		// TODO: define status and reply messages here
 	}
 
-	tiles {
+	tiles(scale: 2) {
+     	multiAttributeTile(name:"rich-control"){
+			tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
```
mrnohr: do hue bridges have a switch capability?

juano2310:
There were part of the original one... I can remove them. I saw you did this one too but there were some changes on the API that handles alerts and transitions, that's why I continued using the old one. I will remove the extra one.

devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
```diff
@@ -23,8 +23,8 @@ metadata {
 		// TODO: define status and reply messages here
 	}
 
-	tiles (scale: 2){
-		multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
+	tiles(scale: 2) {
+		multiAttributeTile(name:"rich-control", type:"lighting", canChangeIcon: true) {
```
mrnohr: We specifically had the name stay as "switch" to maintain compatibility with older versions. It probably doesn't matter now since we added a fix to the platform to handle this potential error case.

juano2310:
Thanks @mrnohr, I fixed all the comments and since I reverted to your previous version of the tiles the switch comment was fixed too.

juano2310:
@mrnohr is this good to go? I was hopping to get it to QA soon. Thanks again!

devicetypes/smartthings/hue-bridge.src/hue-bridge.groovy
```diff
@@ -15,19 +15,27 @@ metadata {
 		// TODO: define status and reply messages here
 	}
 
-	tiles {
+	tiles(scale: 2) {
+     	multiAttributeTile(name:"rich-control"){
+			tileAttribute ("", key: "PRIMARY_CONTROL") {
```
mrnohr, line 20: Have you tested this without a name here? I'm afraid it will break. It would be good testing on older clients too.

juano2310, line 20: I did test and it worked :) but haven't tested in older clients. Should I ask QA specifically about this? I don't have older clients

mrnohr, line 20: OK, I'm not sure how concerned we are with older clients, but that is used to display tiles on older clients.

mrnohr:
Other than the question about older clients, this seems OK to me

juano2310:
Thanks! I will merge and create a ticket highlighting this. Thanks again!

PR #112, opened 2015-09-09
Improving the stability and reliability of the temp humidity sensor

workingmonk:
1. Humidity and Temp reports back at least every hour IF no change, every 30 second if changes are continuous
2. Battery reports every 6 hours. can't set to 24hours because of int16 limit


tpmanley:
:+1:

PR #115, opened 2015-09-10
Giving nice names to NYCE sensors

workingmonk:
Updating the custom names for models after I added the deviceJoinName capability to fingerprint last week.

cc @tylerlange


tylerlange:
:+1:

PR #116, opened 2015-09-10
atomicState to state

juano2310:


smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -258,7 +259,7 @@ def manualRefresh() {
 }
 
 def uninstalled(){
-	atomicState.bridges = [:]
```
bflorian, line 261: Clearing the state in the uninstalled() method is moot. There's no way to ever get to it again. Shouldn't hurt though

juano2310:
How about now :)

Yaima:
:+1: Good to go

PR #118, opened 2015-09-11
New capabilities; buffering; scheduling

davidsulpy:
This pull request contains major enhancements to the Initial State Event Streamer.
Enhancements regarding ST's request for less network resource usage:
- Events are now buffered using `atomicState`
- When there are 10 or more events in the `atomicState` buffer, the buffer is flushed and the events are sent to Initial State
- Every 15 minutes a schedule task is triggered to flush the buffer as long as the buffer is not empty. This solves for the case where very few events are being triggered.

Enhancements based on User's feedback and demand:
- Now allows users to subscribe to `capability.energyMeter` and `capability.powerMeter` among many other new capabilities a user can subscribe to

Other Enhancements:
- Switched all uses of `state` to `atomicState` since buffering was required
- Added link to Initial State's terms of service in comments
- Added ability to define the Initial State streaming service's subdomain for better testing
- Added a version to the `atomicState` to help troubleshoot the SmartApp in both logs and state view.

### DIFF
--- a/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
+++ b/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
@@ -239,6 +239,7 @@ def installed() {
 }
 
 def updated() {
+	atomicState.version = "1.0.18"
 	unsubscribe()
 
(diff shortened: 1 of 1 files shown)


kris-schaller:
:+1:

kris-schaller:
:+1:

PR #135, opened 2015-09-24
Fibaro Smoke Sensor: initial device type

kwarodom:
@mckeed Can you review this? @workingmonk 
Changes have been made as follows:
1. remove CO capability
2. change default temperature report interval to 30 mins
3. change default excess temperature signaling interval to 30 mins
4. change default lack of Z-Wave range indication interval to 6 hours
5. check MSR before sending configuration commands to device
6. add instruction how to check state of the smoke sensor

### DIFF
--- a/devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
+++ b/devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
@@ -177,7 +177,7 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityCommandsSupported
     log.info "Executing zwaveEvent 98 (SecurityV1): 03 (SecurityCommandsSupportedReport) with cmd: $cmd"
     setSecured()
     log.info "checking this MSR : ${getDataValue("MSR")} before sending configuration to device"
-    if (getDataValue("MSR").startsWith("010F-0C02")){
+    if (getDataValue("MSR")?.startsWith("010F-0C02")){
         response(configure()) //configure device using SmartThings default settings
     }
 }
@@ -188,7 +188,7 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.NetworkKeyVerify cmd) {
     //after device securely joined the network, call configure() to config device
     setSecured()
     log.info "checking this MSR : ${getDataValue("MSR")} before sending configuration to device"
-    if (getDataValue("MSR").startsWith("010F-0C02")){
+    if (getDataValue("MSR")?.startsWith("010F-0C02")){
         response(configure()) //configure device using SmartThings default settings
     }
 }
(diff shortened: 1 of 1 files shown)


workingmonk:
once @mckeed approves this, we can merge it

mckeed:
Looks good to me. We should make sure to communicate that this handler shouldn't be used with the older version of the device that was sold in Europe.

workingmonk:
@mckeed is there a way to limit that using the fingerprint? applicationId ?

mckeed:
We checked and it doesn't look like it will match the fingerprint, so the issue would just be people thinking it's supported and buying it, or manually setting the device type in graph and having problems. Warodom added code to prevent sending the configuration commands to the wrong model which could mess it up, but the configuration won't do anything and we don't know if the basic features will work properly.

workingmonk:
even if the configuration is prevented, if the fingerprint matches and DTH shows up in the connect page, you will have users believing that a critical device like the smoke detector will work but in reality, we wouldn't have configured it.

Just wanted to ask you if we could block it in the fingerprinting stage so that it doesn't show up in the Connect page.

kwarodom:
I've just made update on tiles (based on Gilbert comment). Previously, tiles are details(["smoke","temperature","heatAlarm", "tamper"]). The new update i hide "heatAlarm", "tamper", and just show these attributes status on the activity feeds.

kwarodom:
@mckeed Could you review SHA: a80bb9fa53ea05c53d21df8ad70578fc96db821b ? I have updated the device type to  send commands to update battery and temperature tiles during joining process. I also add save nav operator to MSR when it's value get called.

devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -0,0 +1,453 @@
...
+    def result = [createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)]
+    /* check MSR = "manufacturerId-productTypeId" to make sure configuration commands are sent to the right model */
+    if (!isConfigured() && getDataValue("MSR")?.startsWith("010F-0C02")) {
+        result << response(configure()) // configure a newly joined device or joined device with preference update
+    } else {
+        //Only ask for battery if we havn't had a BatteryReport in a while
+        if (!state.lastbatt || (new Date().time) - state.lastbatt > 24*60*60*1000) {
+            log.debug("Device has been configured sending >> batteryGet()")
+            result << response(zwave.securityV1.securityMessageEncapsulation().encapsulate(zwave.batteryV1.batteryGet()).format())
+            result << response("delay 1200")  // leave time for device to respond to batteryGet
```
mckeed: If I recall correctly, this delay isn't going to work because it's not in the same HubAction as the other commands. Because of how these delays are processed, you have to get all of the commands in the same call to response() to make sure they'll be considered one command list:

```
cmds << zwave.security...format()
cmds << "delay 1200"
...
cmds << zwave.wakeUp...format()
result += response(cmds)
```

Maybe that issue has been fixed, not sure, but I've run into it a couple times before.

mckeed:
Other than that issue, which isn't a big deal if you're getting the battery reports fine anyway, the code still looks good to me.

:+1:

kwarodom:
@mckeed  Thanks for your comment. I checked again, you are right. I need to add all commands in the same response() call. 72899ee fixed the issue (i tested with the smoke sensor again and crosscheck with zniffer). Please let me know if it's good to merge.

PR #136, opened 2015-09-24
CREX-3129 Update service manager apps to be singleton

mrobinet:


workingmonk:
@mrobinet Who do you want to review these changes?

charliek:
If we have more than one of these smartapps installed now and then we mark this as a singleton what happens?  Has that scenario been tested?

mrobinet:
The user will still have two apps installed. They can fix it manually by uninstalling and reinstalling. I'm investigating the number of duplicate installations that currently exist and that will dictate whether we attempt to dedup these for customers.

charliek:
Ok. As long as that is being accounted for this looks ok to me. But I'm not the smart app authority so take that for what it is :)

mrobinet:
Hold on merging this for now. There is still testing being done.

PR #143, opened 2015-09-28
Z-Wave Water Valve: initial device type

kwarodom:
@mckeed Can you review this device handler?

### DIFF
--- a/devicetypes/smartthings/fortrezz-water-valve.src/fortrezz-water-valve.groovy
+++ b/devicetypes/smartthings/fortrezz-water-valve.src/fortrezz-water-valve.groovy
@@ -19,7 +19,6 @@ metadata {
 		capability "Sensor"
         
 		fingerprint deviceId: "0x1000", inClusters: "0x25,0x72,0x86,0x71,0x22,0x70"
-		fingerprint deviceId: "0x1006", inClusters: "0x25"
 	}
 
 	// simulator metadata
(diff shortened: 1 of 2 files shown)


mckeed:
This fingerprint is already in fortrezz-water-valve, so you should take it out of that device handler as part of this PR.

Instead of the paragraph in the preferences, do this

```
def updated() {
    response(refresh())
}
```

Which will automatically call refresh() on join.

Why did you use a multiAttributeTile, isn't there just one attribute?

I think there is no need to handle DeviceResetLocallyNotification because the device will be deleted right after.

The parts where you compare `cmd.value` to a string don't make sense, because that variable is typed as a Short, so things like `cmd.value == "on"` are impossible.

Sending manufacturerSpecificGet from poll() is unnecessary and wasteful. I know zwave-switch does that but poll() isn't actually called for zwave-switch so I haven't gotten around to taking that out. I'd also prefer if it wasn't sent from refresh() once the MSR is already in the device data, but that's not a huge deal.

kwarodom:
@mckeed thanks for your feedback. I've updated the device handler as follows:
1. The common fingerprint has been removed from fortrezz-water-valve.
2. Automatically call refresh() when join to update device status
3. MultiAttributeTile is used to make water valve looks like other devices such as z-wave switch, z-wave lock etc.
4. DeviceResetLocallyNotification handler method is removed.
5. cmd.value that is compared to string is removed.
6. poll() will not send manufacturerSpecificGet. refresh() will not sent manufacturerSpecificGet once MSR is already updated.

mckeed:
:+1:

workingmonk:
@kwarodom please include ticket number in the PR for us to keep track for deployment. Same question, has this been QA approved?

PR #149, opened 2015-09-29
Master

ianisms:
+Unlock it, Light it, Announce Arrival SmartApp:

On arrival: unlocks the door and after door opens, turns on lights and announces arrival


ianisms:
Sorry for all the changes, had a snafu with Git, really just adding a single SmartApp

PR #156, opened 2015-09-29
MSA-592: WeatherBug Home Connect

SmartThings:
Submitting 'WeatherBug Home Connect' for publication


smartapps/v1/weatherbug-home.src/weatherbug-home.groovy
```diff
@@ -0,0 +1,306 @@
+/**
+ *  WeatherBug Home
+ *
+ *  Copyright 2015 WeatherBug
+ *
+ */
+definition(
+    name: "WeatherBug Home",
+    namespace: "v1",
```
kris-schaller: namespace is wrong

PR #170, opened 2015-10-05
Fixed a potential bug if a InitialState SmartApp hasn't been updated

davidsulpy:
Added a line to check if `atomicState.eventBuffer` is not initialized and if so, initialize it with an empty array before appending to the array in the eventHandler. This is to mitigate an error being thrown in some scenarios where the `atomicState.eventBuffer` is somehow not properly initialized.


kris-schaller:
:+1:

PR #177, opened 2015-10-09
Adds capabilities "Temperature Measurement" and "Relative Humidity Me…

juano2310:
…asurement"


bflorian:
It doesn't look like the Zen DTH generates relative humidity, does it?

bflorian:
:+1:

PR #191, opened 2015-10-14
CREX-2981 Update the background image for arrival devices

mrnohr:
Currently the white circle background on a white tile makes it look empty. This will make the circle gray. This will be redesigned at some point, but for now at least it won't look like an empty tile.


workingmonk:
simple fix  :+1:

PR #192, opened 2015-10-14
Fix requirement for SMS

tslagle13:
Removed requirement to provide a SMS number is the user does not have contacts. Add logic to verify a number was provided before sending SMS.


kris-schaller:
:+1:

PR #194, opened 2015-10-14
Improved dimmer

workingmonk:
Improving the DTH for dimmer. This will be our new base for dimmers and basic (non-color and non-CCT) bulbs.

This is making use of the changes that I have added to the cloud for the parsing and config improvements.

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
@@ -13,28 +13,32 @@
  */
 
 metadata {
-    definition (name: "ZigBee Dimmer", namespace: "smartthings", author: "SmartThings") {
+    definition (name: "ZigBee Dimmer Power", namespace: "smartthings", author: "SmartThings") {
         capability "Actuator"
         capability "Configuration"
         capability "Refresh"
+        capability "Power Meter"
         capability "Sensor"
         capability "Switch"
         capability "Switch Level"
 
-
-        fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008"
+        fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0B04"
+        fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0702"
     }
 
     tiles(scale: 2) {
         multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
             tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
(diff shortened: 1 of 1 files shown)


workingmonk:
@tpmanley can you look at this now? I have separated out the DTH.

devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // simulator metadata
+    simulator {
+        // status messages
+        status "on": "on/off: 1"
+        status "off": "on/off: 0"
+
+        // reply messages
+        reply "zcl on-off on": "on/off: 1"
```
tpmanley: These reply messages should be changed or removed since we don't use the zcl commands anymore

devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
```diff
@@ -0,0 +1,112 @@
...
+
+    def resultMap = zigbee.getKnownDescription(description)
+    if (resultMap) {
+        log.info resultMap
+        if (resultMap.type == "update") {
+            log.info "$device updates: ${resultMap.value}"
+        }
+        else if (resultMap.type == "power") {
+            def powerValue
+            if (device.getDataValue("manufacturer") != "OSRAM") {       //OSRAM devices do not reliably update power
```
tpmanley, line 64: Is the manufacturer set even if not specified in any of the fingerprints? Also, I don't think I'm familiar with getDataValue - what does it do?

workingmonk, line 64: getDataValue can be used to retrieve values stored in the data array: manufacturer, model, endpoint, etc

workingmonk, line 64: if the manufacturer info is not set, the getDataValue should return null

devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
workingmonk: these are simulated replies for the IDE. i agree that we need to revisit the code for simulator

devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
tpmanley, line 64: Will manufacturer be set since it isn't specified in any of the fingerprints? I thought it would only read those attributes if it was specified in a fingerprint.

workingmonk, line 64: Correct. The idea is to use this device type moving forward for all the devices/bulbs and wanted to document the OSRAM issue in the code

tpmanley:
:+1:

PR #195, opened 2015-10-15
Improved Switch DTH

workingmonk:
Updates based on the cloud methods

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -1,0 +1,101 @@
+/**
+ *  Copyright 2015 SmartThings
+ *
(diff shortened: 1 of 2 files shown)


workingmonk:
cc @tpmanley

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    // simulator metadata
+    simulator {
+        // status messages
+        status "on": "on/off: 1"
+        status "off": "on/off: 0"
+
+        // reply messages
+        reply "zcl on-off on": "on/off: 1"
```
tpmanley: We're not using the zcl commands anymore so these reply lines should be changed or even removed

tpmanley:
Otherwise looks good

:+1:

PR #196, opened 2015-10-15
Master

arkansas-coder:


workingmonk:
why the presence optional for a smartapp that uses presence ?

workingmonk:
Also, there is no description and the title of the PR is "Master"??

arkansas-coder:
Hello Vinay,

This is the only app I found that can turn off a switch after a specified
elapsed time. (There may be others, but I did not find them.) Making the
presence sensor optional allows the app to be used by persons without a
presence dongle, but users can still use them, if desired.

I'm sorry for the lack of a description, etc. This is the first time that I
used GitHub. One of its tutorials said that there is default text, but this
appears to be inaccurate. Is there something that I should do to correct
the lack of description, etc.?

Any additional suggestions that will allow me to be more professional will
be appreciated.

Thanks,
Ron Aronica

workingmonk:
Hey Ron,
there is a smartapp under Lights & Switches called "Smart Light Timer". We are continuing our efforts to make the discovery of SmartApps better.

arkansas-coder:
Thanks. It does what I want, and yes, a better way to find these apps would
be a great help for both Smart Things and its users.

PR #199, opened 2015-10-16
Replaced the hardcoded endpoint with variable

sticks18:
Replaced all instances of the harcoded endpoint Id with the function as in newer zigbee devices, so this device type can be used as a generic zigbee dimmable bulb device until device specific device types are created.


tpmanley:
I realize this is a super old PR, but just trying to do some cleanup. Most of these things have been fixed. There are still a few remaining `st wattr` commands that hardcode the endpoint and those should be changed to `zigbee.writeAttribute`.

PR #209, opened 2015-10-20
DVCSMP-668: Aeon Multi Sensor 6

kwarodom:
@mckeed  Could you review this for the Aeon Multi Sensor 6?
The following changes has been made to the original Aeon Multisensor device type handler to improve and modernize it:
1. Add "powerSupply" attribute to be able to tell power source (USB Cable/Battery)
2. Add preference page for user to customize "motion delay time", "motion sensitivity", and "sensor report interval"
3. Add color backgroud of "illuminance" value tile
4. Add tile for "ultravioletIndex"
5. Add tile for "powerSupply"
6. Modify updated() to be able to send configuration commands to sensor whether it is powered by USB cable or battery
7. When battery operated, send command to get update battery level if it hasn't been reported for a while
8. Report MSR of the sensor
9. Add handle for ConfigurationReport command class to update the "powerSupply" tile and change opetion mode of the sensor accordingly
10. Update configure() to configure parameters changed by user in preference page
11. Take out the "Configure" tile and instead send the configuration commands on every wakeup (sensor is battey powered)
Note: [this community blog post](https://community.smartthings.com/t/aeotech-multisensor-6-gen-5-zwave-plus-model-zw100-a/17732/210?) refers to the issue of getting UV reading as the only time the sensor pickup UV index value is when it is exposed to full sunlight. This issue might has to be investigated further with manufacturer.

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
@@ -24,6 +24,7 @@ metadata {
 		capability "Battery"
 
 		attribute "tamper", "enum", ["detected", "clear"]
+		attribute "powerSupply", "enum", ["USB Cable", "Battery"]
 
 		fingerprint deviceId: "0x2101", inClusters: "0x5E,0x86,0x72,0x59,0x85,0x73,0x71,0x84,0x80,0x30,0x31,0x70,0x7A", outClusters: "0x5A"
 	}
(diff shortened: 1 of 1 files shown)


kwarodom:
@mckeed I have updated the DTH (794ff6b) based on your comments as follows:
1. There shouldn't be a tile for battery vs. powered. It's redundant and pointless. 
   - Show batteryStatus tile insteady of battery tile to be able to display both when sensor is USB powered or battery powered. See below for the revised tiles.
2. Another issue is configuration appears to be broken when the device is included non-secure, which is a valid option for this device.  That non-secure config issue is also preventing the WakeUpNoMoreInformation from being sent which will kill the battery.
   - Update configure() to  send both unsecure and secure configuration commands when sensor is joined normally or securely. NWI command is sent with the non-secure config.
3. I don't think the powerSupply:failed is the best way to handle a config update when you don't know the power supply. Couldn't you just setConfigured(false) and then make some request to test if the sensor is listening? Like do a configurationGet or something and trigger configure() from the report.
   - Instead of using powerSupply:failed, configurationGet cmd is sent and then the configure() is triggered by configurationReport to determine powerSupply (USB Cable/Battery)
4. If you're already requesting battery on wake up, I don't think it needs to be also reported periodically. Requesting that the battery be every 8 minutes is a waste (of battery). 
   - Instead of querying battery level on wake up, battery report is put in association group 2 that is configured to report every 6 hours by default
5. I find it very hard to read the white text on the #F3F2E9 background of the lux tile on iOS. 
   - Remove background for illuminance. This can be added when we have best practice of showing colors for lux.

The updated screenshots are as follows:
198a4bbe-d559-4fd9-8c8c-0db027eac41c

a5a15696-45c0-4037-96f0-ef7367aac6e9

24f835d3-6a8a-4e5b-976f-2edddbb75956

mckeed:
:+1: There appears to be a race condition issue with secure configure and I have some other minor cosmetic concerns, but this is definitely better than what we have right now. The current handler is throwing exceptions so I think we should merge it and fix it up later.

PR #210, opened 2015-10-20
MSA-632: Climax Power Outlet

SmartThings:
Submitting 'Climax Power Outlet' for publication


workingmonk:
As @tpmanley pointed out, this should be part of https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy

PR #212, opened 2015-10-22
Sensibo integration with Smartthing

EricG66:


EricG66:
This a Sensibo integration for SmartThings

PR #225, opened 2015-10-27
Rename Sonos SmartApps

Yaima:
https://smartthings.atlassian.net/browse/DVCSMP-607


Yaima:
@bflorian, @thegilbertchan, @workingmonk, @juano2310, @tpmanley

workingmonk:
Do not merge!!!

The changes look good, but we need to account for the file name changes. The webhook won't take care of that.

juano2310:
@workingmonk which webhooks? just curious not to do something bad in the future.
What needs to be done?

workingmonk:
in dev and staging, we have webhooks that auto deploy after merge. in this case, instead of a rename it will create a new copy of the speaker files.

juano2310:
@workingmonk make sense... thanks!!!

Yaima:
Who can fix the hooks @workingmonk, @remasse ? We had the same issue with the SamartCam connect app in staging.

workingmonk:
@bflorian is exploring if there is a way to get the rename info out of webhook. until then, we need to manually rename before we merge this in

workingmonk:
@Yaima @juano2310 Is there a test plan for this to QA?

The input variable name has been updated in all the SmartApps. What happens to the users who already have this installed? For them, the new variable will have no data causing NPE, right?

workingmonk:
I have made the necessary changes to the filename in development for this to be merged in, but i'll apply the do not merge tag for now, until we resolve the issue with possible NPE

juano2310:
@workingmonk Do we need to write an script in this cases?

workingmonk:
either that or keep the variable name same.

Yaima:
In my opinion, since these apps are part of SmartThingsPublic, I'd change the variables as well. It is a faster approach leaving the names with sonos. I guess it is up to @thegilbertchan to decide if we want to keep the name sonos in the variables or if you want to remove it.

thegilbertchan:
Preferably I would change the variables but I'm assuming that itll have an impact on backwards compatibility. Can you guys make it happen without great complexity. Otherwise we can deploy a new speaker smartapp and reference those in the node graph and remove the old from node graph. We can keep it in the production so that it'll work for those already installed. 

I would defer to you guys I how we should deploy.. 

Warm regards,
Gilbert Chan
(415)378-4369
TheGilbertChan@gmail.com | Facebook.com/gilbertchan | Twitter.com/thegilbertchan | LinkedIn.com/in/thegilbertchan
"There's no substitute for hard work"

Yaima:
I think that's a good approach, similar to what we did with the old legacy solutions. We can deploy the speaker apps as new ones. @workingmonk, @juano2310, @thegilbertchan ?

workingmonk:
But if we keep the variable names as 'sonos', it does NOT impact the user. From a user's point of view, the preference, the name of the smart app, etc. all say 'Speaker'.

I am reluctant with the new copy with a different name approach, because in the long term, this app will be neglected causing support issues or worse copy paste issues.

Yaima:
It will if we get a new customer creating a device type and getting one of those apps as an example for their integration.

Yaima:
It is not a copy with a different name approach. It will be 4 new apps in nodegraph. We will keep the existing ones hidden, so all the new installations will use the speaker apps instead of sonos. Something similar to what we are doing with the old solutions.

workingmonk:
At that point it is not a customer, it a developer who has to navigate the variable names. The variables names will be consistent, just that the name is called 'sonos'.

the customer is impacted when they have an existing smart app called '... with sonos' which is neglected in the long run.

Yaima:
I think that leaving the name sonos is putting a patch over another patch just like everything we have been doing.

thegilbertchan:
I agree that it is potentially a problem with those calling into support not knowing there's a different version in the marketplace. I too don't like having neglected smartapps in prod. 

@bflorian any ideas?

Yaima:
Can we leave them just as is (Sonos), and create a new reference for speaker until we come out with a solution like we did for Lighting and SHM? In that case a Bose customer will not have Sonos in the reference and for new integrations, they will not see sonos if they are looking for smartapps examples.

thegilbertchan:
The reason we're doing the rename now is because we support the Samsung speaker today but all use cases reference Sonos. 

​

​If the problem is variables and we can properly name the smartapps then I'm fine moving forward with this. It's not an issue if variables are named sonos since only devs looking at the code will see it - right?

On Friday, Nov 13, 2015 at 8:50 PM, Yaima notifications@github.com, wrote:

Can we leave them just as is (Sonos), and create a new reference for speaker until we come out with a solution like we did for Lighting and SHM? In that case a Bose customer will not have Sonos in the reference and for new integrations, they will not see sonos if they are looking for smartapps examples.

workingmonk:
right

Yaima:
As long as you are ok with this, I can keep the name of the variables as they were initially. And yes, you are right, only devs will see it.

thegilbertchan:
Okay. Let me know if there are any outstanding issues to resolve to move forward.

Yaima:
I dont think there is, as long as you agree with it, I can keep the old names in place.

thegilbertchan:
Let me know if I understand what it'll be once we deploy
1) the smartapp name will say speaker instead of sonos
2) the preferences will ask you to pick your speakers instead of sonos and sonos will not be shown anywhere in the mobile app for the smartapp 
3) the smartapp code will still have the vars sonos
4) _guessing here_ sonos will be shown in the smartapp preferences within the ide <- this is not desirable but okay to move forward 
5) this will be backwards compatible and not lose functionality for any users
6) ttl will be removed from preference selection
7) because of #6 this could mean that someone that goes into prefs for a configure smartapp could break their functionality <- again not desirable but fine.

Yaima:
That's right.

workingmonk:
cool. in that case, can you make the change and squash the commits? I change is already ready in dev, but NOT in prod yet. Once we review it, we can merge this one in. 

@thegilbertchan can you follow up on whether this is going to be tagged for singleton changes to apps?

thegilbertchan:
Not singleton. A user should be able to install multiple of these smartapps since they're automations. 

On Saturday, Nov 14, 2015 at 12:12 AM, Vinay Rao notifications@github.com, wrote:

cool. in that case, can you make the change and squash the commits? I change is already ready in dev, but NOT in prod yet. Once we review it, we can merge this one in. 

@thegilbertchan can you follow up on whether this is going to be tagged for singleton changes to apps?

workingmonk:
singleton is allowing the user to install multiple of the automation without creating multiple records. like Smart Lighting, users can create multiple automation, but they will all be nested under one.

We need singleton for this, right?

workingmonk:
what is 'ttl will be removed from preference selection'?

thegilbertchan:
Should NOT be marked as singleton. It should create multiple records since a user may want to install these for various triggers. Smart Lights is different; the parent is singleton but the children are not (thus a user can install various automations).

We’re removing custom text to speech capability since our TTS is not working. Sorry mistyped TTL vs TTS. 

singleton is allowing the user to install multiple of the automation without creating multiple records. like Smart Lighting, users can create multiple automation, but they will all be nested under one.

We need singleton for this, right?

workingmonk:
cool. yup ttl is what confused me.

Yaima:
@workingmonk, @thegilbertchan var names are reversed

workingmonk:
everything looks good, can you squash the 4 commits

PR #230, opened 2015-10-28
Fix DVCSMP-1227 & DVCSMP-1232

juano2310:
https://smartthings.atlassian.net/browse/DVCSMP-1227

if you use the hue native app to adjust the dimmer setting, smartthings will reset the dimmer to previous value when toggling from ST app (and automations)

https://smartthings.atlassian.net/browse/DVCSMP-1232

a bug where if you remove a bulb from hue and not from the service manager it would fail while creating child devices. Is possible this was the issue. I sent a fix with the PR from dimming during on command.


juano2310:
@Yaima, @workingmonk, @bflorian

Yaima:
:+1:

PR #232, opened 2015-10-28
MSA-652: OAuth Endpoint

SmartThings:
Submitting 'OAuth Endpoint' for publication


imbrianj:
How does one submit new community apps or update old community apps to the platform?  I've had some changes to apps, but there's no documentation or support that I can see.  I've pretty much abandoned it as a result.  Those that contact me to report feature requests, pull requests or bug reports - I basically give them the support email and let it get ignored.

workingmonk:
cc @tslagle13

tslagle13:
Instructions on how you would publish code can be found here: http://docs.smartthings.com/en/latest/publishing/index.html#for-public-distribution

Although at the moment we are not accepting new SmartApps but will definitely take submissions for current code that needs to be updated.

PR #235, opened 2015-10-29
SHARD-211: DST fix for Button Controller

lbredeso:
Button Controller is not currently passing the location's time zone into the `timeToday` method, which does not provide enough information to properly execute across DST boundaries.

/cc @bflorian


workingmonk:
@lbredeso something that we still want to fix?

lbredeso:
@workingmonk yes

workingmonk:
change looks good. 👍 
If you can add a ticket number to the title, it will be great.

lbredeso:
@workingmonk https://smartthings.atlassian.net/browse/SHARD-211

lbredeso:
@workingmonk added, will merge now

PR #237, opened 2015-10-30
DVCSMP-1189: Wemo refactor final

juano2310:
https://smartthings.atlassian.net/browse/DVCSMP-1189

Detect and mark device offline within 5 minutes.
Show Device offline in device tile.
Show Device offline in Recent Activity.
Log the current IP address to Recent Activity.
Log the changed IP address to Recent Activity.
Support 'Turning on' and 'Turning off' (blindly changing the state of
device to ON or OFF without confirming bulb responded correctly)
Turn on / off through Wemo-App reflected timely in SmartThings
App/Ecosystem.
Manual turn on / off of device is reflected timely in SmartThings
App/Ecosystem.
Lower case createEvent
Bug Fixes

### DIFF
--- a/devicetypes/smartthings/wemo-light-switch.src/wemo-light-switch.groovy
+++ b/devicetypes/smartthings/wemo-light-switch.src/wemo-light-switch.groovy
@@ -85,7 +85,7 @@ def parse(String description) {
 	def result = []
 	def bodyString = msg.body
 	if (bodyString) {
-		unschedule("isOffline")
+		unschedule("setOffline")
 		def body = new XmlSlurper().parseText(bodyString)
 
 		if (body?.property?.TimeSyncRequest?.text()) {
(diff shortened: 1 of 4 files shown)


workingmonk:
why is isOffline a command? you are using it to just do a sendEvent. do you plan to send a command using that method in the future?

everything else looks great :+1:

juano2310:
isOffline is the function that sets the device offline if after polling the parser doesn't get a response from the device and cancels it.
    +def isOffline() {
-   //sendEvent(name: "currentIP", value: "Offline", displayed: false)
-    sendEvent(name: "switch", value: "offline", descriptionText: "The device is offline")
  +}
  
  def poll() {
  log.debug "Executing 'poll'"
  +if (device.currentValue("currentIP") != "Offline")
-    runIn(10, isOffline)

workingmonk:
yes, but i don't think we need to define it as a command. you have defined it as a custom command only for Wemo Switch.

workingmonk:
just noticed that "Wemo Switch" is the only DTH that defines isOffline as custom command.

nitpicking here, but isOffline seems like it is checking if the device is offline, what it is actually doing is setting it. setOffline seems a better name.

juano2310:
Will take care of that, a little bug in motion sensor, squash and re submit

workingmonk:
awesome :)

juano2310:
@workingmonk here are all the changes + squash https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/238

juano2310:
@workingmonk if you have a minute please check this new, squashed PR it has all the changes you requested https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/239
Thanks a lot!

PR #242, opened 2015-10-31
Fixed errors in setup screens, made people setting optional

tpmanley:
Fixed a few typos on the setup screens and comments
Made people settings optional because you might not always want to use that setting
Removed extra "paragraph" keywords that were messing up the setup screens

cc @tslagle13

### DIFF
--- a/smartapps/tslagle13/vacation-lighting-director.src/vacation-lighting-director.groovy
+++ b/smartapps/tslagle13/vacation-lighting-director.src/vacation-lighting-director.groovy
@@ -25,7 +25,7 @@ definition(
     namespace: "tslagle13",
     author: "Tim Slagle",
     category: "Safety & Security",
-    description: "Randomly turn on/off lights to simulate the appearance of a occupied home while you are away.",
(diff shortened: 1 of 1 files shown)


tpmanley:
@tslagle13 can you review when you get a chance?

tslagle13:
People option should be required. This keeps the logic from turning off lights if someone transitions from a "away state" to a "present state" and the timer is still running. 

Another option would be to kill the timer if someone comes home.

tpmanley:
For people like me that don't use presence sensors I am unable to use this app. Is there any harm in making it optional?

tslagle13:
yeah, that is true. What happens now is if you change your mode from your away mode to your home mode the timer is still running, notices you aren't in the mode once the timer executes and then turns off lights if you aren't in the away mode. 

We could change this and add some clarification so that you can specify if I am in X mode don't take any actions or kill the timer.

tslagle13:
I made a whole bunch of changes to my app. I am going to create a new PR because my changes are pretty expansive. Made people input optional.

PR #246, opened 2015-11-02
Fix 'Low Battery Handler' exception caused by non-integer battery events

tpmanley:
ZigBee locks report battery percentage remaining in .5% increments. However
the Low Battery Handler Smart App in Hello Home expects it to be an integer.

cc @workingmonk


workingmonk:
The changes look good, but we haven't handled the default value/incorrect value that can be returned for 
1. the attribute
2. the dataType

tpmanley:
I agree, but I think that's more of a pervasive problem with many of our DTH that we should address in the ZigBee API, and merging this fix as is will fix an exception that is occurring with all zigbee locks. Do you agree it's OK to merge as is?

workingmonk:
:+1:

workingmonk:
@tpmanley is it okay if this change is addressed with my open PR? my changes are in the same block of code

tpmanley:
Any reason not to just merge both prs as is? They are different lines of code so I don't expect a merge conflict

workingmonk:
in that case lets go ahead and merge this one first. i am waiting on cert to fix mine

PR #247, opened 2015-11-03
New fingerprints for Yale lock and fixing the battery issue

workingmonk:
Since Yale is going to fix the issue with OTA and new hardware, we will undo this special handling later.

cc @tpmanley


tpmanley:
+1

workingmonk:
@dkirker please review cert

workingmonk:
DEVC-250

dkirker:
:shipit:

PR #251, opened 2015-11-04
Jawbone Global Oauth

juano2310:
@llinder do you mind checking this?

### DIFF
--- a/smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
+++ b/smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
@@ -27,7 +27,7 @@ mappings {
 	path("/receivedToken") { action: [ POST: "receivedToken", GET: "receivedToken"] }
 	path("/receiveToken") { action: [ POST: "receiveToken", GET: "receiveToken"] }
 	path("/hookCallback") { action: [ POST: "hookEventHandler", GET: "hookEventHandler"] }
-    path("/oauth/initialize") {action: [GET: "oauthInitUrl"]}
+  path("/oauth/initialize") {action: [GET: "oauthInitUrl"]}
 	path("/oauth/callback") { action: [ GET: "callback" ] }
 }
 
@@ -280,10 +280,9 @@ def validateCurrentToken() {
 
 def initialize() {
     log.debug "Callback URL - Webhook"
-	def localServerUrl = getApiServerUrl()
-	def hookUrl = "${localServerUrl}/api/token/${state.accessToken}/smartapps/installations/${app.id}/hookCallback"
+		def hookUrl = buildActionUrl("hookCallback")
     def webhook = "https://jawbone.com/nudge/api/v.1.1/users/@me/pubsub?webhook=$hookUrl"
-	httpPost(uri: webhook, headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ])
+		httpPost(uri: webhook, headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ])
 }
 
 def setup() {
--- a/smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
+++ b/smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
@@ -27,15 +27,12 @@ mappings {
 	path("/receivedToken") { action: [ POST: "receivedToken", GET: "receivedToken"] }
 	path("/receiveToken") { action: [ POST: "receiveToken", GET: "receiveToken"] }
 	path("/hookCallback") { action: [ POST: "hookEventHandler", GET: "hookEventHandler"] }
-  path("/oauth/initialize") {action: [GET: "oauthInitUrl"]}
+    path("/oauth/initialize") {action: [GET: "oauthInitUrl"]}    
 	path("/oauth/callback") { action: [ GET: "callback" ] }
 }
 
-def getSmartThingsClientId() { return appSettings.clientId }
-def getSmartThingsClientSecret() { return appSettings.clientSecret }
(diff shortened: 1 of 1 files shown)


smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -287,9 +279,10 @@ def validateCurrentToken() {
 }
 
 def initialize() {
-	def hookUrl = "${serverUrl}/api/token/${state.accessToken}/smartapps/installations/${app.id}/hookCallback"
+    log.debug "Callback URL - Webhook"
+	def localServerUrl = getApiServerUrl()
+	def hookUrl = "${localServerUrl}/api/token/${state.accessToken}/smartapps/installations/${app.id}/hookCallback"
```
jimmyjames, line 275: why can't we use `buildActionUrl("hookCallback")` here, instead of building it manually?

juano2310:
Totally, Done!

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310, line 275: I was testing different alternatives but I just replaced

jimmyjames:
I have some other questions, but they are more around the proper way to do this OAuth flow. I'll wait for @llinder to review.

It would be awesome if this could be thoroughly vetted, so that we can point to this as a reference implementation of supporting the OAuth flow (but Lance or others should review to make sure this is the case).

juano2310:
@llinder, sorry I close it by mistake but it is still valid :)

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -328,7 +320,7 @@ def setup() {
 
 def installed() {
 	enableCallback()
```
llinder: I'm not sure that we need to make this call anymore.  Its effectively doing the same thing as /oauth/initialize except it won't work across shards.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -28,16 +27,16 @@ mappings {
...
-   return appSettings.clientId
-}
-
-def getSmartThingsClientSecret() {
-   return appSettings.clientSecret
-}
+def getSmartThingsClientId() { return appSettings.clientId }
+def getSmartThingsClientSecret() { return appSettings.clientSecret }
+def getServerUrl() { return "https://graph.api.smartthings.com" }
+def getShardUrl() { return getApiServerUrl() }
```
llinder: This is purely opinionated but since getShardUrl and getApiServerUrl are synonymous here I think we should just remove this method and access `apiServerUrl` everywhere.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -28,16 +27,16 @@ mappings {
...
+def getBuildRedirectUrl() { "${serverUrl}/oauth/initialize?appId=${app.id}&access_token=${state.accessToken}&apiServerUrl=${shardUrl}" }
+def buildRedirectUrl(page) { return buildActionUrl(page) }
```
llinder, line 36: Why define a function that defers to another function?  Again a purely opinionated comment but I think its clearer if we stick to method names from the smart app executor base and not redefine them in a smart app.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -77,20 +76,21 @@ def callback() {
...
+        log.debug "RedirectURL = ${redirectUrl}"
+        def donebutton= state.JawboneAccessToken != null
+        return dynamicPage(name: "Credentials", title: "Jawbone UP", nextPage: null, uninstall: true, install: donebutton) {
+               section { href url:redirectUrl, style:"embedded", required:true, title:"Jawbone UP", state: hast ,description:description }
         }
     } else {
-        description = "Jawbone Credentials Already Entered." 
+        description = "Jawbone Credentials Already Entered."
         return dynamicPage(name: "Credentials", title: "Jawbone UP", uninstall: true, install:true) {
                section { href url: buildRedirectUrl("receivedToken"), style:"embedded", state: "complete", title:"Jawbone UP", description:description }
```
llinder, line 91: I'm a bit confused as to why users are redirected to an endpoint that claims and access token if they already have an access token.

llinder:
Mostly just comments about things that are not necessary or a little confusing.  Other than that I don't see anything fundamentally wrong with the changes.

juano2310:
Thanks a lot! I will implement all the recommendations just in case they use this code as a reference.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310, line 36: This was inherited when we removed the client and secret from the code. They just replaced the constant with the call to app.setting but I can update it to clean everything a bit :)

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310, line 91: We don't redirect there, it is the Service Manager settings. That is what you see if you go back to the SmartApp. Usually you don't go there but you can use it to uninstall it.

juano2310:
@llinder thanks a lot for the comments! I cleaned as you suggested. if you give me :+1: I will proceed to merge :smile:

llinder:
:100:  looks good

PR #252, opened 2015-11-04
Force Level = 1% to 1

juano2310:
This ensures that the bulb will be able to dim to it minimum


juano2310:
@Yaima do you mind quick checking this?

Yaima:
:+1:

PR #253, opened 2015-11-04
CREX-1094 Delete stale device subscriptions on IFTTT app update

mrobinet:
If a device is removed from the IFTTT app in the mobile client it leaves an event subscription hanging around that can prevent devices from being deleted. This change removes all orphaned event subscriptions during app update.


mostlylikeable:
:+1: support will be so happy. :)

PR #261, opened 2015-11-05
Harmony Global Oauth

juano2310:
Global Oauth for Harmony. @llinder do you mind checking this one too? Thanks a lot!


llinder:
:+1:

PR #263, opened 2015-11-06
Updating selected thermostat device handlers to adopt multi-attribute tiles

pennig:


joneilloine:
This can't be merged until all mobile clients are able to handle it (or gracefully degrade if not fully handle), right?

PR #266, opened 2015-11-09
Replaced atomicState with State

juano2310:
This bug was discovered by Shodana Song. @bob, @Yaima, @workingmonk do you mind taking a look :) Thanks!


Yaima:
Since we are not using atomicState.accessToken anywhere else, :+1:

PR #270, opened 2015-11-10
[DVCSMP-1107] OG Multi Garage

workingmonk:
changes to support garage (tilt) sensor in old multi using preference


tpmanley:
:+1:

PR #272, opened 2015-11-11
MSA-677: Priv...

SmartThings:
Submitting 'Priv...' for publication


workingmonk:
The PR has no content.

kris-schaller:
@workingmonk Hey Vinay, can you close out the submission in the IDE when you close this out? :smile: Hopefully we can do this automatically in the future but for the time being it would be helpful

PR #277, opened 2015-11-13
[DEVC-242] Adding fingerprints for osram and sengled dimmable bulbs

workingmonk:


workingmonk:
@dkirker @tpmanley 
Added the OSRAM bulb to non-power DTH as their power metering is unreliable

tpmanley:
:+1:

PR #278, opened 2015-11-13
[DEVC-231] Moving re-certified device to generic zigbee DTH.

workingmonk:
@tpmanley @dkirker 

GE Link now has the full fingerprint so that it is not considered for generic devices.


workingmonk:
we have decided not to move GE Link to the generic DTH because of the bulb's unreliability and the pending OTA. we can move it to generic DTH then.
Also, the DTH has a lot of custom dimming data.

tpmanley:
:+1:

PR #279, opened 2015-11-13
[DEVC-226] Cree DTH refresh

workingmonk:
Updating tiles and code

Not moving this to generic DTH as the zigbee model and manufacturer name sent by the device has parsing issues that doesn't have a quick fix for the v1 hub.

@tpmanley once we have the fix for v1 and v2 hub, we can move this to generic DTH


tpmanley:
:+1:

tylerlange:
@workingmonk @tpmanley bump on this one

PR #280, opened 2015-11-13
temporary changes to the lock DTH 

workingmonk:
This is to account for the upcoming zigbee library changes

@tpmanley Can you review this? I have already tested it on my end for lock, unlock, refresh and config.


tpmanley:
:+1:

PR #285, opened 2015-11-14
Zigbee Lock Missing Remote Lock Code ability like the Zwave Lock has

flynhawaiian:
The Zigbee Lock is Missing the Remote Lock Code ability like the Zwave Lock has.  Can this be fixed?


workingmonk:
this is a feature request, not a PR

PR #241, opened 2015-10-31
LiFX: update oauth/callback url, use getApiServerUrl() for proxying to corresponding shard

kwarodom:
@juano2310 and @workingmonk  Can you review this?

This update is based on the [PR#207](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/207) LIFX SmartApp Update.

The [PR#1605](https://github.com/PhysicalGraph/data-management/pull/1605) "cross shard proxy logic for /oauth/callback" allows 3rd party OAuth flow to the correct shard by proxying the oauth/callback request after user authorization process. 

P.S. This change has not been tested in prod as it has to wait for the PR#1605 to get deployed in prod on  11/02/2015.

PR #291, opened 2015-11-17
LiFX - LiFX: update oauth/callback url using getApiServerUrl() for shard proxying and change shardUrl param to apiServerUrl

kwarodom:
@workingmonk Can you review this ?

This pull request is base on [PR#241](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/241) taking into account @llinder comments and adding "singleInstance: true" to the service manager definition in order to make sure the service manager is singleton.


dkirker:
:shipit:

PR #292, opened 2015-11-17
fingerprint spacing consistency, additional fingerprint color

workingmonk:
@tpmanley 

removed the fingerprint from centralite dimmer
removed unnecessary capability "Sensor" from switch, switch w/ power, dimmer, dimmer w/ power DTH
spacing consistency in the fingerprint


devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
```diff
@@ -18,15 +18,14 @@ metadata {
         capability "Configuration"
         capability "Refresh"
         capability "Power Meter"
-        capability "Sensor"
```
tpmanley: I think we should leave the sensor capability here because this device can detect power usage.

devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
```diff
@@ -17,15 +17,14 @@ metadata {
         capability "Actuator"
         capability "Configuration"
         capability "Refresh"
-        capability "Sensor"
```
tpmanley, line 20: Makes sense to remove it here because the device cannot sense anything.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -18,13 +18,12 @@ metadata {
         capability "Configuration"
         capability "Refresh"
         capability "Power Meter"
-        capability "Sensor"
```
tpmanley: I think we should leave sensor here too.

workingmonk:
@tpmanley updated based on your comments. i'll squash the commits if there are no further changes

tpmanley:
:+1:

PR #297, opened 2015-11-19
Z-Wave Lock: fix Security Exception DVCSMP-1265

mckeed:
The encryptCodes() is no longer needed as all codes are encrypted as they arrive.

Also includes a temporary fix for DVCSMP-1059 until INSIDE-214 is finished. In my testing, it is now checking battery level on my Kwikset lock.


tylerlange:
@workingmonk we need to deploy this, can you take a look?

workingmonk:
deploying this now

PR #339, opened 2015-12-07
Ecobee: retries to get refresh tokens if exceptions are caught 

kwarodom:
@juano2310 Can you review this PR?

In this change, before sending push and notification message to user to re-login to their Ecobee account, the service manager will make attempts 3 times each at every 5 minutes to get the access token using the stored refresh token (expire within one year). 

For the thermostat DTH, i just added (default) to holdType option so that users know which option is the default if they don't provide input.


juano2310:
:+1:

PR #355, opened 2015-12-11
multi: Fix invalid accelerometer readings during refresh

tpmanley:
The code can only correctly handle the accelerometer readings when they
are all received in the same message. Individual reads of the attributes
results in individual read attribute responses which led to events with
incorrect values. The most straightforward fix is to not read the attributes
in the first place and rely on attribute reporting to report the values
at the appropriate time.

Resolves:
https://smartthings.atlassian.net/browse/DVCSMP-1315


tpmanley:
@workingmonk @tylerlange

tpmanley:
@Yaima could review this when you get a chance?

Yaima:
This looks good and I agree with the fix :+1:

PR #365, opened 2015-12-14
Improve performance when streaming to Inital State and prevent buffer growing unbounded

rappleg:
Reduce DB hits to access atomicState by several times per execution of the genericHandler.  This greatly reduces SmartApp execution latency and allows the events to be streamed to Initial State faster.

More importantly this fixes several edge cases where the eventBuffer can grow unbounded and once it becomes too large will always fail when trying to post to Initial State.

/cc @bflorian

### DIFF
--- a/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
+++ b/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
@@ -11,9 +11,9 @@
  *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
  *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
  *  for the specific language governing permissions and limitations under the License.
- *  
+ *
  *  SmartThings data is sent from this SmartApp to Initial State. This is event data only for
- *  devices for which the user has authorized. Likewise, Initial State's services call this 
+ *  devices for which the user has authorized. Likewise, Initial State's services call this
  *  SmartApp on the user's behalf to configure Initial State specific parameters. The ToS and
  *  Privacy Policy for Initial State can be found here: https://www.initialstate.com/terms
  */
@@ -90,7 +90,7 @@ def subscribeToEvents() {
 	if (beacons != null) {
 		subscribe(beacons, "presence", genericHandler)
 	}
-	
+
 	if (cos != null) {
 		subscribe(cos, "carbonMonoxide", genericHandler)
 	}
@@ -218,7 +218,7 @@ def setAccessKey() {
 		atomicState.grokerSubdomain = "$newGrokerSubdomain"
 		atomicState.isBucketCreated = false
 	}
-    
+
 	if (newAccessKey && newAccessKey != atomicState.accessKey) {
 		atomicState.accessKey = "$newAccessKey"
 		atomicState.isBucketCreated = false
@@ -262,7 +262,7 @@ def uninstalled() {
 }
 
 def tryCreateBucket() {
-	
+
 	// can't ship events if there is no grokerSubdomain
 	if (atomicState.grokerSubdomain == null || atomicState.grokerSubdomain == "") {
 		log.error "streaming url is currently null"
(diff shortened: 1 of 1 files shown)


smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -331,58 +330,57 @@ def genericHandler(evt) {
...
 
 // a helper function for shipping the atomicState.eventBuffer to Initial State
 def tryShipEvents() {
-
 	// can't ship events if there is no grokerSubdomain
-	if (atomicState.grokerSubdomain == null || atomicState.grokerSubdomain == "") {
+	if (state.grokerSubdomain == null || state.grokerSubdomain == "") {
 		log.error "streaming url is currently null"
+		// Avoid the eventBuffer from growing unbounded if streaming url isn't set
+		atomicState.eventBuffer = []
```
kris-schaller: shouldn't this be state?

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -331,58 +330,57 @@ def genericHandler(evt) {
...
 		return
 	}
 	// can't ship if access key and bucket key are null, so finish trying
-	if (atomicState.accessKey == null || atomicState.bucketKey == null) {
+	if (state.accessKey == null || state.bucketKey == null) {
+		// Avoid the eventBuffer from growing unbounded if no accessKey or bucketKey is set
+		atomicState.eventBuffer = []
```
kris-schaller: shouldn't this be state?

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -331,58 +330,57 @@ def genericHandler(evt) {
 // after a specified amount of time to reduce the load on ST servers
 def flushBuffer() {
 	log.trace "About to flush the buffer on schedule"
-	if (atomicState.eventBuffer != null && atomicState.eventBuffer.size() > 0) {
+	if (state.eventBuffer != null && state.eventBuffer.size() > 0) {
```
kris-schaller: @rappleg state.eventBuffer is not created, so it will always be null

davidsulpy:
I originally wrote this using `atomicState` because the SmartThings documentation says:

> It’s also worth noting that you should not use both state and atomicState in the same SmartApp. Doing so will likely cause inconsistencies in in state values.

In the last sentence of the last note of this page: http://docs.smartthings.com/en/latest/smartapp-developers-guide/state.html

I would have otherwise been happy with using `state` over `atomicState` here.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -331,58 +330,57 @@ def genericHandler(evt) {
...
+	def eventBuffer = atomicState.eventBuffer
 	def eventPost = [
-		uri: "https://${atomicState.grokerSubdomain}.initialstate.com/api/events",
+		uri: "https://${state.grokerSubdomain}.initialstate.com/api/events",
 		headers: [
 			"Content-Type": "application/json",
-			"X-IS-BucketKey": "${atomicState.bucketKey}",
-			"X-IS-AccessKey": "${atomicState.accessKey}",
+			"X-IS-BucketKey": "${state.bucketKey}",
+			"X-IS-AccessKey": "${state.accessKey}",
```
davidsulpy: If the `bucketKey` and `accessKey` are going to be changed to be on `state` as opposed to `atomicState` they need to be changed also in the `setAccessKey`, `getAccessKey`, `setBucketKey`, and `getBucketKey` methods as well.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -331,58 +330,57 @@ def genericHandler(evt) {
...
 		tryShipEvents()
 	}
 }
 
 // a helper function for shipping the atomicState.eventBuffer to Initial State
 def tryShipEvents() {
-
 	// can't ship events if there is no grokerSubdomain
-	if (atomicState.grokerSubdomain == null || atomicState.grokerSubdomain == "") {
+	if (state.grokerSubdomain == null || state.grokerSubdomain == "") {
```
davidsulpy: if the `grokerSubdomain` is going to be used here in the evaluation, it needs to be set at some point in the app, this commit is only setting the `atomicState.grokerSubdomain` so this conditional will always evaluate as true and no events will ever be shipped because the url will not be correct.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -330,57 +330,53 @@ def genericHandler(evt) {
 // This is a handler function for flushing the event buffer
 // after a specified amount of time to reduce the load on ST servers
 def flushBuffer() {
+	def eventBuffer = atomicState.eventBuffer
```
rappleg, line 333: Pull eventBuffer off of atomicState here and assign to variable here to avoid hitting the DB twice to check != null and size.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -330,57 +330,53 @@ def genericHandler(evt) {
...
 	log.trace "About to flush the buffer on schedule"
-	if (atomicState.eventBuffer != null && atomicState.eventBuffer.size() > 0) {
-		tryShipEvents()
+	if (eventBuffer != null && eventBuffer.size() > 0) {
+		atomicState.eventBuffer = []
```
rappleg, line 336: Clear the eventBuffer here shortly after we retrieve it so that we don't miss events between now and when we actually post the current eventBuffer to Initial State.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -330,57 +330,54 @@ def genericHandler(evt) {
...
+		tryShipEvents(eventBuffer)
 	}
+	log.debug "Event added to buffer: " + eventBuffer
 }
 
 // a helper function for shipping the atomicState.eventBuffer to Initial State
-def tryShipEvents() {
+def tryShipEvents(eventBuffer) {
 
+	def grokerSubdomain = atomicState.grokerSubdomain
```
rappleg, line 359: Pull grokerSubdomain off of atomicState here and assign to variable here to avoid hitting the DB twice to check != null and size.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -330,57 +330,54 @@ def genericHandler(evt) {
...
 	// can't ship events if there is no grokerSubdomain
-	if (atomicState.grokerSubdomain == null || atomicState.grokerSubdomain == "") {
+	if (grokerSubdomain == null || grokerSubdomain == "") {
 		log.error "streaming url is currently null"
 		return
 	}
+	def accessKey = atomicState.accessKey
+	def bucketKey = atomicState.bucketKey
```
rappleg, line 366: Pull off accessKey and bucketKey off here to avoid hitting the DB twice.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -389,13 +386,10 @@ def tryShipEvents() {
 			log.debug "shipped events and got ${resp.status}"
 			if (resp.status >= 400) {
 				log.error "shipping failed... ${resp.data}"
-			} else {
-				// clear the buffer
-				atomicState.eventBuffer = []
```
rappleg, line 394: No need to clear the buffer here on success since it will have already of been cleared before calling tryShipEvents to remove the chance of it possibly growing unbounded if config is bad or if the eventBuffer grows too large to be sent to Initial State without error.

rappleg:
I changed back to use all atomicState to avoid the confusion and just assigned the duplicated access keys to use local variables instead since it is a best practice to use all atomicState rather than a mix of both.

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
kris-schaller, line 333: :+1:

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
kris-schaller, line 336: :+1:

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
kris-schaller, line 359: :+1:

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
kris-schaller, line 366: :+1:

smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
kris-schaller, line 394: :+1:

kris-schaller:
:+1: Thanks @rappleg

PR #366, opened 2015-12-14
Updated the every element app to better represent what each element can and can't do

vlaminck:
I went through every element and actually thought through what each one does to try and better represent each element.
- Added `paragraph` elements to help explain differences and point out subtle or non-intuitive aspects of the element. 
- Grouped elements into sections to help organize. EX: grouped by style or functionality
- Separated input elements into their own pages since there is enough differentiation between them.
- Used `groupedOptions` where possible because a future platform change will merge `options` and `groupedOptions`.


kris-schaller:
:+1:

PR #367, opened 2015-12-15
Make sure we persist the updated eventBuffer with the new event added back to atomicState

rappleg:
This got missed on the last PR, but making sure to update the eventBuffer with the newly added event back to atomicState.

/cc @kris-schaller


kris-schaller:
:+1:

PR #369, opened 2015-12-15
Ecobee Update

snakedog116:
Update ecobee-thermostat.groovy …
Added some description to the temporary hold to help explain what kind of temp hold it is.

Update ecobee-connect.groovy …
Replace the licensing information as previously deleted by 1b9d2fe @kwarodom. I believe this was removed in error. The original code was under Apache, so the resulting code should be too.


kwarodom:
:+1:

PR #375, opened 2015-12-18
add event translation

Bijnagte:


Bijnagte:
good point, I will eliminate the need for translation then. It will be a little larger change

bflorian:
:+1:

PR #379, opened 2015-12-19
MSA-753: Nobody Home

SmartThings:
Submitting 'Nobody Home' for publication


imbrianj:
This is an update to this existing smart app: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/smartapps/imbrianj/nobody-home.src/nobody-home.groovy

I wasn't sure if it was preferred to publish a new one or file a PR against the existing.  Courtesy of @ai7

PR #381, opened 2015-12-21
Adding paths to locally hosted images for bose sound control

kris-schaller:
@tylerlange Changing icon paths for locally hosted images.


tylerlange:
:+1:

PR #383, opened 2015-12-21
Fine-tuning the Motion Threshold and Motion Threshold multiplier 

umangparekh:
@bflorian  Fine-tuning the Motion Threshold and Motion Threshold multiplier for manufacturer : Smarthings.

### DIFF
--- a/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
@@ -371,21 +371,50 @@ def getTemperature(value) {
 
 	def refresh() {
 		log.debug "Refreshing Values "
-		def refreshCmds = [
         
-        /* sensitivity - default value (8) */
+        def refreshCmds = []
         
-        "zcl mfg-code ${manufacturerCode}", "delay 200",
-        "zcl global write 0xFC02 0 0x20 {02}", "delay 200",
-        "send 0x${device.deviceNetworkId} 1 1", "delay 400",
+		if (device.getDataValue("manufacturer") == "SmartThings") {
+        	
+			log.debug "Refreshing Values for manufacturer: SmartThings "
+         	refreshCmds = refreshCmds + [
 
-		"st rattr 0x${device.deviceNetworkId} 1 0x402 0", "delay 200",
-		"st rattr 0x${device.deviceNetworkId} 1 1 0x20", "delay 200",
+	            /* These values of Motion Threshold Multiplier(01) and Motion Threshold (D200) 
+	               seem to be giving pretty accurate results for the XYZ co-ordinates for this manufacturer. 
+	               Separating these out in a separate if-else because I do not want to touch Centralite part 
+	               as of now. 
+	            */
 
-        "zcl mfg-code ${manufacturerCode}", "delay 200",
-        "zcl global read 0xFC02 0x0010",
-        "send 0x${device.deviceNetworkId} 1 1","delay 400"
-	]
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
+	            "zcl global write 0xFC02 0 0x20 {01}", "delay 200",
+	            "send 0x${device.deviceNetworkId} 1 1", "delay 400",
+	            
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
+	            "zcl global write 0xFC02 2 0x21 {D200}", "delay 200",
+            	"send 0x${device.deviceNetworkId} 1 1", "delay 400",
+                
+            ]
+            
+        
+        } else {
+             refreshCmds = refreshCmds + [
+
+                /* sensitivity - default value (8) */
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
+                "zcl global write 0xFC02 0 0x20 {02}", "delay 200",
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -371,21 +371,67 @@ def getTemperature(value) {
...
+	            "send 0x${device.deviceNetworkId} 1 1", "delay 400",
+
+	            "st rattr 0x${device.deviceNetworkId} 1 0x402 0", "delay 400",
+	            "st rattr 0x${device.deviceNetworkId} 1 1 0x20", "delay 400",
+
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
+	            "zcl global read 0xFC02 0x0010",
+	            "send 0x${device.deviceNetworkId} 1 1","delay 400",
+
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
```
tpmanley: Reading the x,y,z attributes individually won't achieve the desired result. See https://smartthings.atlassian.net/browse/DVCSMP-1315 for more information. The reads should be removed from here like they were removed for Centralite.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -371,21 +371,67 @@ def getTemperature(value) {
...
 		log.debug "Refreshing Values "
-		def refreshCmds = [
         
-        /* sensitivity - default value (8) */
+        def refreshCmds = []
         
-        "zcl mfg-code ${manufacturerCode}", "delay 200",
-        "zcl global write 0xFC02 0 0x20 {02}", "delay 200",
-        "send 0x${device.deviceNetworkId} 1 1", "delay 400",
+		if (device.getDataValue("manufacturer") == "SmartThings") {
```
tpmanley: I think it would be preferable to just put the motion threshold related commands in an if/else block, since all the other commands are the same for both manufacturers. That way there is less duplicated code.

umangparekh:
@tpmanley I have verified at my end that the Garage Door use case works + implemented previous code review feedback. Let me know what you think.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -461,19 +489,34 @@ private Map parseAxis(String description) {
...
             log.debug "X Part: ${signedX}"
         }
+        // Y and the Z axes are interchanged between SmartThings's implementation and Centralite's implementation
         else if (part.startsWith("13")) {
 			def unsignedY = hexToInt(part.split("13")[1].trim())
 			def signedY = unsignedY > 32767 ? unsignedY - 65536 : unsignedY
-			xyzResults.y = signedY
-            log.debug "Y Part: ${signedY}"
+			if (device.getDataValue("manufacturer") == "SmartThings") {
+                xyzResults.z = signedY
```
tpmanley: Based on my testing, `signedY` needs to be negated in order to match the Centralite implementation. As it is currently implemented, when the Z value is positive on Centralite it is negative on the other one.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -371,21 +371,49 @@ def getTemperature(value) {
 
 	def refresh() {
 		log.debug "Refreshing Values "
-		def refreshCmds = [
         
-        /* sensitivity - default value (8) */
+        def refreshCmds = [
+	            "zcl mfg-code ${manufacturerCode}", "delay 200",
```
tpmanley: Since the `zcl mfg-code` is part of the WriteAttribute transaction including the `zcl global write` and the `send` it is OK and preferable to put it in the if/else block below even though it means there is a little bit of duplicate code.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -371,21 +371,49 @@ def getTemperature(value) {
...
-		"st rattr 0x${device.deviceNetworkId} 1 0x402 0", "delay 200",
-		"st rattr 0x${device.deviceNetworkId} 1 1 0x20", "delay 200",
+                /* sensitivity - default value (8) */
+                "zcl global write 0xFC02 0 0x20 {02}", "delay 200",
+            ]
+        }
+        
+        //Common refresh commands
+        refreshCmds = refreshCmds + [
+            "send 0x${device.deviceNetworkId} 1 1", "delay 400",
```
tpmanley: Similar to the `zcl mfg-code`, the `send` is part of the WriteAttribute transaction and should be moved to the if/else block

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -553,3 +596,5 @@ private byte[] reverseArray(byte[] array) {
 	return array
 }
 
+
```
tpmanley, line 600: Minor, but these extra newlines can be deleted

PR #384, opened 2015-12-22
Cleanup and refactoring of Ecobee device and app prior to adding new features

StrykerSKS:
This is a cleanup of the Ecobee device and SmartApp prior to introducing additionally features and capabilities.


StrykerSKS:
I'm currently doing a fairly major rewrite on top of this, so perhaps it makes sense to hold off on this one and wait for the final version?

PR #385, opened 2015-12-22
Adding multiAttributeTile definition to Simulated Thermostat device type handler

pennig:
As we make further developments in the area of Multi-Attribute Tiles, we owe it to our developer community to provide a good example of them in practice. This is such an example.


KyleLeneau:
:+1:

PR #386, opened 2015-12-23
Fine-tuning the Motion Threshold and Motion Threshold multiplier for …

umangparekh:
…manufacturer Smarthings.

Garage Door use case works + implemented code review feedback

Implementing code review feedback + inverted the signedY to match Centralite like behavior.


tpmanley:
:+1: tested x,y,z reported values and garage door use case and compared against Centralite sensor, looks equivalent

PR #394, opened 2015-12-24
Allowing the Zen Thermostat to be a temperature measurement device

terafin:
Just added a simple capability here.


terafin:
Going to fix the squash.

PR #399, opened 2015-12-29
Changing icon paths to hosted files

kris-schaller:


kris-schaller:
Simple path change, :+1:

PR #407, opened 2015-12-31
multi: ignore attribute reports that don't include all three axis

tpmanley:
Resolves:
    https://smartthings.atlassian.net/browse/DVCSMP-1366

@workingmonk @Yaima, could one of you review please?


Yaima:
:+1:

PR #411, opened 2016-01-04
multi: Fix x,y,z parsing error

tpmanley:
The previous axis value parsing code was very fragile. For example, this
code block:

```
def unsignedY = hexToInt(part.split("13")[1].trim())
```

would fail when `part` was "13xx13", where "xx" is any value. The split
assumed the value "13" was present only once in the string, and everything
after the "13" was the value. When "13" was part of the value this code
would interpret only "xx" as the value, instead of "xx13".

The new parsing code is not fragile like this. It knows exactly what bytes
of the string are X, Y, and Z and parses the values correctly.


tpmanley:
Please review @Yaima or @workingmonk 
FYI @tylerlange

Yaima:
:+1:

PR #419, opened 2016-01-06
multi: Adjust threshold multiplier to prevent motion detection while …

tpmanley:
…idle

The previous threshold multiplier was found to be too low so motion was
being detected while the sensor was idle. This new value of 630 seems to
produce better results.


tpmanley:
@umangparekh can you please review?

umangparekh:
:+1:

PR #426, opened 2016-01-08
Increased delay to mark device as offline

juano2310:


juano2310:
@Yaima @workingmonk do you mind checking this simple PR. It's just an increase in the delay time. Thanks!

Yaima:
:+1:

forrp2:
Do we know why 10 minute checking worked well in Oct 2015 with the beta, but did not work well in Jan 2016 in production? 

Was 30 minutes the shortest time that would work?  ie. was 15 minutes, 20 minutes, 25 minutes no good?

juano2310:
Those are no minutes. Those are seconds.

PR #430, opened 2016-01-11
PROB-537 - Fix error in line 337

juano2310:


juano2310:
@Yaima @workingmonk quick PR just moving a log inside the condition.

Yaima:
:+1:

PR #441, opened 2016-01-15
arrival: Keep 'read attr' messages from being turned into events

tpmanley:
Previously parse was returning null which causes the platform to create an event
using the message passed to parse. We don't want that to happen so return
an empty list instead.

Resolves:
    https://smartthings.atlassian.net/browse/SMJN-38


tpmanley:
please review @lbredeso @workingmonk

lbredeso:
:+1:

PR #443, opened 2016-01-18
Feature/sensor updates

tpmanley:
Please review @workingmonk @umangparekh

### DIFF
--- a/devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
+++ b/devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
@@ -88,31 +88,37 @@ private handleReportAttributeMessage(String description) {
     }
 }
 
-private handleBatteryEvent(rawValue) {
-    def linkText = getLinkText(device)
-
-    def eventMap = [
-        name: 'battery',
-        value: '--'
-    ]
-
-    def volts = rawValue / 10
-    if (volts > 0){
-        def minVolts = 2.0
-        def maxVolts = 2.8
+/**
+ * Create battery event from reported battery voltage.
+ *
+ * @param volts Battery voltage in .1V increments
+ */
+private handleBatteryEvent(volts) {
+    if (volts == 0 || volts == 255) {
+        log.debug "Ignoring invalid value for voltage (${volts/10}V)"
+    }
+    else {
+        def batteryMap = [28:100, 27:100, 26:100, 25:90, 24:90, 23:70,
+                          22:70, 21:50, 20:50, 19:30, 18:30, 17:15, 16:1, 15:0]
+        def minVolts = 15
+        def maxVolts = 28
 
         if (volts < minVolts)
             volts = minVolts
         else if (volts > maxVolts)
             volts = maxVolts
-        def pct = (volts - minVolts) / (maxVolts - minVolts)
-
-        eventMap.value = Math.round(pct * 100)
-        eventMap.descriptionText = "${linkText} battery was ${eventMap.value}%"
+        def pct = batteryMap[volts]
+        if (pct != null) {
+            def linkText = getLinkText(device)
+            def eventMap = [
+                name: 'battery',
+                value: pct,
+                descriptionText: "${linkText} battery was ${pct}%"
+            ]
+            log.debug "Creating battery event for voltage=${volts/10}V: ${eventMap}"
+            sendEvent(eventMap)
+        }
     }
-
-    log.debug "Creating battery event: ${eventMap}"
-    sendEvent(eventMap)
 }
 
 private handlePresenceEvent(present) {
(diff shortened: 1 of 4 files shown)


tpmanley:
This PR includes a whitespace change. It was just getting to be too difficult to make changes because of all the inconsistent whitespace issues. To see the non-whitespace changes just add `?w=1` to the end of the PR URL.

tpmanley:
I found a bug in the new battery code for multi/moition/moisture that I need to fix and will update the PR.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -302,132 +302,141 @@ def getTemperature(value) {
...
-		log.debug "Temperature"
-		def linkText = getLinkText(device)
-		if (tempOffset) {
-			def offset = tempOffset as int
-			def v = value as int
-			value = v + offset
+			if (device.getDataValue("manufacturer") == "SmartThings") {
+				volts = rawValue // For the batteryMap to work the key needs to be an int
+				def batteryMap = [28:100, 27:100, 26:100, 25:90, 24:75, 23:50,
+								  22:30, 21:15, 20:5, 19:3, 18:0, 17:0, 16:0]
```
workingmonk: centralite sensors tend to go bust at 23

tpmanley: Yeah, this doesn't apply to CL. Check out the Jira for more info

workingmonk: oops, overlooked your manufacturer check there

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -215,24 +215,47 @@ def getTemperature(value) {
...
+		value: '--'
+	]
+
 	def volts = rawValue / 10
-	def descriptionText
-	if (volts > 3.5) {
-		result.descriptionText = "${linkText} battery has too much power (${volts} volts)."
-	}
+
+	if (rawValue == 255) {}
```
tpmanley: Previously this sensor wasn't checking to make sure the battery attribute was valid (!= 0xFF)

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -230,30 +230,46 @@ def getTemperature(value) {
...
 	def result = [
 		name: 'battery',
 		value: '--'
 	]
 
 	def volts = rawValue / 10
-	def descriptionText
 
-	if (rawValue == 0) {}
+	if (rawValue == 255) {}
```
tpmanley: Technically 0 is a valid value, but if it is actually 0 then it's wrong because the sensor couldn't possibly be powered. I think I'm going to update all the sensors to report '--' if rawValue is 0 or 255

umangparekh:
Pardon my ignorance about the DTH framework. But, Is there a way all the sensors could be derived from a common base class? If this can be done, then I suggest we move the batteryMap change for SmartThings sensors in the base class. Let me know what you guys think. 

Other than that: :+1:

tpmanley:
@umangparekh , there isn't really a good way to do that unfortunately

devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
```diff
@@ -88,27 +88,33 @@ private handleReportAttributeMessage(String description) {
...
-
+/**
+ * Create battery event from reported battery voltage.
+ *
+ * @param volts Battery voltage in .1V increments
+ */
+private handleBatteryEvent(volts) {
     def eventMap = [
         name: 'battery',
         value: '--'
```
workingmonk: not sure if this is a good idea. I have seen devices sometimes inadvertently report 0 or FF in the middle of their life cycle. Setting it to '--' causes two issues:
1. In the device tile you suddenly lose the currently battery level of the device. And probably won't get an update on its battery for probably a day
2. Battery level is tied to Low Battery Monitoring, not sure how well it is going to accept the non-int value

tpmanley: I did it that way because that's the way the other sensors were implemented. Do you think I should change them all to not report any battery event if the value is 0 or 255?

workingmonk: That might make more sense. This along with the map work that you've done for battery is something that we should standardize for other battery devices

tpmanley:
@workingmonk, I fixed the issue for the arrival. I think it's a bigger change to the other sensors which I'm not sure should be made as part of this PR. What do you think?

workingmonk:
agree. we should have a separate PR for other battery methods

workingmonk:
looks good :+1:

PR #454, opened 2016-01-20
deprecating this DTH as the fingerprints have been moved to ZigBee RGBW

workingmonk:
cc @tylerlange @tpmanley 

commenting out the fingerprint so that device pairing for the flex strip can move to zigbee rgbw


tylerlange:
:+1:

PR #467, opened 2016-01-23
New Initial State Event Streamer: removed buffering and sched-tasks

davidsulpy:
Now Initial State Event Streamer doesn't use the `atomicState` for event storage any more for buffered events. Circumvents potential race condition with `atomicState.eventBuffer` and issues with scheduling buffer flushes. Also, significantly reduces write load on `atomicState`.

### DIFF
--- a/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
+++ b/smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
@@ -1,7 +1,7 @@
 /**
  *  Initial State Event Streamer
  *
- *  Copyright 2015 David Sulpy
+ *  Copyright 2016 David Sulpy
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  *  in compliance with the License. You may obtain a copy of the License at:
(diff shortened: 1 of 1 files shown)


smartapps/initialstate-events/initial-state-event-streamer.src/initial-state-event-streamer.groovy
```diff
@@ -327,37 +325,17 @@ def genericHandler(evt) {
...
+	def event = new JsonSlurper().parseText("{\"key\": \"$name\", \"value\": \"$value\", \"epoch\": \"$epoch\"}")
+
+	tryShipEvents(event)
+	
+	log.debug "Shipped Event: " + event
 }
 
-// a helper function for shipping the atomicState.eventBuffer to Initial State
-def tryShipEvents(eventBuffer) {
+def tryShipEvents(event) {
```
kris-schaller, line 338: Wouldn't this try to post the event every time rather than batching the requests? My only fear with this is that we'll be creating a lot of HTTP requests. @davidsulpy

davidsulpy, line 338: @kris-schaller Yes, this is correct, it will make an HTTP request for every event that the SmartApp is configured to listen to. This is opposed to the previous version that was using an array on atomicState in order to store un-shipped events in a buffer/queue. The buffering (batching) was removed due to issues with the atomicState for storing and retrieving event arrays as well as issues with scheduled tasks to flush buffers that aren't full but are aging. These issues made the SmartApp unusable for most if not all of it's users as of late.

kris-schaller:
This is the lesser of two evils. atomicState flushing is causing race conditions. We're going to merge this in and work on a better solution. Since the current integration is broken this needs to be deployed to fix it.

PR #474, opened 2016-01-29
multi: Fix occasional failures in sending ZigBee commands during configure

tpmanley:
The missing delay after some of the commands was sometimes causing the
ZigBee command to fail.

Resolves:
    https://smartthings.atlassian.net/browse/DVCSMP-1225


tpmanley:
@workingmonk @tylerlange

workingmonk:
:+1:

PR #480, opened 2016-02-02
Bugfixes for codeReports and AlarmReport

rboy1:
1. Missing break in case 32 for AlarmReport
2. Bugfix - Not reporting all codes programmed into the lock causing SmartApps to fail while awaiting a codeReport notification (once a code is programmed into a lock it should be notified because there is no way to read back codes from the lock, apps awaiting a codeReport will go into an infinite programming loop to trying to update the codes and awaiting a codeReport notification indicating a successful programming)

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -275,6 +275,7 @@ def zwaveEvent(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd) {
 		case 32:
 			map = [ name: "codeChanged", value: "all", descriptionText: "$device.displayName: all user codes deleted", isStateChange: true ]
 			allCodesDeleted()
+			break
 		case 33:
 			map = [ name: "codeReport", value: cmd.alarmLevel, data: [ code: "" ], isStateChange: true ]
 			map.descriptionText = "$device.displayName code $cmd.alarmLevel was deleted"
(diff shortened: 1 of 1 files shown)


tylerlange:
@mckeed @dkirker please review

mckeed:
Thanks for the submission RB – sorry I haven't had time to do this myself.

I would prefer the "was reset" event to be displayed: false because I think it will be confusing for people to see it in their device activity stream, but since it's mostly people using your app that are seeing these it's not a big deal to me.

:+1:

@workingmonk can you deploy this at some point

workingmonk:
@rboy1 are you planning on working on making changes based on @mckeed suggestion? If not we can merge this in.

@tylerlange can you link this to a ticket so that I can followup with this during deployment and have this QA approved?

tylerlange:
https://smartthings.atlassian.net/browse/DVCSMP-1476

rboy1:
I think you may be right @mckeed
Let me take another look at it.
It should be a state change absolutely. Should it be exposed to the user? Question. What is the downside of not showing it?

rboy1:
So I checked it out, the reset is one of the three options depending upon the type of lock. I think it should be showed on the screen. However I agree maybe the verbiage is not the best. Instead of reset it should say deleted. (that's really what's is going right?)?

mckeed:
It's like that because it's not for when you're trying to delete a code. Schlage locks don't send the actual code in their reports so if we just set and then get, we can't tell if the code was actually changed because the report looks the same. To get around that, I first delete the code and make sure it's blank (that's the cause of the 'was reset' event) and then set it to the intended code. So it's just an implementation detail that shouldn't be exposed to the user.

rboy1:
I got that from the code, so I'm suggesting we report it as deleted to the user, keep it consistent. The live logs can indicate the reset for the devs. It's important to report it since it was deleted (assume if the z-wave transactions are incomplete, the last event should be reported)

rboy1:
Need a little assistance here, how do I made edits to the pull request? Or @mckeed can you do it? Just need to change "reset" to "deleted" in the map.descriptionText and it's good to go.

rboy1:
@mckeed please go ahead with this code as is. I'll initiate a separate pull request to change "reset" to "deleted" (unless you can change it). I would like to get this out at the earliest please.

rboy1:
@mckeed sorry to keep bugging you but would like to get this patch out please, too many users are pinging me about issues with the stock st handler.

mckeed:
I already gave my +1

@workingmonk can you merge and deploy this please

rboy1:
thanks @mckeed 
@workingmonk any luck getting this merged / deployed?

tylerlange:
@rboy1 we're deploying today. I'll get it in.

rboy1:
Awesome!

PR #481, opened 2016-02-02
MSA-862: EyXAr Notifications

SmartThings:
Submitting 'EyXAr Notifications' for publication


rmasangkay:
For testing and simulation voice notification it will require For ST app, currently supporting only Android devices. Without the For ST app, it will still send notifications to devices with SmartThings user only..

For ST app is now uploaded in Google Play Store for Beta Testing.

Use this link for testing the app For ST:
https://play.google.com/apps/testing/eyxar.com.forst

PR #483, opened 2016-02-04
DVCSMP-399 Add initial jammed state.

lucaschristian:
[Review on Reviewable](https://reviewable.io/reviews/smartthingscommunity/smartthingspublic/483)


dkirker:
@mckeed I was asked by @tylerlange to have you take a quick look if you think this makes sense for showing jammed for locks. I guess we just didn't have that state.

mckeed:
The state needs to be added to the Capability too.

PR #494, opened 2016-02-09
Fixed ArrayIndexOutOfBoundsException

larsfinander:
Fixed ArrayIndexOutOfBoundsException from events that lack values to some fields.

We are seeing around 500k ArrayIndexOutOfBounds exceptions from a number of the different XXXX (Connect) SmartApps that control LAN devices. The reason is that events are received that are missing values for certain fields and that causes an exception when a non existing value is getting accessed.

Below is an example that would cause an exception when parsing the "ssdpPath:" field (.split(":")[1] fails)

devicetype:04, mac:00166C8C6901, networkAddress:00000000, deviceAddress:0000, stringCount:04, ssdpPath:, ssdpUSN:uuid:Upnp-BasicDevice-1_0-00166C8C6901::upnp:rootdevice, ssdpTerm:, ssdpNTS:ssdp:aliv

This is a fix to that problem.

### DIFF
--- a/smartapps/smartthings/samsung-tv-connect.src/samsung-tv-connect.groovy
+++ b/smartapps/smartthings/samsung-tv-connect.src/samsung-tv-connect.groovy
@@ -316,60 +316,40 @@ private def parseEventMessage(String description) {
 	parts.each { part ->
 		part = part.trim()
 		if (part.startsWith('devicetype:')) {
-			def valueString = part.split(":")[1].trim()
-			event.devicetype = valueString
+			part -= "devicetype:"
+			event.devicetype = part.trim()
 		}
 		else if (part.startsWith('mac:')) {
-			def valueString = part.split(":")[1].trim()
-			if (valueString) {
-				event.mac = valueString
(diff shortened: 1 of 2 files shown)


joneilloine:
Can you check if Sonos (Connect) has the same issue? It is also throwing a ton of AIOOB exceptions.

tylerlange:
@juano2310

juano2310:
:+1:

larsfinander:
@joneilloine 
I am making a separate PR for :
intel-device-connect.groovy
samsung-audio-connect.groovy
sonos-connect.groovy

lbredeso:
FYI, I'm going to make PR's for Wemo (Connect), Hue (Connect), and Sonos (Connect) that will hopefully also help with this error, as they will cause the SSDP alive and byebye events to never be sent to these "Service Manager" apps to begin with.  But I probably can't get those done as quickly as this fix, so no reason not to proceed with this in my opinion.

juano2310:
:+1:

PR #513, opened 2016-02-16
Removed canInstallLabs()

juano2310:
I just removed canInstallLabs() to avoid error generated by platform @workingmonk @Yaima @munds

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -35,23 +35,11 @@ preferences {
 }
 
 def mainPage() {
-	if(canInstallLabs()) {
-		def bridges = bridgesDiscovered()
-		if (state.username && bridges) {
-			return bulbDiscovery()
-		} else {
-			return bridgeDiscovery()
-		}
+	def bridges = bridgesDiscovered()
+	if (state.username && bridges) {
+		return bulbDiscovery()
 	} else {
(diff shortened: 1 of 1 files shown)


Yaima:
:thumbsup:

joshualyon:
Have you considered filtering out the `VIRTUAL` hubs from the `getRealHubFirmwareVersions`? From what I could tell, it looks like that's where the issue is as the virtual hubs return the firmware version as `0`.

Here's an example of how I implemented it in my Kodi SmartApp/DTH:
https://github.com/iamcanadian2222/ST-Kodi/commit/e724fa404460e1b28ee1738aaf736e119a971a5f

juano2310:
Hi @iamcanadian2222, to be honest version "000.011.00603" is super old, there aren't any hubs with that version anymore. That was right at the beginning, before CES 2014, but it didn't hurt so we never removed it but I think is about time :)
There is a fix coming in the platform if someone needs to use canInstallLabs() but Hue doesn't need it :)

joshualyon:
I had been wondering the same thing... if the issue was really around firmware and SmartThings hubs are cloud connected, then shouldn't almost all of the hubs be updated! :smile: 

I'm curious what the platform change is with `canInstallLabs()` since it basically just gets the firmware version for all the hubs and checks to see if they are over a certain version.

I was surprised that `location.hubs` is returning virtual hubs... is the fix somehow related to changing that property/implementation?

juano2310:
As far as I know one of the functions called will return null if it is a virtual hub :) but I can't imagine anyone needs this at all unless there is something release in the future that requires the last firmware to work.

PR #514, opened 2016-02-16
PROB-870 - Harmony fails to save credentials

juano2310:
It seams like the user removed some activities on Harmony side without removing them from SmartThings. This is causing an issue when adding new activities. This fix checks if the activity still exists before creating a new device.

@Yaima


Yaima:
:thumbsup:

PR #518, opened 2016-02-17
INSIDE-787: Improve Wemo (Connect) event subscriptions

lbredeso:
This is part 1 of changes that take advantage of new, more efficient event subscriptions for SSDP search responses, in addition to new HubAction callbacks. This allows us to avoid the current inefficient location null subscriptions.

Part 2, which will come in a later PR, will involve simply removing the methods I marked as deprecated.

/cc @juano2310 @bflorian

### DIFF
--- a/smartapps/smartthings/wemo-connect.src/wemo-connect.groovy
+++ b/smartapps/smartthings/wemo-connect.src/wemo-connect.groovy
@@ -16,14 +16,14 @@
  *  Date: 2013-09-06
  */
 definition(
-    name: "Wemo (Connect)",
-    namespace: "smartthings",
-    author: "SmartThings",
-    description: "Allows you to integrate your WeMo Switch and Wemo Motion sensor with SmartThings.",
-    category: "SmartThings Labs",
-    iconUrl: "https://s3.amazonaws.com/smartapp-icons/Partner/wemo.png",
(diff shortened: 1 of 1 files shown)


juano2310:
That's awesome! Thanks for sharing!

lbredeso:
@juano2310 @darksun any feedback or objections to me merging this?  Would like to get some testing on this, so it'd be good to merge.

darksun:
:+1: no objections from me

juano2310:
Not at all :+1: looking forward to test it myself too :)

lbredeso:
Awesome! Please test and let me know of any bugs you may find. I wasn't able to test with the in-wall light switch as I don't have one, but was able to test the motion and switch plugs at least.

juano2310:
@lbredeso I will, can this be tested in prod or it requieres other things that aren't in that environment yet?

lbredeso:
@juano2310 everything is already in prod, so go for it, I already verified it there. :-)

juano2310:
awesome! Thanks!

juano2310:
@lbredeso one thing to keep in mind when this goes out is to update users account so it subscribes to ssdpSubscribe() since this only happens on installed() or updated()

PR #524, opened 2016-02-18
MSA-884: This is a quick update to a previously published device hand…

dkirker:
…ler to accept the "update" packet that occurs when the zigbee pairs with the hub to stop an exception due to it not being in the map format that other packets are in.


workingmonk:
:+1:

PR #525, opened 2016-02-18
Better exception handling for Ecobee

Yaima:
Refreshing only if status code 14 - Authentication token has expired.
Refresh your tokens.


Yaima:
@juano2310

juano2310:
:+1:

PR #532, opened 2016-02-19
DEVC-246: DEVC-246 "Enerwave motion doesn't always get the associationSet that the hub sends on join"

SmartThings:
Submitting 'DEVC-246 "Enerwave motion doesn't always get the associationSet that the hub sends on join"' for publication


workingmonk:
@mckeed

dkirker:
:+1: @mckeed 's commit

dkirker:
@workingmonk Anything more needed?

workingmonk:
If @mckeed is good with these changes, :+1:

PR #536, opened 2016-02-22
MSA-895: Nodon : Octan Remote device handler

SmartThings:
Submitting 'Nodon : Octan Remote device handler' for publication


tylerlange:
@mckeed can you take a look at this when you get a chance? They're sending us samples soon.

PR #539, opened 2016-02-22
Fix First Alert smoke alarm check-in handling

mckeed:
DEVFRWK-78
The Z-Wave Smoke Alarm handler hasn't been creating an event on the custom hourly check-in that the First Alert ZSMOKE and ZCOMBO send, so they're not showing the last activity time that they should be.

Also adds a battery read on wake-up in case we're not getting a response from the one sent after check-in, and sets all battery reports to isStateChange:true for peace of mind.


joneilloine:
Yay!

tylerlange:
@dkirker can you review today and merge? support isn't happy with this bug.

dkirker:
@mckeed :+1:

PR #542, opened 2016-02-22
DVCSMP-1505 LIFX setColor() throws exception

larsfinander:
-Added empty return to command methods since sendEvent is used
-Added null check for hex values

This is causing exceptions for anyone using Smart Lights automations with set color + LIFX bulbs. I causes around 10k exceptions a day and breaks functionality


larsfinander:
@bflorian Can you code review this?

bflorian:
:+1:

PR #548, opened 2016-02-24
DVCSMP-1517 Wemo (Connect) SmartApp throws NPE

larsfinander:
This is one of the biggest offenders in number of exceptions per day (around 100k), this fix addresses the nullpointer exception, however, we still dont know why or when the child devices can become null to begin with. That requires more investigation.


larsfinander:
@juano2310 can you take a look at this?

juano2310:
:+1:

PR #551, opened 2016-02-25
callbackUrl naming conflict

juano2310:
With the changes we introduced in the new authentication flow we added the function 
def getCallbackUrl() { "https://graph.api.smartthings.com/oauth/callback" }
By Groovy magic you can invoke this function by calling callbackUrl as a variable.
Later we define a variable with the same name in the function that it is currently not working. 
def addSubscription() {
    def data = request.JSON
    def attribute = data.attributeName
    def callbackUrl = data.callbackUrl
generating a conflict on naming. This commit renames getCallbackUrl() and the only place that it is called from. 

@Yaima @workingmonk


Yaima:
:+1:

PR #553, opened 2016-02-25
DVCSMP-1521 NPE in Bose (Connect)

larsfinander:
-NPE would occur if parsed LAN message is missing content-type header


larsfinander:
@st-andersson Please code review!

mrworf:
borat-thumbs-up

PR #555, opened 2016-02-26
[DEVTOOLS-506] Fixing lock fingerprint causing Fingerprint Tool issue

workingmonk:
@tpmanley 
The logic built into the fingerprint tool is inserting the new fingerprint at the line after the instance of 'fingerprint profile'. In case of the this file, it causes the new fingerprint to be inserted in the same as the manufacturer info of the last fingerprint.


tpmanley:
:+1:

PR #560, opened 2016-02-26
PROB-528 - Commented singleInstance: true

juano2310:
This will allow any one with more than one Hue Bridge to install an instance for each Bridge.
@Yaima


juano2310:
Duplicated https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/561

PR #561, opened 2016-02-26
PROB-528 - Commented singleInstance: true

juano2310:
This will allow any one with more than one Hue Bridge to install an instance for each Bridge.
@yaima


Yaima:
:+1:

PR #567, opened 2016-02-29
DVCSMP-1531 switch hue command methods to void

bflorian:
Eliminates exception thrown because platform was trying to turn the return value into a hub command


juano2310:
Interesting I will keep it in mind :+1:

PR #572, opened 2016-03-01
[DVCSMP-1255] Fixing issue with weird battery values

workingmonk:


workingmonk:
cc @tpmanley 
fixing issue right now for battery values 0, 255 and voltage value over 3.5

tpmanley:
:+1:

PR #573, opened 2016-03-01
[DVCSMP-1463] Deprecating copied DTH

workingmonk:
removing the fingerprints from the additional copy of DTH

cc @tpmanley


tpmanley:
:+1:

PR #576, opened 2016-03-02
device.refresh change for tile

workingmonk:
cc @tpmanley @tylerlange @jimmyjames 
Tagging all three of you because I noticed the discrepancy in the doc vs what we code


tpmanley:
:+1:

jimmyjames:
@workingmonk these changes make sense, are you saying that we document to use `"device.switch"` no matter the attribute?

workingmonk:
@jimmyjames no i am saying the document seems like with "device.refresh", but for some reason all the DTH files (including ones i wrote) have "device.(main attribute)" for refresh

PR #577, opened 2016-03-02
Including unknown temperature values as part of the response for sensors

Yaima:
https://smartthings.atlassian.net/browse/DVCSMP-1511
@juano2310 @dsainteclaire


juano2310:
:+1:

PR #587, opened 2016-03-06
INTL-292 i18n alignment for smartsense-moisture-sensor

twack:


Bijnagte:
:+1:

workingmonk:
value will be null on line 268

PR #589, opened 2016-03-06
INTL-290 i18n alignment for smartsense-multi-sensor

twack:


Bijnagte:
:+1:

PR #591, opened 2016-03-06
INTL-387 i18n alignment for mobile-presense

twack:


Bijnagte:
:+1:

PR #593, opened 2016-03-06
INTL-353 create i18n korean translation for smartpower-outlet

twack:
Created i18n translation for smartpower-outlet (110-Volt version)


Bijnagte:
:+1:

PR #597, opened 2016-03-08
[DVCSMP-1442] Cree bulb does not properly reflect dimming value

workingmonk:
cc @tpmanley @tylerlange


workingmonk:
updated zigbee api to getEvent. demoting log.warn as it is now an expensive operation.

workingmonk:
also updated the spacing and copyright year. use ?w=1 in url to see non-space changes

workingmonk:
cc @Yaima

Yaima:
:+1:

PR #624, opened 2016-03-10
Fixed Ecobee polling

Yaima:
https://smartthings.atlassian.net/browse/DVCSMP-1511
Ecobee was polling once per device, creating multiple API calls. Now it
is polling once at the smartapp level every 5 minutes and sending the
response to each child device.

### DIFF
--- a/devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
+++ b/devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
@@ -20,7 +20,6 @@ metadata {
 		capability "Temperature Measurement"
 		capability "Motion Sensor"
 		capability "Refresh"
-		capability "Polling"
 	}
 
 	tiles {
@@ -68,6 +67,6 @@ def refresh() {
 
 void poll() {
 	log.debug "Executing 'poll' using parent SmartApp"
-	parent.pollChild(this)
+	parent.pollChild()
 
 }
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -463,29 +463,33 @@ def pollChildren(child = null) {
 }
 
 // Poll Child is invoked from the Child Device itself as part of the Poll Capability
-def pollChild(child){
+def pollChild(){
 
-	if (pollChildren(child)){
-		if (!child.device.deviceNetworkId.startsWith("ecobee_sensor")){
-			if(atomicState.thermostats[child.device.deviceNetworkId] != null) {
-				def tData = atomicState.thermostats[child.device.deviceNetworkId]
-				log.info "pollChild(child)>> data for ${child.device.deviceNetworkId} : ${tData.data}"
-				child.generateEvent(tData.data) //parse received message from parent
-			} else if(atomicState.thermostats[child.device.deviceNetworkId] == null) {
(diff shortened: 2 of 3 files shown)


Yaima:
@juano2310, @dsainteclaire  Can you take a look at this PR?

StrykerSKS:
Why is Polling removed as a Capability? (Perhaps I misunderstand how that capability works.)

Also, is there any reason to keep both pollChild() and pollChildren()?

StrykerSKS:
Oh, and one other comment. It can still be very useful to pass in the child instance (this) when the poll is initiated from the device (from a refresh for example) from a debugging/trace perspective. Without that child reference it is hard to create log entries back through the different threads.

Yaima:
Polling is not working properly right now and for cloud devices if we poll at the device level we will be duplicating the calls to the api. I am still generating events at the service manager level that create logs entries per child.

devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
```diff
@@ -68,6 +67,6 @@ def refresh() {
 
 void poll() {
 	log.debug "Executing 'poll' using parent SmartApp"
-	parent.pollChild(this)
+	parent.pollChild()
```
dsainteclaire, line 70: Just because I'm new.  Why aren't you passing the device type to poll anymore?  What kind of effect does that have?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -463,29 +463,33 @@ def pollChildren(child = null) {
...
-				log.error "ERROR: Device connection removed? no data for ${child.device.deviceNetworkId}"
-				return null
+def pollChild(){
+
+	def devices = getChildDevices()
+
+	if (pollChildren()){
+		devices.each { child ->
+			log.info "***found $child"
+			if (!child.device.deviceNetworkId.startsWith("ecobee_sensor")){
```
dsainteclaire, line 473: So, as I understand.  You just iterate over all the devices looking for ecobees, instead of accepting a child device and generating the event directly?

dsainteclaire:
Just a few questions, so I can understand the nature of the solution.  Otherwise, from a new guy's perspective, LGTM

devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
StrykerSKS, line 70: @dsainteclaire This doesn't impact the polling at all. Since this is a cloud integration, the API calls can return the data for all the devices in the same API call. So there is no need to pass the information directly related to which child made the call since we will get the data back for all children at the same time.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
StrykerSKS, line 473: This is the same premise as above, no need to only poll for a single child device, instead the API will return information on all of the child devices. Which means that it is more efficient to update the information for all of the children at the same time as well.

The nice side effect of this is if you were to manually use "refresh" from the Things page that it will not only refresh the data for that sensor/thermostat but will also update the data for all of the other sensors/thermostats at the same time! This also means a lot fewer API calls are needed.

StrykerSKS:
Would be happy to continue the discussions over at the developers forum if you have more questions.

Yaima:
Thanks @StrykerSKS! I talked to @dsainteclaire today. He seats by me :dancer:

PR #625, opened 2016-03-10
EX-45: Wemo (Connect) device addition fails on update for old devices

lbredeso:
Details here:  https://smartthings.atlassian.net/browse/EX-45

Basically, the method "dni" doesn't exist, and returns nulls. Usually not a problem, but for really old devices where there is no "mac" in device data, the app fails to update.

/cc @juano2310


juano2310:
:+1:

PR #628, opened 2016-03-11
Rolling up master to staging

workingmonk:


workingmonk:
@mitchpond fixed that issue while merging in from staging

PR #631, opened 2016-03-11
[DVCSMP-1107] Prepping up v1 multi for garage door functionality

workingmonk:
This is the stage 1 phase. This will enable the 'status' attribute. Once this goes out in the cloud and AppEngine end, we can open the PR for phase two to enable the 'status' tile and garage preference.

cc @tpmanley @lbredeso

### DIFF
--- a/devicetypes/smartthings/smartsense-multi.src/smartsense-multi.groovy
+++ b/devicetypes/smartthings/smartsense-multi.src/smartsense-multi.groovy
@@ -22,6 +22,8 @@ metadata {
 		capability "Battery"
 
 		fingerprint profileId: "FC01", deviceId: "0139"
+
+		attribute "status", "string"
 	}
 
 	simulator {
@@ -102,7 +104,7 @@ def parse(String description) {
 
 }
 
-private Map parseSingleMessage(description) {
+private List parseSingleMessage(description) {
 
 	def name = parseName(description)
 	def value = parseValue(description)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-multi.src/smartsense-multi.groovy
```diff
@@ -120,8 +123,18 @@ private Map parseSingleMessage(description) {
...
+	)
+
+	results << createEvent(
+		name: "status",
+		value: value,
+		unit: null,
+		linkText: linkText,
+		descriptionText: descriptionText,
+		handlerName: handlerName,
+		isStateChange: isStateChange,
```
tpmanley, line 135: this will be for `name` which I assume is `contact`. is that what you intended?

workingmonk, line 135: it is creates both the contact and status event

tpmanley, line 135: I'm specifically talking about `isStateChange`. It's basically assuming that a state change for contact is a state change for status, which I imagine is normally true, but is it always true?

workingmonk, line 135: it is always true for phase 1, where status is in sync with the contact attribute. for phase 2, they will be independent events in garage door mode.

tpmanley:
:+1:

workingmonk:
thanks @tpmanley

PR #633, opened 2016-03-14
Fix text-to-speech feature

tslagle13:
Refactor case statement so custom message works. Remove comments from input selection.

cc @bflorian @workingmonk @Yaima


workingmonk:
the default value was set to Custom before @Yaima made the change in december. do you want to change that too?

tslagle13:
The "bell" sound was wrapped into the default case so it would evaluate the if/else and then the last line of the default case was to set the set the sound to "bell". Splitting custom message into a separate case and making the "bell" sound default allowed the custom message to be successfully set.

workingmonk:
:+1:

PR #641, opened 2016-03-15
DVCSMP-1615 & DEVC-372

juano2310:
DVCSMP-1615  - Fix exception if Hue, Saturation or Hex is sent to setColor
DEVC-372 - Improves readability of Activity Feed

### DIFF
--- a/devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
+++ b/devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
@@ -3,6 +3,7 @@
  *
  *  Author: SmartThings
  */
+
 // for the UI
 metadata {
 	// Automatically generated. Make future change here.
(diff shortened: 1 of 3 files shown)


juano2310:
cc @larsfinander @munds

munds:
👍

larsfinander:
@juano2310 For setColor() and activity feed, is it possible to translate the color values into a human readable name, like blue, red... rather than showing nothing in stream?

juano2310:
Any recommendation? How can we do that?

larsfinander:
Something along the lines of this perhaps: https://gist.github.com/XiaoxiaoLi/8031146  
But I think we could get away with using a fewer number of colors... And ideally, this should be a "platform" function so that all our smartapps and device types could do it..

juano2310:
We have a color picker. The combination are a lot more than those. What if the user chose a color outside those?

larsfinander:
I just figured whatever color they pick would be closest to one of the basic colors.

munds:
Isn't that what we kinda had before? Like a pre-populated list of colors that we could choose from a dropdown?

juano2310:
@munds that's in apps but not in the device type. 
@larsfinander how do you determine proximity to a color? For example #a73a43

PR #645, opened 2016-03-16
INTL-292 undo non i18n changes

Bijnagte:
cc @workingmonk @tpmanley

### DIFF
--- a/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
@@ -3,15 +3,15 @@
  *  Copyright 2016 SmartThings
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
- *  use this file except in compliance with the License. You may obtain a copy 
+ *  use this file except in compliance with the License. You may obtain a copy
  *  of the License at:
  *
  *      http://www.apache.org/licenses/LICENSE-2.0
  *
- *  Unless required by applicable law or agreed to in writing, software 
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -256,8 +257,7 @@ private Map getBatteryResult(rawValue) {
 				def pct = batteryMap[volts]
 				if (pct != null) {
 					result.value = pct
-                    def value = pct
```
tpmanley, line 259: Unless I'm missing something, this line needs to be here otherwise the line below won't have access to `value`

tpmanley:
This doesn't undo the changes that were accidentally reverted. These are the changes that were lost:
0c87601c64f3cf74c346199f4f2b8e0e3cc9967e
12220da75081ad533b24e09a99803feae0582719
5ec82217acd76d4db94c9585cb72d041d640e472

tpmanley:
My mistake, I thought I was reviewing the multi sensor

tpmanley:
:+1:

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
workingmonk, line 259: not sure what the resolution here was. i can still see no `value` defined in this method. wouldn't that result in the description text always being null?

tpmanley, line 259: Sorry, @Bijnagte stopped by and explained it. The value isn't inserted here, it's inserted later using the value stored in the event.

PR #646, opened 2016-03-16
INTL-290 revert non i18n changes

Bijnagte:
@workingmonk @dkirker


tpmanley:
:+1:

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -341,8 +340,7 @@ private Map getBatteryResult(rawValue) {
 				def pct = batteryMap[volts]
 				if (pct != null) {
 					result.value = pct
-                    def value = pct
```
workingmonk, line 344: without `value` being defined in the method, again description text would be null

workingmonk, line 344: you could change the value field to result.value in the description text

PR #647, opened 2016-03-16
INTL-353 revert non i18n changes

Bijnagte:
cc @workingmonk


workingmonk:
the changes look fine here. i would have prefered not having the description text here at all, but this is great for now.
:+1:

PR #649, opened 2016-03-16
multi: Fix occasional error with threeAxis attribute reporting

tpmanley:
Resolves:
     https://smartthings.atlassian.net/browse/DVCSMP-1623


workingmonk:
:+1:

PR #654, opened 2016-03-20
Fix the handling for setFanMode

StrykerSKS:
Included extra parameters (isTemperatureRelative: "false", isTemperatureAbsolute: "false") so that the temperature values are ignored and only the fan mode is changed. 

As par the discussion at the Ecobee developer site:
http://developer.ecobee.com/api/topics/qureies-related-to-setfan


tylerlange:
@dsainteclaire @Yaima

Yaima:
:+1:  I am good merging this change. I already talked to @StrykerSKS about it a couple of weeks ago.

PR #660, opened 2016-03-21
Added hubVerification()

juano2310:
If the hub exist as a thing parsing the response from description.xml
was lost. Now it is sent back to the parent were the device can be
marked as verified if the modelName starts with "Philips hue bridge"


larsfinander:
:+1:

PR #661, opened 2016-03-21
Turns off is value switch = off

juano2310:
After turning color the bulb turns off is value switch = off. This is used by Smart Home Monitor to restore bulb to the values before Notification with light was activated.

cc @larsfinander


juano2310:
@larsfinander do you know if this change was added to hue at any giving point? Can I close this one?

larsfinander:
@juano2310 that functionality should already be in there

PR #665, opened 2016-03-22
Fixing wrong cases of the WeMo brand

munds:
@juano2310 Fixing case issues for DVCSMP-439.

Didn't change function names, because that is not displayed.


juano2310:
:+1:

PR #674, opened 2016-03-23
[DPROT-20] ZigBee Refactor

workingmonk:
cc @tpmanley

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -51,22 +51,15 @@ metadata {
 // Parse incoming device messages to generate events
 def parse(String description) {
     log.debug "description is $description"
-
-    def resultMap = zigbee.getKnownDescription(description)
-    if (resultMap) {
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -51,22 +51,15 @@ metadata {
...
-        log.info resultMap
-        if (resultMap.type == "update") {
-            log.info "$device updates: ${resultMap.value}"
-        }
-        else if (resultMap.type == "power") {
+    def event = zigbee.getEvent(description)
+    if (event) {
+        if (event.name == "power") {
             def powerValue
-            if (device.getDataValue("manufacturer") != "OSRAM") {       //OSRAM devices do not reliably update power
```
tpmanley, line 63: Don't we still need this check of manufacturer?

workingmonk, line 63: Not really. We have moved all the OSRAM products to non-power DTH

tpmanley:
Looks good then :+1:

PR #687, opened 2016-03-23
MSA-993: 데스트

SmartThings:
Submitting '데스트' for publication


workingmonk:
file names needs to be in english

workingmonk:
otherwise the system interprets it as null causing the weird directory structure

PR #691, opened 2016-03-23
[DVCSMP-1657] Bug/fix multi

workingmonk:
cc @tpmanley


devicetypes/smartthings/smartsense-multi.src/smartsense-multi.groovy
```diff
@@ -203,7 +203,8 @@ private List parseContactMessage(String description) {
 	parts.each { part ->
 		part = part.trim()
 		if (part.startsWith('contactState:')) {
-			results << getContactResult(part, description)
+			results << getContactResult(part, description)[0]
```
tpmanley, line 206: Rather than calling getContactResult twice, how about just doing `results.addAll(getContactResult(part, description))`?

PR #692, opened 2016-03-23
[DVCSMP-1657] Bug/hotfix multi

workingmonk:
cc @tpmanley


jonsie:
That should fix it, although it would be nice to not have to spend time reasoning about why.  You could change the return type of `getContactResult()` to `List<Map>`, and then add `Map` return type to `getAccelerationResult`, `getTempResult`, `getXyzResult`, `getBatteryResult`, `getRssiResult`, and `getLqiResult`.  

Either way, :+1:

workingmonk:
@jonsie added return types

tpmanley:
:+1:

workingmonk:
thanks @tpmanley and @jonsie

PR #696, opened 2016-03-24
Revert "Revert "DVCSMP-1615 & DEVC-372""

workingmonk:
Reverts SmartThingsCommunity/SmartThingsPublic#668

This is basically undoing the temporary fix that went in


workingmonk:
cc @larsfinander

PR #701, opened 2016-03-25
Initial commit of Blue Iris integration

pursual:


pursual:
Does anybody even look at Pull requests?

jasonbio:
@pursual we do :) and we're working on a better way to get these QA'd and approved in a timely manner

PR #710, opened 2016-03-28
MSA-1006: Aeon Home Energy Meter

SmartThings:
Submitting 'Aeon Home Energy Meter' for publication


tylerlange:
@mckeed can you please review?

mckeed:
:-1: We can't add these custom attributes to all our device types just to make mfr info available – I'm planning to make this available a different way.

tylerlange:
@mckeed this is to support Astralink. Is it possible for us to add these to this DTH and perhaps one other? If not, please let me know and we'll communicate.

tylerlange:
cc @st-andersson

PR #711, opened 2016-03-28
Restrict colorTemperature to valid values

KristopherKubicki:
Removed the hacky 0-100 method and added  a restriction on color temperature.  If the bulb gets a color temperature outside of its nominal range, the bulb will turn off.


workingmonk:
this would work in this case, but the osram bulbs have moved to [zigbee-white-color-temperature-bulb](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-white-color-temperature-bulb.src/zigbee-white-color-temperature-bulb.groovy). 

what was the issue that you saw? i initially put in the hack for sliders that could only send values from 0 to 100.

workingmonk:
trying to understand the deeper issue here. that was the 'lower' value that you sent that caused the bulb to turn off

KristopherKubicki:
Hi Vinay - Yes - sending the bulb a CT < 2700 caused it to turn off.

PR #714, opened 2016-03-29
Update zigbee-dimmer.groovy

Sumit-B-Singh:
Add support for LIGHTIFY CLASSIC A 60 WHITE


workingmonk:
hey sumit,
is this the uk model? the profile id shows that it is a ZLL bulb

Sumit-B-Singh:
Yes, this is for the EU Lightify Classic A 60 White: http://www.osram.com/osram_com/products/led-technology/lightify/lightify-home/lightify-lamps/lightify-classic-a-60-white/index.jsp

workingmonk:
@Sumit-B-Singh sorry for the long delay. We've added support for the osram lights under the zll-_.groovy dth for EU.
I took the fingerprint from your PR, but immediately had a problem with the model name again. I've added support for both the models. Can you review the zll-_.groovy dth and tell us if there are both combination of model names supported by osram? the zll code will be in prod on/around 17th may

PR #715, opened 2016-03-29
DVCSMP-1672 Phillips HUE: Mismatched on button  color for hue lux

larsfinander:
… lux

-Changed color from green to blue to match other Hue lights

All devicetypes should be migrated to blue color according to Samsung


larsfinander:
@munds Please review!

munds:
You have my approval good sir. Let the new colors flash!

PR #716, opened 2016-03-29
DVCSMP-1597 Bloom and Strip lights need temperature control removed

larsfinander:
-Added new device type for Hue lights that have color control but no temperature control (Bloom/Strip V1)
-Add missing event to setLevel

### DIFF
--- a/devicetypes/smartthings/hue-bloom.src/hue-bloom.groovy
+++ b/devicetypes/smartthings/hue-bloom.src/hue-bloom.groovy
@@ -1,7 +1,7 @@
 /**
- *  Hue Bulb
+ *  Hue Bloom
  *
- *  Philips Hue Type "Extended Color Light"
+ *  Philips Hue Type "Color Light"
  *
  *  Author: SmartThings
  */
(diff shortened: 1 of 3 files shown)


larsfinander:
@munds @juano2310 Please review, there is a slight chance that this will change tomorrow after Hue meeting in case their device classes will not match what I have here.

juano2310:
@larsfinander I saw that you added modelid: v.modelid. Any change there has to be replicated on line 868. What happens is that if the bridge doesn't exist the service manager catch it and parse it but if the bridge exists the information is sent back to the Service manager and processed there convertBulbListToMap(). That's why we weren't discovering the bridge if the bridge already existed. The parse method wasn't sending it back to the parent, that's why I created hubVerification() which basically replicates the verification in the parse method in the Service Manager.

larsfinander:
Thanks, will update!

juano2310:
@lbredeso I think Hue would benefit a lot from the new subscription infrastructure. Hue is the most popular LAN device.

lbredeso:
@juano2310 Yes, you are correct. Hue is especially bad because the Hue Bridge responds to every M-SEARCH request, even those with search targets for different devices.

I've already completed the work for Hue (Connect), but need a platform change to make it into production before I submit a PR for it.  I'll cc you in that PR.

juano2310:
Awesome!!! Looking forward for that!

larsfinander:
@juano2310 made your suggested change

larsfinander:
@lbredeso I am interested  in seeing that PR too

lbredeso:
@larsfinander Sure, I'll loop you in when I submit it.

juano2310:
:+1:

PR #720, opened 2016-03-30
DVCSMP-1667 Authentication needs to be updated for Philips Hue

larsfinander:
-Philips updaded their API and current ST implementation will stop
working when next Hue firmware is released without this change


workingmonk:
ok to merge to staging

PR #733, opened 2016-03-31
INTL-309 Translating Tiles

sabersd:


sabersd:
wrong environment,I'm the worst

PR #669, opened 2016-03-22
MSA-975: Iris Key Fob

SmartThings:
Submitting 'Iris Key Fob' for publication

PR #743, opened 2016-03-31
Iris Smart Fob

mitchpond:
Resubmission of #669


workingmonk:
@mitchpond thanks for this PR. i have changed a lot of how thing works, but wanted to make sure that your work is appreciated and acknowledged. one of the other changes is the changes to the button capability to communicate with the smartapp how many buttons the `button device` has. in development and testing it right now. i'll have a new pr to enable the fingerprint and that change when it is ready. thanks!

PR #761, opened 2016-04-05
EX-3: Improve Hue (Connect) event subscriptions

lbredeso:
This work enables the use of the new ssdpTerm event subscriptions for Hue (Connect).

Please note that this code uses a `HubAction` constructor that doesn't exist yet in production, so it'll be a couple days until this can be verified there; I just wanted to get this out for review sooner than later.

/cc @larsfinander @juano2310

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -60,7 +60,7 @@ def bridgeDiscovery(params=[:])
         app.updateSetting("selectedHue", "")
     }
 
-	subscribe(location, null, locationHandler, [filterEvents:false])
+	ssdpSubscribe()
 
 	//bridge discovery request every 15 //25 seconds
 	if((bridgeRefreshCount % 5) == 0) {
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -516,11 +630,7 @@ def doDeviceSync(){
 	log.trace "Doing Hue Device Sync!"
 	convertBulbListToMap()
 	poll()
-    try {
-		subscribe(location, null, locationHandler, [filterEvents:false])
-    } catch (all) {
```
lbredeso, line 521: I'm assuming this try/catch is here because the platform used to have a problem with duplicate subscriptions at some time in the past. As of right now, we have handled this on the platform side, so the try/catch should be unnecessary.

juano2310, line 521: Not sure I did that but I see no reason for that unless it was added in response to @larsfinander investigating exceptions for Hue.

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -747,15 +857,11 @@ private getId(childDevice) {
...
 	def host = getBridgeIP()
 	def uri = "/api/${state.username}/lights/"
-    try {
-		sendHubCommand(new physicalgraph.device.HubAction("""GET ${uri} HTTP/1.1
+	log.debug "GET: $host$uri"
+	sendHubCommand(new physicalgraph.device.HubAction("""GET ${uri} HTTP/1.1
 HOST: ${host}
 
 """, physicalgraph.device.Protocol.LAN, selectedHue))
-	} catch (all) {
```
lbredeso, line 755: I followed the code path for sendHubCommand and can't see how an exception could be thrown here. I'm assuming this was in an attempt to be thorough (and the log message is copy/pasted and doesn't apply to this situation), but it looks to be unreachable. I also don't see this being done elsewhere for sendHubCommand, so feel ok removing it.

juano2310, line 755: :+1:

PR #772, opened 2016-04-12
Update tile-multiattribute-thermostat.groovy

KyleLeneau:


workingmonk:
need this against production branch.

PR #775, opened 2016-04-13
Added fingerprint for NYCE door hinge sensor

motley74:
To support the NYCE Door Hinge Sensor (model 3010)

### DIFF
--- a/devicetypes/smartthings/nyce-open-closed-sensor.src/nyce-open-closed-sensor.groovy
+++ b/devicetypes/smartthings/nyce-open-closed-sensor.src/nyce-open-closed-sensor.groovy
@@ -24,6 +24,7 @@ metadata {
         command "enrollResponse"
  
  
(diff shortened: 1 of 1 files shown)


tslagle13:
Cc @workingmonk

workingmonk:
question is for @tylerlange on the certification load for this. do we just add the support so that the device works for now?

tylerlange:
No, we wouldn't typically add a fingerprint for a device that we're not planning on certifying. I understand the impact here is pretty minimal though. What's your opinion @workingmonk

workingmonk:
yes, i agree. this way, if someone tries to add the devices, it will at least work rather than show up as Thing.

tylerlange:
@workingmonk if you can verify the fingerprint is correct please proceed

workingmonk:
👍

motley74:
Thanks to all for working on this.

PR #776, opened 2016-04-13
MSA-1178: OSRAM Lightify Dimming Switch

motley74:
Submitting 'OSRAM Lightify Dimming Switch' for publication


workingmonk:
See PR #850 and #851

mjkoster:
I think this would be more useful if the button release event could be subscribed to. Is there some gating reason this code is commented out?

motley74:
The code for button release is commented out because I am not sure what you would subscribe the button release event to.  As far as I know there is not a good way to get dimming while held to work so not sure what the release event would get you.  I could uncomment the code in the device type so that the event can be subscribed to via a smart app but my intention was to uncomment that once it can support actual dimming functions.

PR #806, opened 2016-04-19
custom device types for the CoopBoss hardware version 3

dkirker:
Prepping for submission to SmartThings.  Removed extra comments.

Submitting to public store for approval

Updated for new SmartThings interface

Release Candidate for CoopBoss H3V9

save

Update to fix invalid temperature reporting and added maxCurrentNE methods and attributes.

Made changes to defalutState:true to state "default" on smart tiles that had only one default state.  This was causing the Android version of the app to crash.

Updated comments

Added test for null values that were causing errors during join process

Fixed log errors during join process by testing for null values before init process.

Update fingerprint with Manufacturer data

SmartApps for CoopBoss hardware version 3

Updated Icons.

1

Updates to Prep for RC

Clean up

save


workingmonk:
permission to merge to staging 👍

PR #807, opened 2016-04-19
MSA-1202: Revision to ST z-Wave door and window sensor to work with Garage door and have garage door icons

joelw135:
Submitting 'Revision to ST z-Wave door and window sensor to work with Garage door and have garage door icons' for publication


tylerlange:
@mckeed is this publishable? i think the fingerprint conflicts

PR #808, opened 2016-04-19
DVCSMP-1716 Philips Hue: Invalid hue parameter used in setColor

larsfinander:
-Replaced transition with transitiontime according to Hue API


larsfinander:
@juano2310 Please review!

juano2310:
:+1:

PR #813, opened 2016-04-20
SHARD-159: Wemo (Connect) updated() fails in certain IP change scenarios

lbredeso:
There is an uncommon situation that can occur where a device is selected in Wemo (Connect) settings in the past that later does not appear in state. I'm still unsure how the app state gets into this situation for Wemo (Connect), but I do know that that this change will keep updated() calls from failing. This is directly related to SHARD-157.

/cc @bflorian @charliek @kyleboon


charliek:
+1

kyleboon:
Could this be from an inconsistent read?

either way 👍

lbredeso:
@kyleboon hmmm, that's actually plausible!

PR #821, opened 2016-04-22
MSA-1219: 이현정

hyunjoung:
Submitting '이현정' for publication


workingmonk:
name and namespace needs to be in english

PR #824, opened 2016-04-22
MSA-1223: 이은하앱

Leeeunha:
Submitting '이은하앱' for publication


workingmonk:
name and namespace needs to be in english

PR #833, opened 2016-04-25
Rolling down hotfix for deploy tool to staging

workingmonk:


workingmonk:
cc @jord98716 @jeff-blaisdell

PR #839, opened 2016-04-27
Update zigbee-white-color-temperature-bulb.groovy

KristopherKubicki:
Add events to commands, in case the parse() method does not get the response


workingmonk:
This is not the right approach. Only parse should generate event messages. If the action methods generate the events, there are two issues:
1. There is false confirmation of action
2. The health framework is fooled into thinking that the device is healthy

PR #841, opened 2016-04-28
SHARD-174: Hue (Connect) updated() calls can fail in certain situations

lbredeso:
If a Hue Bulb has previously been discovered and persisted in the app state of Hue (Connect), and it doesn't have "type" information in state, and doesn't get updated due to the hub being offline (or other situations), the update will fail with a NullPointerException.


larsfinander:
👍

PR #850, opened 2016-05-03
DVCSMP-1741 Changes to button controller to read number of buttons from the dth

workingmonk:
cc @tylerlange 
This should be merged in as part of the device certification work for the new zigbee DTH

### DIFF
--- a/smartapps/smartthings/button-controller.src/button-controller.groovy
+++ b/smartapps/smartthings/button-controller.src/button-controller.groovy
@@ -27,10 +27,9 @@ definition(
 
 preferences {
 	page(name: "selectButton")
-	page(name: "configureButton1")
-	page(name: "configureButton2")
-	page(name: "configureButton3")
-	page(name: "configureButton4")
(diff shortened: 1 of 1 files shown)


workingmonk:
Adding a do not merge as this is pending certification

motley74:
I understand that #851 will add update an existing device type to work with the OSRAM Lightify dimming switch but I am curious about the SmartApp side.  The SmartApp I wrote is more geared towards using the switch as a dimmer switch as originally intended and as far as I have seen the button controller app does not provide the ability to set a light level percentage for the held events (or pushed events for that matter).  The SmartApp I wrote provides this ability and I plan to improve it later to support actual dimming/brightening as long as long as the buttons are held once the platform better supports that ability.

workingmonk:
@tylerlange you will need this code to test button devices with non standard (4) number of buttons

workingmonk:
cc @marstorp

workingmonk:
Verified and tested by Steven.

PR #857, opened 2016-05-05
fix parse signature for generic multiattribute example tile

jimmyjames:
@KyleLeneau this is needed for the generic tile to install through the controller


pennig:
👍

workingmonk:
add ticket number to the PR title

PR #859, opened 2016-05-06
Update ready-for-rain.groovy

imbrianj:
Added ability to set delay before attempting to send message, will cancel alert if contacts closed within delay.

@motley74

### DIFF
--- a/devicetypes/imbrianj/foscam-local.groovy
+++ b/devicetypes/imbrianj/foscam-local.groovy
@@ -1,0 +1,287 @@
+/**
+ *  Foscam
+ *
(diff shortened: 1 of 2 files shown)


imbrianj:
Tried to get the feature set of the local foscam controller closer to parity with the standard one. Areas still needing some help:

Only tested on 8910W - stripped out most of the logic to try and detect device version when I was trying to narrow down some issues.
No polling - so no way of knowing if alarm is on or off - also cannot currently toggle alarm status.
Removed the setting of presets.
Still not perfect, but it's good enough for me that I'm switching to it and closing my firewall.

PR #867, opened 2016-05-09
DVCSMP-1699 Na02>Missing/Failed Oauth Tokens - Netatmo

rohandesai:
- refactored the oauth component for the app to work in different shards, if supported
- fixed some white spacing issues and indents

@llinder @jasonbio

### DIFF
--- a/smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
+++ b/smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
@@ -4,29 +4,33 @@
 import java.text.DecimalFormat
 import groovy.json.JsonSlurper
 
-private apiUrl() 			{ "https://api.netatmo.com" }
-private getVendorName() 	{ "netatmo" }
-private getVendorAuthPath()	{ "https://api.netatmo.com/oauth2/authorize?" }
-private getVendorTokenPath(){ "https://api.netatmo.com/oauth2/token" }
+private getApiUrl()			{ "https://api.netatmo.com" }
+private getVendorName()		{ "netatmo" }
+private getVendorAuthPath()	{ "${apiUrl}/oauth2/authorize?" }
+private getVendorTokenPath(){ "${apiUrl}/oauth2/token" }
 private getVendorIcon()		{ "https://s3.amazonaws.com/smartapp-icons/Partner/netamo-icon-1%402x.png" }
-private getClientId() 		{ appSettings.clientId }
-private getClientSecret() 	{ appSettings.clientSecret }
-private getServerUrl() 		{ "https://graph.api.smartthings.com" }
+private getClientId()		{ appSettings.clientId }
+private getClientSecret()	{ appSettings.clientSecret }
+private getServerUrl() 		{ appSettings.serverUrl }
+private getShardUrl()		{ return getApiServerUrl() }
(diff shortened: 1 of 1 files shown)


smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
```diff
@@ -11,19 +11,22 @@ private getVendorTokenPath(){ "https://api.netatmo.com/oauth2/token" }
 private getVendorIcon()		{ "https://s3.amazonaws.com/smartapp-icons/Partner/netamo-icon-1%402x.png" }
 private getClientId() 		{ appSettings.clientId }
 private getClientSecret() 	{ appSettings.clientSecret }
-private getServerUrl() 		{ "https://graph.api.smartthings.com" }
+private getServerUrl() 		{ return "https://graph.api.smartthings.com" }
```
llinder: This doesn't appear to be used anywhere could we delete it?

smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
```diff
@@ -11,19 +11,22 @@ private getVendorTokenPath(){ "https://api.netatmo.com/oauth2/token" }
...
+private getShardUrl()		{ return getApiServerUrl() }
+private getCallbackUrl()	{ "https://graph.api.smartthings.com/oauth/callback" }
```
llinder: Maybe reuse getServerUrl here instead of duplicating the hard coded value.  Then there will only be one item to make configurable.

smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
llinder: Might be good to make this configurable so it can more easily be tested in lower environments when that becomes more common place.

smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
```diff
@@ -35,35 +38,53 @@ preferences {
...
+	if (!state.accessToken) {
+		log.debug "About to create access token."
+		state.accessToken = createAccessToken()
+	}
 
-			return dynamicPage(name: "Credentials", title: "Authorize Connection", nextPage:"listDevices", uninstall: true, install:false) {
-				section { href url:buildRedirectUrl("auth"), style:"embedded", required:false, title:"Connect to ${getVendorName()}:", description:description }
-			}
-		} else {
+	log.debug "STATE: ${state.authToken}"
```
llinder: Might be a good idea to not log this in production since it pretty sensitive security info

smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
```diff
@@ -340,7 +310,8 @@ def refreshToken() {
 
 	log.debug "State: $state"
```
llinder: Might be a good idea not to log this in production.

llinder:
The oauth callback changes look good.  I made a few other comments which might be worth considering.

rohandesai:
@llinder addressed comments. Can you please take a look?

jasonbio:
👍

PR #868, opened 2016-05-10
DVCSMP-1699 Na02>Missing/Failed Oauth Tokens - Life360

rohandesai:
- now calling the method to get the right shard url for Life360

@llinder @jasonbio

### DIFF
--- a/smartapps/smartthings/life360-connect.src/life360-connect.groovy
+++ b/smartapps/smartthings/life360-connect.src/life360-connect.groovy
@@ -27,7 +27,6 @@ definition(
 ) {
 	appSetting "clientId"
 	appSetting "clientSecret"
-    appSetting "serverUrl"
 }
 
 preferences {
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/life360-connect.src/life360-connect.groovy
```diff
@@ -192,7 +192,7 @@ def getSmartThingsClientId() {
    return "pREqugabRetre4EstetherufrePumamExucrEHuc"
 }
 
-def getServerUrl() { appSettings.serverUrl }
```
llinder, line 195: should we remove the old serverUrl from the smartapp settings?

llinder:
Seems fine as long as Life360 doesn't start validating redirect URLs in their OAuth flow.  If they ever do this will break and we will need to make further updates.

rohandesai:
@llinder addressed comment for this too.

jasonbio:
👍

PR #872, opened 2016-05-11
fix exception caused by typo

tslagle13:
added paren to end of "someoneIsHome" check on accident.

@juano2310 can you review?

cc @workingmonk


juano2310:
:+1:

workingmonk:
i hope this is not the norm. we can't just call everything priority and use the staging branch for every update. these small bugs is exactly what testing time is for.
👍 to merge to staging

PR #887, opened 2016-05-12
DVCSMP-1770 Add MSR to zwave-switch and configure at pairing

larsfinander:
-Make ZWave switch and Dimmer switch behave the same and save MSR and
manufacturer name in data


larsfinander:
@mckeed For a first Amazon Echo rollout, PM wants to play it save and just enable for Leviton and two previously Echo certified zigbee devices. If so we need MSR in ZWave switch as wel.

Here is a proposed change, kept manufacturername in both as well in case someone actually wrote code dependent on that.
Is there a way we can have refresh() called automatically on pairing?

mckeed:
👍

PR #890, opened 2016-05-12
DPROT-94 moving the cree specific generic fingerprint back to cree dth 

workingmonk:
because of cree manufacturerInfo spacing issue

cc @tpmanley


workingmonk:
closed this one and created a new one against staging

PR #893, opened 2016-05-13
DOCS-284 - update tiles for best practices, add setColor to colorWheel

jimmyjames:
Updating state and attributeState to use `defaultState: true` instead of `"default"` name per best practices.

Also added `setColor()` command to the color wheel device tile.

@KyleLeneau or @pennig one of you should review.


pennig:
👍

PR #897, opened 2016-05-13
PENG-158 UBI should not allow undefined commands

rohandesai:
- now validating commands per capability of the device in the smartapp

@jasonbio @larsfinander

### DIFF
--- a/smartapps/smartthings/ubi.src/ubi.groovy
+++ b/smartapps/smartthings/ubi.src/ubi.groovy
@@ -107,8 +107,8 @@ mappings {
 	path("/locks") {
 		action: [
 			GET: "listLocks",
-			PUT: "updateLock",
-			POST: "updateLock"
+			PUT: "updateLocks",
+			POST: "updateLocks"
 		]
 	}
 	path("/locks/:id") {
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/ubi.src/ubi.groovy
```diff
@@ -442,31 +442,90 @@ def executePhrase() {
...
 	def command = request.JSON?.command
-	if (command)
-	{
-		command = command.toLowerCase()
-		devices."$command"()
+	if (!devices) {
+		httpError(404, "Devices not found")
+	}
+	if (command){
+		//command = command.toLowerCase()
```
larsfinander: Leftover comment?

smartapps/smartthings/ubi.src/ubi.groovy
```diff
@@ -442,31 +442,90 @@ def executePhrase() {
...
+	if (!device) {
 			httpError(404, "Device not found")
-		}
-		else
-		{
-			device."$command"()
-		}
 	}
+
+	//def command = params.command
```
larsfinander: leftover comments

smartapps/smartthings/ubi.src/ubi.groovy
```diff
@@ -442,31 +442,90 @@ def executePhrase() {
...
+	if (command) {
+		//command = command.toLowerCase()
```
larsfinander: leftover comments

larsfinander:
👍

rohandesai:
Tested this with @igalburt , and as discussed with him this is good to go. Merging this.

bflorian:
I don't see any problem with this change but it does seem to be duplicating much of the logic that is already in the platform in the [DeviceWrapper](https://github.com/PhysicalGraph/data-management/blob/master/src/groovy/physicalgraph/app/DeviceWrapper.groovy#L44-L67) class that will throw an error if a SmartApp tried to call a command that the device itself does not support. 

Can't this logic be simplified to just constrain the /switches path to only execute on and off commands, the /locks path to only support lock/unlock, etc., and leave the checking of wether the device itself supports a given command to the platform?

PR #899, opened 2016-05-16
[CHF-74] Add category to SmartSense DTH

jackchi:
Here's a guideline for defining the Device Category. Currently we will put it in the metadata. The thinking is that eventually data-mgmt IDE can pick this info up. 

@unixbeast


ndasu:
:thumbsup:

PR #900, opened 2016-05-16
DPROT-94 removing cree fingerprint from zll dimmer until the fix in trim 

workingmonk:
for deviceJoinName

cc @tpmanley


tpmanley:
👍

PR #902, opened 2016-05-17
DVCSMP-1775 - Call poll() every 5 minutes to reduce duplicated calls

juano2310:
Call poll() every 5 minutes instead of discovery() to reduce duplicated getActivityList() call.

### DIFF
--- a/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
+++ b/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
@@ -339,7 +339,7 @@ def initialize() {
 	state.aux = 0
 	if (selectedhubs || selectedactivities) {
 		addDevice()
-        runEvery5Minutes("discovery")
+        runEvery5Minutes("poll")
 	}
 }
 
@@ -474,7 +474,7 @@ def activity(dni,mode) {
 def poll() {
 	// GET THE LIST OF ACTIVITIES
     if (state.HarmonyAccessToken) {
-    	getActivityList()
+    	  getActivityList()
         def Params = [auth: state.HarmonyAccessToken]
         def url = "https://home.myharmony.com/cloudapi/state?${toQueryString(Params)}"
         try {
@@ -520,14 +520,17 @@ def poll() {
                 return "Poll completed $map - $state.hubs"
             }
         } catch (groovyx.net.http.HttpResponseException e) {
-            if (e.statusCode == 401) { // token is expired
-                state.remove("HarmonyAccessToken")
-                return "Harmony Access token has expired"
-            }
-		} catch(Exception e) {
-        	log.trace e
-		}
-	}
+              if (e.statusCode == 401) { // token is expired
+                  state.remove("HarmonyAccessToken")
+                  log.warn "Harmony Access token has expired"
+              }
+        } catch (java.net.SocketTimeoutException e) {
+        	log.warn "Connection to the hub timed out. Please restart the hub and try again."
+              state.resethub = true
+        } catch (e) {
+        	log.warn "Hostname in certificate didn't match. Please try again later."
+        }
+    }
 }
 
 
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -474,7 +474,7 @@ def activity(dni,mode) {
 def poll() {
 	// GET THE LIST OF ACTIVITIES
     if (state.HarmonyAccessToken) {
-    	getActivityList()
+    	  getActivityList()
```
dsainteclaire, line 477: Just nitpicking, but this formatting is weird

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -520,14 +520,17 @@ def poll() {
...
         } catch (groovyx.net.http.HttpResponseException e) {
-            if (e.statusCode == 401) { // token is expired
-                state.remove("HarmonyAccessToken")
-                return "Harmony Access token has expired"
-            }
-		} catch(Exception e) {
-        	log.trace e
-		}
-	}
+              if (e.statusCode == 401) { // token is expired
```
dsainteclaire, line 523: Does Harmony support refresh tokens? If their token expired, so they are logged out, are they being alerted somewhere else that they need to login? It looks like poll() will just keep silently failing until the user logs back in

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -520,14 +520,17 @@ def poll() {
...
+                  state.remove("HarmonyAccessToken")
+                  log.warn "Harmony Access token has expired"
+              }
+        } catch (java.net.SocketTimeoutException e) {
+        	log.warn "Connection to the hub timed out. Please restart the hub and try again."
+              state.resethub = true
```
dsainteclaire, line 529: Nitpick.  Weird formatting. Also, is the only place where the user would see the warning about restarting the hub in the auth page?  It's my first time looking at this code, so I only scanned it quickly

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310, line 523: It might. @sterlingz I think we should create a ticket for that but  don't think that should hold this PR. @dsainteclaire

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -520,14 +520,17 @@ def poll() {
...
+        } catch (e) {
+        	log.warn "Hostname in certificate didn't match. Please try again later."
```
dsainteclaire: Is this log statement accurate?  Because in this case e could be any type of exception that hasn't already been caught, which could make debugging ambiguous if you his this block

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -339,7 +339,7 @@ def initialize() {
 	state.aux = 0
 	if (selectedhubs || selectedactivities) {
 		addDevice()
-        runEvery5Minutes("discovery")
+        runEvery5Minutes("poll")
```
dsainteclaire, line 342: Is there more details about what the goal of this change is?  Because switching from one method call to another seems to accomplish different goals, because it looks like discovery initially hits a completely different API compared to what gets hit by poll(), and then discovery() calls poll() on it's own once it goes through it's own initial API call.  I just want to make sure that some of that logic isn't necessary that is now being excluded by calling a different method.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310, line 529: Yes and in the IDE log. This would help support during debugging at least. I would suggest to create tickets for every improvement suggested.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310, line 342: We only need to discover at the beginning and we have to poll every 5 minutes to keep states up to date. Poll calls getActivityList() that calls https://home.myharmony.com/cloudapi/activity/all as discovery generating a duplication.

dsainteclaire:
👍

PR #908, opened 2016-05-18
INTL-527 Temperature Notifications Should Respect Location Temp Scale

ch3lsea:
For SmartApps It's Too Hot/Cold the temperature in the notification was defaulting to fahrenheit even if the user's location setting was celsius. This PR makes the app check the location setting and send the correct temp scale in the notification.

### DIFF
--- a/smartapps/smartthings/its-too-cold.src/its-too-cold.groovy
+++ b/smartapps/smartthings/its-too-cold.src/its-too-cold.groovy
@@ -73,7 +73,8 @@ def temperatureHandler(evt) {
 			// TODO: Send "Temperature back to normal" SMS, turn switch off
 		} else {
 			log.debug "Temperature dropped below $tooCold:  sending SMS to $phone1 and activating $mySwitch"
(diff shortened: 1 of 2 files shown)


smartapps/smartthings/its-too-cold.src/its-too-cold.groovy
```diff
@@ -73,7 +73,8 @@ def temperatureHandler(evt) {
 			// TODO: Send "Temperature back to normal" SMS, turn switch off
 		} else {
 			log.debug "Temperature dropped below $tooCold:  sending SMS to $phone1 and activating $mySwitch"
-			send("${temperatureSensor1.displayName} is too cold, reporting a temperature of ${evt.value}${evt.unit?:"F"}")
+			def tempScale = location.temperatureScale == "C" ? "C" : "F"
```
sabersd: this could just be "C"?: "F"

sabersd:
👍

PR #914, opened 2016-05-24
Newer Fibaro motion sensor handler needs motion state labels

jazzslider:


workingmonk:
@tylerlange @mckeed this seems a valid change. Small but effective.

tylerlange:
👍

stevenchitov:
We have now "Motion" and "No motion" labels. Tested on Staging

PR #931, opened 2016-05-26
INTL-626 Add mfr fingerprint for Fibaro Smoke Sensor

mckeed:
To solve some fingerprinting discrepancies.


mckeed:
See https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/941

PR #942, opened 2016-05-31
DVCSMP-1801: mfgCode in the ZigBee library for smartsense-multi-sensor

varzac:
With the change to the ZigBee library in appengine to add
the optional manufacturers code, the custom writing of
ZigBee commands can be replaced with calls to the library

@tpmanley

### DIFF
--- a/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
@@ -403,39 +403,21 @@ def refresh() {
 
 	if (device.getDataValue("manufacturer") == "SmartThings") {
 		log.debug "Refreshing Values for manufacturer: SmartThings "
-		refreshCmds = refreshCmds + [
-			/* These values of Motion Threshold Multiplier(01) and Motion Threshold (7602)
-				seem to be giving pretty accurate results for the XYZ co-ordinates for this manufacturer.
-				Separating these out in a separate if-else because I do not want to touch Centralite part
-				as of now.
-			*/
-
-			"zcl mfg-code ${manufacturerCode}", "delay 200",
-			"zcl global write 0xFC02 0 0x20 {01}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-
-			"zcl mfg-code ${manufacturerCode}", "delay 200",
-			"zcl global write 0xFC02 2 0x21 {7602}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-		]
+		/* These values of Motion Threshold Multiplier(0x01) and Motion Threshold (0x0276)
+            seem to be giving pretty accurate results for the XYZ co-ordinates for this manufacturer.
+            Separating these out in a separate if-else because I do not want to touch Centralite part
+            as of now.
+        */
+		refreshCmds += zigbee.writeAttribute(0xFC02, 0x0000, 0x20, 0x01, [mfgCode: manufacturerCode])
+		refreshCmds += zigbee.writeAttribute(0xFC02, 0x0002, 0x21, 0x0276, [mfgCode: manufacturerCode])
 	} else {
-		refreshCmds = refreshCmds + [
-			/* sensitivity - default value (8) */
-			"zcl mfg-code ${manufacturerCode}", "delay 200",
-			"zcl global write 0xFC02 0 0x20 {02}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-		]
+		refreshCmds += zigbee.writeAttribute(0xFC02, 0x0000, 0x20, 0x02, [mfgCode: manufacturerCode])
 	}
 
 	//Common refresh commands
-	refreshCmds = refreshCmds + [
-		"st rattr 0x${device.deviceNetworkId} 1 0x402 0", "delay 200",
-		"st rattr 0x${device.deviceNetworkId} 1 1 0x20", "delay 200",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global read 0xFC02 0x0010",
-		"send 0x${device.deviceNetworkId} 1 1","delay 400"
-	]
+	refreshCmds += zigbee.readAttribute(0x0402, 0x0000) +
+			zigbee.readAttribute(0x0001, 0x0020) +
+			zigbee.readAttribute(0xFC02, 0x0010, [mfgCode: manufacturerCode])
 
 	return refreshCmds + enrollResponse()
 }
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -403,39 +403,21 @@ def refresh() {
...
-			"zcl global write 0xFC02 2 0x21 {7602}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-		]
+		/* These values of Motion Threshold Multiplier(01) and Motion Threshold (7602)
+            seem to be giving pretty accurate results for the XYZ co-ordinates for this manufacturer.
+            Separating these out in a separate if-else because I do not want to touch Centralite part
+            as of now.
+        */
+		refreshCmds += zigbee.writeAttribute("0xFC02", "0", "0x20", "01", [mfgCode: manufacturerCode])
+		refreshCmds += zigbee.writeAttribute("0xFC02", "2", "0x21", "0276", [mfgCode: manufacturerCode])
```
tpmanley: The arguments are allowed to be strings for backwards compatibility, but they really should be integers. Also the attributes should be 0x0000 and 0x0002 to make it more clear what they refer to since attributes are 16-bits.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0013 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0014 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500"
-		]
+	def configCmds = enrollResponse()
```
tpmanley: We typically create lists of commands and concatenate them all at once. Not a huge deal, but doing it that way will match most of the other DTH code we've written.

tpmanley:
Put the Jira in the PR title

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
-		"zcl global send-me-a-report 0xFC02 0x0013 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0014 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500"
-		]
+	def configCmds = enrollResponse() +
+			zigbee.configureReporting(0x0001, 0x0020, 0x20, 30, 21600, 1) +
+			zigbee.configureReporting(0x0402, 0x000, 0x29, 30, 3600, 6400) +
```
tpmanley: Missing one 0 in the attribute

tpmanley: 6400 -> 0x0064

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
+			zigbee.configureReporting(0xFC02, 0x0010, 0x18, 10, 3600, 0x01, [mfgCode: manufacturerCode]) +
+			zigbee.configureReporting(0xFC02, 0x0012, 0x29, 1, 3600, 0x0100, [mfgCode: manufacturerCode]) +
```
tpmanley: 0x0100 -> 0x0001 here and next three lines

tpmanley:
@workingmonk can you review this too?

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
workingmonk: or you can change it to 100 (more readable)

workingmonk: library will convert it to 0x0064 and then 6400 before sending it to the chip

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0014 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500"
-		]
+	def configCmds = enrollResponse() +
+			zigbee.configureReporting(0x0001, 0x0020, 0x20, 30, 21600, 0x0100) +
+			zigbee.configureReporting(0x0402, 0x0000, 0x29, 30, 3600, 100) +
+			zigbee.configureReporting(0xFC02, 0x0010, 0x18, 10, 3600, 0x0100, [mfgCode: manufacturerCode]) +
```
tpmanley: The length of datatype 0x18 is 1 so the payload should just be 0x01

workingmonk:
the getManufacturerCode should probably have `if else (manufacturer == "CentraLite")` rather than just `else`.
After all these changes after the comments, @varzac can you pair a new sensor if this new code and see if the attributes are reporting fine.

workingmonk:
👍  after the test

varzac:
@tpmanley @workingmonk I have paired a device using this DTH and everything reports.  The only thing is that the battery level doesn't show up right away, or on a refresh, but it did show up after 5 or so minutes.  Is this normal?  

Also if you could just take a look at the configure reporting values one more time to make sure everything is correct with regards to the length and big/little endianness.  I tried looking up the cluster definitions, but I couldn't find the relevant information, so I was not looking in the right place.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -403,39 +403,21 @@ def refresh() {
...
-
-			"zcl mfg-code ${manufacturerCode}", "delay 200",
-			"zcl global write 0xFC02 0 0x20 {01}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-
-			"zcl mfg-code ${manufacturerCode}", "delay 200",
-			"zcl global write 0xFC02 2 0x21 {7602}", "delay 200",
-			"send 0x${device.deviceNetworkId} 1 1", "delay 400",
-		]
+		/* These values of Motion Threshold Multiplier(01) and Motion Threshold (7602)
```
tpmanley: The values in this comment should be updated to match the code.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
 	String zigbeeEui = swapEndianHex(device.hub.zigbeeEui)
```
tpmanley: This line isn't used anymore so can be removed

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0013 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0014 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500"
-		]
+	def configCmds = enrollResponse() +
+			zigbee.configureReporting(0x0001, 0x0020, 0x20, 30, 21600, 0x01) +
```
tpmanley: This can be replaced with a call to `batteryConfig`

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -446,35 +428,13 @@ def configure() {
...
+			zigbee.configureReporting(0x0402, 0x0000, 0x29, 30, 3600, 100) +
```
tpmanley: This can be replaced with a call to `temperatureConfig`

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -569,7 +529,7 @@ private Map getXyzResult(results, description) {
 private getManufacturerCode() {
 	if (device.getDataValue("manufacturer") == "SmartThings") {
 		return "0x110A"
-	} else {
+	} else if (manufacturer == "CentraLite") {
```
tpmanley: I'm not sure this is a good idea. If a device was joined a long time ago it may not have manufacturer set in device data. I think it's safer to leave it as it was.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -403,78 +403,37 @@ def refresh() {
...
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0013 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500",
-
-		"zcl mfg-code ${manufacturerCode}", "delay 200",
-		"zcl global send-me-a-report 0xFC02 0x0014 0x29 1 3600 {0100}", "delay 200",
-		"send 0x${device.deviceNetworkId} 1 ${endpointId}", "delay 500"
-		]
+	def configCmds = enrollResponse() +
```
workingmonk, line 430: don't we have enrollResponse() in the zigbee library?

varzac, line 430: We do have an enrollResponse in the zigbee library.  However, after testing I have found that if I use the zigbee library version we do not receive open/close messages back.  This is interesting as the only difference I can see between the two is that we are using the st wattr instead of the raw write command like we are using here.  I'm not really sure why this would be.  I'm working on setting up the sniffer so I will see what is actually going over the air.

varzac, line 430: So after some investigation it looks like the enrollResponse version that is implemented in the library is incorrect, which was why using it caused problems here.  The bytes for the EUI get double swapped. https://smartthings.atlassian.net/browse/DPROT-114

For now I will leave the version here.

workingmonk:
looks good 👍 
needs to make sure that this gets regression testing before going out to prod. If you merge it in before EOD tomm, this should be in next week's release

tpmanley:
👍

PR #943, opened 2016-05-31
DVCSMP-1802 - Removing "motion sensor" capability from ecobee remote sensors

tylerlange:
These devices aren't meant to be used as security sensors. They have
inconsistent response times, and can sometimes send events quite a bit
after they actually happen. Suggested fix is to remove

### DIFF
--- a/devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
+++ b/devicetypes/smartthings/ecobee-sensor.src/ecobee-sensor.groovy
@@ -18,7 +18,6 @@ metadata {
 	definition (name: "Ecobee Sensor", namespace: "smartthings", author: "SmartThings") {
 		capability "Sensor"
 		capability "Temperature Measurement"
-		capability "Motion Sensor"
 		capability "Refresh"
 	}
 


tylerlange:
@dsainteclaire @jasonbio

jasonbio:
Might want to make sure the parent SmartApp isn't making motion sensor requests if it's dealing with this specific device type. If it isn't then 👍

tylerlange:
@jasonbio we want to keep the motion events exposed but we don't want to be able to subscribe to them in other SmartApps.

tylerlange:
@dsainteclaire bump

dsainteclaire:
Since you removed the motion sensor capability, I think that it makes sense to also remove the motion tile from the DTH, because that feels kind hanky.  Also, I was looking at the Ecobee SmartApp and this line (573) would probably need to be removed because the device handler would no longer support motion:
d.sendEvent(name:"motion", value: occupancy)

tylerlange:
We want to keep the tile, but not allow SmartApps to subscribe to this as a standard motion sensor.

jasonbio:
@dsainteclaire please also double check the parent connector app and make sure that if it polls this DTH with any motion related stuff, it won't throw an exception with the capability removed (I don't have time to check this myself 😢 )

dsainteclaire:
@jasonbio @tylerlange I got side-tracked today with other things, but I'll try this out first thing tomorrow morning

dsainteclaire:
@tylerlange after asking about it in the developers channel it sounds like you'd need to add a custom attribute to the DTH to handle the motion events, and the service manager would need to update that attribute instead of device.motion, because removing the capability causes the plumbing to stop working because the motion attribute would not longer be there.

@jasonbio JFYI

jasonbio:
@dsainteclaire okay so just double checking that this PR needs an additional change to the service manager correct?

PR #948, opened 2016-06-01
Revert "DVCSMP-1107 Stage 2 of adding preferences to support garage s…

workingmonk:
…ensor in v1 safari mutli sensor"


workingmonk:
reverting the change as there is some issue with AppEngine

PR #958, opened 2016-06-02
Removed prempoint because it was published to the wrong repository

rohandesai:
@tylerlange @larsfinander 

@workingmonk Prempoint made a publication request to wrong repository. This PR will revert the merge in public. We will also need to remove this from staging.


workingmonk:
since this is already in staging. create the pr in staging.

PR #962, opened 2016-06-03
DVCSMP-1676 Philips Hue: Need to provide timeout message if search

larsfinander:
takes too long.

-Shows timeout screen after 5 minutes
-Removed old image code that didnt work for one of the pages
-Changed to show bridge serial number instead of IP
-Handled null pointer when adding unsupported devices


larsfinander:
@juano2310 Please review, not sure what is up with the weird whitespace layout in github.. Looks fine in editor

juano2310:
I think images are an interesting way of guiding the user. Are we planing on restoring or adding new images for each step?

Code is 👍

larsfinander:
We do have official images from hue and I am considering readding that later.

workingmonk:
edited title

PR #982, opened 2016-06-09
NPE fix for NetAtmo

rohandesai:
NetAtmo was failing on NPE's. Added required null safe operators. Tested it by forcing NULL values, and it does not error out any more on NPE for line 495 it was errorring out on.


rohandesai:
@workingmonk @larsfinander

larsfinander:
👍

PR #990, opened 2016-06-14
Add re-type for Vision Motion Sensor Plus.

dkirker:


dkirker:
@workingmonk  @mckeed fishing for a +1

mckeed:
+1

PR #995, opened 2016-06-15
DVCSMP-1727 Philips Hue: Update status on response

larsfinander:
-Only send events when feedback is recevied from bridge
-Cleaned up color conversions
-DVCSMP-1728 Philips Hue: Handle new LAN polling
-DVCSMP-1763 Philips Hue: Bad rounding can cause level to be 0, light on
-Updated some strings

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -53,7 +53,6 @@ def bridgeDiscovery(params=[:])
 
 	def options = bridges ?: []
 	def numFound = options.size() ?: 0
-
 	if (numFound == 0) {
 		if (state.bridgeRefreshCount == 25) {
 			log.trace "Cleaning old bridges memory"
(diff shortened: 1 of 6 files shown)


larsfinander:
DO NOT MERGE

larsfinander:
@juano2310 Please review

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -661,10 +665,63 @@ def locationHandler(evt) {
...
 	log.trace "Doing Hue Device Sync!"
+
+	// Check if state.updating failed to clear
+	if (state.lastUpdateStarted < (now() - 20 * 1000) && state.updating) {
+		state.updating = false
+		log.warn "state.updating failed to clear"
+	}
+
 	convertBulbListToMap()
 	poll()
```
juano2310, line 674: Do we still need this poll() there every 5 minutes?

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -661,10 +665,63 @@ def locationHandler(evt) {
...
 	ssdpSubscribe()
 	discoverBridges()
```
juano2310, line 676: I used discoverBridges() to monitor if the IP changed. Now that we have local polling. Don't we have a better way to do this?

smartapps/smartthings/hue-connect.src/hue-connect.groovy
larsfinander, line 674: We do because even if hub polls, it only sends us an answer back when data changes unless we explicitly do a poll()

smartapps/smartthings/hue-connect.src/hue-connect.groovy
larsfinander, line 676: Local polling only gives us changes when we query light status. As soon as IP changes for example, local polling will just stop until smartapp triggers a new poll with new IP.

juano2310:
:+1:

PR #1003, opened 2016-06-20
PRP-151 - Harmony Cloud

juano2310:
@lbredeso do you mind taking a look?

### DIFF
--- a/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
+++ b/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
@@ -824,17 +824,51 @@ def deviceHandler(evt) {
 
 def sendToHarmony(evt, String callbackUrl) {
 	def callback = new URI(callbackUrl)
-	def host = callback.port != -1 ? "${callback.host}:${callback.port}" : callback.host
-	def path = callback.query ? "${callback.path}?${callback.query}".toString() : callback.path
-	sendHubCommand(new physicalgraph.device.HubAction(
-		method: "POST",
-		path: path,
-		headers: [
-			"Host": host,
-			"Content-Type": "application/json"
-		],
-		body: [evt: [deviceId: evt.deviceId, name: evt.name, value: evt.value]]
-	))
+  if(isIP(callback.host)){
+  	def host = callback.port != -1 ? "${callback.host}:${callback.port}" : callback.host
+  	def path = callback.query ? "${callback.path}?${callback.query}".toString() : callback.path
+  	sendHubCommand(new physicalgraph.device.HubAction(
+  		method: "POST",
+  		path: path,
+  		headers: [
+  			"Host": host,
+  			"Content-Type": "application/json"
+  		],
+  		body: [evt: [deviceId: evt.deviceId, name: evt.name, value: evt.value]]
+  	))
+  } else {
+    def params = [
+      uri: callbackUrl,
+      body: [evt: [deviceId: evt.deviceId, name: evt.name, value: evt.value]]
+    ]
+    try {
+        log.debug "Sending data to Harmony Cloud: $params"
+        httpPostJson(params) { resp ->
+            log.debug "Harmony Cloud - Response: ${resp.status}"
+        }
+    } catch (e) {
+        log.error "Harmony Cloud - Something went wrong: $e"
+    }
+  }
+}
+
+public static boolean isIP(String str) {
+    try {
+         String[] parts = str.split("\\.");
+         if (parts.length != 4) return false;
+         for (int i = 0; i < 4; ++i) {
+             int p
+             try {
+               p = Integer.parseInt(parts[i]);
+             } catch (Exception e) {
+               return false;
+             }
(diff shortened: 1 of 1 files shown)


lbredeso:
👀

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -813,7 +813,7 @@ def deviceHandler(evt) {
 		}
 	} else if (deviceInfo) {
 		if (deviceInfo.callbackUrl) {
-			sendToHarmony(evt, deviceInfo.callbackUrl)
+			sendToHarmonyCloud(evt, deviceInfo.callbackUrl)
```
lbredeso: It looks like this will work for the Harmony Cloud calls, but the code doesn't appear to be backwards compatible with existing `deviceInfo.callbackUrl` values, which may reference LAN IP addresses that will not work correctly for `httpPostJson`. Since we are using the same SmartApp both for existing and new installations, you will probably need to inspect `deviceInfo.callbackUrl` to decide whether to call `sendToHarmony` or `sendToHarmonyCloud`.

I'm not sure the best way to make that determination. Maybe you could parse it and if it contains an IP address, use `sendToHarmony`, but if it is not an ip address, `sendToHarmonyCloud`?

juano2310: In line 810 it asks for state.harmonyHubs and sends data to sendToHarmony if there is any, so I thought this was trying to determine where to send the information locally or to the cloud. In that case maybe I should ask what is the URL for their cloud and if the callback contains their cloud url then I route it to sendToHarmonyCloud

lbredeso: state.harmonyHubs is the new way of storing hub connection info, while state[evt.deviceId] is the old way.  Both are checked, since we need to stay backwards compatible, but state.harmonyHubs is the "new" way. We should make sure we have the ability to call either method from either piece of data.

If you use their cloud URL to make the determination instead of an IP address, you'll just need to be able to handle the case where they decide to change their cloud URL some day.

juano2310: Do you have a good method to check for IP address in the URL? If not I can
dig for one.

lbredeso: I know you could at least parse out the host with:

```
new URI("http://10.0.1.1/foo").host // 10.0.1.1
```

As for knowing if an IP address is in the host rather than some DNS name, you could try using a Groovy regex for that.

juano2310:
@lbredeso how about now? :)

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -824,17 +824,46 @@ def deviceHandler(evt) {
...
+    }
+  }
+}
+
+public static boolean isIP(String str) {
+    try {
+         String[] parts = str.split("\\.");
+         if (parts.length != 4) return false;
+         for (int i = 0; i < 4; ++i) {
+             int p = Integer.parseInt(parts[i]);
```
lbredeso: Nice, I think this new code should work, but you should probably put the Integer.parseInt inside a try/catch, otherwise I think hypothetical hosts like "smartthings.harmony.logitech.com" would cause a NumberFormatException to be thrown.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -824,17 +824,50 @@ def deviceHandler(evt) {
...
+  }
+}
+
+public static boolean isIP(String str) {
+    try {
+         String[] parts = str.split("\\.");
+         if (parts.length != 4) return false;
+         for (int i = 0; i < 4; ++i) {
+             try {
+               int p = Integer.parseInt(parts[i]);
```
lbredeso: This variable "p" must be declared outside the scope of the try block, then initialized inside, otherwise the subsequent range check will always fail.

juano2310: Would this work? Thanks again!

lbredeso:
👍 I think this looks good, it'll be interesting to see how end-to-end testing goes.

juano2310:
Wondering about that. What do you recommend for QA?

lbredeso:
It needs to be tested before being deployed in production, to avoid inadvertently breaking everything.

You will need to investigate how other integrations do this, I'm curious how we do this for Echo.

One possibility is that you could create a separate copy of the SmartApp specifically for testing upcoming changes. It would then have different OAuth credentials which the Harmony hub could use, but I don't know how you'd get a Harmony hub to work with that. You'll probably need to work with their engineers on a solution.

juano2310:
Thanks a lot! I think we could set up staging as the testing environment and ask them to test against it by providing oauth information they might need.

PR #1007, opened 2016-06-22
[CHF-131] Adding Health Check to V1 SmartSense Open/Close Sensor

jackchi:
Hot-Fix for adding `Health Check` to V1 `SmartSense Open/Close Sensor`

cc
@workingmonk


workingmonk:
why does this need to be a hotfix??

PR #1009, opened 2016-06-22
PRP-151 - Update logitech-harmony-connect.groovy

spurohit1:
If port is present in the callbackUrl then send command to hub else post the request through http.
Removed isIP function since it is no longer required.


lbredeso:
I think this should work, provided there is never an explicit port added to the cloud callbackUrl in the future, but I'm curious: why wouldn't the previous code work?

spurohit1:
While updating the patch, I realized Juan was using host to parse the
ip-address and not callbackUrl. This must have worked but the notification
is broken somewhere else. sendToHarmony function is not triggered during
notifications.

workingmonk:
please add ticket number to this PR cc @lbredeso @juano2310

PR #1015, opened 2016-06-27
New handlers for Vision ZW+ Motion and Door/Window DVCSMP-1826 DVCSMP-1831

mckeed:
These devices are sold under different brand names in different regions: Monoprice and Vision.


dkirker:
+1

PR #1016, opened 2016-06-27
[CHF-138] Z-Wave Aeon MultiSensors are health checked

jackchi:
Most Aeon sensors are configured to report back `5-8 minutes`. 
ZigBee devices are now set to `2hrs`.  I propose setting this to `1hr`. 

cc 
@mckeed @mortent

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
@@ -20,6 +20,9 @@ metadata {
 		capability "Configuration"
 		capability "Sensor"
 		capability "Battery"
+		capability "Health Check"
+
+		command "configureAfterSecure"
 
 		fingerprint deviceId: "0x0701", inClusters: "0x5E,0x86,0x72,0x59,0x85,0x73,0x71,0x84,0x80,0x30,0x31,0x70,0x98,0x7A", outClusters:"0x5A"
 	}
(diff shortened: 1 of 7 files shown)


devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
```diff
@@ -12,14 +12,17 @@
...
 		capability "Motion Sensor"
 		capability "Temperature Measurement"
 		capability "Relative Humidity Measurement"
 		capability "Illuminance Measurement"
 		capability "Configuration"
 		capability "Sensor"
 		capability "Battery"
+		capability "Health Check"
+
+		command "configureAfterSecure"
```
jackchi, line 25: The `Aeon MultiSensor Gen5` valueTile `configureAfterSecure` is the command on the tile. Need to add this or else that configure tile doesn't actually call `configure`. 
@mckeed - is this the right approach?

mckeed:
Looks good except for in the Multi 6 DTH, the report interval can be changed in the preferences, so the checkInterval should be based on that. The others should eventually have the same setting, but it can be changed when that's added.

devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
mckeed, line 25: That's fine. I'm of the opinion that configure tiles should be taken out of all z-wave devices (see DVCSMP-264), but for now this is good.

jackchi:
I see on Aeon MultiSensor 6 we need to have `checkInterval` be variable. 
We currently only send `checkInterval` in configure() upon updated/install. 

Do we want to punt Aeon6 this for now? Or keep 1 hour?

mckeed:
Not sure what you mean? Why can't you just set the value to something like `timeOptionValueMap[reportInterval] * 2`

configure() gets called whenever the setting is changed

jackchi:
All `checkInterval` change to 2\* (periodic reporting interval). Ready to merge. 
@mckeed @mortent

lbredeso:
We shouldn't be adding new categorization mechanisms to DeviceTypes. If we need a capability more specific than "healthCheck", we should create multiple capabilities and implement them where appropriate.

👎

PR #1023, opened 2016-06-30
PRP-172 Fix Hue Connect parse errors on newer versions of Groovy

rappleg:
#1022, but against production

/cc @workingmonk


workingmonk:
👍

PR #1030, opened 2016-07-06
Revert "PRP-172 Fix Hue Connect parse errors on newer versions of Groovy"

rappleg:
Reverting this commit for now to get a new deploy out and taking a different approach since LazyMap is not available in the current version of Groovy we run so not backwards compatible.

/cc @workingmonk


workingmonk:
👍

PR #1034, opened 2016-07-06
DVCSMP-1879 Fix error in Z-Wave Door/Window Sensor throwing exceptions on update

mckeed:
Accidentally took out format() without calling commands()


rappleg:
👍

PR #1041, opened 2016-07-11
PRP-172 Fix Hue Connect parse errors on newer versions of Groovy

rappleg:
Backwards compatible fix for issue in #1022, but will work for both Grails 2.3.11 and Grails 2.5.4.

/cc @workingmonk


workingmonk:
👍

PR #1053, opened 2016-07-14
PRP-192 Fix false low battery notifications on Groovy 2.4.4

rappleg:
In Groovy 2.4.4 when casting BigDecimal to int evaluates to 0, where in Groovy 2.1.9 it gets calculated correctly.  This is causing false battery notifications when trying to upgrade Grails versions.  This change rounds the percentage before getting the min to make sure it will work on both versions of Groovy.

/cc @workingmonk @tslagle13


workingmonk:
👍 
how urgent is this? I have a prod deploy coming up tomm

PR #1063, opened 2016-07-18
Add capability sensor to SS Open/Closed

tslagle13:
There are some integrations out there using the "Actuator" and "Sensor" Capabilities and this doesn't show up for them.

cc @workingmonk


tylerlange:
:+1: @tslagle13

CosmicPuppy:
**Thanks for the fast response!**

I hate to say this, though, but really, someone needs to do a QA scan on all the published DTHs to make sure each claims at least one or both of "sensor" or "actuator"... _Please?_

PR #1085, opened 2016-08-02
SSVD-1823 No parenthesis

juano2310:


gausnes:
👍  looks good from a data-management side

PR #1098, opened 2016-08-04
SSVD-1605 and SSVD-1897 Fix for update of switch tile

marstorp:
Fix for update of switch tile when speaker is turned on/off from routines and adding a intermediate state between on/off to prevent user to switch state before speaker has responded on the switch command.


workingmonk:
👍

Issue #1058, opened 2016-07-15
SmartSense Open/Closed Sensor, SmartSense Moisture, Simulated Switch (and other many other DTHs) is missing standard Capability "sensor" and/or "actuator"


CosmicPuppy:
### All DTHs should claim `capability "sensor"` and/or `capability "actuator"`.

_For reference,_ the **SmartSense _Motion_ Sensor** correctly claims `capability "sensor"` here:
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/2b3a4e1278fbf2838621cf6a419cbbde9d37224c/devicetypes/smartthings/smartsense-motion-temp-sensor.src/smartsense-motion-temp-sensor.groovy#L26

But the **SmartSense _Open/Closed_ Sensor** does not:
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/2b3a4e1278fbf2838621cf6a419cbbde9d37224c/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy#L21

---

This is critical to **SmartTiles V6** because we encourage the users to select _all_ their devices with two general filter input statements (sensor and actuator). As such, Devices of this DTH are currently not selectable in our SmartApp: **SmartTiles V6 Beta (Connect)**.

CosmicPuppy:
`Capability "sensor"` is _**also**_ missing in:  https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/tile-ux/tile-basic-presence.src/tile-basic-presence.groovy#L20

CosmicPuppy:
**And on SmartSense Moisture Sensor:**
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/2b3a4e1278fbf2838621cf6a419cbbde9d37224c/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy#L18

workingmonk:
closing this issue as you've open the PRs to fix the issue

PR #1101, opened 2016-08-05
Added Capability "Sensor" and/or "Actuator" per DTH standards.

CosmicPuppy:
There are some integrations out there using the "Actuator" and "Sensor" Capabilities and instances of these Devices (and perhaps more) do not show up for them. _Example_ SmartApp "SmartTiles V6 Beta".
- Ref Docs: http://docs.smartthings.com/en/latest/device-type-developers-guide/overview.html?highlight=sensor%20actuator#actuator-and-sensor
- Ref Issue #1058.
- Ref @tslagle13 for more info regarding **SmartTiles V6**'s use of these standard Capabilities as device authorization input filters.

**CC:** @tylerlange , @workingmonk 

_Thank-you!_

### DIFF
--- a/devicetypes/smartthings/testing/simulated-color-control.src/simulated-color-control.groovy
+++ b/devicetypes/smartthings/testing/simulated-color-control.src/simulated-color-control.groovy
@@ -1,6 +1,8 @@
 metadata {
 	definition (name: "Simulated Color Control", namespace: "smartthings/testing", author: "SmartThings") {
     	capability "Color Control"
+		capability "Sensor"
+		capability "Actuator"
 	}
 
 	simulator {
--- a/devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy
+++ b/devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy
@@ -16,6 +16,8 @@ metadata {
     definition (name: "Simulated Switch", namespace: "smartthings/testing", author: "bob") {
 		capability "Switch"
         capability "Relay Switch"
+		capability "Sensor"
+		capability "Actuator"
 
 		command "onPhysical"
 		command "offPhysical"
--- a/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
@@ -22,6 +22,7 @@ metadata {
 		capability "Temperature Measurement"
 		capability "Water Sensor"
 		capability "Health Check"
+		capability "Sensor"		
 
 		command "enrollResponse"
 
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -22,6 +22,7 @@ metadata {
 		capability "Temperature Measurement"
 		capability "Refresh"
 		capability "Health Check"
+		capability "Sensor"		
 
 		command "enrollResponse"
 
--- a/devicetypes/smartthings/smartsense-open-closed-accelerometer-sensor.src/smartsense-open-closed-accelerometer-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-open-closed-accelerometer-sensor.src/smartsense-open-closed-accelerometer-sensor.groovy
@@ -24,6 +24,7 @@
  		capability "Refresh"
  		capability "Temperature Measurement"
 		capability "Health Check"
+		capability "Sensor"
 
 		command "enrollResponse"
  	}
--- a/devicetypes/smartthings/smartsense-temp-humidity-sensor.src/smartsense-temp-humidity-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-temp-humidity-sensor.src/smartsense-temp-humidity-sensor.groovy
@@ -21,6 +21,7 @@ metadata {
 		capability "Temperature Measurement"
 		capability "Relative Humidity Measurement"
 		capability "Health Check"
+		capability "Sensor"
 
 		fingerprint endpointId: "01", inClusters: "0001,0003,0020,0402,0B05,FC45", outClusters: "0019,0003"
 	}
--- a/devicetypes/smartthings/testing/simulated-alarm.src/simulated-alarm.groovy
+++ b/devicetypes/smartthings/testing/simulated-alarm.src/simulated-alarm.groovy
@@ -16,6 +16,8 @@
 metadata {
 	definition (name: "Simulated Alarm", namespace: "smartthings/testing", author: "SmartThings") {
 		capability "Alarm"
+		capability "Sensor"
+		capability "Actuator"
 	}
 
 	simulator {
--- a/devicetypes/smartthings/testing/simulated-contact-sensor.src/simulated-contact-sensor.groovy
+++ b/devicetypes/smartthings/testing/simulated-contact-sensor.src/simulated-contact-sensor.groovy
@@ -15,6 +15,7 @@ metadata {
 	// Automatically generated. Make future change here.
 	definition (name: "Simulated Contact Sensor", namespace: "smartthings/testing", author: "bob") {
 		capability "Contact Sensor"
+		capability "Sensor"
 
 		command "open"
 		command "close"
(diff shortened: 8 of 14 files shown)


tslagle13:
@workingmonk, Can we merge these changes? Seems really straight forward to me.

devicetypes/smartthings/testing/simulated-color-control.src/simulated-color-control.groovy
```diff
@@ -1,6 +1,8 @@
 metadata {
 	definition (name: "Simulated Color Control", namespace: "smartthings/testing", author: "SmartThings") {
     	capability "Color Control"
+		capability "Sensor"
```
workingmonk, line 4: should color control have capability sensor?

devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy
```diff
@@ -16,6 +16,8 @@ metadata {
     definition (name: "Simulated Switch", namespace: "smartthings/testing", author: "bob") {
 		capability "Switch"
         capability "Relay Switch"
+		capability "Sensor"
```
workingmonk, line 19: sensor for switch?

workingmonk:
@tslagle13 added comments

devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy
CosmicPuppy, line 19: ## **Yes!!!**
- _Any_ Device that _sends / reports_ Events for Attribute values (i.e., reports a key State, but not counting some non-key function like "Battery"), _is_ a **Sensor**.

A **Switch** _reports_ when it turned `on` or `off`. Perhaps better explained that _most_ (or at least _many_) Switches can be _physically_ activated; thus, the Switch is "sensing" the activation and reporting it.

Per these lines of Code; the Simulated Switch does **_send / report**_ Events (including simulated "physically **_sensed_**" events:  https://github.com/CosmicPuppy/SmartThingsPublic/blob/7beb2e390520609dc4f8088b6d6507499017a8fe/devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy#L47-L65

---

By contrast, an "Actuator" is any Thing that _receives_ and acts upon Commands (again, except for secondary Commands like Configure, Poll, and Refresh). Obviously a smart Switch is **_also_** an Actuator, as it _receives_ and acts upon the Commands `on()` and `off()`.

A smart Thermometer (not a _thermostat_) is **_not**_ an "Actuator" because it has no Commands. It only senses and reports temperature, it does not receive Commands to _set_ Temperature. If it accepts configure commands (e.g., perhaps to configure the scale C/F, or refresh), those don't count as actuator qualifying properties. Similarly for a Moisture/Leak Sensor (unless it _also_ is a part of a Valve control / actuation device).

tslagle13:
Agreed. @CosmicPuppy can you update the PR please to remove sensor for the switch and color control DHs?

devicetypes/smartthings/testing/simulated-color-control.src/simulated-color-control.groovy
CosmicPuppy, line 4: - Any Device that sends / reports Events / Attribute values (i.e., reports a key State ... not counting something secondary like "Battery"), is a Sensor.

So... If the Device "reports" if something else changes it's color, then it is acting as a Sensor. 

(_I'm checking the code now to see if this is true_ ... **Update:** Well; perhaps since it is a "simulated boilerplate device", this particular DTH does not report externally initiated changes. So this _particular_ code does not provide any Sensor functionality. I included "Sensor" since _real_ Color DTH's probably do provide color feedback (e.g., if a Hue Bridge connected Hue Bulb is updated from the Hue App, SmartThings updates the Color Tile with the latest real value, right? Or hopefully, it should. My Hue Bridge connected bubs are not great from a synchronization of state viewpoint, unfortunately; but I think that's a bug.).

**EDIT**: Just confirmed that Hue bulbs indeed report Color change Events; thus Hue Bulbs are "Sensors" in that they "Sense" if someone has changed their color, and report what they _sensed_.

```
2016-08-10 1:11:33.806 PM PDT
moments ago DEVICE      color   #3B21FF     Den Centre color is #3B21FF
2016-08-10 1:11:33.345 PM PDT
moments ago DEVICE      color   #3C21FF     Den Centre color is #3C21FF
```

---

**BTW:** For the purpose of SmartTiles V6 SmartApp (and many similar SmartApps), it actually doesn't matter if Sensor and/or Actuator are properly used, as long as one and/or the other Capability is claimed for every single Device. Every DTH must be one and/or the other; unless it is a very, very weird DTH -- I'd love to know of any such example.

Our SmartApp just uses Sensor / Actuator as master input filters and then query the selected Devices for their detailed _specific_ Capabilities in order to know what services to provide for them.

CosmicPuppy: - Any Device that sends / reports Events / Attribute values (i.e., reports a key State ... not counting something secondary like "Battery"), is a Sensor.

So... If the Device "reports" if something else changes it's color, then it is acting as a Sensor. (I'm checking the code now to see if this is true).

---------
**BTW:** For the purpose of SmartTiles V6 SmartApp (and many similar SmartApps), it actually doesn't matter if Sensor and/or Actuator are properly used, as long as one and/or the other Capability is claimed for every single Device. Every DTH must be one and/or the other; unless it is a very, very weird DTH -- I'd love to know of any such example.

Our SmartApp just uses Sensor / Actuator as master input filters and then query the selected Devices for their detailed *specific* Capabilities in order to know what services to provide for them.

CosmicPuppy:
Per Line Note response above, _nearly all_ **Switches** are **Sensors**. --> https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/1101#discussion_r74313629
- [ ] I can update "**Simulated** Color Control" DH ... but let's double-check them all for a moment, first.

workingmonk:
from that point of view, anything with a state is a sensor. so except for buttons, everything will have the sensor capability

CosmicPuppy:
@workingmonk: I agree with you at least 80%.

While the DTH functionality and it's state, etc., is important in the definition, it is important to also consider the nature and purpose of the Physical Device.

For example, I mentioned that a Device which "happens" to report its Battery level, doesn't have to claim Capability Sensor, as that is an internal non-environmental function of the Device. If it were monitoring the Battery on your Car or UPS system ... well, then, definitely, it is a "Sensor".

---

Ironically, your **Button** example is actually incorrect, I think...
- A **Button** is definitely a **_Sensor_**. It _reports / sends_ to SmartThings when it senses it has been "`pushed`" or "`held`".
- A **Button** however, is not necessarily an **_Actuator**_! An Aeon Minimote, for example, does not directly / physically do anything and, it does not have Commands because of this. (Reference [Docs](http://docs.smartthings.com/en/latest/capabilities-reference.html#button) ... Button has no Commands).  **NB:** Some implementations of Button are written to have allow "virtual presses" from SmartApps or Sub-Tiles. This does not make Button an Actuator.

---

Perhaps the simplest _objective_ way to explain it (which you can then _think about and apply to the real world_) is:
1. If a Device has Commands to make it "do something", then it is an **Actuator**. I exclude counting internal Commands like `refresh(), poll(), configure()`. The except is if these Commands don't make the Device _do something_, such as "virtual (non-physical) presses" on a Button.
2. If a Device reports (via Attribute Events) something either physically or from some other automated source, then it is a **Sensor**. I exclude counting sensed internal stuff like Battery ... but it is very rare that a Device which reports Battery isn't also reporting something external like a Switch being flipped, a Temperature, Contact, Motion, Wind Speed, Shock, Button Pressed, Tile Clicked, etc.).

A Siren is an Actuator (has Commands). A glass break detector is a Sensor (no Commands, just Attribute events). .... etc., etc..

workingmonk:
oops. i meant the other way around for the button example. i agree with your theory too, but let me confirm this internally on the actuator v/s sensor debate.

i don't mean to push out this PR. this definitely is one of the things that we want to fix. but we are going to fix it all, lets get it right

CosmicPuppy:
OK. @jimmyjames and a few of us had this discuss a long while back because he was trying to ensure the [Documentation](http://docs.smartthings.com/en/latest/device-type-developers-guide/overview.html?highlight=sensor%20actuator#actuator-and-sensor) was accurate.

The discussion eventually focused on whether or not a DTH had Commands ... which really is a great way to define Actuators (except for the ones that cheat and the internal Commands).

But, as mentioned, Sensors are also defined by their code ... i.e., do they send/report sensed Events. I'm repeating myself; but **_Switch**_ is an excellent example. If a Switch _only_ accepted Commands and turned on/off the power to something, but **_never_** reported its State, then, and only then, would it _not_ be a Sensor. Such a Switch would be unusual, and annoying, but it could exist. Luckily, that's not what we're dealing with here.

---

Please don't let the distinction distract from the **important and urgent** goal that SmartTiles needs every DTH to be one and/or the other ... we don't care at all if it is "correct".

workingmonk:
👍 
since we all agreed to the change internally, i am approving and merging this change in.

workingmonk:
@tslagle13 can you add a ticket number to the PR?

PR #1116, opened 2016-08-12
[CHF-204] Implementation of PING in SmartPower Outlet

jackchi:
Implement `PING` to read on/off attribute from Outlet

PING also artificially create `deviceWatch-lastActivity` event in cases of `isStateChange:false` cases 

Fix refresh() to use `zigBee.electricMeasurementPowerRefresh` 

screen shot 2016-08-12 at 2 34 35 pm

cc @ShunmugaSundar @mortent @darksun


jackchi:
I talked to Tom Manley and the wrapper substitution is good. I'm merging this now.

PR #1128, opened 2016-08-17
SSVD-2465 Philips Hue No description for slider

larsfinander:
-Changed Kelvin tile to Whites (same as Philips Hue app) to reduce user confusion


larsfinander:
@workingmonk  Just waiting for QA

PR #1130, opened 2016-08-18
SSVD-2465 changing the kelvin tile to name tiles for OSRAM

workingmonk:
@tpmanley please review the change
cc @larsfinander for possibly changing tile for hue


workingmonk:
approved by Tom

PR #1138, opened 2016-08-23
SSVD-2532 setting up the generic name at install

workingmonk:
cc @tpmanley


tpmanley:
:+1:

Commit message doesn't seem to match the code so double check that this is what you intended

PR #1141, opened 2016-08-23
  Revert DPROT-2: "Update DTHs to use ZigBee library ZoneStatus"

varzac:
This reverts commit 7bfa0304afccb6b39e967eb7be80219b19f448f2.

@tpmanley @workingmonk

PR #1144, opened 2016-08-23
DPROT-2 Reverting again to put the change back " Revert "Update DTHs to use ZigBee library ZoneStatus""

workingmonk:
Reverts SmartThingsCommunity/SmartThingsPublic#1141

cc @tpmanley @varzac


varzac:
👍

PR #1148, opened 2016-08-24
[DVCSMP-1980] some cleanup of Ecobee Connect SmartApp

jimmyjames:
General cleanup and slight refactoring of the Ecobee Connect SmartApp and Thermostat Device Handler. This PR is a prerequisite for a later PR to convert synchronous HTTP requests to be asynchronous where possible.

Changes in this PR include:
- Whitespace changes to make the source consistent and more readable
- Removal of unused methods
- Removal of unused method parameters
- Extraction of some logic to their own methods to decrease the size of some methods, especially where this can be used in later async changes
- Instead of building a JSON string to send to Ecobee when invoking commands, use a map then convert to JSON so that the parameters are easier to read/understand and change later if necessary
- Removal of storing entire response from Ecobee in atomicState, since it was never used
- Added comments to some methods for clarity and to help future developers/maintainers

There are other refactorings that could be done to improve this code. This PR does not attempt to address everything that could/should be done, but rather makes a set of high-value/low-risk changes to make converting to async http easier and a more isolated change.

@dsainteclaire @jasonbio you guys probably have the most domain knowledge to review.

/cc @DougSven

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -294,10 +294,10 @@ Map sensorsDiscovered() {
 }
 
 def getThermostatDisplayName(stat) {
-	if(stat?.name) {
+    if(stat?.name) {
         return stat.name.toString()
     }
-	return (getThermostatTypeName(stat) + " (${stat.identifier})").toString()
+    return (getThermostatTypeName(stat) + " (${stat.identifier})").toString()
 }
 
 def getThermostatTypeName(stat) {
@@ -384,8 +384,8 @@ def pollHandler() {
 }
 
 def pollChildren(child = null) {
-	def thermostatIdsString = getChildDeviceIdsString()
-	log.debug "polling children: $thermostatIdsString"
+    def thermostatIdsString = getChildDeviceIdsString()
+    log.debug "polling children: $thermostatIdsString"
 
     def requestBody = [
         selection: [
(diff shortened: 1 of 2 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -595,54 +562,26 @@ private refreshAuthToken() {
...
-				path  : "/token",
-				query : [grant_type: 'refresh_token', code: "${atomicState.refreshToken}", client_id: smartThingsClientId],
+			method: 'POST',
+			uri   : apiEndpoint,
+			path  : "/token",
+			query : [grant_type: 'refresh_token', code: "${atomicState.refreshToken}", client_id: smartThingsClientId],
 		]
 
-		log.debug refreshParams
+		log.debug "making refresh request with paramaters: $refreshParams"
```
dsainteclaire: It probably doesn't matter, but I'm not super enthusiastic about logging refresh tokens or auth tokens

jimmyjames: I agree. There are other places as well tokens are logged but I don't think should be. I'll update that.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -381,75 +384,41 @@ def pollHandler() {
 }
 
 def pollChildren(child = null) {
-	def thermostatIdsString = getChildDeviceIdsString()
-	log.debug "polling children: $thermostatIdsString"
-	def data = ""
+    def thermostatIdsString = getChildDeviceIdsString()
+    log.debug "polling children: $thermostatIdsString"
+
+    def requestBody = [
```
dsainteclaire, line 387: I'm very supportive of this, because hardcoding the JSON makes the code hard to read, and even harder when you need to add more params.  Have you tested to make sure it works?  It definitely should, but I would just want to verify it

dsainteclaire:
I would just want to run it through some tests or have QA run it through regression since the way the API payload is generated fundamentally changed (for the better).  I like all the cleanup though.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
jimmyjames: Yes I've tested it and it works, but QA should definitely do a regression test.

jimmyjames:
pushed another commit with removal of logging token info. I'll work with QA to ensure full regression is done, and once thumbed-up, will squash the commits before merging.

dsainteclaire:
Sounds good to me, thanks for the effort! :shipit:

PR #1149, opened 2016-08-24
DPROT-153 Fix event map name in NYCE Motion sensor

varzac:
There was a null pointer as a result of using the wrong map name

This resolves: https://smartthings.atlassian.net/browse/DPROT-153

@tpmanley @workingmonk


workingmonk:
👍

PR #1158, opened 2016-08-26
SSVD-2534 Set temperature unit/scale when sending temperature events

mrobinet:
When sending temperature events, we need to ensure that the unit is also set. This allows us to better handle when the temperature scale is changed for a the location as we are capturing what the scale was when the event occurred.

@mrnohr made changes to update device tiles appropriately, providing the unit is capture on the events. See SSVD-2534 for related PR.

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -152,11 +152,11 @@ def generateEvent(Map results) {
 				sendValue =  location.temperatureScale == "C"? roundC(sendValue) : sendValue
 				isChange = isTemperatureStateChange(device, name, value.toString())
 				isDisplayed = isChange
-				event << [value: sendValue, isStateChange: isChange, displayed: isDisplayed]
+				event << [value: sendValue, unit: temperatureScale, isStateChange: isChange, displayed: isDisplayed]
 			}  else if (name=="maxCoolingSetpoint" || name=="minCoolingSetpoint" || name=="maxHeatingSetpoint" || name=="minHeatingSetpoint") {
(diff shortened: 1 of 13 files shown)


mrnohr:
Do we need to make the changes for heatingSetpoint and coolingSetpoint? We only look for temperature events on the platform.

mrobinet:
Yeah, perhaps something more generic and not based on attribute name at all. 

`unit == CELSIUS || unit == FAHRENHEIT`

mrnohr:
👍

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -235,8 +235,8 @@ void setHeatingSetpoint(setpoint) {
 
 	def sendHoldType = holdType ? (holdType=="Temporary")? "nextTransition" : (holdType=="Permanent")? "indefinite" : "indefinite" : "indefinite"
 	if (parent.setHold(heatingValue, coolingValue, deviceId, sendHoldType)) {
-		sendEvent("name":"heatingSetpoint", "value":heatingSetpoint)
-		sendEvent("name":"coolingSetpoint", "value":coolingSetpoint)
+		sendEvent("name":"heatingSetpoint", "value":heatingSetpoint, "unit":location.temperatureScale)
+		sendEvent("name":"coolingSetpoint", "value":coolingSetpoint, "unit":location.temperatureScale)
```
workingmonk, line 239: why not sure `"unit": temperatureScale` for consistency? or is there a different reason for using location.temperatureScale?

mrobinet: I did it mostly because it was consistent with the rest of the references here. Look at lines 233 and 234 just above this.

PR #1164, opened 2016-08-29
DVCSMP-1954 Routines turning Bose speaker on/off, toggles the power

marstorp:
DVCSMP-1954 Routines turning Bose speaker on/off, toggles the power state to off/on


workingmonk:
👍

PR #1166, opened 2016-08-29
Device watch test

larsfinander:


larsfinander:
@workingmonk

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -816,7 +817,7 @@ private sendBasicEvents(device, param, value) {
 
 	switch (param) {
 		case "on":
-			device.sendEvent(name: "switch", value: (value == true) ? "on" : "off")
+			device.sendEvent(name: "switch", value: (value == true) ? "on" : "off", isStateChange: true)
```
workingmonk, line 820: any reason for forcing the isStateChange?

larsfinander: If I dont, I dont think lastactivity will update since cloud will not send event if state is the same

workingmonk: Yes, but forcing the isStateChange value will cause smartapps to fire repeatedly if they are listening to the on or off event on the bulb

larsfinander:
Replaced

PR #1169, opened 2016-08-29
SSVD-2613 UX for Hue when adding lights is confusing

larsfinander:
-Split lights list in SmartApp into already added lights and new lights

Problem was that user would expect devices to be removed if you uncheck them in the list of light but they arent.

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -133,14 +133,23 @@ def bulbDiscovery() {
 	state.inBulbDiscovery = true
 	def bridge = null
 	if (selectedHue) {
-        bridge = getChildDevice(selectedHue)
-        subscribe(bridge, "bulbList", bulbListData)
+		bridge = getChildDevice(selectedHue)
+		subscribe(bridge, "bulbList", bulbListData)
 	}
-    state.bridgeRefreshCount = 0
-	def bulboptions = bulbsDiscovered() ?: [:]
-	def numFound = bulboptions.size() ?: 0
-    if (numFound == 0)
-    	app.updateSetting("selectedBulbs", "")
+	state.bridgeRefreshCount = 0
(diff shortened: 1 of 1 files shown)


larsfinander:
@juano2310 @workingmonk Does one of you have time for another VD code review?

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -133,29 +133,49 @@ def bulbDiscovery() {
...
+				 existingLightsDescription += ", ${it.value}"
+			}
+		}
+	}
 
 	return dynamicPage(name:"bulbDiscovery", title:"Light Discovery Started!", nextPage:"", refreshInterval:refreshInterval, install:true, uninstall: true) {
 		section("Please wait while we discover your Hue Lights. Discovery can take five minutes or more, so sit back and relax! Select your device below once discovered.") {
-			input "selectedBulbs", "enum", required:false, title:"Select Hue Lights (${numFound} found)", multiple:true, options:bulboptions
+			input "selectedBulbs", "enum", required:false, title:"Select Hue Lights to add (${numFound} found)", multiple:true, options:newLights
+			paragraph title: "Previously added Hue Lights (${existingLights.size()} added)", existingLightsDescription
```
workingmonk, line 175: did you want the comma between the title and the object?

larsfinander: Yes, it does not work properly without it!

workingmonk: `paragraph title: "Previously added Hue Lights (${existingLights.size()} added) $existingLightsDescription"`
^^ this won't work?

larsfinander: If done like it is now, the description is black and listed devices are in grey which i think is good

workingmonk:
👍

PR #1174, opened 2016-08-30
SSVD-2632 The Device Details Recently tab does not log any 

workingmonk:
activity regarding OSRAM Lightify RGWB bulb color change

cc @larsfinander making this similar to what hue does


larsfinander:
👍

PR #1175, opened 2016-08-31
WWST-40 Philips Hue: Implement device watch

larsfinander:
@workingmonk @mortent Can you guys have a look? This is a first version of device watch. Will need to update the other DTHs to just like bulb but please have  a look.

### DIFF
--- a/devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
+++ b/devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
@@ -17,6 +17,7 @@ metadata {
 		capability "Switch"
 		capability "Refresh"
 		capability "Sensor"
+		capability "Health Check"
 
 		command "setAdjustedColor"
         command "reset"
@@ -167,6 +168,10 @@ void setColorTemperature(value) {
     }
 }
 
+void installed() {
+	// setup device watch
+	sendEvent(name: "checkInterval", value: 60 * 30, data: [protocol: "lan"], displayed: false)
+}
 void refresh() {
     log.debug "Executing 'refresh'"
     parent.manualRefresh()
@@ -181,4 +186,8 @@ def verifyPercent(percent) {
         log.warn "$percent is not 0-100"
         return false
     }
+}
+
+def ping() {
+    log.trace "${parent.ping(this)}"
 }
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -95,8 +95,7 @@ def bridgeDiscoveryFailed() {
 	}
 }
 
-def bridgeLinking()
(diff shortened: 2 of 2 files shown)


devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
```diff
@@ -167,6 +168,10 @@ void setColorTemperature(value) {
     }
 }
 
+void installed() {
+	// setup device watch
+	sendEvent(name: "checkInterval", value: 60 * 6, displayed: false)
```
mortent: You should add `data: [protocol: "lan"]` to this event. It tells device-watch that this device is lan connected, so it will not be marked offline whenever the TV extend is unplugged.

mortent: Also, what is the thinking behind the 6 minute interval here?

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -931,23 +932,31 @@ private handlePoll(body) {
...
 
 	def bulbs = getChildDevices()
 	for (bulb in body) {
 		def device = bulbs.find{it.deviceNetworkId == "${app.id}/${bulb.key}"}
 		if (device) {
 			if (bulb.value.state?.reachable) {
+				if (state.bulbs[bulb.key]?.online == false) {
+					// light just came back online, notify device watch
+					def lastActivity = now()
+					sendEvent(name: "deviceWatch-lastActivity", value: lastActivity, description: "Last Activity is on ${new Date((long) lastActivity)}", displayed: false, isStateChange: true)
```
mortent: I would prefer if we found another name and value for this event. This combination is something we chose to signal that a device/dth was pinged by device-watch when it actually did have activity inside the reporting interval. So part of the workaround for the isStateChange events on Kafka problem.

mortent:
I think this should work and it is probably the correct solution right now. We should think about how device-watch should handle devices like this where the DTH knows when a device goes offline. Philips Hue is a little different than the other devices that have used the health check capability.

devicetypes/smartthings/hue-bulb.src/hue-bulb.groovy
workingmonk: for now the agreed interval for switches is 30 minutes. is the 6 minutes for testing?

larsfinander: oops, 6 min was indeed for testing, will change to 30.

larsfinander: add data: to checkInterval

larsfinander:
Replaced by https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/1181

PR #1176, opened 2016-08-31
DVCSMP-1999: Use proper IAS Zone alarm for triggering motion

varzac:
Previously we were incorrectly looking at the alarm1 bit when we should
be looking at the alarm2 bit.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-1999

@tpmanley @workingmonk


tpmanley:
👍

tpmanley:
@varzac put the Jira in the title of the PR too

workingmonk:
If this needs to go out next week, point the PR to staging

varzac:
@workingmonk is next week the earliest this could go out?  Anyone using that DTH doesn't have working motion sensors.

workingmonk:
In that case, this is a hotfix. Point the PR towards the production branch and we can release it after qa approval.

PR #1179, opened 2016-08-31
DVCSMP-1999 Use proper IAS Zone alarm for triggering motion

varzac:
Previously we were incorrectly looking at the alarm1 bit when we should
be looking at the alarm2 bit.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-1999

@tpmanley @workingmonk


tpmanley:
:+1:
Put the Jira in the PR title too

workingmonk:
👍

PR #1180, opened 2016-08-31
[DVCSMP-1979] Use async http for polling and refresh tokens.

jimmyjames:
This change makes certain HTTP requests asynchronous to reduce SmartApp execution time spent waiting for HTTP responses. The following requests are made to be asynchronous:
- Scheduled polling (polling happens every 5 minutes)
- Refresh token requests (Ecobee tokens last one hour)

The following HTTP requests have NOT been made asynchronous:
- OAuth request during install. This cannot be async since response is used to drive dynamic pages flow.
- Initial poll. Changing this to async would change the current behavior (currently if initial poll fails, it will cause initialize() to fail. Perhaps this can be changed, but left it as-is currently since it preserves existing behavior, and really there's not much of a performance gain since it is a low-frequency occurence)
- Command execution. Changing command execution will require a larger refactoring and has impacts not yet fully understood. Will do this in a later PR if possible.

This change depends on SHARD-33 and SAPLAT-7.

@dsainteclaire @DougSven

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -20,6 +20,8 @@
  *      JLH - 02-15-2014 - Fuller use of ecobee API
  *      10-28-2015 DVCSMP-604 - accessory sensor, DVCSMP-1174, DVCSMP-1111 - not respond to routines
  */
+include 'asynchttp_v1'
+
 definition(
 		name: "Ecobee (Connect)",
 		namespace: "smartthings",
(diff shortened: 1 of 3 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -826,6 +943,11 @@ def sendPushAndFeeds(notificationMessage) {
  * Stores data about the thermostats in atomicState.
  * @param thermostats - a list of thermostats as returned from the Ecobee API
  */
+ /**
+ * Stores data about the thermostats in atomicState.
+ * @param thermostats - a list of thermostats as returned from the Ecobee API
+ */
```
DougSven: Extra comment here.

DougSven:
👍 Looks good. Glad to see you could at reuse a lot of code between the async/sync processes.

PR #1181, opened 2016-08-31
WWST-40 Philips Hue: Implement device watch

larsfinander:
@workingmonk @mortent  switched branch to staging and added remaining devices

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -95,8 +95,7 @@ def bridgeDiscoveryFailed() {
 	}
 }
 
-def bridgeLinking()
-{
+def bridgeLinking() {
 	int linkRefreshcount = !state.linkRefreshcount ? 0 : state.linkRefreshcount as int
 	state.linkRefreshcount = linkRefreshcount + 1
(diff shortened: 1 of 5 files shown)


smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -879,36 +877,40 @@ private handleCommandResponse(body) {
...
 	updates.each { childDeviceNetworkId, params ->
 		def device = getChildDevice(childDeviceNetworkId)
-		sendBasicEvents(device, "on", params.on)
-		sendBasicEvents(device, "bri", params.bri)
-		sendColorEvents(device, params.xy, params.hue, params.sat, params.ct)
-                    }
+		def id = getId(device)
+		// If device is offline, then don't send events which will update device watch
+		if (isOnline(id)) {
+			sendBasicEvents(device, "on", params.on)
```
mortent, line 907: So would these events be sent on some kind of schedule as long as a device remain online?

larsfinander: no those events here are only sent as a response to the user requesting a state change. I.e. user tries to turn bulb on. There are no regular scheduled events.

mortent:
👍

PR #1182, opened 2016-08-31
SSVD-2631 Hue color change events twice

larsfinander:
-Only show one color change event to user when
both hue and sat changes


larsfinander:
@workingmonk In case they want to hotfix this

workingmonk:
👍 
do you want to deploy this?

PR #1193, opened 2016-09-06
CREX-5826 - 2 SMS bug fix

juano2310:


smartapps/smartthings/notify-me-when.src/notify-me-when.groovy
```diff
@@ -111,16 +111,22 @@ private sendMessage(evt) {
...
+				options.method = 'both'
+			else {
+				log.debug 'Sending SMS'
+				options.method = 'phone'
+			}
+		else if (pushAndPhone != 'No') {
+			log.debug 'Sending push'
+			options.method = 'push'
+		} else {
+			log.debug 'Sending nothing'
```
bflorian, line 127: I think you may have to set `options.method = 'none'` here or you'll still get a push notification

bflorian:
👍

PR #1194, opened 2016-09-06
revert changes for DVCSMP-1980 and SSVD-2534 on staging

jimmyjames:
reverting changes on staging for the Ecobee DTH and Connect app until QA can complete their regression testing.

@workingmonk @mrobinet


workingmonk:
@mrobinet can you approve the changes?

mrobinet:
I'm fine with the temp changes being reverted, so long as they get put back in for the next release. I don't believe Ecobee is one of the supported devices for TV, but the temp unit change should really go in as the bug is not specific to TV hubs.

PR #1211, opened 2016-09-07
DVCSMP-1959 removed log messages from smartapp that may print sensitive information

dsainteclaire:
See title.
@juano2310 can you review? :)


juano2310:
👍

PR #1217, opened 2016-09-07
DVCSMP-1959: Replace log logging personal phone number

AndrewBresee:
"$phone1" contains personal phone number. This PR removes this log and replaces it with a similar one that does not contain personal information. 
@juano2310

### DIFF
--- a/smartapps/smartthings/text-me-when-it-opens.src/text-me-when-it-opens.groovy
+++ b/smartapps/smartthings/text-me-when-it-opens.src/text-me-when-it-opens.groovy
@@ -48,7 +48,8 @@ def updated()
 
 def contactOpenHandler(evt) {
 	log.trace "$evt.value: $evt, $settings"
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/text-me-when-it-opens.src/text-me-when-it-opens.groovy
```diff
@@ -48,7 +48,8 @@ def updated()
 
 def contactOpenHandler(evt) {
 	log.trace "$evt.value: $evt, $settings"
-	log.debug "$contact1 was opened, texting $phone1"
+  log.debug "$contact1 was opened, sending text"
+	// log.debug "$contact1 was opened, texting $phone1"
```
juano2310: I would remove this commented line. We can leave the corrected one only :)

juano2310:
:+1:

PR #1221, opened 2016-09-08
DEVTOOLS-748 Configure gradle to compile devicetypes and smartapps.

jterhune:
Looking for comments on this. One criticism is that the dependencies will not be kept in sync with data management, but this seems better than nothing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ buildscript {
                 username smartThingsArtifactoryUserName
                 password smartThingsArtifactoryPassword
             }
-            url "http://artifactory.smartthings.com/libs-release-local"
+            url "https://artifactory.smartthings.com/libs-release-local"
         }
     }
 }
@@ -27,9 +27,37 @@ buildscript {
 repositories {
     mavenLocal()
     jcenter()
+    maven {
+        credentials {
+            username smartThingsArtifactoryUserName
+            password smartThingsArtifactoryPassword
+        }
+        url "https://artifactory.smartthings.com/libs-release-local"
+    }
+}
+
+sourceSets {
+    devicetypes {
+        groovy {
+            srcDirs = ['devicetypes']
+        }
+    }
+    smartapps {
+        groovy {
+            srcDirs = ['smartapps']
+        }
+    }
 }
 
 dependencies {
+    devicetypesCompile 'org.codehaus.groovy:groovy-all:2.4.7'
+    devicetypesCompile 'smartthings:appengine-z-wave:0.1.2'
+    devicetypesCompile 'smartthings:appengine-zigbee:0.1.11'
+    smartappsCompile 'org.codehaus.groovy:groovy-all:2.4.7'
+    smartappsCompile 'smartthings:appengine-common:0.1.8'
+    smartappsCompile 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
+    smartappsCompile 'org.grails:grails-web:2.3.11'
+    smartappsCompile 'org.json:json:20140107'
 }
 
 slackSendMessage {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -27,9 +27,37 @@ buildscript {
...
 
 dependencies {
+    devicetypesCompile 'org.codehaus.groovy:groovy-all:2.4.7'
+    devicetypesCompile 'smartthings:appengine-z-wave:0.1.2'
+    devicetypesCompile 'smartthings:appengine-zigbee:0.1.11'
+    smartappsCompile 'org.codehaus.groovy:groovy-all:2.4.7'
+    smartappsCompile 'smartthings:appengine-common:0.1.8'
+    smartappsCompile 'org.codehaus.groovy.modules.http-builder:http-builder:0.7.1'
+    smartappsCompile 'org.grails:grails-web:2.3.11'
+    smartappsCompile 'org.json:json:20160810'
```
sjurgemeyer: Datamgmt seems to be using version 20140107 of this library.

sjurgemeyer:
👍 This seems like a worthwhile change to me.  It won't be able to catch a lot of things due to the dynamic nature of the smartapps, but it'll at least catch some of the basic stuff.

build.gradle
```diff
@@ -27,9 +27,37 @@ buildscript {
 repositories {
     mavenLocal()
     jcenter()
+    maven {
+        credentials {
+            username smartThingsArtifactoryUserName
+            password smartThingsArtifactoryPassword
+        }
+        url "http://artifactory.smartthings.com/libs-release-local"
```
lbredeso: Should be https

robzienert: I'd like to not add more builds and artifacts into the deprecated Artifactory instance. In my opinion, migrating artifacts away from this instance is a mandatory prerequisite to this PR.

smartapps/smartthings/energy-alerts.src/energy-alerts.groovy
```diff
@@ -64,7 +64,7 @@ def meterHandler(evt) {
     def lastValue = atomicState.lastValue as double
     atomicState.lastValue = meterValue
 
-    def dUnit ? evt.unit : "Watts"
+    def dUnit = evt.unit ?: "Watts"
```
robzienert, line 67: Pretty sure we should have this as a separate PR.

jterhune: The code doesn't compile, so it would have to be merged before the compile checking. Why do you want to separate it?

build.gradle
jterhune: I know how stuff like that tends to slip, but I don't want to shave the yak too far on this one. We currently aren't even _compiling_ this code before we try to integrate it.

jterhune:
I'm going to merge this today. I addressed Shaun and Luke's comments, and I've pushed PRs for getting the dependencies into new artifactory, but doing a release for those deps has to wait for other project schedules, etc. I'll remind myself to circle back, but feel free to nag me in a week. :D

smartapps/smartthings/energy-alerts.src/energy-alerts.groovy
workingmonk: FYI, this was fixed late last night by Juan in staging.

workingmonk:
@jterhune This change will be in staging by EOD today and will be in prod by next Tuesday. Also please add the ticket number to this PR. No other way to keep track of the changes for next week during deploy

jterhune:
@workingmonk I need to create a ticket -- what JIRA project should I use? Also, I want to be sure you're aware this only changed the build process, it doesn't affect what gets deployed to staging/prod.

workingmonk:
i thought the build was being used for deployments. Are you using it for something else? The ticket number should what the driver for the change is. Usually it is a DEVTOOLS ticket to improve the deployment, but if this is for something else, isn't there on what is driving this

jterhune:
The `deployments` section of circle.yml deals with deployments, but this PR left that section alone... it just added some additional validation steps before doing a deployment. I've created DEVTOOLS-748 for this work.

PR #1233, opened 2016-09-09
[CHF-201] Removing DTH workaround now that all events go to Kafka

jackchi:
One against staging

@mortent @workingmonk


mortent:
👍

PR #1245, opened 2016-09-13
[CHF-353] Cree Bulb polling fix; reads status every 5 minutes

jackchi:
Sets polling to 5 min for Cree Bulbs

https://smartthings.atlassian.net/browse/CHF-353

@mortent @workingmonk


workingmonk:
Do you want to try with cmds.each as Bob mentioned?
Also, do we want to remove the poll method and capability?

jackchi:
Changed to `each` and removed `polling` capability

workingmonk:
👍

PR #1250, opened 2016-09-15
Update README.md

aaron-smartthings:


aaron-smartthings (commented):
fixed branding

aaron-smartthings (commented):
added older compatible sensor versions

PR #1252, opened 2016-09-15
Update README.md

aaron-smartthings:


aaron-smartthings (commented):
updated copy to reflect compatible devices

PR #1257, opened 2016-09-15
MSA-1474: Danfoss DTH as discussed with Tim

SmartThings:
Submitting 'Danfoss DTH as discussed with Tim' for publication

### DIFF
--- a/devicetypes/richdp/danfoss-lc13-trv-v0-7.src/danfoss-lc13-trv-v0-7.groovy
+++ b/devicetypes/richdp/danfoss-lc13-trv-v0-7.src/danfoss-lc13-trv-v0-7.groovy
@@ -1,0 +1,284 @@
+/**
+ *  Copyright 2015 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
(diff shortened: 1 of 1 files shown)


devicetypes/richdp/danfoss-lc13-trv-v0-7.src/danfoss-lc13-trv-v0-7.groovy
```diff
@@ -0,0 +1,284 @@
...
+		
+		attribute "thermostatFanState", "string"
+
+		command "switchMode"
+		command "switchFanMode"
+        command "quickSetCool"
+        command "quickSetHeat"
+
+		fingerprint deviceId: "0x0804"
+		fingerprint inClusters: "0x80, 0x46, 0x81, 0x72, 0x8F, 0x75, 0x43, 0x86, 0x84"
```
dkirker, line 35: These should be combined on one line to avoid the risk of other devices matching the fingerprint. Like:

fingerprint deviceId: "0x0804", inClusters: "0x80, 0x46, 0x81, 0x72, 0x8F, 0x75, 0x43, 0x86, 0x84"

There is a new format that is used, but I am not sure that we have documentation yet.

The format, for z-wave devices based on manufacturer specific info, is:

fingerprint mfr:"0109", prod:"2002", model:"0205"

dkirker (changes requested):
The fingerprint will need to be fixed.

PR #1260, opened 2016-09-16
SSVD-2736 Philips Hue: Color Coordinator does not work

larsfinander:
-SSVD-2631 Double color events
-SSVD-2601 Color picker control does not show the current color
-Changed color model for Philips Hue to use hue/sat instead of x/y
-Added color events in hex
-Added HSV color conversion algorithms


larsfinander:
@workingmonk Same PR as I had open to production before (it was approved by juan), lets get it into staging and merge on tuesday like usual.

workingmonk:
👍

PR #1263, opened 2016-09-17
DVCSMP-2057 Philips Hue: Correct incorrect bridge mac

larsfinander:
@tpmanley @workingmonk @juano2310 Please review. This is to handle current issue where some customers have an incorrect mac address set for their hue bridge (we still dont know why this happens)

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -333,9 +333,9 @@ def bulbListHandler(hub, data = "") {
     def bridge = null
 	if (selectedHue) {
 		bridge = getChildDevice(selectedHue)
+		bridge?.sendEvent(name: "bulbList", value: hub, data: bulbs, isStateChange: true, displayed: false)
 	}
-    bridge.sendEvent(name: "bulbList", value: hub, data: bulbs, isStateChange: true, displayed: false)
-    msg = "${bulbs.size()} bulbs found. ${bulbs}"
+	msg = "${bulbs.size()} bulbs found. ${bulbs}"
 	return msg
 }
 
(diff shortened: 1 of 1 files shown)


tpmanley:
Looks good to me

👍

larsfinander:
I just realized that if user has multiple bridges the selectedhue will keep shifting, will have to thing about that for a sec.

larsfinander:
Updated the code, now it only changes selectedHue if the current selectedHue is invalid. If the user had multiple bridges, my previous code would keep switching between them every time SSDP updates came in.

larsfinander:
@tpmanley @workingmonk @juano2310 Please re-review. I think the code works as intended now

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -490,24 +490,25 @@ def ssdpBridgeHandler(evt) {
...
-					if (newDNI != it.deviceNetworkId) {
-						def oldDNI = it.deviceNetworkId
-						log.debug "updating dni for device ${it} with $newDNI - previous DNI = ${it.deviceNetworkId}"
-						it.setDeviceNetworkId("${newDNI}")
-						if (oldDNI == selectedHue) {
-							app.updateSetting("selectedHue", newDNI)
-						}
-						doDeviceSync()
-					}
+			// There might be a mismatch between bridge DNI and the actual bridge mac address, correct that
```
tpmanley, line 498: Will this handle the case where we are connected to bridge A, and then bridge B is added to the network? It looks this might switch over to Bridge B's MAC address

juano2310:
LGTM 👍

tpmanley (changes requested):

juano2310:
Right now we don't support multiple bridges. Ideally we would store this data in the DTH so we can have multiple
.

smartapps/smartthings/hue-connect.src/hue-connect.groovy
juano2310: Right now we don't support multiple bridges. Ideally we would store this data in the DTH so we can have multiple, right @larsfinander?

tpmanley: I think you're right, but we should make sure that a second bridge wouldn't screw up the first one. Maybe we should be dropping any SSDP messages if the USN doesn't match the one already in the bridge list. Also should we be checking the model name to be equal to "Philips hue bridge" to make sure it's not some other device that uses the same URN?

larsfinander: THe intent of this line is to cover that case, def isSelectedValid = bridges?.find {it.value?.mac == selectedHue}
IF selectedHue is a valid bridge, we wont update it.

larsfinander: @juano2310 Right, think Vinay changed that in his superconnect app

workingmonk: yup. the superconnect supports multiple bridges

tpmanley (approved):

PR #1264, opened 2016-09-17
PROB-1373: hue: fix HTTP request headers

posborne:
Previously, the whitespace characters from the file were used for
newlines in HTTP headers.  In order for the HTTP headers sent to
the hue to be valid, line separators must always be \r\n.  Oddly, the
hue seemed to accept and respond to requests with the invalid header
that was being sent but it would cause increased latency for all
other API clients.

In addition to the missing carriage returns, the GET request was also
missing the required blank line which marks the end of the request
headers.

https://smartthings.atlassian.net/browse/PROB-1366
http://status.smartthings.com/incidents/13j8g8g2w7ly
https://community.smartthings.com/t/new-hue-delay/57569

@larsfinander @tpmanley @juano2310

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -1175,9 +1175,8 @@ private poll() {
 	def host = getBridgeIP()
 	def uri = "/api/${state.username}/lights/"
 	log.debug "GET: $host$uri"
-	sendHubCommand(new physicalgraph.device.HubAction("""GET ${uri} HTTP/1.1
-HOST: ${host}
-""", physicalgraph.device.Protocol.LAN, selectedHue))
+	sendHubCommand(new physicalgraph.device.HubAction("GET ${uri} HTTP/1.1\r\n" +
+		"HOST: ${host}\r\n\r\n", physicalgraph.device.Protocol.LAN, selectedHue))
 }
 
 private isOnline(id) {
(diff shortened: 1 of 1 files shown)


varzac (approved):

tpmanley:
+1

dburman:
+1

lbredeso (commented):
Though this apparently fixes the issue w/ the new Hue bridge firmware, do we know if previous Hue bridge firmware is ok with the additional carriage returns?

posborne:
> Though this apparently fixes the issue w/ the new Hue bridge firmware, do we know if previous Hue bridge firmware is ok with the additional carriage returns?

We don't for certain.  Requests from, e.g., the hue app do include carriage returns (as should be the case).  I do not have any bridges with the old firmware at this point to test with.

rappleg:
👍

smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -1182,9 +1182,8 @@ private poll() {
 	def host = getBridgeIP()
 	def uri = "/api/${state.username}/lights/"
 	log.debug "GET: $host$uri"
-	sendHubCommand(new physicalgraph.device.HubAction("""GET ${uri} HTTP/1.1
-HOST: ${host}
-""", physicalgraph.device.Protocol.LAN, selectedHue))
+	sendHubCommand(new physicalgraph.device.HubAction("GET ${uri} HTTP/1.1\r\n" +
```
tpmanley, line 1178: You may want to use the alternate signature for `HubAction` that takes a Map instead of a string so you don't have to explicitly write the newlines. See `discoverHueBulbs` for a function that does basically the same thing as `poll`.

posborne:
As per direction from @tpmanley I have move this commit to be based off and into the production branch.

smartapps/smartthings/hue-connect.src/hue-connect.groovy
posborne: I think I'll leave this commit as-is for now as it has been tested.  I agree that using the alternate signature likely makes sense as a change to put in place down the road.

larsfinander:
👍

workingmonk:
👍  to merge to production. deploying this now as hotfix.

PR #1266, opened 2016-09-19
Revert "[DVCSMP-1979] Use async http for polling and refresh tokens."

jimmyjames:
This reverts commit 826993cc4528223486ad73b92faca82c1177f4ff.

Reverting async http changes to Ecobee to allow more time for testing.

@workingmonk I targeted this against staging, feel free to merge if it looks good.

/cc @dsainteclaire this will also allow you some more time to review the changes in-depth.


workingmonk:
👍

PR #1279, opened 2016-09-21
DVCSMP-2076 - Async Update

juano2310:


smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -349,67 +352,49 @@ def uninstalled() {
...
+      headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ]
+    ]
+    asynchttp_v1.get('responseGoals', params, childDevice)
+
+		def params = [
+			uri: 'https://jawbone.com',
+			path: '/nudge/api/users/@me/moves',
+			headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ]
+		]
+		asynchttp_v1.get('responseMoves', params, childDevice)
```
jimmyjames: Is this second http request dependent on the first in any way? Since the request is made asynchronously, the second request may be made before the response comes back from the first. If the second request should only be made when the first request completes, consider moving the second request in to the response handler of the first.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -349,67 +352,49 @@ def uninstalled() {
...
```
jimmyjames: Same question here regarding the second request being dependent on the first

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -349,67 +352,49 @@ def uninstalled() {
...
-    }
-    catch (e) {
-            // eat it
-    }
+		def params = [
+      uri: 'https://jawbone.com',
+      path: '/nudge/api/users/@me/goals',
+      headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ]
+    ]
+    asynchttp_v1.get('responseGoals', params, childDevice)
```
jimmyjames: It looks like `generateInitialEvent()` is called as part of the setup process. Since the requests are async and the response won't be processed synchronously, are we sure that anything that happens after calling `generateInitialEvent()` is OK to happen before this response is processed? Otherwise the setup will continue even though this request has not been processed.

jimmyjames (commented):
I don't know the details of the Jawbone integration, but I do have some general comments/questions related to changing some requests to be async, just to ensure we've thought through how the behavior may be different.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310: No dependency :)

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310: No dependency :)

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310: Yes, no rush at all, basically populating the initial set of values.

juano2310:
Non of those request have any specific urgency nor dependency so I think they will be fine. I have a question can I pass "childDevice" to the responseHandler like I did? @jimmyjames @DougSven

jimmyjames:
I doubt it will work to pass in the childDevice to the response handler. Since the data is serialized to JSON, I'm pretty sure it works just like how state values are [serialized](http://docs.smartthings.com/en/latest/smartapp-developers-guide/state.html#what-can-be-stored-in-state-and-atomic-state). Did you try it?

juano2310:
Yes, I just did and it didn't work :( How can I pass "childDevice" then?

jimmyjames:
For one thing, the data passed to the handler should be a map. That won't solve the issue of passing a child device object, but it's still worth noting.

The short answer is I don't think you can pass the child device to the response handler method. The data is intended to be a simple map of JSON-serializable values. You could pass an ID or something that could be used in the response handler method to get the child device, if that is possible. If not, I think you'd need to consider further refactoring. @DougSven may have other insights or ideas.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
jimmyjames: before being async would a failure response cause the install to fail? And if it did, is that desired functionality? If not, or if we don't care, then I think this would be fine to make async.

juano2310: No, this is just to update values

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -349,67 +353,39 @@ def uninstalled() {
...
- 	def sleeps = null
-    httpGet(uri: urlgoals, headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ]) {response ->
-        goals = response.data.data
-    }
-    httpGet(uri: urlmoves, headers: ["Authorization": "Bearer ${state.JawboneAccessToken}" ]) {response ->
-        moves = response.data.data.items[0]
-    }
+def responseGoals(response, dni) {
+		def childDevice = getChildDevice(dni.value)
+		def goals = response.json.data
```
jimmyjames: if the response is not successful (non 2xx) or has some other error, or if the json cannot be parsed from the response, a runtime exception would be thrown here. Not sure if you want to handle those cases and log them or try to recover somehow; if not and you are fine with a runtime exception here, then looks good.

jimmyjames (commented):
Just a few more comments/questions, but otherwise looks good provided it's been tested and working.

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
juano2310: How about now?

smartapps/juano2310/jawbone-up-connect.src/jawbone-up-connect.groovy
```diff
@@ -349,67 +353,67 @@ def uninstalled() {
...
+			log.error "response has error: $response.errorMessage"
+	} else {
+			def moves
+			try {
+					// json response already parsed into JSONElement object
+					moves = response.json.data.items[0]
+			} catch (e) {
+					log.error "error parsing json from response: $e"
+			}
+			if (goals) {
```
jimmyjames: guessing `goals` should be `moves` here

jimmyjames (changes requested):
One little copy/paste error to clean up then this looks good. thanks Juan!

jimmyjames:
👍

PR #1284, opened 2016-09-22
SSVD-2798 Philips Hue: Bridge keeps getting unchecked during discovery

larsfinander:
Additional fix, my original fix only covered bulbs but apparently issue affected bridge input as well.


juano2310:
👍

PR #1292, opened 2016-09-26
CHF-406

pchomal:
Added HealthCheck Implementation for Z-wave Dimmer (GE 12718, GE 12724, GE 12729) with checkInterval of 16min

### DIFF
--- a/devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
+++ b/devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
@@ -20,6 +20,7 @@ metadata {
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
+		capability "Health Check"
 
 		fingerprint mfr:"0063", prod:"4457", deviceJoinName: "Z-Wave Wall Dimmer"
 		fingerprint mfr:"0063", prod:"4944", deviceJoinName: "Z-Wave Wall Dimmer"
@@ -82,6 +83,8 @@ metadata {
 }
 
 def updated(){
+	// Device-Watch simply pings if no device events received for 32min(checkInterval)
+	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
   switch (ledIndicator) {
         case "on":
             indicatorWhenOn()
@@ -213,6 +216,13 @@ def setLevel(value, duration) {
 
 def poll() {
 	zwave.switchMultilevelV1.switchMultilevelGet().format()
+}
+
+/**
+ * PING is used by Device-Watch in attempt to reach the Device
+ * */
+def ping() {
+	refresh()
 }
 
 def refresh() {
--- a/devicetypes/smartthings/dimmer-switch.src/.st-ignore
+++ b/devicetypes/smartthings/dimmer-switch.src/.st-ignore
@@ -1,0 +1,2 @@
+.st-ignore
+README.md
--- a/devicetypes/smartthings/dimmer-switch.src/README.md
+++ b/devicetypes/smartthings/dimmer-switch.src/README.md
@@ -1,0 +1,45 @@
+# Z-wave Dimmer Switch
+
+
+
+Works with:
+
+* [GE Z-Wave In-Wall Smart Dimmer (GE 12724)](http://products.z-wavealliance.org/products/1197)
+* [GE Z-Wave In-Wall Smart Dimmer (Toggle) (GE 12729)](http://products.z-wavealliance.org/products/1201)
+* [GE Z-Wave Plug-in Smart Dimmer (GE 12718)](http://products.z-wavealliance.org/products/1191)
+
+## Table of contents
+
+* [Capabilities](#capabilities)
+* [Health](#device-health)
+* [Troubleshooting](#Troubleshooting)
+
+## Capabilities
+
+* **Switch Level** - it's defined to accept two parameters, the level and the rate of dimming
+* **Actuator** - represents that a Device has commands
+* **Indicator** - gives you the ability to set the indicator LED light on a Z-Wave switch
+* **Switch** - can detect state (possible values: on/off)
+* **Polling** - represents that poll() can be implemented for the device
+* **Refresh** - _refresh()_ command for status updates
+* **Sensor** - detects sensor events
+* **Health Check** - indicates ability to get device health notifications
+
+## Device Health
+
+Z-Wave Smart Dimmers (In-Wall, In-Wall(Toggle), Plug-In) are polled by the hub.
+As of hubCore version 0.14.38 the hub sends up reports every 15 minutes regardless of whether the state changed.
+Check-in interval = 32 mins.
+Not to mention after going OFFLINE when the device is plugged back in, it might take a considerable amount of time for
+the device to appear as ONLINE again. This is because if this listening device does not respond to two poll requests in a row,
+it is not polled for 5 minutes by the hub. This can delay up the process of being marked ONLINE by quite some time.
+
+## Troubleshooting
+
+If the device doesn't pair when trying from the SmartThings mobile app, it is possible that the device is out of range.
+Pairing needs to be tried again by placing the device closer to the hub.
+Instructions related to pairing, resetting and removing the device from SmartThings can be found in the following link:
+* [General Z-Wave Dimmer/Switch Troubleshooting Tips](https://support.smartthings.com/hc/en-us/articles/200955890-Troubleshooting-GE-in-wall-switch-or-dimmer-won-t-respond-to-commands-or-automations-Z-Wave-)
+* [GE Z-Wave In-Wall Smart Dimmer (GE 12724) Troubleshooting Tips](https://support.smartthings.com/hc/en-us/articles/200902600-GE-In-Wall-Paddle-Dimmer-Switch-GE-12724-Z-Wave-)
+* [GE Z-Wave In-Wall Smart Dimmer (Toggle) (GE 12729) Troubleshooting Tips](https://support.smartthings.com/hc/en-us/articles/207568463-GE-In-Wall-Smart-Toggle-Dimmer-GE-12729-Z-Wave-)
+* [GE Z-Wave Plug-in Smart Dimmer (GE 12718) Troubleshooting Tips](https://support.smartthings.com/hc/en-us/articles/202088474-GE-Plug-In-Smart-Dimmer-GE-12718-Z-Wave-)


pchomal:
@jackchi @ShunmugaSundar 
Please review and merge.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
```diff
@@ -12,15 +12,19 @@
...
 		capability "Actuator"
 		capability "Switch"
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
+		capability "Configuration"
+		capability "Health Check"
 
 		fingerprint inClusters: "0x26", deviceJoinName: "Z-Wave Dimmer"
+		fingerprint mfr:"0063", prod:"4944", deviceJoinName: "Z-Wave In-Wall Dimmer"
```
jackchi: Hi @juano2310 - 

This work is to add Device Health capability and support to Z-Wave dimmers. 

Can you elaborate on the difference between this DTH and `dimmer-switch.groovy`? 
My first impression is for us to change both DTHs. If you can confirm this, we can have Piyush change both `zwave-dimmer-switch-generic.groovy` & `dimmer-switch.groovy`. 

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy#L25

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
```diff
@@ -195,11 +206,24 @@ def refresh() {
...
 }
 
+def healthPoll() {
+	def cmds = refresh()
+	cmds.each{ sendHubCommand(new physicalgraph.device.HubAction(it))}
+}
+
+def configure() {
+	unschedule()
+	schedule("0 0/5 * * * ? *", "healthPoll")
```
jackchi: The DTH scheduling method `runEvery5Minutes()` should be going out to data-mgmt production this Wed. 9/29. Let's test this method in dev now and have one PR to merge rather than 2. 
https://smartthings.atlassian.net/wiki/display/DATAMGMT/2016/09/26/2016-09-29+Release

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
juano2310: The difference is that for known devices we show the real name when it pairs and tiles depending of the device. Yes we should change both :)

jackchi: @pchomal  - We confirmed. The fingerprints you added here is already covered in `dimmer-switch.groovy` DTH. Please `Health Check` that DTH and take out these fingerprints.

jackchi (changes requested):
Remove fingerprints and Health Check `dimmer-switch.groovy` DTH

ShunmugaSundar:
@jackchi 
Incorporated the changes. Please review.

pchomal:
@jackchi @ShunmugaSundar 
Incorporated the changes. Please review.

jackchi (approved):

jackchi:
Everything looks good. As soon as the SchedulingApi `runEvery5Minutes` get into Production. We will push this out. 

👍

jackchi:
@workingmonk @mckeed says you need to get hands on actual devices to test before this can be merged.

pchomal:
@jackchi 
This code has been tested with GE 12724.
So I'll update when it's tested with the other two dimmers.

jackchi:
let's also remove the `Category` from these DTHs.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
```diff
@@ -225,6 +234,19 @@ def refresh() {
 	delayBetween(commands,100)
 }
 
+def healthPoll() {
+	def cmds = refresh()
+	cmds.each{ sendHubCommand(new physicalgraph.device.HubAction(it))}
```
mckeed: This is bad form – when sending multiple commands they must all be sent in the same object, a HubMultiAction, or the delays don't actually do anything.

The correct form would be sendHubCommand(response(refresh())) – response() is a helper that converts commands to HubAction objects. Unfortunately it looks like sendHubCommand only accepts single actions, so that has to be fixed first.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
```diff
@@ -225,6 +234,19 @@ def refresh() {
...
+}
+
+def configure() {
+	unschedule()
+	runEvery5Minutes(healthPoll)
```
mckeed: Z-Wave switches and dimmers are polled as needed from the hub and should **not** be polled regularly from the cloud. Although they have the Polling capability, the polling job is set to exclude them because it causes big problems for large and unhealthy Z-Wave networks. We should be relying on the ping() method to check that these devices are online.

The spec of the hub-based polling system is here: https://smartthings.atlassian.net/wiki/display/HUBPLT/Z-Wave+and+SmartThings#Z-WaveandSmartThings-Hub-BasedSwitchPolling

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
```diff
@@ -225,6 +234,19 @@ def refresh() {
...
+	log.debug "Configuring Reporting and Bindings."
+	// Device-Watch allows 2 check-in misses from device
+	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
```
mckeed: The hub-based polling system sends switch events every 15 minutes in the absence of state changes to let the cloud know that the device is still being polled. The checkInterval should be at least 16 minutes.

The hub also sends an offline alert approx. 5 minutes after a polled switch stops responding, so we could have better responsiveness by making use of those events to mark the device offline.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
```diff
@@ -19,6 +19,8 @@ metadata {
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
+		capability "Configuration"
```
mckeed: Why are we adding this capability instead of using the installed() or updated() callbacks?

mckeed (changes requested):

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
pchomal: Yes, sendHubCommand() only accepts single actions, so that needs to be fixed.
But, it is not required as of now in the current implementation as the healthPoll() method has been removed.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
pchomal: I agree that Z-Wave dimmers are being polled as needed from the hub, so shouldn't be polled from the cloud regularly, that's why I think of using a simplistic approach towards ping().
Hence, removed the scheduling of healthPoll().

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
pchomal: As we had been using the configure() method previously to send the checkInterval Event in the Health Check Implementation of zigbee devices, we followed a similar approach for Z-Wave devices. And as configure() wasn't present, we added it.
In this particular DTH updated() method is present, so it's correct to use it as mentioned by you to send checkInterval Event from here.
But what about the case where neither of installed()/updated() method are present(as in "Z-Wave Dimmer Switch Generic.groovy")?

pchomal:
@mckeed @jackchi @ShunmugaSundar 
Changes have been made to reflect the comments and have been tested with Z-Wave Dimmer(GE 12724).
Please review and merge.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
pchomal: checkInterval changed to 16min.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
mckeed: Actually, I would prefer using updated() in _all_ cases for setting checkInterval, since it is called after changing a device's type (manually via the web interface). At least when it's not tied to specific configuration commands.

cc @jackchi

mckeed (approved):

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
pchomal: @jackchi @juano2310 
When we pair GE12724, the device type shown in staging IDE is **Z-wave Dimmer Switch Generic**.
So which DTH will the device refer to when paired?
Though I've made the changes in **dimmer-switch.groovy**, will the device refer to it?
As the _in-clusters_ is matching in **zwave-dimmer-switch-generic.groovy**
and the _mfr_ & _prod_ is matching in **dimmer-switch.groovy**

mckeed: That device should not pair as Z-Wave Dimmer Switch Generic. Mfr and prod fingerprinting should always take precedence over inClusters. Can you share the account and shardthat this device is on so I can look at why it got the wrong device type?

pchomal: **User** - sundar.rs@samsung.com
**Shard** - https://graph-na01s-useast1.smartthingsgdev.com
I tried pairing it again, the device type is dimmer switch now.

pchomal:
Tested with all 3 devices, but devices are taking time to come back online.
It maybe due to range/queuing issues, not sure though.

jackchi (changes requested):
In keeping consistency with 2x misses, ZWave Switches are hub polled every 15 minutes. Let's make it 2x15min + 2 minutes jitter time. Totaling `32min`. Also spoke with @surfous and this would satisfy the Product delivery of within 35min. 

@pchomal @mckeed

mckeed:
@jackchi this is a different case. The hub polls z-wave switches every 10 seconds - 5 minutes. The 15 minute check-in is just a summary that already makes up for single/double missed responses. For this reason I think we do not need to double it and 16 minutes will be right.

pchomal:
We tested with device moving it closer to the Hub, and now it takes about **11min** to come back online, 2*5 min + 1 min Jitter time.
But as 15min check-in summary makes up for missed responses, so `16min` seems right.

ShunmugaSundar:
@jackchi, Please let us know the decision towards check-in.  I request you to merge this PR if there is no change required,

mckeed:
@pchomal The time it takes for the device to come back online after plugging it in is unrelated to the checkInterval. Once a device is offline, Device Watch just waits for activity from the device.

The Z-Wave polling system slows down polling after a few missed responses because repeatedly trying to contact unresponsive Z-Wave devices can have a detrimental effect on the network.

jackchi (approved):

PR #1299, opened 2016-09-27
DVCSMP-2089 - Harmony - Make HTTP Requests Async

juano2310:


smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
-    } catch(Exception ex) {
-    	msg = ex
-    }
-    runIn(10, "poll", [overwrite: true])
-    return msg
+  def params = [
+      uri: url,
+      contentType: 'application/json'
+  ]
+  asynchttp_v1.post('activityResponse', params, [:])
```
jimmyjames: you can just omit the empty map (it's an empty map by default if not passed)

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
+}
+
+def activityResponse(response, data) {
+	if (response.hasError()) {
+			log.error "Logitech Harmony - response has error: $response.errorMessage"
+      if (response.status == 401) { // token is expired
+          state.remove("HarmonyAccessToken")
+          log.warn "Logitech Harmony - Access token has expired"
+      }
+	} else {
```
jimmyjames, line 581: This whole logic block is hard to read because of the indentation. I know the IDE does horrible things to tabs/spaces, but worth cleaning this up IMO for readability.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
-                                    runIn(5, "poll", [overwrite: true])
-                            }
-                        }
+        def tokenParam = [auth: state.HarmonyAccessToken]
+        def params = [
+            uri: "https://home.myharmony.com/cloudapi/state?${toQueryString(tokenParam)}",
+            headers: ["Accept": "application/json"],
+            contentType: 'application/json'
+        ]
+        asynchttp_v1.get('pollResponse', params, [:])
```
jimmyjames: no need to pass the empty map if not passing data to the response handler

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -380,8 +383,6 @@ def discovery() {
             	log.debug "valid Token"
                 state.Harmonydevices = response.data
                 state.resethub = false
-                getActivityList()
-                poll()
```
jimmyjames, line 384: curious if this is related to async changes at all, or why we don't need to call these methods as part of discovery now?

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
-		} catch(Exception e) {
-        	log.trace e
-		}
+      def tokenParam = [auth: state.HarmonyAccessToken]
+      def params = [
+  				uri: "https://home.myharmony.com/cloudapi/activity/all?${toQueryString(tokenParam)}",
+  				headers: ["Accept": "application/json"],
+  				contentType: 'application/json'
+  		]
+  		asynchttp_v1.get('activityListResponse', params, [:])
```
jimmyjames: no need to pass the empty map

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310: Those actually update a device that already exist. During discovery we still don't have the devices so no point in trying to find data yet.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
+	} else {
+			def ResponseValues
+			try {
+					// json response already parsed into JSONElement object
+					ResponseValues = response.json
+			} catch (e) {
+					log.error "Logitech Harmony - error parsing json from response: $e"
+			}
+			if (ResponseValues) {
+          if (ResponseValues.code == 200) {
```
jimmyjames: The sync response handler checked for 200 or dni == "all". Here we only check the status, not sure if that's an intentional change.

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -430,142 +431,182 @@ def addDevice() {
...
+              	log.trace "Command sent succesfully"
+              	poll()
```
jimmyjames: This just calls poll() when Harmony responds successfully after the activity api call, whereas before we updated something in state. I'm just curious that we are changing that behavior intentionally/for good reason ;)

jimmyjames (commented):
Just a couple little things like formatting and other questions, but the async changes look pretty good assuming it's been tested (especially when tokens expire)

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310: yes, so the activity device type and the harmony hub DTH gets updated with the correct value and status based on real data and not on assuming that the command was successful

jimmyjames:
👍

PR #1311, opened 2016-10-03
CHF-413 Philips Hue bulb shows unavailable after 16minutes and

larsfinander:
CHF-412 Hue Bridge shows OFFLINE instead of "Unavailable"


larsfinander:
@juano2310 review please

juano2310:
It looks good to me :+1:  @larsfinander tested most of the possible scenarios affected by removing bulbListHandler() both in V1 and V2

juano2310:
👍

PR #1317, opened 2016-10-04
[CHF-417] [CHF-416] Health Check devices enroll with appropriate checkInterval

jackchi:
When we enroll existing Health Check devices in our platform, some devices will be offline. Since calling `configure()` will update the periodic reporting intervals for online devices, Device-Watch will have an expectation of how often they are checking in according to the newer value. We want to keep the checkInterval to the older value - `3 * 60 * 60 + 60` seconds until DTH confirms that the newer config is successful. 

By parsing out the response in `parse()` for successful configuration, DTH will update Device-Watch with the correct `checkInterval` value. 
- Also addresses the fix for CHF-416 `zigbee-dimmer` / `zigbee-rgbw-bulb` / `zigbee-white-color-temperature-bulb` on `refresh()` it configures to the new onOffConfig intervals. 

updatedcheckinterval

CHF - @mortent @ShunmugaSundar @parijatdas @skt123 @pchomal @dchem @darksun 
Device Protocols @tpmanley @workingmonk

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ buildscript {
                 username smartThingsArtifactoryUserName
                 password smartThingsArtifactoryPassword
             }
-            url "http://artifactory.smartthings.com/libs-release-local"
+            url "https://artifactory.smartthings.com/libs-release-local"
         }
     }
 }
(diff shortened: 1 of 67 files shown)


workingmonk:
@jackchi you will either have to rebase your branch or cherry-pick your changes to a new branch created off of staging

PR #1320, opened 2016-10-04
[CHF-417] [CHF-416] Health Check devices enroll with right checkInterval

jackchi:
When we enroll existing Health Check devices in our platform, some devices will be offline. Since calling `configure()` will update the periodic reporting intervals for online devices, Device-Watch will have an expectation of how often they are checking in according to the newer value. We want to keep the checkInterval to the older value - `3 * 60 * 60 + 60` seconds until DTH confirms that the newer config is successful. 

By parsing out the response in `parse()` for successful configuration, DTH will update Device-Watch with the correct `checkInterval` value. 
- Also addresses the fix for CHF-416 `zigbee-dimmer` / `zigbee-rgbw-bulb` / `zigbee-white-color-temperature-bulb` on `refresh()` it configures to the new onOffConfig intervals. 

updatedcheckinterval

CHF - @mortent @ShunmugaSundar @parijatdas @skt123 @pchomal @dchem @darksun 
Device Protocols @tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
+++ b/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
@@ -93,7 +93,7 @@ def ping() {
 }
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.onOffConfig() + zigbee.levelConfig()
+    zigbee.onOffRefresh() + zigbee.levelRefresh()
 }
 
 def healthPoll() {
@@ -103,10 +103,8 @@ def healthPoll() {
 
 def configure() {
     unschedule()
-    schedule("0 0/5 * * * ? *", "healthPoll")
-    log.debug "Configuring Reporting and Bindings."
+    runEvery5Minutes("healthPoll")
     // Device-Watch allows 2 check-in misses from device
     sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
-    // minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-    zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + zigbee.onOffRefresh() + zigbee.levelRefresh()
+    zigbee.onOffRefresh() + zigbee.levelRefresh()
 }
--- a/devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
+++ b/devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
@@ -16,7 +16,7 @@
 
 metadata {
 	// Automatically generated. Make future change here.
-	definition (name: "SmartPower Outlet", namespace: "smartthings", author: "SmartThings", category: "C1") {
+	definition (name: "SmartPower Outlet", namespace: "smartthings", author: "SmartThings") {
 		capability "Actuator"
 		capability "Switch"
 		capability "Power Meter"
@@ -104,8 +104,21 @@ def parse(String description) {
 		}
 	}
 	else {
-		log.warn "DID NOT PARSE MESSAGE for description : $description"
-		log.debug zigbee.parseDescriptionAsMap(description)
+		def cluster = zigbee.parse(description)
+
+		if (cluster.clusterId == 0x0006 && cluster.command == 0x07){
+			if (cluster.data[0] == 0x00) {
+				log.debug "ON/OFF REPORTING CONFIG RESPONSE: " + cluster
+				sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+			}
+			else {
+				log.warn "ON/OFF REPORTING CONFIG FAILED- error code:${cluster.data[0]}"
+			}
+		}
+		else {
+			log.warn "DID NOT PARSE MESSAGE for description : $description"
+			log.debug "${cluster}"
+		}
 	}
 }
 
@@ -128,8 +141,10 @@ def refresh() {
 }
 
 def configure() {
-	// Device-Watch allows 2 check-in misses from device
-	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch allows 3 check-in misses from device (plus 1 min lag time)
+	// enrolls with default periodic reporting until newer 5 min interval is confirmed
+	sendEvent(name: "checkInterval", value: 3 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
 	zigbee.onOffConfig(0, 300) + powerConfig() + refresh()
 }
--- a/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
@@ -17,7 +17,7 @@ import physicalgraph.zigbee.clusters.iaszone.ZoneStatus
 
 
 metadata {
-	definition (name: "SmartSense Moisture Sensor",namespace: "smartthings", author: "SmartThings", category: "C2") {
+	definition (name: "SmartSense Moisture Sensor",namespace: "smartthings", author: "SmartThings") {
 		capability "Configuration"
 		capability "Battery"
 		capability "Refresh"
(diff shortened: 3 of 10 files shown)


devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
```diff
@@ -93,7 +93,7 @@ def ping() {
 }
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.onOffConfig() + zigbee.levelConfig()
+    zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.onOffConfig(0, 300) + zigbee.levelConfig()
```
tpmanley: Need the handling of the configure reporting response instead of generating the checkInterval event in configure

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
```diff
@@ -128,8 +136,9 @@ def refresh() {
 }
 
 def configure() {
-	// Device-Watch allows 2 check-in misses from device
-	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch allows 3 check-in misses from device (plus 3 min lag time)
+	sendEvent(name: "checkInterval", value: 3 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee"])
```
tpmanley: I think this should be removed since the event will be created when the configure reporting response is received

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -122,10 +122,17 @@ private Map parseCatchAllMessage(String description) {
 				break
```
tpmanley, line 125: Since command 0x07 isn't ignored anymore you need to make a change to the other case in this switch statement so that it only handles the correct commands. One option is just make sure the command is not 0x07

tpmanley (changes requested):
Made some comments on the first few but the same comments will apply to the rest of the device type handlers

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
jackchi: We discussed this and the plan is to still enroll devices when they are offline. If we don't do this, the stats in Device-Watch will be inaccurate

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
jackchi: We want both.

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
jackchi: Ah gotcha. So put everything in the `else` statement

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -292,8 +298,8 @@ def refresh() {
 }
 
 def configure() {
-	// Device-Watch allows 2 check-in misses from device
-	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch allows 3 check-in misses from device (plus 3 min lag time)
```
parijatdas: Is the lag time kept as 3min or 1min coz in the value parameter of sendEvent it is '1*60'?

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
tpmanley: Is it true that the checkInterval has not changed in relation to a prior reporting configuration like it has with some of the other devices? Just wondering why you're not handling the configure reporting response in this DTH like in the others.

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
tpmanley: I understand. 

The checkInterval calculation doesn't look right. The report config is 5 minutes, so shouldn't it be `3 * 5 * 60 + 3 * 60` in order to match the comment?

tpmanley: Reread the PR description and it looks like the calculation is the older value. So maybe it's just the comment that's wrong.

jackchi: Yeah comment was wrong. I'll update it.

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
jackchi: The `checkInterval` in the first configure() assumes that the Device was not able to be configured (offline). When the 0x07 bind command comes back with success we update `Device-Watch` with the newer `12min` checkInterval value.

jackchi:
@tpmanley PR updated to reflect the comments. Please review!

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
tpmanley: This DTH doesn't have the handling of the 0x07 command yet

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
```diff
@@ -128,8 +136,9 @@ def refresh() {
 }
 
 def configure() {
-	// Device-Watch allows 2 check-in misses from device
-	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch allows 3 check-in misses from device (plus 1 min lag time)
```
tpmanley, line 144: I think it will be more clear to make a comment that the checkInterval is being set to a long interval and will be updated to the correct value when the device responds to the configure reporting command. It's just confusing since 3600 seconds doesn't appear to be the max report interval.

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -118,14 +118,25 @@ private Map parseCatchAllMessage(String description) {
 	if (shouldProcessMessage(cluster)) {
 		switch(cluster.clusterId) {
 			case 0x0001:
-				resultMap = getBatteryResult(cluster.data.last())
+				if (cluster.command == 0x007) {
```
tpmanley: This is minor, but commands are 8-bits long and we prefer to write hex values with the appropriate number of digits for the data type so this should be `0x07`. Also the comment should say this is the Configure Reporting Response, not bind msg.

tpmanley: Or you could just change this to
`if (cluster.command != 0x07) {
     resultMap = getBatteryResult(cluster.data.last())
}`

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
jackchi: Yeah this is intentional. Since the cree-bulb actually doesn't do periodic reporting we will rely on the cloud to schedule a healthPoll()

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
jackchi: I'll change to 8-bit `0x07`. In the original DTH, `//bind msg` was the comment used in `shouldProcessMessage`, but I'll change it to the way you suggested.

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
tpmanley: That's a good point. We probably shouldn't even have the config calls since it doesn't support attribute reporting.

mortent:
Apart from the comments already in here, this looks good to me.

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
jackchi: The `cree-bulb` DTH is updated without `levelConfig` & `onOffConfig` both in configure() and refresh()

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
jackchi: Comments added.

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -118,14 +118,23 @@ private Map parseCatchAllMessage(String description) {
...
+				}
 				break
 
             case 0x0402:
-                // temp is last 2 data values. reverse to swap endian
-                String temp = cluster.data[-2..-1].reverse().collect { cluster.hex1(it) }.join()
-                def value = getTemperature(temp)
-                resultMap = getTemperatureResult(value)
+				if (cluster.command == 0x07 && cluster.data[0] == 0x00) {
+					log.debug "TEMP REPORTING CONFIG RESPONSE" + cluster
```
tpmanley: I didn't notice this before, but there will be a problem if the command is 0x07 and the status is not 0x00. You could move the check of `cluster.data[0]` inside this if statement to fix it.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -122,19 +122,31 @@ private Map parseCatchAllMessage(String description) {
...
+					resultMap = getBatteryResult(cluster.data.last())
+				}
 				break
 
 			case 0x0402:
-				// temp is last 2 data values. reverse to swap endian
-				String temp = cluster.data[-2..-1].reverse().collect { cluster.hex1(it) }.join()
-				def value = getTemperature(temp)
-				resultMap = getTemperatureResult(value)
+				if (cluster.command == 0x07 && cluster.data[0] == 0x00) {
```
tpmanley: Same comment here

tpmanley: There were a few other places too that need to be updated

devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
jackchi: Good catch. Checking for configuration success is inside `0x07` block now. Also logs warnings.

tpmanley (approved):
Looks good. Would probably be a good idea to log the status if the configure reporting fails

PR #1321, opened 2016-10-05
DVCSMP-2104 - Harmony - Fix Exceptions

juano2310:
- MissingPropertyException: No such property: code for class: java.lang.String
- NullPointerException: Cannot get property 'activities' on null object
- NullPointerException: Cannot get property 'device' on null object

### DIFF
--- a/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
+++ b/smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
@@ -362,7 +362,7 @@ Map discoverDevices() {
             def hubname = getHubName(it.key)
             def hubvalue = "${hubname}"
             hubs["harmony-${hubkey}"] = hubvalue
-        	it.value.response.data.activities.each {
+        	  it.value.response.data.activities.each {
                 def value = "${it.value.name}"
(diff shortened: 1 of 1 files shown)


juano2310:
@jimmyjames @DougSven please take a look. This should solve most of the biggest exceptions for this Smartapp. Thanks!

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -557,58 +550,6 @@ def pollResponse(response, data) {
 	}
 }
 
-def getActivityList() {
```
DougSven, line 560: It's not necessary to remove this method unless it's no longer needed. The method missing exceptions were related to executing an old version of the installed smart app

juano2310: Is not longer needed. It was updating a list that was not being use.

DougSven: right on

DougSven:
👍 Looks good.

workingmonk:
please add the ticket number to the PR title

PR #1322, opened 2016-10-06
DVCSMP-2104 - Harmony - Fix Exceptions, improve logging

juano2310:
@DougSven @jimmyjames do you mind checking this again. I just added Harmony to the logs for better tracking in sumo and all the new changes now to staging. I also brought getActivitylist() back but I only call it on updated() so I remove async for that one.


DougSven:
Looks good. Yeah makes sense to run activity list synchronously on install and update. 👍

PR #1327, opened 2016-10-07
DVCSMP-2043 OSRAM: Saturation values are not updated in older firmwares

larsfinander:


devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -177,5 +177,5 @@ def setHue(value) {
 
 def setSaturation(value) {
     def scaledSatValue = zigbee.convertToHexString(Math.round(value * 0xfe / 100.0), 2)
-    zigbee.command(COLOR_CONTROL_CLUSTER, SATURATION_COMMAND, scaledSatValue, "0500")      //payload-> sat value, transition time
+    zigbee.command(COLOR_CONTROL_CLUSTER, SATURATION_COMMAND, scaledSatValue, "0500") + "delay 300" + zigbee.readAttribute(0x0300, ATTRIBUTE_SATURATION)
```
tpmanley: The transition time used for the Move to Saturation command is 0x0005 which translates to 1/2s so the delay before reading the Saturation attribute should be at least 500ms.

larsfinander: THanks, done!

larsfinander:
@workingmonk @tpmanley Review please, I was considering trying to add code to detect firmware and only do this for older ones but not sure if it is worth it. My first attempt at reading firmware didnt succeed and I am not sure if it is exposed to DTH.

tpmanley (approved):
For consistency should use `COLOR_CONTROL_CLUSTER` instead of hardcoding `0x0300` but otherwise looks good

PR #1334, opened 2016-10-11
CHF-420 Add device health check to Bose DTH

marstorp:
Added health check to Bose DTH and updated connect app with Bose icons, fix for keeping the selected list during setup (submitOnChange: true), and syncing DTH with device after creation to update DTH tiles.

### DIFF
--- a/devicetypes/smartthings/bose-soundtouch.src/bose-soundtouch.groovy
+++ b/devicetypes/smartthings/bose-soundtouch.src/bose-soundtouch.groovy
@@ -27,7 +27,7 @@ metadata {
         capability "Switch"
         capability "Refresh"
         capability "Music Player"
-        capability "Polling"
+        capability "Health Check"
 
         /**
          * Define all commands, ie, if you have a custom action not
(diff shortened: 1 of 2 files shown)


marstorp:
@juano2310 can you code review this?
Tyler Gustafson have tested and given it an OK

devicetypes/smartthings/bose-soundtouch.src/bose-soundtouch.groovy
```diff
@@ -837,6 +855,10 @@ def boseRefreshNowPlaying(delay=0) {
     return boseGET("/now_playing")
 }
 
+def boseSendGetNowPlaying() {
```
juano2310, line 858: Is this related to Health Check?

juano2310 (approved):

juano2310:
👍

devicetypes/smartthings/bose-soundtouch.src/bose-soundtouch.groovy
marstorp: Yes, it is, it's being setup by the cron job which is polling the device every two minutes, that drives the event that health check is using to see if everything is ok.
If there's events reported in the time configured health check will make a final call to poll() to see if device is around.

PR #1341, opened 2016-10-12
[DPROT-177] Call refresh before configure

varzac:
With the changes in https://smartthings.atlassian.net/browse/DPROT-168
it can take a little longer for the configuration process for a device
to complete.  As a result sensor values may not be available immediately
by calling the refresh commands before the configure commands we can
populate those values sooner.

This resolves: https://smartthings.atlassian.net/browse/DPROT-177

@tpmanley


tpmanley (approved):
Looks good to me. @workingmonk you want to take a quick look too?

workingmonk:
Any reason all the enrollCmds have been taken out from configure?

tpmanley:
He was able to remove it because refresh already calls the enrollCmds

PR #1342, opened 2016-10-12
 DVCSMP-2013 - Temperature Scale Change bug fix

james-smartthings:
Switching between different temperature scales of the hub, via the SmartApp, would cause the SmartApp break. There were several issues that were causing the SmartApp to break:
- When changing the temperature scale, the setpoints did not update to the correct scale unless the user hit refresh. Leaving the app and re-entering the app did not fix this.
- Upon changing the temperature scale from F to C, the setpoints did not convert to the correct scale even with a refresh. eg If the cooling setpoint was 65F and you converted to C the cooling setpoint would be 65C.
- Certain variables were being cast as the wrong type when switching between temperature scales, which caused errors in the SmartApp

This fix will now update the setpoints if a user refreshes the SmartApp, or by raising/lowering the cooling, heating, aux heat setpoints. This fix also eliminates the type errors that occurred when switching between temperature scales.

This fix does not automatically refreshes the setpoints on the change of the temperature scale. There is no documented way to subscribe to changes to a location’s temperature scale. So if a user changes the temperature scale, the SmartApp will still display the incorrect scale on the "current status" tile. That tile will be updated once the user changes the setpoint for either the cooling, heating or aux heating setpoints, refreshing the app, or on the auto refresh of the app every 5 minutes.

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -148,14 +148,12 @@ def generateEvent(Map results) {
 						 handlerName: name]
 
 			if (name=="temperature" || name=="heatingSetpoint" || name=="coolingSetpoint" ) {
-				def sendValue = convertTemperatureIfNeeded(value.toDouble(), "F", 1) //API return temperature value in F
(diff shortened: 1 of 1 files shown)


dsainteclaire:
This change LGTM, let's send to QA

dsainteclaire:
LGTM

workingmonk:
👍  to merge to staging

PR #1343, opened 2016-10-13
[DPROT-180] Correctly parse zone status in zigbee-button

varzac:
Previously the DTH didn't handle the extended status from zone status
reporting.  This moves to use the library for parsing zone status
which will handle the extended status properly.

This resolves: https://smartthings.atlassian.net/browse/DPROT-180

@tpmanley

### DIFF
--- a/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
+++ b/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
@@ -89,14 +89,8 @@ def parse(String description) {
 }
 
 private Map parseIasButtonMessage(String description) {
-    int zoneInt = Integer.parseInt((description - "zone status 0x"), 16)
-    if (zoneInt & 0x02) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -89,14 +89,13 @@ def parse(String description) {
...
 
 private Map parseIasButtonMessage(String description) {
-    int zoneInt = Integer.parseInt((description - "zone status 0x"), 16)
-    if (zoneInt & 0x02) {
-        resultMap = getButtonResult('press')
-    } else {
-        resultMap = getButtonResult('release')
+    def zs = zigbee.parseZoneStatus(description)
+
+    if (zs == null) {
```
tpmanley: I don't think this null check is necessary because this function is only called if the description starts with 'zone status'

tpmanley (approved):

PR #1348, opened 2016-10-14
Update netatmo-connect to use non-deprecated api

Dianoga:
**Note** I haven't actually tested this change due to time limitations. Someone should check this out and verify it behaves properly before merging.

devicedata has been deprecated in favor of getstationsdata and will be removed at the end of November.

### DIFF
--- a/smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
+++ b/smartapps/dianoga/netatmo-connect.src/netatmo-connect.groovy
@@ -393,12 +393,12 @@ def getDeviceList() {
 			deviceList[key] = "${value.station_name}: ${value.module_name}"
 			state.deviceDetail[key] = value
             state.deviceState[key] = value.dashboard_data
-		}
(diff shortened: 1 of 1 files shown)


strelitzia123:
This doesn't quite work, as it fails to update additional modules, only the main station.

I don't know how to submit a change on github, but these are the extra changes that need to be made:

Replace lines 396:401 

```
}
response.data.body.modules.each { value ->
def key = value._id
deviceList[key] = "${state.deviceDetail[value.main_device].station_name}: ${value.module_name}"
state.deviceDetail[key] = value
state.deviceState[key] = value.dashboard_data
```

With the following lines:

```
value.modules.each { value2 ->            
def key2 = value2._id
deviceList[key2] = "${value.station_name}: ${value2.module_name}"
state.deviceDetail[key2] = value2
state.deviceState[key2] = value2.dashboard_data            
}
```

In effect, the .each loops picking up the modules need to be nested due to the change in structure of the json response.

PR #1367, opened 2016-10-19
DVCSMP-2134, DVCSMP-2135 ZLL dimmable and white color temp bulb health check

surfous:
- configure a healthPoll every 5 min
  - dimmable refreshes on/off state and switch level (DVCSMP-2135)
  - white color temperature refreshes on/off state, switch level, and white color temperature (DVCSMP-2134)
  - set up checkInterval from both configure and updated
- checkInterval set to 12 min (2 misses + leeway)
- ping refreshes level setting

### DIFF
--- a/devicetypes/smartthings/zll-dimmer-bulb.src/zll-dimmer-bulb.groovy
+++ b/devicetypes/smartthings/zll-dimmer-bulb.src/zll-dimmer-bulb.groovy
@@ -11,6 +11,9 @@
  *  for the specific language governing permissions and limitations under the License.
  *
  */
+import groovy.transform.Field
+
+@Field Boolean hasConfiguredHealthCheck = false
 
 metadata {
     definition (name: "ZLL Dimmer Bulb", namespace: "smartthings", author: "SmartThings") {
@@ -21,6 +24,7 @@ metadata {
         capability "Refresh"
         capability "Switch"
         capability "Switch Level"
+        capability "Health Check"
 
         //fingerprint profileId: "C05E", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 1000", outClusters: "0000,0019"
         fingerprint profileId: "C05E", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 1000", outClusters: "0019"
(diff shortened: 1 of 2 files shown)


surfous:
@workingmonk and/or @jackchi, please code review or suggest other(s) who should review. 

I will ping Lakshmy to arrange QA for these two handlers once it passes code review.

workingmonk (approved):
looks good. but also get @jackchi  to approve from a health pov

devicetypes/smartthings/zll-dimmer-bulb.src/zll-dimmer-bulb.groovy
```diff
@@ -96,7 +97,35 @@ def poll() {
...
+    unschedule("healthPoll")
+    runEvery5Minutes("healthPoll")
+    // Device-Watch allows 2 check-in misses from device
+    sendEvent(name: "checkInterval", value: 60 * hcIntervalMinutes, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
     zigbee.onOffConfig() + zigbee.levelConfig() + zigbee.onOffRefresh() + zigbee.levelRefresh()
 }
+
+def configure() {
+    log.debug "configure()"
+    configureHealthCheck()
```
jackchi: I see you added `configureHealthCheck()` to both configure() and updated(). Is this necessary? What was the thought behind this? Thanks

surfous: Necessary? I'm not sure. This was a pattern I'd seen in the docs, and now that I look, it was  used in SmartApps calling a common initialize() method from both configure() and updated():
http://docs.smartthings.com/en/latest/smartapp-developers-guide/anatomy-and-life-cycle-of-a-smartapp.html#smartapp-structure

Looking at your comment on CHF-429, it seems that updated() is called if the DH for a device is changed in the IDE. What about in the case of deploying a new version of the **same** DH which changes the checkInterval?

Regardless, the best answer I can give is another question: For a DH which implements capability configurable, can configure() ever be invoked in an execution where updated() would not be?

If the answer is a definitive no, then there's no need to call it from both places.

Finally, it seems that I should leave the ZigBee.onOffConfig()... line with configure() for these DHs as that's configuring the device, which is the intent of the capability.configurable - do you agree?

jackchi: After talking to you this is the pattern we want to go with. Looks good! 👍

surfous: Great - my subsequent changes implemented running configureHealthCheck() only once-per-execution as we'd discussed before merging on your approval.

jackchi:
👍 
Let's have QA confirm all the supported models.

jackchi:
@surfous I tried the Field variable but it doesn't seem to work as expected here. 
The difference between your script and the one I tried is that I put the `configureHealthCheck` in the `updated()` lifecycle method.

surfous:
@jackchi How is it working that seems inconsistent with expectations? I would like to understand what it's doing or not doing.

jackchi:
By adding this the Field variable, my `configureHealthCheck` ran 3 times when I put it in the `updated()` method. Also Morten pointed out that there isn't a good way of knowing the execution lifecycle of a DT / SA. Usually this is used as scripts. Also there isn't harm in sending the `checkInterval` multiple times.

surfous:
@jackchi I understand what I need to understand here. 

Also, Field is working exactly as expected ;) We can talk about that later if you want - it's a solid tool. But you're right, since sending the `checkInterval` event more than once isn't detrimental, there's no need to use it here.

PR #1372, opened 2016-10-19
[DPROT-186] Update CentraLite DTHs to reset to factory default

varzac:
This change adds a call to reset the devices to factory default before
the rest of the normal configuration.  This is to work around a bug in
the CentraLite firmware where the binding table is sometimes not getting
properly cleared through other methods.

This resolves: https://smartthings.atlassian.net/browse/DPROT-186

This can't be merged until the version of appengine-zigbee with the new method are live.

@tpmanley


tpmanley (approved):

varzac:
@tpmanley Is this still something we want to move forward with?  I was waiting for the other big DTH changes to go out, and it needed library updates, but it should be good to go now if we want to move forward.

varzac:
We've decided not to fix this problem in this way.

PR #1374, opened 2016-10-20
MSA-1540: Device handle for OSRAM Lightify ON/OFF/Dim light.

yafenzhang:
Submitting 'Device handle for OSRAM Lightify ON/OFF/Dim light.' for publication


yafenzhang:
Close it because it does not contain some new code that Smartthing Team updated .

PR #1378, opened 2016-10-22
DVCSMP-2155 ZigBee switch power DH health check

surfous:
- Add capability Health Check
- specify 5 minute reporting interval to `zigbee.onOffConfig()` in `refresh()`
- call `refresh()` then set the checkInterval attribute in `configureHealthCheck()` 
- call `configureHealthCheck()` from both `configure()` and `updated()`
- implement `ping()` by sending the `zigbee.onOffRefresh()` command

For the Google Home project, supports specifically
- GE/Jasco ZigBee Plug-in Smart Switch (45853)
- GE/Jasco ZigBee In-Wall Smart Switch (45856)

The DH itself supports those, plus the Smartthings Outlet V4
- `fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 000F, 0B04", outClusters: "0019"`

and two other generic fingerprints:
- `fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0B04"`
- `fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702"`

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -20,6 +20,7 @@ metadata {
         capability "Power Meter"
         capability "Sensor"
         capability "Switch"
+        capability "Health Check"
 
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0B04"
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702"
@@ -77,10 +78,26 @@ def on() {
 }
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig() + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
+    Integer reportIntervalMinutes = 5
+    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig(0,reportIntervalMinutes * 60) + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
 }
 
 def configure() {
-    log.debug "Configuring Reporting and Bindings."
+    log.debug "in configure()"
+    configureHealthCheck()
+}
+
+def configureHealthCheck() {
+    Integer hcIntervalMinutes = 12
     refresh()
+    sendEvent(name: "checkInterval", value: hcIntervalMinutes * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+}
+
+def updated() {
+    log.debug "in updated()"
+    configureHealthCheck()
+}
+
+def ping() {
+    return zigbee.onOffRefresh()
 }


surfous:
For this device, I didn't follow the pattern we established for the non-reporting device because of Duncan's 10/19 comment here: 
https://smartthings.atlassian.net/wiki/pages/viewpage.action?pageId=72974565
saying, 

> The checkInterval should only be sent in configure() if the interval is linked to some configuration command sent from configure().

The checkInterval is indeed based around the `zigbee.onOffConfig(0, 300)` sent in `refresh()` which is called from `configure()`.

surfous:
@jackchi @workingmonk Sorry, forgot to tag you guys on this PR on Friday.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -77,10 +78,16 @@ def on() {
...
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig() + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
+    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig(0,300) + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
 }
 
 def configure() {
     log.debug "Configuring Reporting and Bindings."
+    Integer hcIntervalMinutes = 12
+    sendEvent(name: "checkInterval", value: hcIntervalMinutes * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
jackchi: Can we put this in `updated()` lifecycle method?

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -77,10 +78,16 @@ def on() {
 }
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig() + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
+    zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh() + zigbee.onOffConfig(0,300) + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
```
jackchi: Can we separate out the `config` commands from the `refresh` commands ?

jackchi (changes requested):
We might get into the same issue of needing to update `parse()` if we have to enroll existing ZigBee Switches. But for now, that's fine.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
surfous: I get your reasoning from a code organization angle. I just want to be careful not a alter the DH in ways that don't affect the intent of the ticket, especially since I'm not as familiar with ZigBee and our utility class and any pitfalls around returning a list of commands to issue to the device.  
Is it equivalent to say the following as the body of refresh():

```
List zigbeeRefresh = zigbee.onOffRefresh() + zigbee.simpleMeteringPowerRefresh() + zigbee.electricMeasurementPowerRefresh()
zigbeeRefresh += zigbee.onOffConfig(0,300) + zigbee.simpleMeteringPowerConfig() + zigbee.electricMeasurementPowerConfig()
return zigbeeRefresh
```

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
surfous: I put this here because of Duncan's comment on the confluence page I link to in my 10/21 comment. Now, maybe his advice only applies to Z-Wave and not ZigBee DHs, but I'd like to square your request with it before making any changes.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
jackchi: nvm here. I think you're good here.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
jackchi: @mckeed - I think we need to put this in `updated()` for reasons you told me before - 1. Manual update via web would have DTH enroll into Device-Watch

surfous:
@jackchi Is this what you had in mind? I think this should cover installation and DH changes via the IDE ('this' being the changes in the amended commit I pushed to this PR).

jackchi:
👍

jackchi (approved):

PR #1383, opened 2016-10-24
CHF-435 Implementation of Health Check for Zigbee Dimmer (GE Plug-In/In-Wall Smart Dimmer)

pchomal:
Implementation of Health Check for Zigbee Dimmer (GE Plug-In/In-Wall Smart Dimmer)

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/.st-ignore
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/.st-ignore
@@ -1,0 +1,2 @@
+.st-ignore
+README.md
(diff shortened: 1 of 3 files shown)


pchomal:
@ShunmugaSundar @jackchi 
Please review the changes and merge.

jackchi:
In enrolling existing devices, we've discovered that for Devices that executes locally (on the Hub's AppEngine), changes to `PARSE()` only goes out to AppEngine after a Hub firmware release. Because of this, all devices that executes locally, gets enrolled with the longer checkInterval since the confirmation subsequent checkInterval never gets sent. 

Some questions to be answered:
1) Does this execute locally?
2) If yes, then perhaps the best option here is to set the device's original checkInterval to the shorter one. 

Thoughts?

jackchi:
It looks like this executes in the cloud, then we should be good here.

jackchi:
Confirmed that this will execute on the cloud. We won't run into issues with those locally executed devices:
https://smartthings.atlassian.net/wiki/pages/viewpage.action?pageId=36045346

PR #1384, opened 2016-10-24
MSA-1549: New Device Handle for Osram Lightify RGBW Light

yafenzhang:
Submitting 'New Device Handle for Osram Lightify RGBW Light' for publication


yafenzhang:
Close it because it does not contain some new code that Smartthing Team updated .

PR #1395, opened 2016-10-25
[CHF-429] Device Health enrollment refactored into updated()

jackchi:
- Removed `Configuration` capability from Cree Bulb
- In `zigbee-dimmer.groovy` & `zigbee-rgbw-bulb.groovy` & `zigbee-white-color-temperature.groovy` DTH, the configuration is removed from `refresh()`. 

Upside: 
- Manually switching Device Types would work with Device Health
- Not all devices will have `Configuration` as a Capability. 

Downside:
- `updated()` method gets called multiple (3) times. But this shouldn't affect Device-Watch compatibility

https://smartthings.atlassian.net/browse/CHF-429

@tpmanley @mckeed Can you quickly look over this?

cc @mortent 
@parijatdas @ShunmugaSundar @skt123 @pchomal

### DIFF
--- a/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
+++ b/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
@@ -59,6 +59,15 @@ metadata {
     }
 }
 
+def installed() {
+    log.debug "${device} installed"
+}
+
+def updated() {
+    log.debug "${device} updated"
+    configureHealthCheck()
+}
+
 // Parse incoming device messages to generate events
 def parse(String description) {
     log.debug "description is $description"
@@ -96,6 +105,14 @@ def refresh() {
     zigbee.onOffRefresh() + zigbee.levelRefresh()
 }
 
+def configureHealthCheck() {
+    log.debug "configureHealthCheck"
+    unschedule("healthPoll")
+    runEvery5Minutes("healthPoll")
+    // Device-Watch allows 2 check-in misses from device
+    sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+}
+
 def healthPoll() {
     log.debug "healthPoll()"
     def cmds = zigbee.onOffRefresh() + zigbee.levelRefresh()
@@ -103,9 +120,5 @@ def healthPoll() {
 }
 
 def configure() {
-    unschedule()
-    runEvery5Minutes("healthPoll")
-    // Device-Watch allows 2 check-in misses from device
-    sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
-    zigbee.onOffRefresh() + zigbee.levelRefresh()
+    refresh()
 }
--- a/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
@@ -70,6 +70,16 @@ metadata {
     }
 }
 
+
+def installed() {
+    log.debug "${device} installed"
+}
+
+def updated() {
+    log.debug "${device} updated"
+    configureHealthCheck()
+}
+
 //Globals
 private getATTRIBUTE_HUE() { 0x0000 }
 private getATTRIBUTE_SATURATION() { 0x0001 }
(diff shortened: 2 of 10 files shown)


devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
```diff
@@ -96,16 +104,16 @@ def refresh() {
...
     def cmds = zigbee.onOffRefresh() + zigbee.levelRefresh()
     cmds.each{ sendHubCommand(new physicalgraph.device.HubAction(it))}
 }
-
-def configure() {
-    unschedule()
-    runEvery5Minutes("healthPoll")
-    // Device-Watch allows 2 check-in misses from device
-    sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
-    zigbee.onOffRefresh() + zigbee.levelRefresh()
```
tpmanley, line 110: I think we probably do want to keep `configure()` just so that it can call `zigbee.onOffRefresh() + zigbee.levelRefresh()`. Otherwise those events won't get created until on/off/setLevel is called

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -138,17 +148,18 @@ def ping() {
 }
 
 def refresh() {
-    zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_COLOR_TEMPERATURE) + zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_HUE) + zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_SATURATION) + zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + zigbee.colorTemperatureConfig() + zigbee.configureReporting(COLOR_CONTROL_CLUSTER, ATTRIBUTE_HUE, 0x20, 1, 3600, 0x01) + zigbee.configureReporting(COLOR_CONTROL_CLUSTER, ATTRIBUTE_SATURATION, 0x20, 1, 3600, 0x01)
```
tpmanley: Why were the configure commands removed? Many DTHs include configure commands in `refresh` to help recover misconfigured or unconfigured devices

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -138,17 +148,18 @@ def ping() {
...
+def configureHealthCheck() {
     // Device-Watch allows 3 check-in misses from device (plus 1 min lag time)
     // enrolls with default periodic reporting until newer 5 min interval is confirmed
     sendEvent(name: "checkInterval", value: 3 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+}
 
+def configure() {
     // OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-    refresh()
+    zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + zigbee.colorTemperatureConfig() + zigbee.configureReporting(COLOR_CONTROL_CLUSTER, ATTRIBUTE_HUE, 0x20, 1, 3600, 0x01) + zigbee.configureReporting(COLOR_CONTROL_CLUSTER, ATTRIBUTE_SATURATION, 0x20, 1, 3600, 0x01)
```
tpmanley: Even if we do remove the configure commands from `refresh()` we need the refresh commands in `configure()` because otherwise the events for the initial state may not get created

tpmanley (changes requested):
Besides those few comments it looks good

devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
jackchi: Thanks for the review. I'll add this back in.

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
jackchi: Same here.

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
jackchi: Okay changes amended.

tpmanley: This actually needs to be just `refresh()` like it was before. We need to call the refresh commands so that the initiate state is setup correctly when the device is added.

varzac: This can just go back to the call to refresh.  No need to duplicate the code.

devicetypes/smartthings/zigbee-white-color-temperature-bulb.src/zigbee-white-color-temperature-bulb.groovy
```diff
@@ -124,14 +133,15 @@ def refresh() {
...
+def configureHealthCheck() {
     // Device-Watch allows 3 check-in misses from device (plus 1 min lag time)
     // enrolls with default periodic reporting until newer 5 min interval is confirmed
     sendEvent(name: "checkInterval", value: 3 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+}
 
+def configure() {
     // OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-    refresh()
+    zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + zigbee.colorTemperatureConfig()
```
varzac: Same note here.  Just call refresh.

tpmanley (approved):

varzac (approved):

Issue #1400, opened 2016-10-28
zigbee-rgbw-bulb hues are off


joejulian:
scaledHueValue is calculated by taking a byte dividing by 255 and multiplying by 100. A hue should be between 0 and 360 though. Initial tests with lightify bulbs show that changing the multiplier to 360 produces the correct hues.

workingmonk:
cc @larsfinander @tpmanley 
We made the change for hue to represented as 0-100 to get in line with the docs. If that is causing issues in color reproduction, it might be worth looking at this again

joejulian:
Clearly this isn't a priority for ST so I'm just going to keep running my own custom devices and close this. I hate the nagging reminder every time I look at my open issues and PRs.

tpmanley:
@joejulian I apologize for missing or forgetting about this. As far as I can tell our device type handlers and smartapps do consistently expect hue to be 0-100. Are you aware of a smartapp that is using 0-360 instead of 0-100 or are you just saying that hue really should be 0-360 to give the most accurate color?

joejulian:
It has to be 0-360 or you don't get 2/3rds of the colors.

workingmonk:
the expectation is for smartapps to convert 0-360 from the user input to 0-100 for device handler, which converts it to 0-254 range based on the zigbee spec

joejulian:
That's not what I was observing.

tpmanley:
What Smartapp are you using?

joejulian:
Just an MQTT bridge

workingmonk:
me and @tpmanley found the issue with the hue conversion with that bridge smartapp. if you normalize the value with 100 instead of 360, it should help with the problems that you are seeing

joejulian:
Well it has been a year since I debugged this. I'll try again.

PR #1401, opened 2016-10-28
zigbee-rgbw-bulb: Hue should be degrees on a circle.

joejulian:
Convert from a byte to degrees (_360) not percent (_100)

Solves #1400


joejulian:
closing for lack of interest

PR #1404, opened 2016-10-31
Revert "[CHF-429] Device Health enrollment refactored into updated()"

jackchi:
Reverts SmartThingsCommunity/SmartThingsPublic#1395

@tpmanley @mckeed 

We also need to revert this PR. If user updates Device preferences, `update()` gets called without the correct `checkInterval`. 

cc @mortent


tpmanley (approved):
👍

PR #1421, opened 2016-11-02
DPROT-200: Pass unhandled messages up to the cloud

varzac:
Currently these DTHs return null when they parse a message that they
don't have specific handling for.  This ends up sending the message up
to the cloud as an event, which prevented us from potentially parsing
that message in the cloud.  By instead returning an empty map we can
instead send the message up to the cloud to be parsed there.  This
allows us to add handling in the cloud for new message without requiring
and AppEngine deploy for them to work.

This resolves: https://smartthings.atlassian.net/browse/DPROT-200

@tpmanley

### DIFF
--- a/devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
+++ b/devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
@@ -79,6 +79,7 @@ def parse(String description) {
 	log.debug "description is $description"
 
 	def finalResult = zigbee.getKnownDescription(description)
+	def event = [:]
 
 	//TODO: Remove this after getKnownDescription can parse it automatically
 	if (!finalResult && description!="updated")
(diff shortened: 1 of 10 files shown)


tpmanley:
Make a similar change to the additional DTHs that we discussed

devicetypes/smartthings/smartpower-outlet.src/smartpower-outlet.groovy
```diff
@@ -109,17 +111,19 @@ def parse(String description) {
 		if (cluster && cluster.clusterId == 0x0006 && cluster.command == 0x07){
 			if (cluster.data[0] == 0x00) {
 				log.debug "ON/OFF REPORTING CONFIG RESPONSE: " + cluster
-				sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+				event = createEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 			}
 			else {
 				log.warn "ON/OFF REPORTING CONFIG FAILED- error code:${cluster.data[0]}"
+				event = null
```
jackchi, line 118: @varzac  shouldn't this be `[:]`

varzac: This is a message we don't need/or care to propagate up to the cloud so we return null instead.

tpmanley:
@workingmonk can you review as well?

tpmanley:
Didn't realize this has already been merged. Looks good to me but would still be good to get an additional review.

workingmonk (commented):
looks good. it side steps (and solves) the issue that DM handles null differently from empty map

PR #1438, opened 2016-11-08
DVCSMP-2219: GE Link: fixed problem creating level events

tpmanley:
The bug causing level events not to be created happens when the setLevel
command runs in the cloud but the Device Type Handler is running locally
in the hub. Since state is not synced from the cloud to the hub the
`state.trigger` value was never set to `"setLevel"` on the hub which means
no level events would be created. The change is to not be reliant on state.
Instead we just don't create level events if the level is 0 since there is
no legitimate way for the level to be 0.

https://smartthings.atlassian.net/browse/DVCSMP-2219


tpmanley:
@workingmonk @varzac

varzac (approved):

workingmonk (approved):

PR #1448, opened 2016-11-10
DEVC-484 Add alternate fingerprints for Leviton switches

mckeed:
The new Leviton switches were joining as Z-Wave Relay with the old fingerprint format. Also, some of our test devices have different MSR values, so I've added those.


tylerlange:
@dkirker can you review

dkirker:
@mckeed Are any of these still needed? Apparently this slipped through the cracks...

PR #1461, opened 2016-11-14
[DVCSMP-2227] Handle all messages in smartpower dimming outlet

varzac:
Previously the implementation of isKnownDescription didn't cover all
possible messages that could be parsed and this caused null pointer
exceptions with certain messages.  This now handles all the
possibilities.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-2227

@tpmanley @workingmonk 

I have added a task to update this DTH some more with the big DTH update coming with the new getEvent library changes, and when doing so I will fix all the ZigBee strings, but this allows us to get this fix into the next AppEngine release coming next week.

### DIFF
--- a/devicetypes/smartthings/smartpower-dimming-outlet.src/smartpower-dimming-outlet.groovy
+++ b/devicetypes/smartthings/smartpower-dimming-outlet.src/smartpower-dimming-outlet.groovy
@@ -71,7 +71,7 @@ def parse(String description) {
 
 	def event = [:]
 	def finalResult = isKnownDescription(description)
-	if (finalResult != "false") {
+	if (finalResult) {
 		log.info finalResult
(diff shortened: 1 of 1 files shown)


tpmanley (approved):

devicetypes/smartthings/smartpower-dimming-outlet.src/smartpower-dimming-outlet.groovy
```diff
@@ -71,7 +71,7 @@ def parse(String description) {
 
 	def event = [:]
 	def finalResult = isKnownDescription(description)
```
workingmonk, line 73: do we want to change this to zigbee.getEvent?

workingmonk (approved):

PR #1477, opened 2016-11-16
MSA-1590: New Device Handle for Osram Lightify Light(ZHA)

yafenzhang:
Submitting 'New Device Handle for Osram Lightify Light(ZHA)' for publication


yafenzhang:
Close it because it does not include some new modification made by smartthing team.

PR #1493, opened 2016-11-24
PROB-1428 zll-dimmer-bulb: Fix bug in configure preventing devices from joining

tpmanley:
The configureHealthCheck method and therefore the configure method was
returning a boolean which was causing an exception during the initial
join that would prevent the device from showing up to the user.

https://smartthings.atlassian.net/browse/PROB-1428


tpmanley:
@workingmonk @jackchi

workingmonk:
👍

PR #1505, opened 2016-11-29
DPROT-215 Don't use sendEvent for AppEngine parse events

varzac:
With the changes made for
https://smartthings.atlassian.net/browse/DPROT-200 there were a few DTHs
that were using sendEvent to directly send events generated during
parse.  However, because using sendEvent didn't result in parse
returning an event AppEngine would send the description up to the cloud
DTH to be handled.  In some cases this resulted in multiple events for
the same device trigger.

This resolves: https://smartthings.atlassian.net/browse/DPROT-215

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
@@ -161,7 +161,7 @@ private Map parseCatchAllMessage(String description) {
 				if (cluster.command == 0x07) {
 					if(cluster.data[0] == 0x00) {
 						log.debug "TEMP REPORTING CONFIG RESPONSE" + cluster
-						sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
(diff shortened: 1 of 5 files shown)


varzac:
@tpmanley @workingmonk Can I get some :eyes: on this?  I was hoping to have it go out with next week's release so we could do the AppEngine release, but time is short now.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -161,7 +161,7 @@ private Map parseCatchAllMessage(String description) {
 				if (cluster.command == 0x07) {
 					if(cluster.data[0] == 0x00) {
 						log.debug "TEMP REPORTING CONFIG RESPONSE" + cluster
-						sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+						resultMap = createEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
tpmanley: I think it would be a little more clear to just return a map here and let parse create the event

tpmanley (approved):

PR #1512, opened 2016-12-05
Update wattvision-manager.groovy to prevent UI error

zenion:
UI is erroring due to invalid math on 3 hour millisecond calculation.


zenion:
does anyone do anything with pulls here?

axemann (approved):
LGTM

tpmanley (approved):
The fix looks reasonable

PR #1515, opened 2016-12-06
[INTL-657] Localizing Hue (Connect)

gausnes:
This will be the first in a series to localize our service managers. I've created a script todo the majority of the work and am cleaning up edge cases myself. Compared to the localization of previous SAs/DTHs the english strings aren't suppose to be used as keys in this round.

I haven't tested deploying this SA yet

Looking for Feedback
- On key structure
  - Outside of the initial preferences and pages it follow: `methodName.property:countInMethod`
- Relevant fields have been accounted for

CC @larsfinander @bflorian @lbredeso @kelseyterry


bflorian:
I love the approach of using keys such as `mainPage.title`, but the reason that wasn't done before is that string lookup wasn't turned on in production and it still isn't. Is the plan to do in all shards before these changes go out?

gausnes:
@bflorian I believe so @kelseyterry informed me that we wanted to replace the english strings with keys for this round of localization.

lbredeso:
Am I right that if we use these keys instead of English text, in the event of a "miss" (e.g. the translation for zh_CN is missing for a given key), the Localization Service will return the key to the user? Is that expected, or am I misunderstanding the behavior here?

bflorian:
@lbredeso most localization systems can be configured to fall back to a default language if there's no entry for the current locale, but I'm not sure if ours does that. @kelseyterry do you know if that is the case?

In general though I think using context-specific keys like this is much better than using English as the key. One of the problems with that is you don't know when you've missed a string in the default English file and therefore you don't get it translated into all the different languages. It also makes maintenance much harder. If you need to tweak the English phrasing a bit you then must change the key in every other language file or wind up in the weird situation where the English keys are almost but not exactly the same as the actual English text. And finally, there are cases where there isn't a one-to-one mapping between English words or phrases and saying the same thing in another language.

gausnes:
After talking with @staticnull the localization service will need some changes to accommodate keys being added instead of english text into the service manager. 

And after talking to @kelseyterry this PR exceeds the scope of the work that is needed at the moment. So I'm going to close this PR, but this might come up in the near future.

PR #1534, opened 2016-12-14
Update dimmer-switch.groovy

wyliegraham:
removing the turn on/off function when changing the dim level


wyliegraham (commented):
sorry, don't know what i'm doing with github, please ignore/revert this

PR #1545, opened 2016-12-19
[CHF-477] Health Check for Z-Wave Dimmer Switch Generic

pchomal:
Health Check implementation for Z-Wave Dimmer Switch Generic with checkinterval of 32min for following devices:

1. Leviton 600W Incandescent Dimmer
2. Leviton Magnetic Low Voltage Dimmer
3. Leviton 3-Speed Fan Controller
4. Enerwave In-Wall Dimmer (ZW500D)


pchomal:
@jackchi @ShunmugaSundar 

Please review & merge

mckeed:
The join names are a bit long, please change the Leviton ones to:
```
Leviton Fan Controller
Leviton Low Voltage Dimmer
Leviton 600W Incandescent Dimmer
```

jackchi:
Can someone explain about `Works with SmartThings` model number vs. the model # in the fingerprints? 

Who from WWST can verify this change? I am not the best person to review and merge this PR. Perhaps someone from Device Lab's team?

Also 4 of these devices are indeed WWST, but why are they not already named?

@mckeed @tylerlange

mckeed:
We only gained the ability to set join names based on Z-Wave manufacturer info recently – no one else has taken on the task of adding all these manufacturer fingerprints, so it works well for CHF to do it as you check models for device health.

The model code is not usually related at all to the customer-facing model number.

jackchi:
👍  Looks good to me then.

PR #1574, opened 2017-01-09
[PROB-1496] Fix for Disable Push Notification in Working From Home SmartApp

jackchi:
@mrnohr @workingmonk


workingmonk:
@jackchi did you verify the change locally?

mrnohr:
Seems like the correct fix. Since `sendPushMessage` is a `bool` then it will never be "No"

workingmonk (approved):

PR #1605, opened 2017-01-20
DVCSMP-2324 - Centralite Thermostat handle multiple attributes

juano2310:


devicetypes/smartthings/centralite-thermostat.src/centralite-thermostat.groovy
```diff
@@ -81,9 +81,11 @@ metadata {
 // parse events into attributes
 def parse(String description) {
 	log.debug "Parse description $description"
+	def result = null
 	def map = [:]
+ 	def additionalMap = [:]
 	if (description?.startsWith("read attr -")) {
-		def descMap = parseDescriptionAsMap(description)
```
tpmanley: You can remove the definition of `parseDescriptionAsMap` since you're using the `zigbee` version now

juano2310: How about now? I removed the whole if, hope that's ok

devicetypes/smartthings/centralite-thermostat.src/centralite-thermostat.groovy
```diff
@@ -81,41 +81,63 @@ metadata {
...
+		log.debug "COOLING SETPOINT"
+		map.name = "coolingSetpoint"
+		map.value = getTemperature(descMap.value)
+		map.unit = temperatureScale
+        if (descMap.additionalAttrs) {
+        	if (map) {
+            	result = createEvent(map)
+            	log.debug "Parse returned $map"
+            }
+            descMap.additionalAttrs.each{
```
tpmanley: You can remove some duplicate attribute handling code if you start by creating a list of maps containing the info for each attribute, and then you iterate through that list and process each attribute individually.

devicetypes/smartthings/centralite-thermostat.src/centralite-thermostat.groovy
```diff
@@ -81,41 +81,63 @@ metadata {
...
-			map.name = "thermostatMode"
-			map.value = getModeMap()[descMap.value]
-		} else if (descMap.cluster == "0202" && descMap.attrId == "0000") {
-			log.debug "FAN MODE"
-			map.name = "thermostatFanMode"
-			map.value = getFanModeMap()[descMap.value]
-		}
+	def descMap = zigbee.parseDescriptionAsMap(description)
+	log.debug "Desc Map: $descMap"
+	if (descMap.cluster == "0201" && descMap.attrId == "0000") {
```
varzac: I would wrap all this in a check to see if descMap is non-empty.  If the description isn't parseable as an attribute value you could get a null pointer exceptions on these checks.

devicetypes/smartthings/centralite-thermostat.src/centralite-thermostat.groovy
varzac: I agree with Tom here.  Put all the if checks inside a loop like this.

juano2310:
How about now? I think that if attrData is empty each won't execute and won't create any exception, right?

juano2310:
New PR https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/1606

devicetypes/smartthings/centralite-thermostat.src/centralite-thermostat.groovy
```diff
@@ -109,13 +114,11 @@ def parse(String description) {
 			map.name = "thermostatFanMode"
 			map.value = getFanModeMap()[descMap.value]
 		}
+		if (map)
```
varzac, line 117: I like to be explicit and use curly braces even on one line ifs.

PR #1607, opened 2017-01-20
[CHF-508] Hue bulbs hides enrollment events

jackchi:
screenshot_20170120-120224

@larsfinander @tylergustafson


larsfinander (commented):
You should do the same for the LAN-Hue-... device types in SmartThingsPrivate. Thanks for doing this!

PR #1608, opened 2017-01-20
DVCSMP-2325 - Remove parse method from simulated Switch

juano2310:


devicetypes/smartthings/testing/simulated-switch.src/simulated-switch.groovy
```diff
@@ -40,8 +40,6 @@ metadata {
 }
 
 def parse(String description) {
```
bflorian: I think you also need to remove `String` here. The stack trace was complaining about the lack of a method with a `Map` argument. Apparently it's possible for `parse()` to be called with a `Map`

bflorian (approved):

bflorian:
👍

PR #1621, opened 2017-01-26
[DPROT-227] Manually refresh on/off status after setLevel

varzac:
This is because ZLL bulbs do not report or allow for configuration of
reporting of attributes.  As a result if we change a value, we have to
manually call a read of that attribute to update our status.

This is a part of: https://smartthings.atlassian.net/browse/DPROT-227

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
+++ b/devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
@@ -82,7 +82,7 @@ def on() {
 }
 
 def setLevel(value) {
-    zigbee.setLevel(value) + ["delay 500"] + zigbee.levelRefresh()         //adding refresh because of ZLL bulb not conforming to send-me-a-report
+    zigbee.setLevel(value) + zigbee.onOffRefresh() + zigbee.levelRefresh()       //adding refresh because of ZLL bulb not conforming to send-me-a-report
 }
 
 /**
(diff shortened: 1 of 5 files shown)


devicetypes/smartthings/cree-bulb.src/cree-bulb.groovy
```diff
@@ -82,7 +82,7 @@ def on() {
 }
 
 def setLevel(value) {
-    zigbee.setLevel(value) + ["delay 500"] + zigbee.levelRefresh()         //adding refresh because of ZLL bulb not conforming to send-me-a-report
+    zigbee.setLevel(value) + ["delay 500"] + zigbee.levelRefresh() + zigbee.onOffRefresh()       //adding refresh because of ZLL bulb not conforming to send-me-a-report
```
tpmanley: Try removing the extra delay after `zigbee.setLevel` and putting the `zigbee.onOffRefresh` there instead. Then the on/off status will be updated more quickly and we'll still give the device time to transition to the new level before reading the current level again.

tpmanley (approved):

varzac:
@workingmonk We need this to go out with next weeks release so that we can release the zigbee library changes next week.

workingmonk (approved):

workingmonk:
@varzac can you ping Steven to get the QA started on this one? just a quick regression in staging is fine.

PR #1628, opened 2017-01-31
DVCSMP-2290 - Ecobee Localization

juano2310:


bflorian:
You should be able to remove the countries from the individual files if you want to since you're depending on the new release for the input messageArgs stuff with this PR. But if you want to leave them there for now in case we don't get the DM release out in time I'm OK with that too 👍

juano2310:
@bflorian I will marge this one and update the files once I add the remaining languages.

PR #1643, opened 2017-02-03
[DPROT-238] Manually call on with setLevel

varzac:
This is to work around misbehaving devices that don't properly honor the
move to level with on off command.

This resolves: https://smartthings.atlassian.net/browse/DPROT-238

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
@@ -97,7 +97,7 @@ def on() {
 }
 
 def setLevel(value) {
-    zigbee.setLevel(value)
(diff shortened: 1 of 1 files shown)


workingmonk:
shouldn't this be a change in the zigbee library? i am ok with this being a temporary fix, but in the long run, this logic should be in the library, right?

varzac:
@workingmonk This is a fix in the DTH because it is a specific device not following the zigbee spec.  We removed the on call from the zigbee library because it causes issues with regards to setting level and turning off devices.

tpmanley (approved):

workingmonk:
do we need the fix for zigbee-dimmer DTH too?

varzac:
@workingmonk Not that we are aware of.  It appears to be Jasco device specific, and non of the fingerprints in zigbee-dimmer are for Jasco products.

PR #1644, opened 2017-02-03
DPROT-238 Fix typo on toInteger zigbee dimmer power

varzac:
This relates to: https://smartthings.atlassian.net/browse/DPROT-238

@tpmanley @workingmonk 

I initially made this error when I wrote it, and I fixed it while testing with my custom DTH, but I failed to copy it back over correctly :man_facepalming: I don't have a zigbee-dimmer-power at home so I can easily test.  I'll have Steven test before merging this time.

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
@@ -97,7 +97,7 @@ def on() {
 }
 
 def setLevel(value) {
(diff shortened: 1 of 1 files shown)


tpmanley (approved):

varzac:
Steven has verified this is working.

stevenchitov:
Re-tested with Zach fix ( 02/03/17 ) working just great now on V2 and V1 HUBs using both manual level change or setting level using Smart Lighting. Testing  was done by publishing own DTH from https://github.com/varzac/SmartThingsPublic/blob/fix-zigbee-dimmer-set-level/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy

PR #1646, opened 2017-02-04
DPROT-222 smartsense motion sensor: update battery voltage to percentage calculation

tpmanley:
Switch to a table based lookup for all devices that use this DTH. The
table lookup for all devices besides the 3325-S is unchanged from the
previous version. The changes for the 3325-S are to treat 2.4V as the
minimum useable voltage.

https://smartthings.atlassian.net/browse/DPROT-222

@workingmonk @varzac


posborne:
For the 3325-S, do we have data to show that the voltage for a "full" battery will actually be as high as 3V when measured under load which is, as I understand it, what we will now be doing?

Looking at this discharge curve for the Energizer CR2450, the voltage when measured at pulse seems to start at closer to 2.8/2.9V.  Kind of a shame we only get 0.1V/bit resolution.

image

Source: http://data.energizer.com/pdfs/cr2450.pdf

varzac (approved):

PR #1662, opened 2017-02-11
DVCSMP-2425 - LIFX Exceptions

juano2310:


marstorp (changes requested):
Please add a comment explaining the reason for deletion of child devices and the consequences for failing to delete it.

juano2310:
The reason the implementation is trying to delete this bulb is because it is not longer connected to the LIFX location. It is failing because the child device it is still connected to a SmartApp and it is preventing it from being deleted causing an exception. Adding the try will prevent this exception from happening. 
Ideally device health would show to the user that the device is not longer accessible so the user can manage the device either by force deleting it or removing it from the SmartApp.

marstorp (approved):

marstorp:
LGTM 👍

PR #1669, opened 2017-02-13
Put async calls back in the code for polling.  Other interactions use old communication mechanism

dsainteclaire:
@jimmyjames I wanted to get your feedback on this. I added code back to move the polling to use asynchronous API, but left everything else (mode change / set point change / token refresh) to be on-demand.  But I have one question, and maybe you can help me track down an answer.  In the async response handler, I want to be able to figure out if the called failed because the auth token expired so I can kick off a token refresh.  But, I can't really find any documentation of the response object that gets passed to the callback.

In the synchronous way, we're able to catch an HttpResponseException and look at the Json  that got returned [if (e.response.data.status.code == 14)] which lets us know that we need to refresh.  Can you ask someone over there how I'd go about doing the same thing?

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -20,6 +20,8 @@
  *      JLH - 02-15-2014 - Fuller use of ecobee API
  *      10-28-2015 DVCSMP-604 - accessory sensor, DVCSMP-1174, DVCSMP-1111 - not respond to routines
  */
+include 'asynchttp_v1'
+
 definition(
 		name: "Ecobee (Connect)",
 		namespace: "smartthings",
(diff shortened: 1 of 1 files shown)


dsainteclaire:
@jimmyjames bump :)

jimmyjames:
Regarding the error handling, this is possible with async - see the error handling section [here](http://docs.smartthings.com/en/latest/smartapp-developers-guide/async-http.html#error-responses).

You can use `getErrorJson()` in this case and then inspect the response for the specific code. You can see what I did in the original PR in the `asyncPollResponseHandler()` method in my original PR [here](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/1180/files).

jimmyjames:
It's been a while so I've either forgotten or things have changed... do the children still poll as well? The ecobee thermostat DTH still implements a `poll()` method and it's used in the code, but is that more-or-less one off (and still done synchronously)?

PR #1676, opened 2017-02-14
add pushsafer.com

appzer:
add [pushsafer.com](https://www.pushsafer.com) notification service


appzer:
add app thru https://graph.api.smartthings.com/ide/apps

PR #1680, opened 2017-02-15
[PROB-1518] Properly create event for outlet power

varzac:
We were not properly creating the event before returning it, causing us
to not actually generate the power event.

This resolves: https://smartthings.atlassian.net/browse/PROB-1518

@tpmanley @workingmonk


varzac:
This is a problem with the smartpower-dimming-outlet as well.  I will have a fix for that pushed here shortly

tpmanley (approved):

workingmonk (approved):

dbradmit:
Looks like we're already heading that direction but waiting a couple weeks for this to make it to production is probably fine.

varzac:
@dbradmit This should be going out with next Tuesday's release.

PR #1708, opened 2017-02-23
ICP-375 CON-90 fix for wemo dimming issue

workingmonk:


workingmonk:
cc @tpmanley

devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
```diff
@@ -113,6 +113,12 @@ def refresh() {
     zigbee.onOffRefresh() + zigbee.levelRefresh()
 }
 
+def installed() {
+    if ((device.getDataValue("manufacturer") == "MRVL") && (device.getDataValue("model") == "MZ100")) {
```
bakken7: I would add a comment reflecting the fact that this is to accommodate a FW defect and what the case would be for removing this if-block.

tpmanley: Does installed get called every time the DTH is changed? Assuming that's the case it might be safer to only create the event if the current value if 0.

workingmonk: Installed is only called once. When the DTH is changed, the updated method is called. But i will add the check if something goes wrong :)

tpmanley (approved):

PR #1731, opened 2017-03-01
DVCSMP-2484 smartsense-motion-sensor: fix motion event reporting for some devices

tpmanley:
The motion event reporting for any device that used the Occupancy Sensor
cluster for reporting motion was broken. I'm not sure if any devices using
this DTH actually use this cluster. I think all the officially supported
devices report motion throug the IAS Zone cluster.

https://smartthings.atlassian.net/browse/DVCSMP-2484

@varzac 
@workingmonk think its worth pulling into the release this week? I was leaning towards no since I don't know that it actually affects any devices that use this DTH.


varzac (approved):

tpmanley:
@workingmonk pointed at staging

PR #1735, opened 2017-03-01
Revert "MSA-1408: initial submission for Curb"

workingmonk:


workingmonk:
Reverting this based on the conversation I had with @tslagle13

PR #1739, opened 2017-03-02
DVCSMP-2487 OpenT2T: Update to 3/2 submission

larsfinander:


larsfinander:
This is under test by microsoft and not open to public, hence no QA is needed.

PR #1741, opened 2017-03-03
[CHF-532] [CHF-533] Health Check Z-Wave Sleepy Fibaro Sensors

jackchi:
`Health Check` expects timeout to be `8 hrs 2 minutes`. Does not implement `ping`. Since these are battery, Z-Wave, sleepy devices. 

https://smartthings.atlassian.net/wiki/display/CHF/Device+Taxonomy

cc @mortent @magnusekenheim @charliek @ShunmugaSundar @mckeed @workingmonk @dkirker


mortent:
So these devices check in every 4 hours? As far as I can tell, this looks good.

jackchi:
Yes. Checked with @mckeed  and they will generate a WakeUpNotification event

charliek:
@workingmonk @jackchi we may want to consider putting this against production.  Should we discuss the plan in the morning?

workingmonk (approved):

PR #1765, opened 2017-03-11
DVCSMP-3478 Add mfr fingerprints to Fibaro Motion

mckeed:
https://smartthings.atlassian.net/browse/DVCSMP-3478


dkirker:
I'll give these a whirl with the devices tomorrow. I think the big issue was on the non Z-Wave Plus models (the z-wave plus models were pairing fine and I was giving them out as a stop-gap; but now we are out of the z-wave plus fibaro motion sensors haha).

greens (approved):
Worked with my devices.

PR #1783, opened 2017-03-15
Hue: Re-add Hue bridge if deleted

larsfinander:
-Add Hue Bridge if missing during deviceSync

### DIFF
--- a/smartapps/smartthings/hue-connect.src/hue-connect.groovy
+++ b/smartapps/smartthings/hue-connect.src/hue-connect.groovy
@@ -472,8 +472,6 @@ def addBridge() {
 					log.error "Failed to create Hue Bridge device"
 				}
 			}
-		} else {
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/hue-connect.src/hue-connect.groovy
```diff
@@ -472,8 +472,6 @@ def addBridge() {
 					log.error "Failed to create Hue Bridge device"
 				}
 			}
-		} else {
-			log.debug "found ${d.displayName} with id $selectedHue already exists"
```
workingmonk, line 476: why remove this logging?

larsfinander: just didnt want it to print this every single poll for every user since it is the normal case.

workingmonk (approved):

PR #1786, opened 2017-03-15
[CHF-552] Implementation of Health Check for Aeon Siren.

skt123:
Added Health Check for Aeon Siren with checkinterval of 32 mins.

### DIFF
--- a/devicetypes/smartthings/aeon-siren.src/README.md
+++ b/devicetypes/smartthings/aeon-siren.src/README.md
@@ -1,0 +1,37 @@
+# Aeon Siren
+
+Cloud Execution
+
+Works with: 
+
+* [Aeon Labs Siren (Gen 5)](https://www.smartthings.com/works-with-smartthings/aeon-labs/aeon-labs-siren-gen-5)
+
+## Table of contents
+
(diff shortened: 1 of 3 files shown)


skt123:
@jackchi @ShunmugaSundar 
Please review and merge.

jackchi:
Let's quickly add a .README. 
@skt123  Does the `ping` generate an event? If so, what is it?

skt123:
@jackchi Apart from the ping() event itself, no others events are being generated from ping. 
Ping just calls `zwave.basicV1.basicGet()` function to get device status as there is no `refresh()` method.

skt123:
@jackchi 
Added Readme. Please check

devicetypes/smartthings/aeon-siren.src/README.md
```diff
@@ -0,0 +1,37 @@
+# Aeon Siren
+
+Cloud Execution
```
jackchi, line 3: ZWave Execution?

skt123: That refers to the location of the code execution which in this case is **cloud** and not **local**.

PR #1795, opened 2017-03-17
HUE: add code to enable localization for Service Manage

larsfinander:
-Updated strings in app
-Added i18 files


juano2310:
LGTM 👍 I as far as I can tell all the parameters in the connect app were updated correctly and the include 'localization' was included :)

juano2310 (approved):
LGTM 👍 As far as I can tell all the parameters, in the connect app, were updated correctly and "include 'localization'" was included :)

PR #1799, opened 2017-03-19
Merging my SmartThings workflow into the public repository

kewashi:
Hi folks - I have been working for a few years on some ST apps but just recently got them tie into GitHub so now I'm offering them to be included for programmers to see. Most are works in progress and all work to some degree. The most ambitious one is HouseMap which requires web hosting and PHP skills. Enjoy and let me know if you find this useful.


kewashi:
Housemap was renamed to HousePanel and is discussed on the Community Forum

PR #1811, opened 2017-03-21
DVCSMP-2399 Added Zooz ZEN20 DTH

bflorian:
Added DTH for [Zooz ZEN20](http://www.getzooz.com/zooz-zen20-power-strip.html) Power Strip with individually addressable outlets.

### DIFF
--- a/devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
+++ b/devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
@@ -1,0 +1,210 @@
+/**
+ *  Zooz ZEN20 Power Strip Outlet
+ *
+ *  Implementation of the Zooz ZEN20 power strip that uses the new composite device capabilities to provide individual
+ *  control of each outlet from SmartApps as well as the mobile app. Incorporates contributions from:
+ *
+ *  Eric Maycock (https://github.com/erocm123/SmartThingsPublic/blob/master/devicetypes/erocm123/zooz-power-strip.src/zooz-power-strip.groovy)
+ *  Robert Vandervoort (https://github.com/robertvandervoort/SmartThings/blob/master/zooZ-Strip-ZEN20/device_type-zooZ-strip-ZEN20_v1.0)
+ *
+ *  Copyright 2017 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+metadata {
+	definition (name: "Zooz Power Strip", namespace: "smartthings", author: "SmartThings") {
+		capability "Switch"
+		capability "Refresh"
+		capability "Actuator"
+		capability "Sensor"
+
+		fingerprint manufacturer: "015D", prod: "0651", model: "F51C", deviceJoinName: "Zooz ZEN 20 Power Strip"
+	}
+
+	tiles {
+		multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
+			tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
+				attributeState "off", label: '${name}', action: "switch.on", icon: "st.switches.switch.off", backgroundColor: "#ffffff", nextState:"turningOn"
+				attributeState "on", label: '${name}', action: "switch.off", icon: "st.switches.switch.on", backgroundColor: "#00A0DC", nextState:"turningOff"
+				attributeState "turningOn", label:'${name}', action:"switch.off", icon:"st.switches.switch.on", backgroundColor:"#00A0DC", nextState:"turningOff"
+				attributeState "turningOff", label:'${name}', action:"switch.on", icon:"st.switches.switch.off", backgroundColor:"#ffffff", nextState:"turningOn"
+			}
+		}
+		childDeviceTiles("outlets")
+		standardTile("refresh", "device.switch", width: 1, height: 1, inactiveLabel: false, decoration: "flat") {
+			state "default", label:'', action:"refresh.refresh", icon:"st.secondary.refresh"
+		}
+	}
+}
+
+/////////////////////////////
+// Installation and update //
+/////////////////////////////
+def installed() {
+	createChildDevices()
+}
+
+def updated() {
+	if (!childDevices) {
+		createChildDevices()
+	}
+	else if (device.label != state.oldLabel) {
+		childDevices.each {
+			def newLabel = "${device.displayName} (CH${channelNumber(it.deviceNetworkId)})"
+			it.setLabel(newLabel)
+		}
+		state.oldLabel = device.label
+	}
+}
+
+
+//////////////////////
+// Event Generation //
+//////////////////////
+def parse(String description) {
+	trace "parse('$description')"
+	def result = []
+	if (description.startsWith("Err")) {
+		result = createEvent(descriptionText:description, isStateChange:true)
+	} else if (description != "updated") {
+		def cmd = zwave.parse(description, [0x60: 3, 0x32: 3, 0x25: 1, 0x20: 1])
+		if (cmd) {
+			result += zwaveEvent(cmd, 1)
+		}
+		else {
+			log.warn "Unparsed description $description"
+		}
+	}
+	result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.multichannelv3.MultiChannelCmdEncap cmd, ep) {
+	def encapsulatedCommand = cmd.encapsulatedCommand([0x32: 3, 0x25: 1, 0x20: 1])
+	if (encapsulatedCommand) {
+		zwaveEvent(encapsulatedCommand, cmd.sourceEndPoint as Integer)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+ */
+metadata {
+	definition (name: "Zooz Power Strip", namespace: "smartthings", author: "SmartThings") {
+		capability "Switch"
+		capability "Refresh"
+		capability "Configuration"
+		capability "Actuator"
+		capability "Sensor"
+
+		fingerprint deviceId: "0x1004", inClusters: "0x5E,0x85,0x59,0x5A,0x72,0x60,0x8E,0x73,0x27,0x25,0x86", manufacturer: "015D", model: "F51C", prod: "0651", deviceJoinName: "Zooz ZEN 20 Power Strip"
```
mckeed: We're using MSR fingerprints for specific device DTHs now. This should be

    fingerprint mfr:"015D", prod:"0651", model:"F51C"

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+
+def refresh() {
+	def cmds = (1..5).collect { endpoint ->
+		encap(zwave.switchBinaryV1.switchBinaryGet(), endpoint)
+	}
+	delayBetween(cmds, 100)
+}
+
+def configure() {
+	def cmds = [
```
mckeed: Remove all of these commands from configure(). None are necessary and firmwareMdGet isn't supported by this device.

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+			new physicalgraph.device.HubAction(encap(zwave.basicV1.basicSet(value: value), endpoint)),
+			new physicalgraph.device.HubAction(encap(zwave.switchBinaryV1.switchBinaryGet(), endpoint)),
+	]
+	sendHubCommand(actions, 500)
+}
+
+private void createChildDevices() {
+	state.oldLabel = device.label
+	for (i in 1..5) {
+		addChildDevice("Zooz Power Strip Outlet", "${device.deviceNetworkId}-${i}", null,
```
mckeed: Currently we identify ZigBee devices by the length of the dni being 4. Not sure if this could mess with that. For my Multi Channel Control SA I used `"${device.deviceNetworkId}ep${i}"`

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+//////////////////////
+def parse(String description) {
+	log.trace "parse('$description')"
+	def result = []
+	if (description.startsWith("Err")) {
+		result = createEvent(descriptionText:description, isStateChange:true)
+	} else if (description != "updated") {
+		def cmd = zwave.parse(description, [0x60: 3, 0x32: 3, 0x25: 1, 0x20: 1])
+		if (cmd) {
+			result += zwaveEvent(cmd, null)
```
mckeed: Could use 1 as the default endpoint value since all messages to/from the root device appear to be routed to CH1 on this device

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+def zwaveEvent(physicalgraph.zwave.Command cmd, ep) {
+	log.warn("${device.displayName}: Unhandled ${cmd}" + (ep ? " from endpoint $ep" : ""))
+}
+
+/////////////////////////////
+// Installation and update //
+/////////////////////////////
+def on() {
+	def cmds = []
+	cmds << zwave.switchAllV1.switchAllOn().format()
```
mckeed: Switch All is for broadcasts to turn multiple devices on, not for multi-channel.

You can use a bit mask to turn on all the channels:

    def cmd = zwave.switchBinaryV1.switchBinarySet(switchValue: 0xFF)
    cmds << zwave.multiChannelV3.multiChannelCmdEncap(bitAddress: true, destinationEndPoint:0x1F).encapsulate(cmd).format()

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+	response(refresh() + configure())
+}
+
+def updated() {
+	if (!childDevices) {
+		createChildDevices()
+	}
+	else if (device.label != state.oldLabel) {
+		childDevices.each {
+			def newLabel = "${device.label} (CH${channelNumber(it.deviceNetworkId)})"
```
mckeed: Should probably change `device.label` here to `device.displayName` in case they deleted the label.

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd, endpoint) {
+	def cmds = []
+	(1..5).each { n ->
+		cmds << encap(zwave.switchBinaryV1.switchBinaryGet(), n)
+		cmds << "delay 100"
+	}
+	return response(cmds)
```
mckeed: Even though the device doesn't send this normally, we should interpret this event the same as SwitchBinaryReport                                      if it's from an endpoint.

What's the idea behind doing a refresh here?

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
```
mckeed: This device doesn't have the Configuration CC

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+	def childDevice = children.find{it.deviceNetworkId.endsWith("$endpoint")}
+	childDevice.sendEvent(name: "switch", value: cmd.value ? "on" : "off")
+
+	if (cmd.value) {
+		// One on and the strip is on
+		result << createEvent(name: "switch", value: "on")
+	} else {
+		// All off and the strip is off
+		def allOff = true
+		(1..5).each { n ->
```
mckeed: This loop could just be ` ! children.any { it.currentValue("switch") == "on" }`

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
```diff
@@ -0,0 +1,232 @@
...
+		createChildDevices()
+	}
+	else if (device.label != state.oldLabel) {
+		childDevices.each {
+			def newLabel = "${device.label} (CH${channelNumber(it.deviceNetworkId)})"
+			it.setLabel(newLabel)
+		}
+		state.oldLabel = device.label
+	}
+	response(refresh() + configure())
```
mckeed: Better to not send this from updated() because configure() isn't actually dependent on device settings. It's being sent at the same time as configure() when the device is first added and jamming things up

mckeed (changes requested):
Other than the noted code issues, this is awesome. I love how well childDeviceTiles works.

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: So `configure()` but not `refresh()` should be called from `installed()`?

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: So we don't need a configure() method at all?

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made change

devicetypes/smartthings/zooz-power-strip.src/zooz-power-strip.groovy
bflorian: Made this change

bflorian:
How does this look now @mckeed ?

mckeed (approved):
Looks good

PR #1832, opened 2017-03-27
To Tyco Door-Window Sensor DTH: added Capability Sensor

CosmicPuppy:
There are some integrations (SmartApps) out there using the "Actuator" and "Sensor" Capabilities and instances of this Tyco Door-Window Sensor Device (and perhaps more) do not show up for them. _Example_ SmartApp "ActionTiles".
- Ref Docs: http://docs.smartthings.com/en/latest/device-type-developers-guide/overview.html?highlight=sensor%20actuator#actuator-and-sensor
- Ref Issue #1058.
- Ref @tslagle13 for more info regarding **ActionTiles**'s use of these standard Capabilities as device authorization input filters.
- Internal Ref: [`ActionTiles's Topic #1216`](http://support.actiontiles.com/topics/1216-tyco-doorwindow-sensor-missing-from-authorizable-things/).

**CC:** @tylerlange , @workingmonk 

_Thank-you!_

### DIFF
--- a/devicetypes/smartthings/tyco-door-window-sensor.src/tyco-door-window-sensor.groovy
+++ b/devicetypes/smartthings/tyco-door-window-sensor.src/tyco-door-window-sensor.groovy
@@ -23,6 +23,7 @@ metadata {
 		capability "Refresh"
 		capability "Temperature Measurement"
 		capability "Health Check"
+		capability "Sensor"
 
 		command "enrollResponse"
 


CosmicPuppy:
Thanks @tylerlange!

Quick question: How do we estimate / know /determine the release date of the merges so we can follow-up with our Customers to retry Authorizing their now-ActionTiles compatible devices?

tylerlange:
@CosmicPuppy Releases should get out the following Tuesday, but there may be other delays, like hub firmware updates to update local DTHs, etc.

PR #1833, opened 2017-03-27
[CHF-558] Health Check ZigBee Switch

pchomal:
1. Added health check for Leviton Switch.
2. 'checkInterval' is kept at 22min.
3. It is a ZigBee reporting device with 10min check-in.
4. We've tested the checkInterval duration and the devices are marked OFFLINE within the stipulated time.
5. Added the README.md file.
6. Modified Typo changes for Keen Home Smart Vent.

@jackchi @ShunmugaSundar Please check and merge the changes.

### DIFF
--- a/devicetypes/smartthings/zigbee-switch.src/zigbee-switch.groovy
+++ b/devicetypes/smartthings/zigbee-switch.src/zigbee-switch.groovy
@@ -18,6 +18,7 @@ metadata {
         capability "Configuration"
         capability "Refresh"
         capability "Switch"
+        capability "Health Check"
 
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006"
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0006", outClusters: "0003, 0006, 0019, 0406", manufacturer: "Leviton", model: "ZSS-10", deviceJoinName: "Leviton Switch"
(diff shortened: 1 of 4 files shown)


devicetypes/smartthings/zigbee-switch.src/zigbee-switch.groovy
```diff
@@ -62,8 +63,18 @@ def parse(String description) {
         sendEvent(event)
     }
     else {
-        log.warn "DID NOT PARSE MESSAGE for description : $description"
-        log.debug zigbee.parseDescriptionAsMap(description)
+        def cluster = zigbee.parse(description)
```
jackchi: Is this portion necessary if we're not setting the customized `onOffConfig()` interval?

jackchi (changes requested):

devicetypes/smartthings/zigbee-switch.src/zigbee-switch.groovy
pchomal: Was thinking if device is online, then only the response to configure would be received in parse.
So, if device is inactive/offline, then in that case the aggressive checkInterval wouldn't be set.

If you feel it isn't necessary then will remove it.

pchomal:
@jackchi 
Made the requested changes.

jackchi (approved):
👍

PR #1834, opened 2017-03-27
[CHF-569] Health Check zwave-water-valve

parijatdas:
1. Added health check for Leak Intelligence Leak Gopher Water Shutoff Valve.
2. 'checkInterval' is kept at 32min.
3. Ping is implemented using refresh().
4. Added the fingerprint of the Leak Intelligence Leak Gopher Water Shutoff Valve with the manufacturer and product code obtained from the raw description after pairing.
5. We've tested the checkInterval duration and the devices are marked OFFLINE within the stipulated time.
6. Added the README.md file.
@jackchi @ShunmugaSundar Please check and merge the changes.

### DIFF
--- a/devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
+++ b/devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
@@ -14,12 +14,14 @@
 metadata {
 	definition (name: "Z-Wave Water Valve", namespace: "smartthings", author: "SmartThings") {
 		capability "Actuator"
+		capability "Health Check"
 		capability "Valve"
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
 
 		fingerprint deviceId: "0x1006", inClusters: "0x25"
+		fingerprint mfr:"0173", prod:"0003", model:"0002", deviceJoinName: "Leak Intelligence Leak Gopher Water Shutoff Valve"
 	}
 
 	// simulator metadata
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
```diff
@@ -54,9 +56,18 @@ metadata {
 }
 
 def updated() {
+	// Device-Watch simply pings if no device events received for 32min(checkInterval)
+	unschedule()
+	runEvery15Minutes("healthPoll")
```
jackchi: Let's use the Z-Wave reporting interval cmd class:
`zwave.wakeUpV2.wakeUpIntervalSet(...)`

jackchi (changes requested):

devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
mckeed: WakeUp is only available on sleepy devices.

We should just get rid of this healthPoll thing. The whole point of ping() was for devices like this that don't check in regularly or need to be polled for status.

devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
```diff
@@ -54,6 +56,8 @@ metadata {
 }
 
 def updated() {
```
jackchi: Let's have `checkInterval` in installed() method too.

parijatdas: `checkInterval` event put in installed() method as well.

jackchi (approved):

PR #1818, opened 2017-03-22
Added Health Check Implementation for Spruce Controller.

skt123:
@jackchi @ShunmugaSundar 
Added Health check for Spruce Controller with checkInterval of 12 mins.
Please review and merge.

PR #1843, opened 2017-03-28
Revert "Added Health Check Implementation for Spruce Controller."

workingmonk:
Reverts SmartThingsCommunity/SmartThingsPublic#1818

Reverting this to resolve the merge issue and other changes made by Donald.

cc @jackchi @dkirker


dkirker:
@jackchi I will be emailing you and @natec007 about this tomorrow. We will want to bring the changes into the code that they were working to deploy (which is the cause for this revert) and make sure that it is tested. I also want to make sure that We (SmartThings) and the partner (Plaid Systems) are aware of each others' efforts.

PR #1846, opened 2017-03-28
ICP-448 Thermostat resource works weirdly.

marstorp:
Mapping between ST and SC fails at times due to improper implementation of capability attributes.
Some violates the attribute types, others are not initialized to a valid value, most due to not having one.
However, mapping requires right type and values.
This PR fixes Ecobee thermostats attributes.

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -32,7 +32,7 @@ metadata {
 		command "switchMode"
 		command "switchFanMode"
 
-		attribute "thermostatSetpoint", "number"
+		attribute "displayThermostatSetpoint", "string" // Added to be able to show "Auto"/"Off" keeping attribute thermostatSetpoint a number
 		attribute "thermostatStatus", "string"
 		attribute "maxHeatingSetpoint", "number"
 		attribute "minHeatingSetpoint", "number"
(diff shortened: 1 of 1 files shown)


marstorp:
@dsainteclaire @workingmonk @juano2310  please review 👁‍🗨

workingmonk:
will changing the attribute names cause issues for existing smartapps?

marstorp:
I guess it could but in that case those smart apps are not using the capabilities and attributes according to the specifications.

workingmonk:
well considering we didn't document our custom attributes, we should check to see if it hurts any smartapp. also, any chance we can keep the existing custom attribute and create a new one as a separate attribute until we are sure that removing the old one won't hurt anything

marstorp:
I didn't change any custom attributes, but I did add a new so the UI could still be showing a number and a string when needed.

workingmonk (approved):

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -70,7 +70,7 @@ metadata {
 			state "heat", action:"switchMode",  nextState: "updating", icon: "st.thermostat.heat"
 			state "cool", action:"switchMode",  nextState: "updating", icon: "st.thermostat.cool"
 			state "auto", action:"switchMode",  nextState: "updating", icon: "st.thermostat.auto"
-			state "auxHeatOnly", action:"switchMode", icon: "st.thermostat.emergency-heat"
+			state "emergency heat", label:"auxHeatOnly", action:"switchMode", icon: "st.thermostat.emergency-heat" // emergency heat = auxHeatOnly
 			state "updating", label:"Working", icon: "st.secondary.secondary"
```
dsainteclaire, line 74: Does the "updating" state break anything in the mapping since that shouldn't be mapped to an OCF state?

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -32,7 +32,7 @@ metadata {
 		command "switchMode"
 		command "switchFanMode"
 
-		attribute "thermostatSetpoint", "number"
```
dsainteclaire, line 35: Since the capability doesn't expose the existence of this attribute, I assume we should be good removing it because no SA should be trying to use it, and if they are they aren't complying with the model.  Right?

dsainteclaire (approved):
Just clarification questions, but otherwise seems good to me

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: This attribute, "thermostatSetpoint" is part of the Thermostat capability and thus doesn't need to be declared by the DTH. SA can use it and the potential issue with existing SA is if they used it and expect the invalid values "Auto" and "Off" to be used. If they are they don't comply with the current model, however if the still need this they can use the new custom attribute instead.

PR #1857, opened 2017-03-30
DPROT-263 EBF-1205 DVCSMP-2420 Support config on both temp-humidty firmwares

varzac:
This solves an issue with invalid configuration of the reporting for
humidity on the smartsense-temp-humidity-sensor.  This issue was masked
by the fact that the device has a default configuration for reporting
humidity, but the configuration has never been valid as it used an
incorrect dataType for the configuration.

This resolves: https://smartthings.atlassian.net/browse/DPROT-263
This resolves: https://smartthings.atlassian.net/browse/EBF-1205
This resolves: https://smartthings.atlassian.net/browse/DVCSMP-2420

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/smartsense-temp-humidity-sensor.src/smartsense-temp-humidity-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-temp-humidity-sensor.src/smartsense-temp-humidity-sensor.groovy
@@ -135,10 +135,7 @@ def refresh() {
 	return zigbee.readAttribute(0xFC45, 0x0000, ["mfgCode": 0x104E]) +   // New firmware
 			zigbee.readAttribute(0xFC45, 0x0000, ["mfgCode": 0xC2DF]) +   // Original firmware
 			zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000) +
-			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020) +
-			zigbee.configureReporting(0xFC45, 0x0000, DataType.INT16, 30, 3600, 100) +
-			zigbee.batteryConfig() +
-			zigbee.temperatureConfig(30, 300)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-temp-humidity-sensor.src/smartsense-temp-humidity-sensor.groovy
```diff
@@ -135,10 +135,7 @@ def refresh() {
 	return zigbee.readAttribute(0xFC45, 0x0000, ["mfgCode": 0x104E]) +   // New firmware
 			zigbee.readAttribute(0xFC45, 0x0000, ["mfgCode": 0xC2DF]) +   // Original firmware
 			zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000) +
-			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020) +
-			zigbee.configureReporting(0xFC45, 0x0000, DataType.INT16, 30, 3600, 100) +
-			zigbee.batteryConfig() +
-			zigbee.temperatureConfig(30, 300)
```
workingmonk, line 141: any reason you removed the config stuff from refresh? last i remember we added it in refresh so that users can override their old config using refresh button as we no longer have a configure button.

varzac: We have a configure button on the device page for support people to use if that is needed. And users shouldn't​ need to reconfigure their device as a normal thing. If they are writing their own custom DTH they can change it to do whatever they need however they want. It seemed an unnecessary amount of traffic to generate on every refresh.

tpmanley: It looks like we've moved away from this pattern in other device type handlers so I'm OK with doing it here too.

tpmanley (approved):

workingmonk (approved):

workingmonk:
added the ticket number to the title

PR #1866, opened 2017-04-03
ICP-448 Thermostat resource works weirdly.

marstorp:
Fixing change of auxHeatOnly to emergency heat.
Also adding main tile icon so Things view doesn’t show a grey blob

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -43,8 +43,8 @@ metadata {
 	}
 
 	tiles {
-		valueTile("temperature", "device.temperature", width: 2, height: 2) {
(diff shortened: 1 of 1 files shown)


marstorp:
@workingmonk @juano2310 @dsainteclaire , please review 🙄

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -185,6 +185,11 @@ def generateEvent(Map results) {
 				isChange = isStateChange(device, name, value.toString())
 				event['isStateChange'] = isChange
 				event['displayed'] = false
+			} else if (name == "thermostatMode") {
+				def mode = value.toString()
+				mode = (mode == "auxHeatOnly") ? "emergency heat" : mode
```
workingmonk, line 190: this would be better as an if-else. logic is right but makes code reading difficult

PR #1871, opened 2017-04-04
DVCSMP-2552 Added composite version of Aeon SmartStrip

bflorian:
New version of Aeon SmartStrip using composite devices for the switchable outlets. Removed fingerprint from old device but leaving it there for now because users may have smart apps that depend on its use of custom attributes and commands

### DIFF
--- a/devicetypes/smartthings/aeon-smartstrip-v2.src/aeon-smartstrip-v2.groovy
+++ b/devicetypes/smartthings/aeon-smartstrip-v2.src/aeon-smartstrip-v2.groovy
@@ -1,0 +1,370 @@
+/**
+ *  Copyright 2017 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+metadata {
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/aeon-smartstrip-v2.src/aeon-smartstrip-v2.groovy
```diff
@@ -0,0 +1,370 @@
...
+		}
+		children.each {
+			actions << encap(zwave.switchBinaryV1.switchBinaryGet(), channelNumber(it.deviceNetworkId))
+		}
+		actions << zwave.meterV3.meterGet(scale: 2).format()
+	}
+	if (actions) {
+		// Sending read requests
+		log.trace "........actions=$actions"
+		sendHubCommand(delayBetween(actions).collect {response(it)})
```
mckeed, line 181: It looks like `sendHubCommand` already adds delays between a list of actions, so I think also calling delayBetween here isn't a good idea. Also several more places below.

`response` can also take a list of actions, but it looks like sendHubCommand doesn't accept HubMultiAction yet.

devicetypes/smartthings/aeon-smartstrip-v2.src/aeon-smartstrip-v2.groovy
```diff
@@ -0,0 +1,370 @@
...
+
+def refresh() {
+	def actions = [
+			zwave.switchBinaryV1.switchBinaryGet().format(),
+	]
+	//for (ep in 1..4) {
+	//	actions << encap(zwave.switchBinaryV1.switchBinaryGet(), ep)
+	//}
+	actions << zwave.meterV3.meterGet(scale: 0).format()
+	actions << zwave.meterV3.meterGet(scale: 2).format()
```
mckeed, line 263: Looks like `meterGet(scale: 2)` is sent from the switchBinaryReport handler, so this is redundant

mckeed (commented):
Looks good. Sorry for the delay.

PR #1902, opened 2017-04-12
DVCSMP-2568 Arrival-sensor-ha Read battery level on join

workingmonk:
In configure we were setting up the battery level to be reported every
20 seconds but we weren't reading the attribute so it would take 20
seconds before the battery level was reported. Now we read it during
configure so it is updated right away after join.

https://smartthings.atlassian.net/browse/DVCSMP-2568


workingmonk:
cc @tpmanley

PR #1903, opened 2017-04-12
DVCSMP-2570 Adding centralite zigbee button fingerprint

workingmonk:


dbradmit:
Should be fine as is but should we add the number of buttons?

`    else if ((device.getDataValue("manufacturer") == "CentraLite") &&
            ((device.getDataValue("model") == "3130") || (device.getDataValue("model") == "3130"))) {
        sendEvent(name: "numberOfButtons", value: 2)`

workingmonk:
good catch @dbradmit 
I've updated the code to make it more optimized for the Centralite case. 
@tpmanley mind taking a look?

tpmanley (approved):

PR #1912, opened 2017-04-17
[ICP-691] Set level on sengled bulbs configure

varzac:
This works around the sengled bulbs not following the ZCL spec and using
the value 0xFF as a valid level (the spec says it should be invalid).
It does this by manually setting the level of the bulb to
0xFE (SmartThings 100%) when it is configured, or if it previously had a
level set, manually setting it to that value.  This avoids the issue of
the bulb reporting 0xFF and having the library ignore it because it is
an invalid value.

This resolves: https://smartthings.atlassian.net/browse/ICP-691

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
@@ -114,9 +114,17 @@ def refresh() {
 
 def configure() {
     log.debug "Configuring Reporting and Bindings."
-
+    def cmds = []
+    if (device.getDataValue("manufacturer") == "sengled") {
+        def startLevel = 0xFE
+        if ((device.currentState("level")?.value != null)) {
+            startLevel = Math.round(Integer.parseInt(device.currentState("level").value) * 0xFE / 100)
+        }
+        // Level Control Cluster, command Move to Level, level start level, transition time 0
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
```diff
@@ -114,9 +114,17 @@ def refresh() {
...
     log.debug "Configuring Reporting and Bindings."
-
+    def cmds = []
+    if (device.getDataValue("manufacturer") == "sengled") {
+        def startLevel = 0xFE
+        if((device.currentState("level")?.value != null)) {
+            startLevel = Math.round(Integer.parseInt(device.currentState("level").value) * 0xFE / 100)
+        }
+        // Level Control Cluster, command Move to Level, level start level, transition time 0
+        cmds << zigbee.command(0x0008, 0x00, sprintf("%02X0000", startLevel))
```
tpmanley: Can use LEVEL_CONTROL_CLUSTER for the cluster

tpmanley (approved):

workingmonk:
is there a ticket number for this issue? and why is it against production? can it wait till tomm?

varzac:
@workingmonk There is a link in the description.  I can point it against whatever, I don't know how these issues reported from Samsung have been handled.  Tom thought they might have been going out as hotfixes.

varzac:
@workingmonk Oliver will be QAing.  Feel free to merge when you are ready.

PR #1923, opened 2017-04-20
DVCSMP-2585 Fix NullPointerException when pairing because color is unset

dkirker:
Also tweak default config values in config value range checking (affects initial pairing of device)
Move change history to README.md

### DIFF
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -110,7 +110,6 @@ metadata {
 
 } // end metadata
 
-
 // Parse incoming device messages from device to generate events
 def parse(String description){
 	//log.debug "==> New Zwave Event: ${description}"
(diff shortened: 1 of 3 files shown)


dkirker:
@ShilpaMathew @workingmonk This PR fixes a couple of minor bugs, as well as moves the release history to README.md (and formatted to accept device health notes and other notes).

Vinay, I left the commit for the move to README.md separate because I wanted to have it in the commit history since it is, in a sense, a "restructure".

devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -203,7 +181,7 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd) {
-    if (cmd.value == 0 && device.latestState("color").value != "#ffffff") {
+    if (device.latestState("color") == null || (cmd.value == 0 && device.latestState("color").value != "#ffffff")) {
         sendEvent(name: "color", value: "#ffffff", displayed: true)
```
workingmonk: not sure why you are setting the color value to #ffffff every time off is received?

dkirker: It appears that a comment I added in the code did not appear. I am going to update that. Basically, the device changes the color to #ffffff in its firmware after off.

workingmonk: even when the off is sent from the app and not physically turned off?

dkirker: If it gets an off from us, or someone unplugs it, it defaults to white on the next on. It is really meant to be a night-light or a status light, which is why I am not being super picky about it maintaining state.

dkirker: I added a comment just above the code. I am going to double check querying the device for its current color, but if I recall we couldn't query for all color components at once and had to query red, green, and blue separately.

dkirker:
@mckeed Pinging you for a code review

workingmonk (approved):

PR #1927, opened 2017-04-20
[DVCSMP-2595] Set scheduled executions for Lights off with no motion SA to overwrite

aaron-miller:
The "Lights Off with No Motion and Presence" SmartApp will create a scheduled execution each time motion is detected by a sensor. This should be changed to instead overwrite any existing scheduled executions to avoid unnecessary executions.


aaron-miller:
@Little-Johnny That behavior should be the same with this change. When the scheduled execution runs it checks all of the motion sensor's states for presence via the `isActivePresence()` method.

rappleg (approved):

PR #1934, opened 2017-04-21
DVCSMP-2573: Acceleration axis validation (Multi sensor DTH)

swstack:
In certain cases the SmartSense Multi Sensors are
missing the Y and Z axis, causing an exception
during .parseAxis(). This change checks that all
3 axis are present before processing the rest of
the message.

### DIFF
--- a/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
@@ -192,6 +192,10 @@ private List<Map> parseAxis(List<Map> attrData) {
 	def y = hexToSignedInt(attrData.find { it.attrInt == 0x0013 }?.value)
 	def z = hexToSignedInt(attrData.find { it.attrInt == 0x0014 }?.value)
 
+	if ([x, y ,z].any { it == null }) {
+		return []
+	}
+
 	def xyzResults = [:]
 	if (device.getDataValue("manufacturer") == "SmartThings") {
 		// This mapping matches the current behavior of the Device Handler for the Centralite sensors
@@ -372,6 +376,10 @@ def updated() {
 }
 
 private hexToSignedInt(hexVal) {
+	if (!hexVal) {
+		return null
+	}
+
 	def unsignedVal = hexToInt(hexVal)
 	unsignedVal > 32767 ? unsignedVal - 65536 : unsignedVal
 }


devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -188,9 +188,20 @@ private List<Map> handleAcceleration(descMap) {
 
 private List<Map> parseAxis(List<Map> attrData) {
 	def results = []
-	def x = hexToSignedInt(attrData.find { it.attrInt == 0x0012 }?.value)
-	def y = hexToSignedInt(attrData.find { it.attrInt == 0x0013 }?.value)
-	def z = hexToSignedInt(attrData.find { it.attrInt == 0x0014 }?.value)
+
+	def attrsFound = []
+	for (attrId in [0x0012, 0x0013, 0x0014]) {
```
varzac: So, using a for loop is not typically the groovy way, and it also makes debugging from sumo traces tougher (since we will only have the line number, and won't know where in the loop it is).  I would instead, leave the definitions of x, y, and z the same and do something like

```
if ([x,y,z].any { it == null }) {
    return []
}
```

swstack: Yeah we just need to defer the calling of `hexToSignedInt` until after that check.

varzac: Right, I misspoke above about leaving the definitions unchanged.  You're correct that we need to remove the change to int until after this check.  Unfortunately for us, all this handling of acceleration is gross.

swstack: Or I can add a `null` check inside of `hexToSignedInt` and then return `null` instead of stack-tracing ... and then your check will work...what u think?

swstack:
I amended the commit, please re-review.

swstack:
Oh also ... not sure what's going on with tabs vs spaces in this project ... i believe this DTH used tabs so my editor just followed suit.

varzac (approved):

varzac:
@swstack In general everything cloud side is tabs, however, you will find SmartThingsPublic pretty inconsistent as it has depended on who made changes and when.  I would say, feel free to convert any file over to tabs, or if you want to keep your changes small just match what the file you are working in uses.

vl4ds (approved):
:+1:

PR #1940, opened 2017-04-24
[DVCSMP-2597] Fix sengled use of FF for max level

varzac:
This works around the fact that sengled element touches can get back
into a state of reporting an invalid value (of FF) if the physical
button on the bulb is used to cycle back to the 100% state.  Here we
interpret FF as FE for sengled and then issue a move to level command to
put it in a state where it will report FE until the level is changed
again.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-2597

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
@@ -70,19 +70,27 @@ def parse(String description) {
         else {
             sendEvent(event)
         }
-    }
-    else {
-        def cluster = zigbee.parse(description)
-        if (cluster && cluster.clusterId == 0x0006 && cluster.command == 0x07) {
-            if (cluster.data[0] == 0x00){
+    } else {
+        def descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap && descMap.clusterInt == 0x0006 && descMap.commandInt == 0x07) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
```diff
@@ -70,19 +70,22 @@ def parse(String description) {
...
                 log.debug "ON/OFF REPORTING CONFIG RESPONSE: " + cluster
                 sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
-            }
-            else {
+            } else {
                 log.warn "ON/OFF REPORTING CONFIG FAILED- error code:${cluster.data[0]}"
             }
-        }
-        else {
+        } else if (device.getDataValue("manufacturer") == "sengled" && descMap && descMap.clusterInt == 0x0008 && descMap.attrInt == 0x0000) {
```
tpmanley, line 82: It would be good to add a comment here describing why this is being done.

tpmanley (approved):

devicetypes/smartthings/zigbee-dimmer-power.src/zigbee-dimmer-power.groovy
varzac: Sure

PR #1942, opened 2017-04-24
DVCSMP-2598 Fix Wattvision throwing HttpResponseException

dkirker:
Fix start_time and end_time date range
Add descriptive error for Wattvision API errors

### DIFF
--- a/smartapps/smartthings/wattvision-manager.src/wattvision-manager.groovy
+++ b/smartapps/smartthings/wattvision-manager.src/wattvision-manager.groovy
@@ -136,10 +136,20 @@ def getDataForChild(child, startDate, endDate) {
 
 	def wattvisionURL = wattvisionURL(child.deviceNetworkId, startDate, endDate)
 	if (wattvisionURL) {
-		httpGet(uri: wattvisionURL) { response ->
-			def json = new org.json.JSONObject(response.data.toString())
-			child.addWattvisionData(json)
-			return "success"
+		try {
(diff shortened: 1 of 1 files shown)


dkirker:
@workingmonk @joneilloine I also mentioned you both in the ticket. It will have some further comment.

smartapps/smartthings/wattvision-manager.src/wattvision-manager.groovy
```diff
@@ -164,9 +174,14 @@ def wattvisionURL(senorId, startDate, endDate) {
 	if (diff > 259200000) { // 3 days in milliseconds
 		// Wattvision only allows pulling 3 hours of data at a time
 		startDate = new Date(hours: endDate.hours - 3)
+	} else if (diff < 10000) { // 10 seconds in milliseconds
+		// Wattvision throws errors when the difference between start_time and end_time is 5 seconds or less
+		// So we are going to make sure that we have a few more seconds of breathing room
+		use (groovy.time.TimeCategory) {
+			startDate = startDate - 10.seconds
```
bflorian: It's a minor point, but if `diff` is 9 sec, then this logic is going to query for a range of 19 sec rather than the minimum of 10. Probably no one will notice but maybe end date minus 10 would be easier to explain.

dkirker: Thanks! I updated it with the latest commit.

PR #1944, opened 2017-04-25
DVCSMP-2579 Move Centralite Door/Window sensor from multi-sensor DTH to open-closed-sensor DTH

dkirker:
Remove from smartsense-multi-sensor.groovy and add to smartsense-open-closed-sensor.groovy


dkirker:
@workingmonk I tested this out. Basically, the fingerprint should not have been in the multi DTH.

PR #1951, opened 2017-04-28
ICP-770 Feature/color changes

workingmonk:
cc @larsfinander @tpmanley


workingmonk:
@larsfinander for the hue values, do we want the reported value in 0-360 or in 0-100? i don't remember what was finally agreed upon. this mapping will probably affect the OCF mapping too. the zll ones right now have 360 and zigbee ha has 100.

larsfinander:
@workingmonk We always report in 100, dove/dealer takes care of converting it to 360 for OCF

PR #1964, opened 2017-05-02
Add statefulness with override parameter

james-powis:
Many people complained that the virtual thermostat had 2 things lacking:
  - It always turned on or off the switch at every polling interval regardless of the
    current state This was chatty and heavy handed.
  - If someone turned on the switch (or another smartapp did) it would
    immediately get turned off (and vice versa)

To address these issues Andrew Vaughan created the improved virtual
thermostat. Reference: https://community.smartthings.com/t/virtual-thermostat-update/12398/3

However I found the code provided in that link suffered from three deficiencies:
  - It did not leverage the threshold to help with short cycling
  - It did not account for where the lastTemp equals the desiredTemp,
    thus causing the on actions to never trigger as the temperature fell
    below the set point.
  - and under certain circumstances you may want the enforcement (always
    ensure its on or off every poll cycle).

This commit merges the two methods and allows for an override between
the two different behaviors.

### DIFF
--- a/smartapps/smartthings/virtual-thermostat.src/virtual-thermostat.groovy
+++ b/smartapps/smartthings/virtual-thermostat.src/virtual-thermostat.groovy
@@ -30,6 +30,7 @@ preferences {
 	}
 	section("Select the heater or air conditioner outlet(s)... "){
 		input "outlets", "capability.switch", title: "Outlets", multiple: true
+		input "enforce", "bool", title: "Enforce Outlet State", required: true, defaultValue: false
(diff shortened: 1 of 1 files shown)


james-powis:
Reopened accidentally deleted source branch

james-powis:
Can I get a review please this 7 months old today...

james-powis:
Closing due to the absolute awfulness of the smarthings community. I am tired of seeing this open, and have no intention of working to fix this long term.

PR #1967, opened 2017-05-03
DVCSMP-2606 Fix recent regression with setting color temperature

tpmanley:
This fixes a recently introduced bug that caused an exception while
attempting to set the color temperature on any of these three DTHs.

https://smartthings.atlassian.net/browse/DVCSMP-2606


workingmonk (approved):

varzac (approved):
It would be nice to get out my generated code at some point so we don't have a hundred different versions of the same constants across all the DTHs.

PR #2001, opened 2017-05-15
DVCSMP-2493 Notify Multi-Channel Control app when root device is deleted

mckeed:
See https://github.com/SmartThingsCommunity/SmartThingsPrivate/pull/1956


bflorian:
👍

PR #2003, opened 2017-05-16
ICP-781 Fixing DTH for valve capability

unixbeast:
@workingmonk Please review this as I was not able to test it and I'm not sure if this is best practice.

Thanks,
Dave

### DIFF
--- a/devicetypes/smartthings/fortrezz-water-valve.src/fortrezz-water-valve.groovy
+++ b/devicetypes/smartthings/fortrezz-water-valve.src/fortrezz-water-valve.groovy
@@ -12,13 +12,13 @@
  *
  */
 metadata {
-	definition (name: "Fortrezz Water Valve", namespace: "smartthings", author: "SmartThings") {
+	definition (name: "Fortrezz Water Valve", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.watervalve") {
 		capability "Actuator"
 		capability "Health Check"
 		capability "Valve"
 		capability "Refresh"
 		capability "Sensor"
-        
+
 		fingerprint deviceId: "0x1000", inClusters: "0x25,0x72,0x86,0x71,0x22,0x70"
 		fingerprint mfr:"0084", prod:"0213", model:"0215", deviceJoinName: "FortrezZ Water Valve"
 	}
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/zwave-water-valve.src/zwave-water-valve.groovy
```diff
@@ -59,18 +59,22 @@ def updated() {
...
-		result = createEvent(zwaveEvent(cmd))
+		def events = zwaveEvent(cmd)
+		def evt1 = createEvent(events[0])
+		log.debug "Parse returned ${evt1?.descriptionText}"
+		def evt2 = createEvent(events[1])
+		log.debug "Parse returned ${evt2?.descriptionText}"
 	}
-	log.debug "Parse returned ${result?.descriptionText}"
-	return result
+	return [evt1, evt2]
```
workingmonk: the variables are defined inside the if block. you can keep the code as is in the parse method, and createEvent is automatically handle the multiple events in the array

unixbeast: ah. ok. I didn't know if create event could handle the multiple events like that. I'll change it

jimmyjames: I think @workingmonk is saying that `parse()` can return a list of events created via `createEvent()` as it was before. I don't see any `createEvent()` API that accepts a list of event info, it only takes a single event definition via a `Map`.

unixbeast: That's how I had it before. I was following the example outlined here: http://docs.smartthings.com/en/latest/ref-docs/device-handler-ref.html?highlight=createEvent#createevent

@workingmonk, can you clarify what you mean by pointing me at an example maybe?

unixbeast:
@workingmonk @jimmyjames 
Ok. I think I have something that will work now. I was able to successfully run this code in the simulator. 
I also noticed that the fortrezz water valve DTH needs the same update. So I added it.
Let me know what you think.

devicetypes/smartthings/fortrezz-water-valve.src/fortrezz-water-valve.groovy
```diff
@@ -62,22 +62,30 @@ def updated(){
...
+	def events = [[name: "contact", value: value, descriptionText: "$device.displayName valve is $value"],
+	[name: "valve", value: value, descriptionText: "$device.displayName valve is $value"]]
+
+	def eventList = []
+	events.each { event ->
+		def evt = createEvent(event)
+		log.debug "Parse returned ${evt?.descriptionText}"
+		eventList << evt
+	}
+	return eventList
```
jimmyjames: Somewhat semantic but couldn't we just do
```
return [createEvent(...), createEvent(...)]
```
and remove the need to loop?

jimmyjames (commented):
lgtm overall. Added one comment but it should work either way.

unixbeast:
@workingmonk should I also update:
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/ccfe5dabdf9a6504a3615bc08d16e32833d4f512/devicetypes/smartthings/zigbee-valve.src/zigbee-valve.groovy

PR #2023, opened 2017-05-22
[DVCSMP-2668] Fix NPE for smartsense multi

varzac:
There was a common NPE caused by assuming there would be additional
attributes reported with one acceleration value.  This corrects it to
instead properly handle the case where there weren't additional
attributes.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-2668

@tpmanley @workingmonk


tpmanley (approved):
Jira in the title

PR #2024, opened 2017-05-22
[ICP-975] Yale Locks fingerprint updates

jackchi:
Updates Yale lock fingerprints for 2 changes:
* `mfr` 0109 & 0129 compatible
* ignore model generational differences on locks

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -37,15 +37,12 @@ metadata {
 		fingerprint mfr:"003B", prod:"6341", model:"0544", deviceJoinName: "Schlage Camelot Touchscreen Deadbolt Door Lock"
 		fingerprint mfr:"003B", prod:"6341", model:"5044", deviceJoinName: "Schlage Century Touchscreen Deadbolt Door Lock"
 		fingerprint mfr:"003B", prod:"634B", model:"504C", deviceJoinName: "Schlage Connected Keypad Lever Door Lock"
-		// Yale mfr could be both 0109 or 0129 for prod < 0005
-		fingerprint mfr:"0109", prod:"0001", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
-		fingerprint mfr:"0129", prod:"0001", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
-		fingerprint mfr:"0109", prod:"0002", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120, YRD220, YRD240
-		fingerprint mfr:"0129", prod:"0002", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120, YRD220, YRD240
-		fingerprint mfr:"0109", prod:"0002", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
-		fingerprint mfr:"0129", prod:"0002", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
-		fingerprint mfr:"0109", prod:"0004", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110, YRD210
-		fingerprint mfr:"0129", prod:"0004", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110, YRD210
+		fingerprint mfr:"0129", prod:"0002", model:"0800", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120
+		fingerprint mfr:"0129", prod:"0002", model:"0000", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD220, YRD240
+		fingerprint mfr:"0129", prod:"0002", model:"FFFF", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
+		fingerprint mfr:"0129", prod:"0004", model:"0800", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110
+		fingerprint mfr:"0129", prod:"0004", model:"0000", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD210
+		fingerprint mfr:"0129", prod:"0001", model:"0000", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
 		fingerprint mfr:"0129", prod:"8002", model:"0600", deviceJoinName: "Yale Assure Lock with Bluetooth"
 	}
 
@@ -89,13 +86,13 @@ import physicalgraph.zwave.commands.doorlockv1.*
 import physicalgraph.zwave.commands.usercodev1.*
 
 def installed() {
-	// Device-Watch simply pings if no device events received for 32min(checkInterval)
-	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch simply pings if no device events received for 60min(checkInterval)
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 }
 
 def updated() {
-	// Device-Watch simply pings if no device events received for 32min(checkInterval)
-	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch simply pings if no device events received for 60min(checkInterval)
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	try {
 		if (!state.init) {
 			state.init = true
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -37,12 +37,15 @@ metadata {
...
 		fingerprint mfr:"003B", prod:"634B", model:"504C", deviceJoinName: "Schlage Connected Keypad Lever Door Lock"
-		fingerprint mfr:"0129", prod:"0002", model:"0800", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120
-		fingerprint mfr:"0129", prod:"0002", model:"0000", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD220, YRD240
-		fingerprint mfr:"0129", prod:"0002", model:"FFFF", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
-		fingerprint mfr:"0129", prod:"0004", model:"0800", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110
-		fingerprint mfr:"0129", prod:"0004", model:"0000", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD210
-		fingerprint mfr:"0129", prod:"0001", model:"0000", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
+		// Yale mfr could be both 0109 or 0129 for prod < 0005
+		fingerprint mfr:"0109", prod:"0001", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
+		fingerprint mfr:"0129", prod:"0001", deviceJoinName: "Yale Push Button Lever Door Lock" // YRD210
```
surfous: for lines 41 and 42, prod `0001` should be `Yale Touchscreen Lever Door Lock`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -37,12 +37,15 @@ metadata {
...
+		fingerprint mfr:"0109", prod:"0002", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120, YRD220, YRD240
+		fingerprint mfr:"0129", prod:"0002", deviceJoinName: "Yale Touchscreen Deadbolt Door Lock" // YRD120, YRD220, YRD240
```
surfous: lines 43 and 44 for prod `0002` are correct

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -37,12 +37,15 @@ metadata {
...
+		fingerprint mfr:"0109", prod:"0002", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
+		fingerprint mfr:"0129", prod:"0002", deviceJoinName: "Yale Touchscreen Lever Door Lock" // YRD220
```
surfous: Lines 46 and 47 have both the wrong prod and deviceJoinName.
prod should be `0003`
deviceJoinName should be `Yale Push Button Lever Door Lock`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -37,12 +37,15 @@ metadata {
...
+		fingerprint mfr:"0109", prod:"0004", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110, YRD210
+		fingerprint mfr:"0129", prod:"0004", deviceJoinName: "Yale Push Button Deadbolt Door Lock" // YRD110, YRD210
```
surfous: Lines 47 & 48 for prod `0004` are correct

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -37,12 +37,15 @@ metadata {
...
 		fingerprint mfr:"0129", prod:"8002", model:"0600", deviceJoinName: "Yale Assure Lock with Bluetooth"
```
surfous: This line would not pair the Assure T1L that I have, because the model of my device is `1600`, and mine was purchased through the SmartThings store. Please remove the model for now. I don't know if the `0600` is a typo or if `0600` is a product id in addition to `1600`

This is my Assure lock's raw description:
`zw:Fs type:4003 mfr:0129 prod:8002 model:1600 ver:1.74 zwv:4.24 lib:03 cc:5E,72,98,5A,73,86 sec:80,62,85,59,71,70,63,8A,8B,4C,4E,7A role:07 ff:8300 ui:8300`

surfous (changes requested):
The fingerprints need to be modified as directed in the line comments, then I'll approve for merge. 

To nail this (not blocking this merge/deploy, though) we need to get specific advice from Yale about their use of the Product ID component of the MSR (what we call model in the raw description). For example, can we use model to distinguish between the keyed and keyless deadbolts so we can name devices even more accurately?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
jackchi: According to WWST & Z-Wave alliance YRD/YRL 210 are `Yale Push Button Lever Door Lock`. 
https://www.smartthings.com/products/-/filter/brands/yale
http://products.z-wavealliance.org/regions/2/categories/6/products?company=158
Why do you say it's Touchscreen?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
jackchi: Both of these have the `prod` as `0002` here:
* http://products.z-wavealliance.org/products/630
* http://products.z-wavealliance.org/products/631
And the names are `Yale Touchscreen Lever Door Lock` according to:
https://www.smartthings.com/products/-/filter/brands/yale

Which one is right?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
jackchi: Duncan pointed me to:
http://products.z-wavealliance.org/products/1972
for the `prod` & `model` #

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
surfous: Who says product type ID 0001 maps to YR*210? ;)
I'm using Yale's reference for my assertion - that document I sent you, bottom of the first page.

Also, A YRD cannot be a lever lock. That D means Deadbolt.

surfous: Hrm. Yale seem to have rewritten history, but seeing how they've treated model (Product ID), I think they backed off on having that represent generation.  They've also made a bit of a mess of prod (Product Type ID) . Their original mapping was:

```
0x01 (Touch Lever)
0x02 (Touch Deadbolt)
0x03 (Push Button Lever) 
0x04 (Push Button Deadbolt)
```
The Z-Wave allianace spec sheets claim the touch lever lock (YRL220) to have a prod of `0002`, however, I suspect the alliance's spec sheet for the YRL220 is incorrect, and if not incorrect, then there has been a change which is not reflected in the current spec sheets. I'm not quick to point to the body behind the protocol as having wrong information, but evidence seems to point that way. 

I have a YRL220. This is from my YRL220 device via the IDE right now:

```
fw: 21.32
endpointId: 0
MSR: 0129-0001-0000
```

This agrees with the Yale doc I sent you and implies that the ZWA spec sheet for the YRL220 is incorrect. Knock-on is that the YRL210 (Push Button Lever) spec sheet is also incorrect, and the Product Type ID there should be `0003`.  Not sure what to trust, but from what I can see, the original mapping I listed above (from that Yale developer doc I sent you) appears to be correct. That covers the "original four" devices. I guess we trust the ZWA with later ones (like the keyless deadbolts)? 

@mckeed, any thoughts or advice here?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
surfous: I'm confident Z-Wave alliance has errors in some of these spec sheets. See my long comment above.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
surfous: That one in the ZWA doc that Duncan pointed you to is for the push button assure deadbolt, `prod 8004, model 0600`

It looks like all three of the touchscreen assure deadbolts listed have `prod 8002`. All are also listed to have `model 0600` regardless of having a keyway or not. They do not list any `prod 8002, model 1600`. Here's where I call foul. Clearly  the `model 1600` exists. I have one, we sell it, and its raw description is a few inches up the screen from here. Further, they claim that the YRD226-ZW2 is *both* an assure lock with a keyway and _without_ a keyway, both having `prod 8002` and `model 0600`. How does that make any sense?

- http://products.z-wavealliance.org/products/1973
- http://products.z-wavealliance.org/products/1878

Hint: the first link above is for the YRD246. The YRD226 has a keyway, the YRD246 does not. Another mistake in the ZWA data.

Look, the ZWA may have gotten erroneous information - I don't know. I'm guessing the data sheets sent by Yale make have been filled out quickly, with cut and paste errors or the like. In any case, I've seen several provable errors across different yale products at the ZWA product site that I've called out here. I don't trust it. 

After looking at this very closely now, I stand behind the changes I've requested. 
To capture all the Assure locks, use just the mfr and prod with the join names below. I don't know yet how to distinguish the bluetooth vs. the non-bluetooth touchscreen Assure locks, but this will get us reasonably correct in the meantime: 

`prod 8002` `Yale Assure Touchscreen Door Lock`
`prod 8004` `Yale Assure Push Button Door Lock`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -86,13 +86,13 @@ import physicalgraph.zwave.commands.doorlockv1.*
 import physicalgraph.zwave.commands.usercodev1.*
 
 def installed() {
-	// Device-Watch simply pings if no device events received for 32min(checkInterval)
-	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	// Device-Watch simply pings if no device events received for 60min(checkInterval)
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
```
workingmonk, line 90: this is 120 min, not 60 min as mentioned in the comment

mckeed:
I think 1 hour would make more sense for the checkInterval

PR #2029, opened 2017-05-23
Show state of sensor instead of ---

rboy1:
The way the ST app currently works it renders differently on Android and iOS, with iOS it doesn't show the state of the sensor unless the label is explicitly included. This setup works fine on both iOS and Android ST apps version 2.4.0


rboy1:
@workingmonk check out this sensor state display correction

workingmonk:
1. the label should probably be `label:'${currentValue}'` rather than the name
2. the PR should be against the master branch.

PR #2038, opened 2017-05-24
ICP-1038 - Add delay

juano2310:


varzac:
Does this fix the issue?  `zigbee.command` has a default delay added of 2 seconds, so unless the transition time is greater than 2 seconds, I wouldn't expect that to be the issue.  However, with a value of "0000" for the transition time, it should transition immediately.  

EDIT: Oops, was looking at the wrong command, the `setHue` command is OK.

workingmonk:
@varzac are you ok with merging this or want Juan to look into something else? i haven't kept track of the zigbee library changes, so will defer on you for the final approval

varzac:
I'm just skeptical that this actually solves the problem.  Juan said he doesn't have a device to test, and looking at the commands it is using I don't think adding a half second delay will solve it.  I don't think this will break anything, but I would like some confirmation that this is a solution before merging.

PR #2039, opened 2017-05-25
ICP-1075 Set default carbonMonoxide when detector checks in

dkirker:


devicetypes/smartthings/zwave-smoke-alarm.src/zwave-smoke-alarm.groovy
```diff
@@ -146,6 +146,13 @@ def zwaveEvent(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd, results) {
 			if(device.currentValue("smoke") != "clear") {
 				createSmokeOrCOEvents("smokeClear", results)
 			}
+
+			if (device.currentValue("carbonMonoxide") != "clear") {
+				def zwMap = getZwaveInfo()
+				if (zwMap && zwMap.mfr == "0138" && zwMap.prod == "0001" && zwMap.model == "0002") {
+					createSmokeOrCOEvents("carbonMonoxideClear", results)
+				}
```
workingmonk, line 161: should this kind of logic be part of the installed() method?

dkirker: Normally, yes. But this devices is weird. If you request an alarm state from it, it shoots back this "checkin" packet. No matter if you ask smoke state, or CO state, or whatever (except battery, of course). :/

workingmonk: but we want to set it to the default value once, right?

dkirker: Well, in this case, this particular message from the detector is an "all clear here" message. So, if smoke or carbonMonoxide are not marked as clear then we should.

workingmonk: But what happens at install? Is the "all clear" sent? the ticket complains that the attribute value is set to null if a default value is not set

dkirker: The all clear does get sent, which is what this PR is handling. However, I am doing it in the manner that already existed with the smoke state a few lines above it. Unfortunately, this detector does not send individual "clear" updates for each alarm type after install. So, unless we do it this way, we don't actually have an honest state from the alarm. The alternate is that I just send a clear event in the install method, but assuming the state of the device I would think would not be good.

Basically, what would happen before is this:
1. User installs device
2. Device sends this "all clear" packet
3. We set smoke to "clear" as per the logic here, giving the smoke state a non-null value (which would remain null otherwise, but the tile would show clear because that is what the first entry is in the description)
4. CO still remains null because nothing was done for that state
5. Only would CO state get set if CO was detected and the alarm triggered it

What this change does is also updates the CO state when we receive the "all clear" packet, since we receive nothing else for it. If we solicit an alarm state from the detector (so we say "what is your smoke state?" and/or "what is your CO state?" it only ever says "everything is all clear here", not "my smoke state is clear" and/or "my CO state is clear" like we would expect).

dkirker: I also just updated the code to explicitly solicit the alarm states, since we should be doing this regardless if the alarm gives them to us (as other alarms may not).

dkirker:
Since I am technically on vacation, I am going to cc @mckeed for code review and further comment on the conversation with @workingmonk 

My only concern (and it is with the DTH in general) is that this one supports both CO and non-CO supporting detectors, I believe, so that means that a detector that does not support CO will have that capability present (and this might mess up Samsung Connect).

PR #2044, opened 2017-05-25
[DHF-7] [ICP-1084] Adds scheduled 1 hr ping for offline devices; updates checkInterval to 1 hour

jackchi:
The `checkInterval` to 1 hr. It has an hourly `ping` when the lock is OFFLINE.


mckeed (commented):
What's the thinking behind changing the check interval to 2 hrs? I feel like 1 hr would make more sense.

mckeed (approved):

jackchi:
@workingmonk Korean QA tested this and found that it is staying `online`. We can push this as a hotfix.

PR #2047, opened 2017-05-26
ICP-1076 Fix Aeon Door/Window Sensor 6 initial value of null

dkirker:


dkirker:
cc @mckeed The other addition that I did was to change the delayBetween during the configure() from 6 seconds ( :o ) to 1/2 second because the Aeon Door/Window 6 was going to sleep long before the 18 seconds passed.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -192,9 +193,11 @@ def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd)
 	}
 	if (!state.lastbat || now() - state.lastbat > 53*60*60*1000) {
 		cmds << command(zwave.batteryV1.batteryGet())
-	} else {
-		cmds << zwave.wakeUpV1.wakeUpNoMoreInformation().format()
 	}
+	if (device.currentValue("contact") == null) { // Incase our initial request didn't make it
+		cmds << command(zwave.sensorBinaryV2.sensorBinaryGet(sensorType: zwave.sensorBinaryV2.SENSOR_TYPE_DOOR_WINDOW))
+	}
+	cmds << zwave.wakeUpV1.wakeUpNoMoreInformation().format()
```
mckeed: Should only send this when we don't send batteryGet because otherwise we send it from the BatteryReport handler.

dkirker: Ah, I see it in there, however, is that an appropriate place for it? Or does the device go to sleep before sending the battery report?

dkirker: I'll go ahead and re-arrange the code a little so that the contact check happens first

dkirker:
@mckeed Wen't ahead and updated the code

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -114,8 +114,9 @@ def updated() {
 def configure() {
 	commands([
 		zwave.manufacturerSpecificV2.manufacturerSpecificGet(),
-		zwave.batteryV1.batteryGet()
-	], 6000)
+		zwave.batteryV1.batteryGet(),
+		zwave.sensorBinaryV2.sensorBinaryGet(sensorType: zwave.sensorBinaryV2.SENSOR_TYPE_DOOR_WINDOW)
```
mckeed: We might send the wakeUpNoMoreInformation before this gets to the device.

Send sensorBinaryGet first, then manufacturerSpecificGet, and take batteryGet out of this. It will be sent from the ManufacturerSpecificReport handler.

It's not safe to send so many Get commands back to back in a generic DTH because they can back up and collide with some devices.

mckeed (changes requested):

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
dkirker: d'oh! Right.... The battery get has been removed.

mckeed (approved):
Looks good. I haven't tested it.

PR #2058, opened 2017-05-30
ICP-1124 - Sylvania SMART BR30 RGBW ' s color ticker moving randomly

juano2310:


devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -103,12 +103,12 @@ def parse(String description) {
 
         if (zigbeeMap?.clusterInt == COLOR_CONTROL_CLUSTER) {
             if(zigbeeMap.attrInt == ATTRIBUTE_HUE){  //Hue Attribute
-                def hueValue = Math.round(zigbee.convertHexToInt(zigbeeMap.value) / 0xfe * 100)
-                sendEvent(name: "hue", value: hueValue, descriptionText: "Color has changed")
+                state.hueValue = Math.round(zigbee.convertHexToInt(zigbeeMap.value) / 0xfe * 100)
+                runIn(5, updateColor, [overwrite: true])
```
marstorp, line 107: Would be nice with a comment on why runIn() is used and why the delay is set to 5 seconds, for maintainability purpose

marstorp (approved):
Nitpicking but for maintainability comments are always nice to have for the next person when it's not obvious why runIn is used.

juano2310:
@workingmonk do you mind merging this?

juano2310:
@tpmanley what do you think about this solution to prevent the color picker in SC from jumping when getting saturation and hue in separated events? It works but wondering if this can affect something else.

tpmanley:
I don't have any objections

PR #2070, opened 2017-06-06
ICP-1097 Fix Sengled Classic default level on pairing

dkirker:
Code brought oever from @varzac commit 8bcbe7b924eb10214a5a91709fba2779a44c3e39
Original commit message:

> Fix sengled use of FF for max level
> This works around the fact that sengled element touches can get back
> into a state of reporting an invalid value (of FF) if the physical
> button on the bulb is used to cycle back to the 100% state.  Here we
> interpret FF as FE for sengled and then issue a move to level command to
> put it in a state where it will report FE until the level is changed
> again.
> This resolves: https://smartthings.atlassian.net/browse/DVCSMP-2597

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
@@ -66,22 +66,29 @@ def parse(String description) {
         else {
             sendEvent(event)
         }
-    }
-    else {
-        def cluster = zigbee.parse(description)
-
-        if (cluster && cluster.clusterId == 0x0006 && cluster.command == 0x07) {
-            if (cluster.data[0] == 0x00) {
+    } else {
+        def descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap && descMap.clusterInt == 0x0006 && descMap.commandInt == 0x07) {
+            if (descMap.data[0] == "00") {
                 log.debug "ON/OFF REPORTING CONFIG RESPONSE: " + cluster
                 sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
-            }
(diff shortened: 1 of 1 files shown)


dkirker:
@workingmonk That sengled fix we were looking for the other day was in the zigbee-dimmer-power! Anyway, I brought it over to here. We may consider adding it to zigbee-switch if there is a sengled device there. Also, maybe it would be worth looking at the other affected devices? This is a discussion for later.

Let me know your thoughts.

varzac (approved):
I don't love having all this manufacturer specific stuff spread all over the place.  We could consider switching all sengled devices to their own DTH and the only have to worry about this fix one place, but that is another DTH to support.  I know @tpmanley has talked to sengled and they have given us firmware updates, I don't know how receptive they would be to fixing this in their bulb firmware.

varzac:
Actually since this bulb doesn't have a button, you could probably fix this in configure and then you wouldn't have to be doing an extra check every time parse is called.

dkirker:
@varzac Yeah, I definitely agree. I did check and the bulb seems to reset to 100% on re-pairing, so even sticking it in installed() is probably fine. Which, maybe that makes more sense? I just feel weird about making assumptions. It may make sense to put it there, "for now," and then actually get a firmware update (since we can do that these days). Let me know what you think. Either way, I am going to recommend that this not go out in this week's deploy to prod as I want QA to have a look.

dkirker:
@workingmonk Let me know if you think this is good to go? @tpmanley Do you have any comment on this? Since the bulb pairs at 100% brightness, and any other brightness is handled correctly, and since the bulb doesn't have a button to get the brightness up to "full", would it be better to just put a check with the others in installed()? Is there the chance that another device could actuate it outside of ST that would get it to send back FF instead of FE? Even with this code from the PR, the extra logic will only get called when the invalid FF packet comes in, so I don't see this being a computational burden. My concerns are more about consistency (across device types).

PR #2077, opened 2017-06-07
Fix reassociation for first gen Schlage locks PROB-1670

mckeed:
The response to BasicSet that fixes the association was broken. Also now sending the reassociation from installed() so lock/unlock notifications other than keypad unlocks will work right away (won't work without CHAD-1028 fix for raw desc values).

Note: still won't work with 0.17.x HubCore due to CHAD-1002

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -88,11 +88,22 @@ import physicalgraph.zwave.commands.usercodev1.*
 def installed() {
 	// Device-Watch pings if no device events received for 1 hour (checkInterval)
 	sendEvent(name: "checkInterval", value: 1 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+
+	// First generation Schlage locks send lock events to assoc group 2 and only BasicSet to group 1
+	if (zwaveInfo.mfr == "003B" && zwaveInfo.prod == "634B") {
+		def cmds = [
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -88,11 +88,22 @@ import physicalgraph.zwave.commands.usercodev1.*
 def installed() {
 	// Device-Watch pings if no device events received for 1 hour (checkInterval)
 	sendEvent(name: "checkInterval", value: 1 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+
+	// First generation Schlage locks send lock events to assoc group 2 and only BasicSet to group 1
+	if (zwaveInfo.mfr == "003B" && zwaveInfo.prod == "634B") {
```
tpmanley, line 93: Is the mfr check required anywhere else, for example in the BasicSet handler?

tpmanley:
There appears to be a number of places in the DTH that is trying to fix the association for this lock. Do we need all of those checks and if so could they be modified to call a common function to get the commands so there isn't so much duplicate code?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -88,11 +88,22 @@ import physicalgraph.zwave.commands.usercodev1.*
...
+		def cmds = [
+			new physicalgraph.device.HubAction(zwave.associationV1.associationRemove(groupingIdentifier:1, nodeId:zwaveHubNodeId).format()),
```
kianooshkarami, line 95: Do we need to remove group 1?

kianooshkarami (approved):

PR #2078, opened 2017-06-08
ICP-1110 Read valve state during pairing

dkirker:


dkirker:
From my comment on the ticket:

Basically, another device where we needed to check the state after pairing. Nothing too crazy. Also, added a check on the state after calls to on() and off() to a) be more consistent across Z-Wave devices b) make sure that we have the latest state.

Tagging @mckeed for a quick code review

mckeed (approved):

PR #2080, opened 2017-06-09
SLC-82 Marking old connect apps and device types as deprecated

workingmonk:


workingmonk:
FYI @larsfinander @marstorp

PR #2093, opened 2017-06-15
DVCSMP-2698 Temporary bandage to properly show valve states in the device tile of local running devices

dkirker:


dkirker:
@workingmonk I created it, just didn't tag you! 😨

workingmonk (approved):

PR #2094, opened 2017-06-15
IOTEVENT-786 check for null/empty values before sending events

unixbeast:
@twack 
FYI: @workingmonk 
some color, hue, and saturation events (color being the big offender) were being dropped from the event pipeline because they had empty values.

Checking for null/empty values before sending the events.

### DIFF
--- a/devicetypes/smartthings/fibaro-rgbw-controller.src/fibaro-rgbw-controller.groovy
+++ b/devicetypes/smartthings/fibaro-rgbw-controller.src/fibaro-rgbw-controller.groovy
@@ -300,15 +300,21 @@ def setColor(value) {
         value.hex = "#${hex(value.red)}${hex(value.green)}${hex(value.blue)}"
     }
 
-	sendEvent(name: "hue", value: value.hue, displayed: false)
-	sendEvent(name: "saturation", value: value.saturation, displayed: false)
-	sendEvent(name: "color", value: value.hex, displayed: false)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/fibaro-rgbw-controller.src/fibaro-rgbw-controller.groovy
```diff
@@ -300,15 +300,21 @@ def setColor(value) {
...
+    if(value.saturation) {
+        sendEvent(name: "saturation", value: value.saturation, displayed: false)
+    }
+    if(value.hex?.trim()) {
+        sendEvent(name: "color", value: value.hex, displayed: false)
+    }
+    if (value.level) {
+        sendEvent(name: "level", value: value.level)
+    }
+    if (value.switch?.trim()) {
```
workingmonk, line 315: any reason for the trim() operation?

unixbeast: trim() is a way to check for null or empty string

workingmonk: trim() only helps if the string is `"  "`. you don't need trim() if the string is null or `""`

since you null check before using trim() the operation is safe, only question is do you see a case where the value would be a string with spaces

PR #2097, opened 2017-06-19
[DHF-16] Added Health Check Implementation for Zigbee Button.

skt123:
@jackchi @ShunmugaSundar

Implemented Health check with default reporting interval of 10 mins.
CheckInterval : 22 mins

Please review and merge.

### DIFF
--- a/devicetypes/smartthings/zigbee-button.src/.st-ignore
+++ b/devicetypes/smartthings/zigbee-button.src/.st-ignore
@@ -1,0 +1,2 @@
+.st-ignore
+README.md
--- a/devicetypes/smartthings/zigbee-button.src/README.md
+++ b/devicetypes/smartthings/zigbee-button.src/README.md
@@ -1,0 +1,43 @@
+# ZigBee Button
+
+Cloud Execution
+
+Works with: 
+
+* [OSRAM LIGHTIFY Dimming Switch](https://support.smartthings.com/hc/en-us/articles/115000236823-SYLVANIA-Dimming-Switch)
+* [Iris Smart Button](https://support.smartthings.com/hc/en-us/articles/115000190186-Iris-Smart-Button)
+* [Iris KeyFob](https://support.smartthings.com/hc/en-us/articles/217409686-Iris-Smart-Fob)
+
+## Table of contents
+
+* [Capabilities](#capabilities)
+* [Health](#device-health)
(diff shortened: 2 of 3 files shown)


jackchi:
@tpmanley @workingmonk 
The `fingerprints` here looks odd to me. Is it intentional? 
Will the Keyfobs report in on a regular interval like the switches?

jackchi:
@skt123 
Let's test the reporting intervals on the keyfobs & see if these are regularly reported

skt123:
@jackchi 
We don't have Iris Keyfob or any of the other devices mentioned in the DTH here in SRI-B.

workingmonk:
@jackchi what is odd about the fingerprints?

jackchi:
I see in `configure()`. there is `zigbee.onOffConfig()` = max reporting time is 10min
But in Zach's code changes (not sure what this is) `zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x20, DataType.UINT8, 30, 21600, 0x01)` the max reporting time is 21600 or 6 hours. 

However, with Iris Smart Button, i do not see this regular onOff reporting every 10 minutes. Is this because it's a button and not a switch?

Does it make more sense to increase checkInterval to 12 hours then?

Testing the Centralite & Osram Switch now. But they are both powered devices so my suspicion is that they will behave the same and not report in every 10min

cc @varzac  @workingmonk

jackchi:
Just confirmed that `Osram Lightify Dimming Switch` & `Centralite Smart Switch` both are not reporting in every 10 minutes. Also the `Osram Lightify Dimming Switch` with brand new battery just went OFFLINE b/c it missed one ping.

jackchi:
I changed the `checkInterval` to 12hrs+2min & in the list of events I do not see the `ping` being called for `Iris SmartButton` & `Osram Lightify Dimming Switch`. However the `Centralite ZigBee Smart Button` doesn't stay ONLINE. Keeps going OFFLINE. 
@skt123 - can you change the `checkInterval` to 12 hrs 2 min?
@workingmonk as for `configure()`, should we just remove onOffConfig() / levelConfig(). Don't think these are actually being recognized by the Devices.

skt123:
@jackchi updated the `checkInterval` to 12 hrs 2 min.

jackchi:
👍

PR #2113, opened 2017-06-22
DVCSMP-2702 smartpower-outlet DTH incorrectly has capability.light

larsfinander:


workingmonk:
I thought this was intentional

PR #2114, opened 2017-06-23
ICP-1148 Support Thermostat Dynamic data

marstorp:
Adding support for dynamic thermostat and fan modes to Ecobee

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -70,7 +70,7 @@ metadata {
 			state "heat", action:"switchMode",  nextState: "updating", icon: "st.thermostat.heat"
 			state "cool", action:"switchMode",  nextState: "updating", icon: "st.thermostat.cool"
 			state "auto", action:"switchMode",  nextState: "updating", icon: "st.thermostat.auto"
-			state "emergency heat", action:"switchMode", icon: "st.thermostat.emergency-heat" // emergency heat = auxHeatOnly
+			state "auxheatonly", action:"switchMode", icon: "st.thermostat.emergency-heat"
 			state "updating", label:"Working", icon: "st.secondary.secondary"
 		}
 		standardTile("fanMode", "device.thermostatFanMode", inactiveLabel: false, decoration: "flat") {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -156,47 +156,49 @@ void poll() {
...
+		results.each { name, value ->
 			def event = [name: name, linkText: linkText, descriptionText: getThermostatDescriptionText(name, value, linkText),
 						 handlerName: name]
 
 			if (name=="temperature" || name=="heatingSetpoint" || name=="coolingSetpoint" ) {
 				def sendValue =  location.temperatureScale == "C"? roundC(convertFtoC(value.toDouble())) : value.toInteger()
-				isChange = isTemperatureStateChange(device, name, value.toString())
-				isDisplayed = isChange
-				event << [value: sendValue, unit: temperatureScale, isStateChange: isChange, displayed: isDisplayed]
+				event << [value: sendValue, unit: temperatureScale]
```
dsainteclaire, line 170: Just curious why we don't care anymore about whether or not the value of the event is changed, and why we don't send isDisplayed anymore.

dsainteclaire: I guess it could be because you always want to show those events

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -156,47 +156,49 @@ void poll() {
...
-			def isDisplayed = true
+		def supportedThermostatModes = []
+		def thermostatMode = null
+
+		results.each { name, value ->
 			def event = [name: name, linkText: linkText, descriptionText: getThermostatDescriptionText(name, value, linkText),
 						 handlerName: name]
 
 			if (name=="temperature" || name=="heatingSetpoint" || name=="coolingSetpoint" ) {
 				def sendValue =  location.temperatureScale == "C"? roundC(convertFtoC(value.toDouble())) : value.toInteger()
```
dsainteclaire, line 169: It might help to add some defensive programming here.  It's something I've thought about for awhile, but the number of exceptions has never been high enough that I got around to it.  In cases of things like Alexa and Google Home, they send set points as real numbers, not integers, which on occasion causes NumberFormatExceptions because you're assuming it's an int, but get a real number and the parser fails

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -156,47 +156,49 @@ void poll() {
...
-				isChange = isStateChange(device, name, value.toString())
-				event << [value: value.toString(), isStateChange: isChange, displayed: false]
+				if (value == true) {
+					supportedThermostatModes << ((name == "auxHeatMode") ? "auxheatonly" : name - "Mode")
+				}
+				return // as we don't want to send this event here, proceed to next name/value pair
 			}  else if (name=="thermostatFanMode"){
-				isChange = isStateChange(device, name, value.toString())
-				event << [value: value.toString(), isStateChange: isChange, displayed: false]
+				sendEvent(name: "supportedThermostatFanModes", value: fanModes(), displayed: false)
```
dsainteclaire, line 180: Just curious if you need to update the supported fan modes every time you get a fan mode event.  You can say yes, just just curious if it was intentional

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -156,47 +156,49 @@ void poll() {
...
+				return // as we don't want to send this event here, proceed to next name/value pair
 			} else {
-				isChange = isStateChange(device, name, value.toString())
-				isDisplayed = isChange
-				event << [value: value.toString(), isStateChange: isChange, displayed: isDisplayed]
+				event << [value: value.toString()]
 			}
 			sendEvent(event)
 		}
+		if (state.supportedThermostatModes != supportedThermostatModes) {
```
dsainteclaire, line 194: Also, you're sending another supported fan modes event here, so I just want to make sure that none of them are duplicates

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: Unless the state has changed the platform will not create/handle this as a new event, only events truly changed or with 'isStateCange: true' will generate a new event to handle. I'm guessing this was something added to DM after the this was DTH was implemented thus the DTH had to keep track of it. I removed it as I don't think it's necessary anymore
Default is that events will be displayed, unless you specifically add `displayed: false`. Also only events with changed data will generate a new event so unchanged events will not be displayed.

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: That should not affect this method which is handling the poll response from the thermostat

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: I'm actually not, it needs to be sent once but if I add it to installed or a place where it is only called when the DTH is installed we miss all existing DTHs so I made this intentionally this time. It could potentially be moved though after this implementation have been soaking for some time.

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: This is the supportedThermostatModes, different from supportedThermostatFanModes

dsainteclaire (approved):
LGTM 🐑

PR #2124, opened 2017-06-27
Add configurable power activity logging

instanttim:
A lot of folks in the community had been asking to "reduce the noise" of the energy reporting in the feed. This solution allows you to turn the energy reporting on/off in the activity feed and set a threshold. The tile remains live with instantaneous power updates of course.


instanttim:
FWIW I'm NOT a professional programmer, so please let me know if there's anything I did wrong here or could have done in a better way. I won't be offended in the slightest.

PR #2129, opened 2017-06-28
Update ridiculously-automated-garage-door.groovy

stephenmelody:
edited for grammar, replaced "was" with "has"


stephenmelody:
edited for grammar, replaced "was" with "has"

PR #2130, opened 2017-06-28
ICP-1075, ICP-1215 Separate ZSMOKE from ZCOMBO and default detector DTH (assume supports smoke+CO if no explicit fingerprint), set initial states to clear

dkirker:
For ICP-1075 we explicitly set the states to clear in installed()
For ICP-1215 we move the ZSMOKE device into its own DTH zwave-basic-smoke-alarm, with ZCOMBO and any non-defined WWST detector remaining in the zwave-smoke-alarm to eliminate the need to re-type existing detectors and with the assumption that new detectors on the market are likely to include CO.

### DIFF
--- a/devicetypes/smartthings/zwave-basic-smoke-alarm.src/zwave-basic-smoke-alarm.groovy
+++ b/devicetypes/smartthings/zwave-basic-smoke-alarm.src/zwave-basic-smoke-alarm.groovy
@@ -12,34 +12,29 @@
  *
  */
 metadata {
-	definition (name: "Z-Wave Smoke Alarm", namespace: "smartthings", author: "SmartThings") {
+	definition (name: "Z-Wave Basic Smoke Alarm", namespace: "smartthings", author: "SmartThings") {
 		capability "Smoke Detector"
-		capability "Carbon Monoxide Detector"
 		capability "Sensor"
 		capability "Battery"
 		capability "Health Check"
 
-		attribute "alarmState", "string"
-
-		fingerprint mfr:"0138", prod:"0001", model:"0002", deviceJoinName: "First Alert Smoke Detector and Carbon Monoxide Alarm (ZCOMBO)"
+		fingerprint deviceId: "0xA100", inClusters: "0x20,0x80,0x70,0x85,0x71,0x72,0x86"
+		fingerprint mfr:"0138", prod:"0001", model:"0001", deviceJoinName: "First Alert Smoke Detector"
 	}
 
 	simulator {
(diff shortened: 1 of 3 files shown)


dkirker:
@workingmonk @mckeed This is the PR that splits the smoke only detector from the DTH that supports smoke+CO, and then for both sets initial states.

I would like to get this into staging before we get it deployed. We need to acquire a ZSMOKE here, however.

dkirker:
Verified this on a Smoke+CO and a Smoke Only (ZCOMBO and ZSMOKE, respectively). Looks good for getting into staging on Wednesday for QA after a code review. @workingmonk @mckeed 

Vinay, I have this against staging right now. I think re-basing against master won't be a problem (I can do tomorrow night), or we can merge after the deploy on Wednesday.

devicetypes/smartthings/zwave-basic-smoke-alarm.src/zwave-basic-smoke-alarm.groovy
```diff
@@ -0,0 +1,186 @@
...
+ *
+ */
+metadata {
+	definition (name: "Z-Wave Basic Smoke Alarm", namespace: "smartthings", author: "SmartThings") {
+		capability "Smoke Detector"
+		capability "Sensor"
+		capability "Battery"
+		capability "Health Check"
+
+		attribute "alarmState", "string"
```
mckeed: Should take alarmState out of this one completely and just use smoke.

This was only for showing both smoke and carbonMonoxide in the same tile.

dkirker: Fixed in the latest rebased commit!

mckeed (approved):
This looks good. Sorry for the delay

dkirker:
@mckeed Awesome Thanks! (Just saw this.) I did one little change for the Basic Smoke Alarm tile. Basically renamed the state "smoke" to "detected" and then changed the initial events in the event list to show either "smoke clear" or "all clear" and not "test clear." Going to do a little test on Samsung Connect and get @workingmonk  to merge.

workingmonk:
@dkirker let me know when this is ready to merge

dkirker:
@workingmonk let's do it. I can deal with the plugin while we do our QA on this.

dbradmit:
So for Support's knowledge, if the device is just a Smoke, it will pair as **zwave-basic-smoke-alarm.groovy** and if it's a combo will pair as **zwave-smoke-alarm.groovy**?

PR #2140, opened 2017-07-05
ICP-507 - Replace location count null with 0

juano2310:


smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -72,7 +72,7 @@ def authPage() {
 		log.debug "have LIFX access token"
 
 		def options = locationOptions() ?: []
-		def count = options.size()
+		def count = options?.size() > 0 ? options.size()  : 0
```
bflorian: Juan, are you able to reproduce the original problem and verify that this change fixes it? Now that I'm looking at the whole thing I don't see how `count` could ever actually be `null`. If the call to `locationOptions()` on line 74 returns null then `options` is set to an empty list, the size of which would be 0.

bflorian: I just verified that this change indeed will not fix the problem. Line 74 guarantees that `count` will never be null. The problem is that the translation service is improperly checking the numeric value `count` and is interpreting a 0 value as not being set. We'll want to fix that in the platform (DM or the translation service) but for now this hack will get around the problem:
```
def count = options.size().toString()
```

bflorian (changes requested):
See comments on line 75. The actual problem is a platform bug in the translation service, though it is possible to work around it here by setting count to a string. But this change as written won't fix the problem

bflorian (approved):

PR #2146, opened 2017-07-05
ICP-1148 Support Thermostat Dynamic data

marstorp:
Adding support for dynamic thermostat and fan modes to Honeywell Zwave DTH.
Also replaced capability "Polling" with "Refresh" and runEvery5Minutes("refresh") as polling capability is unreliable.
Also removed capability "Relative Humidity Measurement" as Honeywell Z-Wave Thermostat (YTH8320ZW1007/U) doesn't support humidity.

### DIFF
--- a/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
@@ -15,10 +15,9 @@ metadata {
 	definition (name: "Z-Wave Thermostat", namespace: "smartthings", author: "SmartThings") {
 		capability "Actuator"
 		capability "Temperature Measurement"
-		capability "Relative Humidity Measurement"
 		capability "Thermostat"
 		capability "Configuration"
-		capability "Polling"
+		capability "Refresh"
 		capability "Sensor"
 		capability "Health Check"
 		
@@ -117,7 +116,7 @@ metadata {
 			state "cool", label:'${currentValue}° cool', backgroundColor:"#ffffff"
 		}
 		standardTile("refresh", "device.thermostatMode", inactiveLabel: false, decoration: "flat") {
-			state "default", action:"polling.poll", icon:"st.secondary.refresh"
+			state "default", action:"refresh.refresh", icon:"st.secondary.refresh"
 		}
 		main "temperature"
 		details(["temperature", "mode", "fanMode", "heatSliderControl", "heatingSetpoint", "coolSliderControl", "coolingSetpoint", "refresh"])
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -305,26 +313,26 @@ def zwaveEvent(physicalgraph.zwave.commands.thermostatfanmodev3.ThermostatFanMod
...
+	def supportedFanModes = []
+	if(cmd.auto) { supportedFanModes << "auto" } // "fanAuto "
+	if(cmd.circulation) { supportedFanModes << "circulate" } // "fanCirculate"
+	if(cmd.low) { supportedFanModes << "on" } // "fanOn"
 
-	state.supportedFanModes = supportedFanModes
-	// No events to be generated, return empty map
+	state.supportedThermostatFanModes = supportedFanModes
+	sendEvent(name: "supportedThermostatFanModes", value: supportedFanModes, displayed: false)
 	return [:]
```
dsainteclaire, line 336: Does this method need to return anything?

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -305,26 +313,26 @@ def zwaveEvent(physicalgraph.zwave.commands.thermostatfanmodev3.ThermostatFanMod
...
+	def supportedModes = []
+	if(cmd.off) { supportedModes << "off" }
+	if(cmd.heat) { supportedModes << "heat" }
+	if(cmd.cool) { supportedModes << "cool" }
+	if(cmd.auto) { supportedModes << "auto" }
+	if(cmd.auxiliaryemergencyHeat) { supportedModes << "emergency heat" }
+
+	state.supportedThermostatModes = supportedModes
+	sendEvent(name: "supportedThermostatModes", value: supportedModes, displayed: false)
 	return [:]
```
dsainteclaire, line 325: Does this method need to return anything?

dsainteclaire (approved):
The change looks good to me.  I'm okay with removing the humidity capability given that it never was exposed as a tile in the UI and the only impact that I can think of would be if someone wrote a SmartApp that used humidity sensors, they wouldn't see this thermostat, which isn't necessarily a bad thing because the SmartApp would behave unpredictably for devices that do not have a humidity sensor

tylerlange:
@mckeed do these changes potentially negatively impact other thermostat models?

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
marstorp: The return value is handled generically in the parse method.
Alternatively it could return null too.

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
marstorp: return value is handled in parse method

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
mckeed: It doesn't make much sense to use sendEvent and then not return anything. For consistency, should just return `[name: "supportedThermostatFanModes", value: supportedFanModes, displayed: false]` instead of the `sendEvent` call

PR #2147, opened 2017-07-06
CSING-26 Send IAS Zone update request when refreshing for 1st party zigbee sensors

dkirker:


dkirker:
@workingmonk Aiming to get this to staging tonight before 10pm.

cc @varzac @tpmanley

tpmanley (approved):

PR #2148, opened 2017-07-06
MSA-2084: Spruce Sensor DTH new model and health check

natec007:
Submitting 'Spruce Sensor DTH new model and health check' for publication

### DIFF
--- a/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
+++ b/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
@@ -1,5 +1,5 @@
 /**
- *  Spruce Sensor -Pre-release V2 10/8/2015
+ *  Spruce Sensor -updated with SLP model number 5/2017
  *
  *  Copyright 2014 Plaid Systems
  *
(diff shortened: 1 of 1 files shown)


devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
```diff
@@ -14,25 +14,34 @@
...
 		
 		capability "Configuration"
 		capability "Battery"
         capability "Relative Humidity Measurement"
         capability "Temperature Measurement"
         capability "Sensor"
+        capability "Health Check"
         //capability "Polling"
 		
+        attribute "checkInterval", "string"
```
jackchi: No need to add `checkInterval` here. It's actually included as a `NUMBER` in the Capbility `Health Check`

devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
```diff
@@ -293,6 +302,11 @@ def setConfig(){
     sendEvent(name: 'configuration',value: configInterval, descriptionText: "Configuration initialized")
 }
 
+def installed(){
+	//check every 1 hour + 2mins
+    sendEvent(name: "checkInterval", value: 1 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
jackchi, line 306: Let's put this `sendEvent("checkInterval")` in both `installed()` & `updated()`

jackchi:
Is this DTH "pingable()" - meaning, that we can contact the device and device would generate an event in the parse() method?

jackchi (changes requested):
1. Take out `checkInterval` from custom attribute
2. Add `sendEvent("checkInterval")` to both updated/installed
3. Can this device be pinged?

devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
natec007: Its a sleepy ZigBee device, so can't be pinged except when awake.
Made the changes requested

jackchi: So how often does this device check-in (generate an event) when idle? Currently if our platform doesn't hear from this device in 62min (your checkInterval value) it'll mark it `OFFLINE`

natec007: It pings the coordinator (hub) every 10 mins but reports data between 10-120mins.  So if the ZigBee event is caught then 62mins should be fine, if you only look at reported data, it should change to 2 hours.

jackchi: 👍

jackchi (approved):
Everything looks good.

PR #2156, opened 2017-07-12
ICP-1267 Move lock state check code from updated() to installed() to reduce chance that hub has gone back to pjoin mode thus blocking Z-Wave.

dkirker:
If we send the state update request in updated(), depending on how quickly the user reacts to the installation of the device, the hub may have gone back in to pjoin mode and requests to the lock will fail. installed() is called immediately during pairing while the hub is no longer actively searching.


dkirker:
cc @mckeed @workingmonk

dkirker:
I will be grabbing a few Z-Wave locks on Wednesday to do a little bit deeper checking of this (with different timing on actions).

dkirker:
I'm rejecting my own PR and will fix it as it appears that the Schlage touchscreen deadbolt now doesn't update state.

dkirker:
Alright @mckeed This one is ready for you. The timing gets us enough of a wait for the Z-Wave behind-the-scenes to finish and is quick enough to get the lock state and battery state requests and responses in before the hub has gone back to inclusion mode and blocked us.

mckeed (approved):
Probably don't need the `state.init` check anymore since installed() doesn't have the double-call problem, afaik, but it doesn't hurt anything. Thanks for making sure it works for the various models.

PR #2157, opened 2017-07-13
[MOB-3453] Update Spruce sensor icon

jackchi:
Can you reach out to Nathan Cauffman to see if he wants this icon?

### DIFF
--- a/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
+++ b/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
@@ -73,7 +73,7 @@ metadata {
 					[value: 64, color: "#44B621"],
 					[value: 80, color: "#3D79D9"],
 					[value: 96, color: "#0A50C2"]
-				]
+				], icon:"st.Weather.weather12"
 		}
         
         valueTile("maxHum", "device.maxHum", canChangeIcon: false, canChangeBackground: false) {


devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
```diff
@@ -73,7 +73,7 @@ metadata {
 					[value: 64, color: "#44B621"],
 					[value: 80, color: "#3D79D9"],
 					[value: 96, color: "#0A50C2"]
-				]
+				], icon: 'http://www.plaidsystems.com/smartthings/st_spruce_leaf_225_top.png'
```
workingmonk: can we use an existing humidity icon? if they change the link, this won't look pretty on the DTH

jackchi: Yeah Nathan (Spruce device maker) confirmed he wanted this. I'll let him make the call.

workingmonk:
cc @dkirker @tylerlange

devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
natec007: Which humidity icon would you use? Do you have a link so I can see it?

dkirker: There is a list of icons gathered by someone here: http://scripts.3dgo.net/smartthings/icons/
Someone else has also created: https://cdn.rawgit.com/JZ-SmartThings/SmartThings/master/Icons/ST-Icons.html

A good choice might be st.Outdoor.outdoor12 or st.Weather.weather12

That said, I think keeping custom icon assets available if used would fall under the WWST agreement requirement?

Anyway, if the Spruce logo is used, my recommendation is to make sure that it has the same aspect ratio and whitespace, as well as the same color, as the other icons.

natec007: st.Weather.weather12 is fine
Thanks

workingmonk: @jackchi can you change the icon to what @natec007 agreed to?

PR #2158, opened 2017-07-14
[DVCSMP-2811] Update ranges for centralite battery values

varzac:
This just updates the range to be more conservative as well as match the
battery curve better.

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -171,8 +171,8 @@ private Map getBatteryResult(rawValue) {
 			def pct = batteryMap[volts]
 			result.value = pct
 		} else {
-			def minVolts = 2.1
-			def maxVolts = 3.0
+			def minVolts = 2.4
+			def maxVolts = 2.7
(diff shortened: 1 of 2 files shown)


tpmanley (approved):
Changes look good. Needs a ticket and we should point at staging in order to get this out next week.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -171,8 +171,8 @@ private Map getBatteryResult(rawValue) {
 			def pct = batteryMap[volts]
 			result.value = pct
 		} else {
-			def minVolts = 2.1
-			def maxVolts = 3.0
+			def minVolts = 2.4
+			def maxVolts = 2.7
```
workingmonk, line 175: battery values would be 100, 67, 33 and 0. is that ok?

workingmonk:
can you reach out to  Steven for testing? this might need a v1 hub to avoid local DTH overriding your change. Matt, has a h/w battery simulator that Steven can use

vseven:
https://community.smartthings.com/t/st-multi-sensor-battery-reporting-issues/98924

This broke many sensors and should be reverted or fixed properly.

PR #2181, opened 2017-07-21
ICP-1103, ICP-1125, ICP-1194 CT Thermostat issues

marstorp:
- Added modes and changed mode names to reflect specification in thermostat capability
- Changed how device is initialized poll by sending all commands separated by 3 seconds
  instead of sending a request and on its response send the next. This to give all
  commands a chance in case one is dropped by the network.
- Reduced the time from 30 to 1 second between setting and requesting a setpoint
  using the setHeating/setCooling method. This is now the same as changing using
  the ST DTH tiles.
- Also modified the UI to use the multi attribute tile and using arrows instead of sliders
  for changing setpoints.

### DIFF
--- a/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
+++ b/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
@@ -6,7 +6,6 @@ metadata {
 		capability "Relative Humidity Measurement"
 		capability "Thermostat"
 		capability "Battery"
-		capability "Configuration"
 		capability "Refresh"
 		capability "Sensor"
 		capability "Health Check"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
```diff
@@ -15,161 +14,173 @@ metadata {
...
+def initialize() {
 	// Device-Watch simply pings if no device events received for 32min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	// Poll device for additional data that will be updated by refresh tile
+	refresh()
 }
 
 def parse(String description)
 {
-	def result = []
```
dsainteclaire, line 123: Does it break anything if we potentially move from returning an empty array, to now just returning null?  It seems like some of these methods require that you return a collection of some sort (like an empty array or dictionary).  Just checking

dsainteclaire (approved):
Just the one question, otherwise it all looks reasonable to me.  Good job!

devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
marstorp: nope 🤕 
Well Vinay said the message is stored if DTH returns null so I changed it to return an empty list instead.

dsainteclaire:
LGTM

devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
workingmonk: https://github.com/PhysicalGraph/data-management/pull/2606
Empty array would mean no action required. Null causes DM to save the raw description. @marstorp updated the code to return empty array

PR #2196, opened 2017-07-26
DVCSMP-2885 Smooth battery values for SmartSense devices

varzac:
Since we have been receiving a lot of reports and complaints about the
battery readings on the smartsense devices that received battery
updates (multi and motion), we are adding this code to smooth out the
readings so it doesn't constantly oscillate between two values.  The
basic logic is, only create an event if we have two readings in a row
that are the same, or a reading is more than 100 millivolts off of the
current battery percentage.

@tpmanley @workingmonk

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -173,11 +173,27 @@ private Map getBatteryResult(rawValue) {
 		} else {
 			def minVolts = 2.4
 			def maxVolts = 2.7
-			def pct = (volts - minVolts) / (maxVolts - minVolts)
-			def roundedPct = Math.round(pct * 100)
-			if (roundedPct <= 0)
-				roundedPct = 1
-			result.value = Math.min(100, roundedPct)
+			// Get the current battery percentage as a multiplier 0 - 1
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -173,11 +173,22 @@ private Map getBatteryResult(rawValue) {
...
+				def pct = (volts - minVolts) / (maxVolts - minVolts)
+				def roundedPct = Math.round(pct * 100)
+				if (roundedPct <= 0)
+					roundedPct = 1
+				result.value = Math.min(100, roundedPct)
+			} else {
+				// Don't update as we want to smooth the battery values
+				result = null
+			}
+			setState(["lastVolts": volts])
```
workingmonk: is this a custom method? setState? 
`state.lastVolts = volts`

varzac: Oh I can make that change.  It's not custom, it's in the base class, but I don't need to use it here, sort of a remnant of an earlier version.

workingmonk (approved):

workingmonk:
@varzac @tpmanley can you add the ticket number to the PR title

vseven:
https://community.smartthings.com/t/st-multi-sensor-battery-reporting-issues/98924

This broke many sensors and should be reverted or fixed properly.

PR #2202, opened 2017-07-29
ICP-1377 Z-Wave Thermostat incorrectly supports range

marstorp:
Z-Wave Thermostat incorrectly supports range of 0-100 for temp slider
Replaced slider UI with left/right arrow to change setpoints and made the
change methods a bit more robust.
Also replaced temperature tile with multi-attribute tile using thermometer icon.

### DIFF
--- a/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
@@ -523,7 +523,6 @@ def switchFanMode() {
 	if (supportedFanModes) {
 		def next = { supportedFanModes[supportedFanModes.indexOf(it) + 1] ?: supportedFanModes[0] }
 		def nextMode = next(currentMode)
-		setThermostatFanMode(nextMode)
 		runIn(2, "setThermostatFanMode", [data: [nextMode: nextMode], overwrite: true])
 	} else {
 		log.warn "supportedFanModes not defined"
--- a/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
@@ -375,11 +375,11 @@ def alterSetpoint(raise, setpoint) {
 	// also, as runIn's have to overwrite and user can change heating/cooling setpoint separately separate runIn's have to be used
 	if (data.targetHeatingSetpoint) {
 		sendEvent("name": "heatingSetpoint", "value": getTempInLocalScale(data.targetHeatingSetpoint, deviceScale),
-				unit: getTemperatureScale(), eventType: "ENTITY_UPDATE")//, displayed: false)
+				unit: getTemperatureScale(), eventType: "ENTITY_UPDATE", displayed: false)
 	}
 	if (data.targetCoolingSetpoint) {
 		sendEvent("name": "coolingSetpoint", "value": getTempInLocalScale(data.targetCoolingSetpoint, deviceScale),
-				unit: getTemperatureScale(), eventType: "ENTITY_UPDATE")//, displayed: false)
+				unit: getTemperatureScale(), eventType: "ENTITY_UPDATE", displayed: false)
 	}
 	if (data.targetHeatingSetpoint && data.targetCoolingSetpoint) {
 		runIn(5, "updateHeatingSetpoint", [data: data, overwrite: true])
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -25,209 +24,175 @@ metadata {
...
-def installed(){
-	sendHubCommand(new physicalgraph.device.HubAction(zwave.thermostatModeV2.thermostatModeSupportedGet().format()))
-	initialize()
+def installed() {
+	// Configure device
+	def cmds = []
+	cmds << new physicalgraph.device.HubAction(zwave.associationV1.associationSet(groupingIdentifier:1, nodeId:[zwaveHubNodeId]).format())
+	cmds << new physicalgraph.device.HubAction(zwave.manufacturerSpecificV2.manufacturerSpecificGet().format())
+	sendHubCommand(cmds)
+	runIn(3, "initialize", [overwrite: true])  // Allow configure command to be sent and acknowledged before proceeding
```
dsainteclaire, line 112: I assume that this is asynchronous because you want to wait for sendHubCommands to finish?  Will that duration always be long enough?

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -25,209 +24,175 @@ metadata {
...
 }
 
-def updated(){
-	initialize()
+def updated() {
+	// If not set update ManufacturerSpecific data
+	if (!getDataValue("manufacturer")) {
+		sendHubCommand(new physicalgraph.device.HubAction(zwave.manufacturerSpecificV2.manufacturerSpecificGet().format()))
+		runIn(2, "initialize", [overwrite: true])  // Allow configure command to be sent and acknowledged before proceeding
```
dsainteclaire, line 119: Same question about making this asynchronous. The duration here is shorter-than above.  Any reason we wouldn't want to standardize on something that will always complete before the runIn executes?

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -358,141 +342,210 @@ def refresh() {
...
+	def cmds = []
+	if (data.targetHeatingSetpoint) {
+		cmds << new physicalgraph.device.HubAction(zwave.thermostatSetpointV1.thermostatSetpointSet(
+					setpointType: 1, scale: state.scale, precision: state.precision, scaledValue: data.targetHeatingSetpoint).format())
+	}
+	if (data.targetCoolingSetpoint) {
+		cmds << new physicalgraph.device.HubAction(zwave.thermostatSetpointV1.thermostatSetpointSet(
+					setpointType: 2, scale: state.scale, precision: state.precision, scaledValue: data.targetCoolingSetpoint).format())
+	}
+	sendHubCommand(cmds)
```
dsainteclaire, line 470: What happens if you call sendHubCommand with and empty array (targetHeatingSetpoint and targetCoolingSetpoint are null)?

dsainteclaire (approved):
Overall, the code looks a lot cleaner and I'm happy with it.  Good job!  Just a few questions / nitpicks

workingmonk:
once approved, we need to get this QA approved and commits squashed before merging this to staging

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
marstorp: Testing shows it's ok, but if it doesn't then worst case is that manufacturer data isn't set in updated which would cause updated to try get manufacturer data again.

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
marstorp: It is shorter as there's one less command sent, the problem is that updated is not always called. Installed will always be called but updated is only called by the mobile app when the device is not added `silently` which can happen at times, the paring page will not show the device added but you'll see it in the things page.

dsainteclaire:
Based on our chat, I'm good with this, though it might make more sense at some point to consolidate all the runIn calls into a single method (even if the duration ends up being longer-than needed in some cases) 🐑

marstorp:
Updated setHeatingSetpoint(degrees) and setCoolingSetpoint(degrees) to be handled in the same way as for the updated version of CT100 due to "PROB-1767 CT100 Not Responding to Routine Cooling Setpoint Command if Heating Setpoint included as well"

PR #2227, opened 2017-08-09
DVCSMP-2838 Raw Zen Thermostat Setpoint Inconsistent with Device Display

MAblewicz:
1. Added Simulator statuses to simulate Zen Thermostat Behavior
2. Removed decimal places for temperature if temperature scale is Fahrenheit
3. Removed decimal places for Setpoint temperature if temperature scale is Fahrenheit
4. Added new temperature adjustment depended on temperature scale (According to Zen Thermostat specification temperature adjustment step is 1°F or 0.5°C)


MAblewicz:
Already fixed

PR #2228, opened 2017-08-09
[DVCSMP-2957] Adding CRC16 encapsulation for Fibaro Smoke Sensor

skt123:
@jackchi @ShunmugaSundar 

In Samsung Connect App, Fibaro Smoke Sensor reported temperature encapsulated by CRC16 checksum. Hence, at times the correct temperature was not being reflected in the App. Added implementation for the same. 
Also added proper tile changes for ST App.


skt123:
@jackchi @ShunmugaSundar 
Also added the default state as "clear" as earlier the device was initialized to "tested" state which was getting translate to NULL value in SC App.
Please review and merge

jackchi:
👍

PR #2231, opened 2017-08-10
DVCSMP-2973 CT100/Honeywell Z-wave NullPointerException

marstorp:
CT100/Honeywell Z-wave NullPointerException when changing mode
Changed name of private method setThermostatMode(data) to setGetThermostatMode(data)
and method setThermostatFanMode(data) to setGetThermostatFanMode(data) so that platform
calls the exposed methods setThermostatMode(String value) setThermostatFanMode(String value)
as these validates the argument against the supported modes and then calls the private methods.
Also removed some of the leftshift operators that showed errors in Sumo logs.

### DIFF
--- a/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
+++ b/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
@@ -101,9 +101,8 @@ metadata {
 
 def installed() {
 	// Configure device
-	def cmds = []
-	cmds << new physicalgraph.device.HubAction(zwave.associationV1.associationSet(groupingIdentifier:1, nodeId:[zwaveHubNodeId]).format())
-	cmds << new physicalgraph.device.HubAction(zwave.manufacturerSpecificV2.manufacturerSpecificGet().format())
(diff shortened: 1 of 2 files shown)


stevenchitov:
Retested with SmartApp(https://github.com/marstorp/SmartThingsPublic/blob/master/smartapps/dianoga/thermostat-auto-off.src/thermostat-auto-off.groovy) and old locally published DTH and was able to reproduce the Exception : timestamp=2017-08-10 21:09:19.167, level=ERROR, log_id=83$e$4, trace_id=, user=, cat=CompilerService, key=sandbox_error, owner=user_uuid:38806845-108d-432a-b9a3-cc927b7c0ab6, id=30e13929-a1e3-454a-8b0f-2edb585f78d4, deviceTypeName=CT100 Thermostat, gitRepoId=null,
java.lang.NullPointerException: Cannot get property 'nextMode' on null object
	at org.kohsuke.groovy.sandbox.impl.Checker$4.call(Checker.java:115)

After that retested again with smart app and new DTH https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2231) and was not able to reproduce the exception.

dsainteclaire (approved):
LGTM.  Let's get it out the door! 🐑

workingmonk:
try the dropdown squash and merge github option next time. i did the same last week and merged 7 commits from ingvar's pr

PR #2234, opened 2017-08-12
ICP-987 added child device components to Aeon Minimote

bflorian:
Creates 4 child component devices on installation (or update of previously installed device) representing the 4 buttons of the Minimote. This approach allows each button to be individually selected by any SmartApp or rule builder without having the check the button number data field as is required when the primary device is selected.

Preserves functionality of the primary devices so that existing installations and the SmartApps using them will not break.


jackchi:
👍  
screenshot_20170812-063017
screenshot_20170812-095608

cc @ShunmugaSundar @pchomal

jackchi:
👍

PR #2251, opened 2017-08-21
DVCSMP-2973 CT100/Honeywell Z-wave NullPointerException

marstorp:
CT100/Honeywell Z-wave NullPointerException when changing mode
In the event supportedModes and supportedFanModes hasn't been updated from a string to a list the mode isn't validated properly.
Adding check for string and if so updating the supported modes.

### DIFF
--- a/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
+++ b/devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
@@ -625,7 +625,7 @@ def switchMode() {
 	def currentMode = device.currentValue("thermostatMode")
 	def supportedModes = state.supportedModes
 	// Old version of supportedModes was as string, make sure it gets updated
-	if (supportedModes && supportedModes[0].size() > 1) {
+	if (supportedModes && supportedModes.size() && supportedModes[0].size() > 1) {
 		def next = { supportedModes[supportedModes.indexOf(it) + 1] ?: supportedModes[0] }
 		def nextMode = next(currentMode)
 		runIn(2, "setGetThermostatMode", [data: [nextMode: nextMode], overwrite: true])
@@ -638,7 +638,7 @@ def switchMode() {
 def switchToMode(nextMode) {
 	def supportedModes = state.supportedModes
 	// Old version of supportedModes was as string, make sure it gets updated
-	if (supportedModes && supportedModes[0].size() > 1) {
+	if (supportedModes && supportedModes.size() && supportedModes[0].size() > 1) {
 		if (supportedModes.contains(nextMode)) {
 			runIn(2, "setGetThermostatMode", [data: [nextMode: nextMode], overwrite: true])
 		} else {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
```diff
@@ -624,7 +624,8 @@ def ping() {
 def switchMode() {
 	def currentMode = device.currentValue("thermostatMode")
 	def supportedModes = state.supportedModes
-	if (supportedModes) {
+	// Old version of supportedModes was as string, make sure it gets updated
+	if (supportedModes && supportedModes[0].size() > 1) {
```
dsainteclaire: Is it possible for supportedModes to be an empty list, where supportedModes[0] would still throw an NPE?

devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
```diff
@@ -636,7 +637,8 @@ def switchMode() {
 
 def switchToMode(nextMode) {
 	def supportedModes = state.supportedModes
-	if (supportedModes) {
+	// Old version of supportedModes was as string, make sure it gets updated
+	if (supportedModes && supportedModes[0].size() > 1) {
```
dsainteclaire: Same question about supportedModes containing an empty list.  It's a bit obsessive, but you still might want to check if supportedModes is null and it's size > 0

devicetypes/smartthings/ct100-thermostat.src/ct100-thermostat.groovy
```diff
@@ -657,7 +659,8 @@ def getSupportedModes() {
 def switchFanMode() {
 	def currentMode = device.currentValue("thermostatFanMode")
 	def supportedFanModes = state.supportedFanModes
-	if (supportedFanModes) {
+	// Old version of supportedFanModes was as string, make sure it gets updated
+	if (supportedFanModes && supportedFanModes[0].size() > 1) {
```
dsainteclaire: Same comment as above about supportedFanModes could be an empty list?  It will it always contain at least "Off"?

devicetypes/smartthings/zwave-thermostat.src/zwave-thermostat.groovy
```diff
@@ -507,7 +507,8 @@ def ping() {
 def switchMode() {
 	def currentMode = device.currentValue("thermostatMode")
 	def supportedModes = state.supportedModes
-	if (supportedModes) {
+	// Old version of supportedModes was as string, make sure it gets updated
+	if (supportedModes && supportedModes[0].size() > 1) {
```
dsainteclaire: Same comment as above about whether or not it's possible for this list to be an empty list (which means it would pass if(supportedModes) but supportedModes[0] could still throw an NPE?

dsainteclaire (commented):
Just a few comments, but if they aren't an issue, then I can approve it

dsainteclaire (approved):
🐑

PR #2257, opened 2017-08-23
GWWST-70 Add latest OSRAM ZLL bulbs

dkirker:
Update to include the fingerprints


dkirker:
Poland has checked. The goal is for deploy on Tuesday.

PR #2258, opened 2017-08-23
[DVCSMP-2841] Onboard Aeotec Panic Button

skt123:
@jackchi @bflorian @ShunmugaSundar 
Added child devices components similar to Aeon Minimote. It should take care of other devices in this DTH as well that has multiple buttons.
Please review and merge.


jackchi:
👍

PR #2260, opened 2017-08-25
GWWST-65, GWWST-61: Update Aeon Multisensor 6 to use Power Source capability, Tamper Alert capability

dkirker:


dkirker:
@workingmonk @mckeed  This is a pretty "small" change. Either way, I must beg for a code review. "Will do tricks for code review!" 🐶

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -138,17 +138,17 @@ def updated() {
 // Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	log.debug "Updated with settings: ${settings}"
-	log.debug "${device.displayName} is now ${device.latestValue("powerSupply")}"
+	log.debug "${device.displayName} is now ${device.latestValue("powerSource")}"
 
-	if (device.latestValue("powerSupply") == "USB Cable") {  //case1: USB powered
```
mckeed, line 143: Won't this break for already-added devices? Maybe leave the `powerSupply` attribute and use it to populate `powerSource` if it isn't set?

dkirker: @mckeed In that case, it would just fall through to the else statement below which would trigger a new request to check the power source. And to be honest, I don't think that code ever worked, as I tried following the old logic and it didn't check out, and the multisensor that I paired before always had that as a non-existent attribute (while also printing the debug that it couldn't get the power source). :/ I think that was generally ignored in the past because it wasn't really exposed to the user and we didn't have a capability.

dkirker: Going to also look at seeing if we were on Battery with the old attribute name just to make sure we aren't supposed to be waiting for a wakeup to allow us to call configure() (though, I fear that value isn't necessarily reliable).

dkirker: @mckeed Alright! Tweaked the code to take in to account previous setups

dkirker: Also, I added an initial clear tamper event.

mckeed (approved):
It looks like the configuration is sent twice on join with this updated() implementation, but it seems to work fine.

I tested with a device that had been added earlier and had the powerSupply attribute and it was converted properly when updated() was called.

dkirker:
Thanks @mckeed !
Yeah, between this device and a couple of Fibaro devices, I have been considering refactoring configuration... I just feel like there are too many assumptions made when using configure() on sleepy devices. I'll go over those with you next week.

dkirker:
I added a tamper tile at @workingmonk 's suggestion, and also did a quick update so both tamper detected and clear events show in the mobile events view (currently only detected show).

PR #2261, opened 2017-08-25
DVCSMP-3242 Update ecobee-sensor.groovy

vseven:
Update old style tile to newer style multiAttributeTile and allow the icon to be changed.  The current device type is very outdated.


vseven:
Same change - update standard tile to multi attribute and cleaned up formatting.  @workingmonk

PR #2263, opened 2017-08-25
DVCSMP-3242 Update smartsense-temp-humidity-sensor.groovy

vseven:
Add the ability to offset humidity with a preference identical to temperature.


vseven:
Have found many of these sensors need a offset on humidity just like temperature.  This adds that ability.  Language is the same as for temperature (copy/paste) and the offsetting was standard like the temperature.  

May or may not want to keep the ability to change the icon.  I personally think every tile should hae that ability but maybe there is a reason you guys prevent it.

@workingmonk

workingmonk:
👍 
Great work on the humidity value. Didn't realize that this would be affecting customers in the field.

cc @tpmanley will need something similar for humidity for local execution?

vseven:
I'm actually using the stock sensor DTH simply because it runs locally even though I know my sensors are off (I have a known good value to judge against).  So the ability to run locally > ability to offset humidity.  If you can do both that would be great.

workingmonk:
it will run locally, just that i am worried that it **may** not take into account the humidityOffset. tagged @tpmanley @varzac to confirm that

PR #2266, opened 2017-08-28
Update ecobee-thermostat.groovy

vseven:
Update the device type to a multi attribute tile per the documentation at http://docs.smartthings.com/en/latest/device-type-developers-guide/tiles-metadata.html#thermostat-multi-attribute-tile to bring this device more in line with current standards.  

THIS IS ONLY A COSMETIC CHANGE.  None of the functionality was change.

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -42,30 +42,72 @@ metadata {
 		attribute "deviceAlive", "enum", ["true", "false"]
 	}
 
-	tiles {
-		standardTile("temperature", "device.temperature", width: 2, height: 2, decoration: "flat") {
-			state("temperature", label:'${currentValue}°', unit:"F", icon: "st.thermostat.ac.air-conditioning",
-					backgroundColors:[
-							// Celsius
-							[value: 0, color: "#153591"],
-							[value: 7, color: "#1e9cbb"],
-							[value: 15, color: "#90d2a7"],
-							[value: 23, color: "#44b621"],
-							[value: 28, color: "#f1d801"],
-							[value: 35, color: "#d04e00"],
-							[value: 37, color: "#bc2323"],
-							// Fahrenheit
-							[value: 40, color: "#153591"],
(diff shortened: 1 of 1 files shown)


vseven:
This ones a little more complicated but again just changing standard tiles to mutli attribute.  This is literally a copy/paste from the SmartThings docs.  I'll understand not committing this one fully but at least the main tile should be switched.  @workingmonk

workingmonk:
@marstorp can you use this as part of the ecobee thermostat changes that you are doing/will do? May be you already have it ready, just wanted to tag you here. :)

vseven:
@marstorp  - Any news on this?  The EcoBee Sensor changes I made look like they went out today but I don't think this has changed.  Granted the sensor was pretty simplistic and this is a bit more complicated but still raising the issue to see if anything has been done.

marstorp:
Hi Allan,
Thank you for your commitment, it's most appreciated 🙂
Regarding thermostats, I've recently been updating the UI for those that I've had to change lately but instead of using the thermostat multiAttributeTile we decided to use a generic instead. This as the background color for this tile is tied to the operating state rather then the temperature that we are used to, it would mostly be gray which IMHO looks rather sad and dull. You can take a look at those that are in the public repo as reference; zen-thermostat.groovy, zwave-thermostat.groovy, ct100-thermostat.groovy.
If you have comments, questions, opinions, please let me know 🙂

vseven:
I have noticed what you are saying with the AC or heat running yet still having a gray background and I've also seen it have a blue background (for cooling) and red (for heating).  I've looked through and everything seems correct per the documentation I linked to in my initial comment.   In fact here is a screen shot of the changes that I put the PR in for when it happened to be cooling and showing in blue correctly:

image

I do like the improved humidity and status on the tile in my device handler, again just based on your guys documentation.  Hopefully that can be retained in whatever you are doing as it looks a lot nicer then the current stock implementation.  

Any time-frame for when the Ecobee Thermostat will be updated?

marstorp:
You can find the PR with the changes aligning Ecobee to our the other thermostats here, including a couple of bug fixes.
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2384

vseven:
Cool.  Thanks!

PR #2271, opened 2017-08-30
Have zigbee setColor honor level settings

varzac:
The ColorControl.setColor command takes a color map as an argument which
can include a level.  This level was not being honored by most of the
zigbee color bulbs.  This shows up as an issue when using the lighting
automation to set color and level, the level portion is not working.

This resolves: https://smartthings.atlassian.net/browse/DVCSMP-3133

@tpmanley


varzac:
Closing as we plan to fix in a different way

PR #2275, opened 2017-09-01
[ICP-1555] Device Health fix for FortezZ Siren

jackchi:
`checkInterval` updated to 8hrs2min


jackchi:
@parijatdas @ShunmugaSundar @mckeed

PR #2277, opened 2017-09-02
ICP-1417 ICP-1437 Initial states for Fibaro devices, and accurate tamper events

dkirker:


devicetypes/fibargroup/fibaro-door-window-sensor-zw5-with-temperature.src/fibaro-door-window-sensor-zw5-with-temperature.groovy
```diff
@@ -142,14 +156,14 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
...
+                map.value = "clear"
+                map.descriptionText = "${device.displayName}: tamper is clear"
 				break
 
         	case 3:
             	map.name = "tamper"
-                map.value = "active"
-                map.descriptionText = "${device.displayName}: tamper alarm activated"
+                map.value = "detected"
+                map.descriptionText = "${device.displayName}: tamper detected"
```
mckeed: This event code specifically means "Product covering removed." Can we say that in the description text, since "tamper detected" isn't very clear to the layperson?

devicetypes/fibargroup/fibaro-door-window-sensor-zw5.src/fibaro-door-window-sensor-zw5.groovy
```diff
@@ -211,7 +242,8 @@ def configure() {
     cmds += zwave.versionV1.versionGet()
```
mckeed: It's not really related to this PR, but can we take out manufacturerSpecificGet and versionGet since that info is in the Raw Description now

devicetypes/fibargroup/fibaro-door-window-sensor-zw5-with-temperature.src/fibaro-door-window-sensor-zw5-with-temperature.groovy
dkirker: I'm going to work out some better text, because it appears that fibaro may be using this as a catch-all (for instance, the motion and flood have "movement" sensors, so if you bump them they send this -- though I will make sure they aren't sending others, but we get this event for sure, and I believe that one can attach the door sensor in a way that depresses the back button so that if it becomes un-depressed it sends this).

Initially I was thinking "$device.displayName} was tampered with", as that would make more sense, but I am not sure about "tamper cleared."

dkirker: @mckeed Alright, I think I found some text that I like and is more descriptive!

"${device name} tampering has stopped"
"${device name} was tampered with or cover removed"

mckeed (approved):

PR #2282, opened 2017-09-05
[GWWST-77] Fibaro Wall Plug WWST Proper Naming

jackchi:
Proper Fibaro Wall Plug WWST naming


dkirker:
@jackchi Since this PR is for the non-ZW5 version and we are focusing on the ZW5 version I am going to close this PR. Also, the fingerprint you added appears to be for the ZW5 version that Poland has, and does not match the non-ZW5 version at my desk.

However! If we are also certifying the non-ZW5 version, we can update the fingerprint and deviceJoinName (so that it is more consistent).

PR #2296, opened 2017-09-08
ICP-1047, ICP-1041, ICP-1148, MOB-3388, DVCSMP-2838, DVCSMP-1492, DVCSMP-1014

marstorp:
…SMP-1014

ICP-1047 - Zen Thermostat doesn't have Temperature unit(C or F).
ICP-1041 - [Zen Thermostat] Fan mode is not displayed in ST Plugin
ICP-1148 - [P170712-02680] Support Thermostat Dynamic data
MOB-3388 - ST app: Zen Thermostat is missing Icon on Android Device
DVCSMP-2838 - Raw Zen Thermostat Setpoint Inconsistent with Device Display
DVCSMP-1492 - Zen Thermostat in Emergency Heat Mode Doesn't Display/Respond
DVCSMP-1014 - Zen thermostat: device handler doesn't handle attribute reports with invalid values
Also changed UI to use multiattribute tile, added battery status, removed config button as it doesn't add any value,
added user configurable thermostat modes in settings page

### DIFF
--- a/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
+++ b/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
@@ -3,546 +3,808 @@
  *
  *  Author: Zen Within
  *  Date: 2015-02-21
+ *  Updated by SmartThings
+ *  Date: 2017-09-07
  */
 metadata {
-  definition (name: "Zen Thermostat", namespace: "zenwithin", author: "ZenWithin") {
-    capability "Actuator"
-    capability "Thermostat"
-    capability "Temperature Measurement"
-    capability "Configuration"
-    capability "Refresh"
-    capability "Sensor"
-    capability "Health Check"
+    definition (name: "Zen Thermostat", namespace: "zenwithin", author: "ZenWithin") {
+        capability "Actuator"
+        capability "Battery"
+        capability "Health Check"
+        capability "Refresh"
+        capability "Sensor"
+        capability "Temperature Measurement"
+        capability "Thermostat"
 
-    fingerprint profileId: "0104", endpointId: "01", inClusters: "0000,0001,0003,0004,0005,0020,0201,0202,0204,0B05", outClusters: "000A, 0019", manufacturer: "Zen Within", model: "Zen-01", deviceJoinName: "Zen Thermostat"
-    
-    //attribute "temperatureUnit", "number"
-      
-    command "setpointUp"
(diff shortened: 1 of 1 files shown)


devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
```diff
@@ -3,546 +3,808 @@
...
+                title: "Thermostat configured modes\nSelect the modes the thermostat has been configured for, as displayed on the thermostat",
+                description: "off, heat, cool", defaultValue: "1", required: true, multiple: false,
+                options:["1":"off, heat, cool", "2":"off, auto, heat, cool", "3":"off, emergency heat, heat, cool"]
+            )
         }
+    }
+}
+
+// Globals
+private getTHERMOSTAT_CLUSTER()                      { 0x0201 }
```
dsainteclaire, line 100: It looks like you decided against using any of these since I don't really see them called anywhere.  Consider removing

dsainteclaire: Nevermind, you're using them in pollDevice().  But why not anywhere else?

dsainteclaire (approved):
Just the one comment.  Otherwise, my head hurts now from looking at such a big change.  🐑

workingmonk:
Since this is the mother of all PR with the number of tickets involved, can we merge to master and give more time to QA for testing. Otherwise they will only have one day to test all the changes

devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
marstorp: Two reasons, when using e.g. THERMOSTAT_CLUSTER it always result in a call to this method, the compilers isn't smart enough to realize that it can substitute with the const returned value, and then it has to be converted to a string as that is what the comparisons are made agains, although I could possible have converted the incoming value to hex too, event so there's a lot of comparisons in the parse method that would generate a lot of calls to theses getters.

marstorp:
Unless there's a urgency from any of the ticket creators we can just wait with merging this until after the deploy on Tuesday.

marstorp:
@pdonovan Hi Paul, thank you for your comment it is most appreciated :)
Regarding iOS, if it hasn't been fixed already it's something the mobile app team is working on.
I've created a ticket for "heat only"/"cool only" thermostats and it should be fixed in a couple of weeks.

PR #2305, opened 2017-09-12
[DVCSMP-2740] Added lock codes functionality to generic Z-Wave and Zi…

vaish-rahul:
…gbee Locks DTH

Added manufacturer specific DTHs for Yale Z-Wave and Zigbee Locks with
Lock Codes functionality and Advanced Settings features.

PR #2310, opened 2017-09-13
Revert "[DVCSMP-2740] Added lock codes functionality to generic Z-Wave and Zi…"

workingmonk:
Reverts SmartThingsCommunity/SmartThingsPublic#2305


workingmonk:
reverting this as @vaish-rahul will have the changes created separately.

PR #2312, opened 2017-09-14
MSA-2219: VISONIC MCT-340

SmartThings:
Submitting 'VISONIC MCT-340' for publication


workingmonk:
@tpmanley there is another fingerprint for Visonic according to this PR

tpmanley:
I didn't add that one in my pr because I have one of those and couldn't get it to even start to join. I am fine with adding it too if we can verify it works. There is a chance there is just something wrong with the one I have.

PR #2313, opened 2017-09-14
[DHF-54] Adding offline pingable for Zwave-generic and dimmer switches

arunrao33:
@jackchi , i made changes only to Zwave switches only.. This will address what Brad was asking.


jackchi (approved):
👍  LGTM

PR #2319, opened 2017-09-15
DVCSMP-3231 Add fingerprint for Visonic MCT-340 E

tpmanley:
This will prevent users from having to self publish the DTH in order to use this popular sensor. In addition it enables this device to run locally on hubs that support it.

https://smartthings.atlassian.net/browse/DVCSMP-3231


varzac (approved):

jmagnuson (approved):

tylerlange:
👍

PR #2322, opened 2017-09-18
GWWST-96 Aeon Siren does not update state when cleared from device

dkirker:


mckeed (approved):

dkirker:
@mckeed  Thanks!
@workingmonk  This will be good to merge after the deploy on 9/26

PR #2346, opened 2017-09-19
Added curb bridge to smartthingspublic

newtectonics:


smartapps/curb/curb-bridge.src/curb-bridge.groovy
```diff
@@ -0,0 +1,605 @@
...
+    unschedule()
+    refreshAuthToken()
+    updateSelectedLocationId()
+	  getDevices()
+    runEvery5Minutes(getHistorical)
+    runEvery3Hours(refreshAuthToken)
+    runEvery3Hours(getKwhr)
+    def rate = settings.samplesPerMinute
+    //log.debug("Sampling at ${rate} samples per minute")
+    schedule("* * * * * ?", doPoll, [data: [cycles: rate]])
```
marstorp: Scheduling and using runIn's this often is not advisable as it generates a lot of load on the platform.
Either provide all desired samples in a poll with max interval of 5 min or have a callback method the server can call to updates the samples.
Also, if possible merge the different runEvery scheduling.

smartapps/curb/curb-bridge.src/curb-bridge.groovy
```diff
@@ -0,0 +1,605 @@
...
+        ]
+        httpPostJson([uri: curbTokenUrl, body: tokenParams]) {
+            resp ->
+                //log.debug "response contentType: ${resp.contentType}"
+            //log.debug("Got POST response: ${resp.data}")
+
+            atomicState.refreshToken = resp.data.refresh_token
+            atomicState.authToken = resp.data.access_token
+
+            getCurbLocations()
```
marstorp: Is it ok if getCurbLocations fails to get any locations?

smartapps/curb/curb-bridge.src/curb-bridge.groovy
```diff
@@ -0,0 +1,605 @@
...
+  </body>
+
+  </html>
+    """
+
+	render contentType: 'text/html', data: html
+}
+
+def getCurbClientId()        { return "R7LHLp5rRr6ktb9hhXfMaILsjwmIinKa" }
+def getCurbClientSecret()    { return "pcxoDsqCN7o_ny5KmEKJ2ci0gL5qqOSfxnzF6JIvwsfRsUVXFdD-DUc40kkhHAZR" }
```
marstorp: I suggest to move the clientId and clientSecret to `App Settings`, clientId is already defined as appSetting but not used anywhere. Either change above method to return e.g. `appSettings.clientID` or just use it in place.

marstorp (approved):
Please check the comments.
 Also, try avoid mixing tabs and spaces when indenting the code.

marstorp (changes requested):
Please check the comments and try avoid mixing tabs and spaces when indenting the code.

smartapps/curb/curb-bridge.src/curb-bridge.groovy
newtectonics: Is the problem here that we're polling our servers too often, and we shouldn't do it more frequently than every five minutes, or is this something wrong with the schedule() function?

I don't totally understand the comment "have a callback method the server can call to updates the samples."

marstorp: The problem is the load that these frequent scheduling are adding to the platform. Only in a very few cases do we poll anything from the cloud more than every 5 minutes.
By defining a callback/subscription you wouldn't need to have any polls scheduled as the server will call the SA whenever there's a reason to update anything.

newtectonics:
@marstorp This should be ready to go.

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+    atomicState.locationLookup.each {
+        // log.debug "location match check: ${it.name} vs selected: ${settings.curbLocation}"
+        if (it.name == settings.curbLocation) {
+            atomicState.location = it.id
+            return it.id
+        }
+    }
+}
+
+def updateChildDevice(dni, label, values) {
```
marstorp: is there a purpose for the argument `label`, I don't see it being used in this method.

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+      </div>
+  </body>
+
+  </html>
+    """
+
+  render contentType: 'text/html', data: html
+}
+
+        def getCurbAuthUrl() {
```
marstorp: indentation...

devicetypes/curb/curb-power-meter.src/curb-power-meter.groovy
```diff
@@ -0,0 +1,168 @@
...
+ *
+ */
+
+metadata {
+    definition(name: "Curb Power Meter", namespace: "curb", author: "Curb") {
+        capability "Power Meter"
+        capability "Energy Meter"
+    }
+
+    simulator {}
```
marstorp: no need to define this if there's no implementation

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+    // }
+
+    def onehparams = [
+        uri: "https://app.energycurb.com",
+        path: "/api/aggregate/${atomicState.location}/1h/m",
+        headers: ["Authorization": "Bearer ${atomicState.authToken}"],
+        requestContentType: 'application/json'
+    ]
+    log.debug onehparams
+    asynchttp_v1.get(processKwh1h, onehparams)
```
marstorp: There are 5 asynchHttp_v1 calls that will result in 5 different callbacks, are the responses spread out in time or is there a possibility that responses arrive more or less at the same time?
Reason I'm asking is that it is preferable to make one request and handled one response, especially since this is scheduled to be done once every minute and asynchttp is used. Each of the response will cause an instant of the app to be spun up which will spin up an instance of the child for each of the call that app instance makes, keep in mind that the state in the DTH is not thread safe.
Granted, if the response for the http get takes to long time it will cause a socket timeout, but this is scheduled to be done every minute so the response should not take that long time...

devicetypes/curb/curb-power-meter.src/curb-power-meter.groovy
```diff
@@ -0,0 +1,168 @@
...
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+metadata {
+    definition(name: "Curb Power Meter", namespace: "curb", author: "Curb") {
+        capability "Power Meter"
+        capability "Energy Meter"
```
marstorp: declare used custom attributes here;
        attribute "kwh30m", "number"
        attribute "kwh1m", "number"
etc...

devicetypes/curb/curb-power-meter.src/curb-power-meter.groovy
```diff
@@ -0,0 +1,168 @@
...
+                        [value: -500, color: "#76ce61"],
+                        [value: -100, color: "#bbedaf"],
+                        [value: 0, color: "#bcbbb5"],
+                        [value: 100, color: "#efc621"],
+                        [value: 1000, color: "#ed8c25"],
+                        [value: 2000, color: "#db5e1f"]
+                    ]
+            }
+        }
+        valueTile("kwh30m", "device.energy", decoration: "flat", width: 1, height: 1) {
```
marstorp: I don't see these valueTiles being used anywhere, do they have any purpose?
Also kwh30m, kwh1m and kwh1 are all using the same attribute, device.energy defined by capability.energyMeter is that intentional or should they use the values for the attributes kwh30m, kwh1m, and kwh1hevent that the corresponding events are generating?

If there's no intention of using any of these attributes they should be removed from both the DTH and the SA.

devicetypes/curb/curb-power-meter.src/curb-power-meter.groovy
```diff
@@ -0,0 +1,168 @@
...
+}
+
+def handleKwh1m(kwh) {
+    sendEvent(name: "kwh1m", value: kwh.round(3))
+}
+
+
+String getDataString()
+{
+  def dataString = ""
```
marstorp: Consider using `dataString = new StringBuilder()` instead, especially if state.values.each() have several iterations.

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+
+            //log.debug("Location ID: ${atomicState.location}")
+        }
+    } catch (e) {
+        log.error "something went wrong: $e"
+    }
+}
+
+def updateSelectedLocationId() {
+    atomicState.locationLookup.each {
```
marstorp: since assigning an atomiState.variable immediately will save the whole state object it's not recommended to loop over the atomicState and make change to it inside the loop. If possible consider to use state instead, just make sure that cases that can cause an update to the state variable can be triggered through multiple instances, otherwise consider changing the code so that atomicState is only update once when the method ends.

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+
+    if(json)
+    {
+        //log.debug "Got Historical Data: ${json}"
+        def total = null
+        def prod = null
+
+        json.each
+        {
+        	if(!it.main && !it.production){
```
marstorp: indentation, tab vs. space

smartapps/curb/curb-energy-monitor.src/curb-energy-monitor.groovy
```diff
@@ -0,0 +1,620 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        updateChildDevice("__NET__", "Main", total.values)
+        if(prod){
+        	updateChildDevice("__PRODUCTION__", "Solar", prod.values)
```
marstorp: indentation, tab vs. space

marstorp (changes requested):
Please check my comments...

newtectonics:
@marstorp Hi Ingvar, I think we have an app that now addresses all of your issues with the previous design. Thanks!

marstorp (approved):
LGTM 👍

workingmonk:
@marstorp can you add the ticket number to this one? and is it certified?

PR #2354, opened 2017-09-21
[DVCSMP-3262] Remove unnecessary write from GE link poll

varzac:
The poll method was doing a write of a transition time value. This is not 
needed and can be done only on the updated call (which it is).

@tpmanley @workingmonk


tpmanley (approved):
Looks good, just add a ticket

workingmonk:
if you are going to remove that, we probably need to fix the updated() method. the reason for it being included in refresh was to handle the preference change.

varzac:
@workingmonk We decided (Tom and I) to leave the write in refresh so it can be used if the initial write fails without having to go through settings.  The problem we want to address was the unnecessary write in the poll.

PR #2369, opened 2017-09-27
PROB-1137 - Skip hub if it is no longer available

juano2310:


smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
```diff
@@ -366,14 +366,18 @@ Map discoverDevices() {
...
         devices.each {
-			def hubkey = it.key
-			def hubname = getHubName(it.key)
-			def hubvalue = "${hubname}"
-			hubs["harmony-${hubkey}"] = hubvalue
-			it.value.response.data.activities.each {
-				def value = "${it.value.name}"
-				def key = "harmony-${hubkey}-${it.key}"
-				activities["${key}"] = value
+			if (it.value.response){
```
marstorp, line 369: there's no need to check for null pointer ?

marstorp (approved):
LGTM if there's no risk of getting null pointers...
⛵️

smartapps/smartthings/logitech-harmony-connect.src/logitech-harmony-connect.groovy
juano2310: Not really, the it and the value doesn't need to be checked because if not the each wouldn't get to this code. The response is being checked by the IF.

PR #2384, opened 2017-10-03
ICP-1501 Ecobee Supported Modes Issue

marstorp:
Adding additional check in SA to verify that both heatStages and coolStages exists before enabling autoMode.
Also changing the exposed "auxheatonly" mode to the capability defined "emergency heat" so that bouncer doesn't
invalidate the event causing timeout/networkErrors in SC when changing to auxheatonly. Emergency Heat is basically
the same as Aux Heat only as both have the "outdoor" heat pump off.
Also changing the DTH UI, aligning it with other thermostats and the way the setpoints are updated.


dsainteclaire (approved):
Looks good to me, but please have QA run a sanity check on it before it goes to prod.

PR #2391, opened 2017-10-04
Initial commit of Ikea wireless dimmer.

tony-trility:


jackchi:
@workingmonk 
Should this be a separate DTH?

jackchi:
@ShilpaMathew `Health Check` is most likely a requirement here.

jackchi:
@tony-trility Please include a JIRA ticket with this commit.

workingmonk:
if this is a dimmer, why does it only have a switch tile?
This should be part of the generic zigbee dth, unless there is a strong reason to not do so

ShilpaMathew:
@tony-trility This shouldn't be a separate dth. Please include it as a part of our generic dimmer dth. Let me know if this change was done for any particular reason.

PR #2399, opened 2017-10-09
[ICP-1824] Zero is displayed in Temperature and Battery state field after paired Fibaro Smoke Sensor

skt123:
@jackchi 
Changing the default value of `secured` as true, so that the device report its battery and temperature value on pairing. Earlier the device was suppose to report battery status only when the user manually presses the B-button. 
Please review.


dkirker:
adding @mckeed for further feedback. This commit doesn't really make sense. We probably shouldn't set the secured variable if the device wasn't paired securely. If we are supporting the device to be paired unsecured we should accommodate for that. Also, we should be able to use the zwaveInfo map object to check for secured state instead of needing to use a state variable, now that we have more pairing data.

mckeed (changes requested):
I agree with Donald, that this isn't the right fix.

PR #2402, opened 2017-10-09
DVCSMP-3309 Virtual Thermostat SmartApp spelling error

marstorp:
A field called "Emer Temp" is misleading.
Changed the text for the field as well as changed the UI flow and adding
a check in updated for changes in 'outlets" settings, not resubscribing
to event's in case it is become null.

### DIFF
--- a/smartapps/smartthings/virtual-thermostat.src/virtual-thermostat.groovy
+++ b/smartapps/smartthings/virtual-thermostat.src/virtual-thermostat.groovy
@@ -25,31 +25,51 @@ definition(
 )
 
 preferences {
-	section("Choose a temperature sensor... "){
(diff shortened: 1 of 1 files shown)


marstorp:
Made an update so that outlets use capability "capability.outlet" instead of "capability.switch"

workingmonk:
capability.outlet may not work with community device handler unless the UI makes that clear

marstorp:
would this be better?
title: "Outlets (must use capability outlet)"

dsainteclaire (approved):
As long as we're reasonably certain that the majority of outlets that were WWST certified use the outlet capability, and not switch, then I'm good with this.  Which, most of them should use the generic DTHs since they would be Zigbee or Z-Wave

PR #2424, opened 2017-10-13
PROB-1869 Zen thermostat connecting with no battery state

marstorp:
Changing max allowed battery voltage from 6.0 to 6.1 as
logs from a user showed the thermostat reporting 6.1V despite
there's only 4 x 1.5V batteries.
Also made a change so that thermostatMode and fanMode only generate
an event when the mode is changing.

### DIFF
--- a/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
+++ b/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
@@ -250,8 +250,11 @@ def parse(String description) {
             case "001c": // ATTRIBUTE_SYSTEM_MODE
                 // Make sure operating state is in sync
                 map.name = "thermostatMode"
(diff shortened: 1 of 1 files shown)


dsainteclaire (approved):
The mode changing code looks fine to me, but I wonder if we really want to make the voltage change?  Different lithium cells may vary slightly from their 1.5v rating, but that's chemistry that we can't control.  There might also be some cells that only get to 1.45v or something

marstorp:
Well either we change the code or the user will see a battery icon showing there's a battery error.
The proper thing to do though would be to contact Zen and ask them what the valid value range for the voltage is.

dsainteclaire (approved):
🐑

workingmonk:
has this been QAed? if not, then I can merge this to master for the release post sdc

PR #2439, opened 2017-10-18
[DVCSMP-3320]. Support for light bulbs.

sagar-trility:


tylerlange:
We're ready to merge this shortly.

cc @workingmonk

tylerlange:
approved

workingmonk:
@sagar-trility can you remove profileId from the fingerprint if it not required? 
Example: 
`fingerprint inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0B05, 1000", outClusters: "0005, 0019, 0020, 1000", manufacturer: "IKEA of Sweden", model: "TRADFRI bulb E26 opal 1000lm", deviceJoinName: "TRADFRI bulb E26 opal 1000lm"`

workingmonk:
i've changed the base to staging, so that it can go to production next Tue 10/31

PR #2483, opened 2017-11-02
DVCSMP-690 Aeon multisensor (original) updates

greens:
@mckeed : Not sure the exact way you guys do pull requests, but I figure this can be a learning experience.

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
+++ b/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
@@ -92,6 +92,8 @@ metadata {
 	}
 
 	preferences {
+		input description: "Please consult the operating manual for advanced setting options. You can skip this configuration to use default settings",
+				title: "Advanced Configuration", displayDuringSetup: true, type: "paragraph", element: "paragraph"
 		// these are cribbed from the newer aeon-multisensor-6
 		input "motionDelayTime", "enum", title: "Motion Sensor Delay Time",
 				options: ["20 seconds", "40 seconds", "1 minute", "2 minutes", "3 minutes", "4 minutes"], defaultValue: "${motionDelayTime}", displayDuringSetup: true
--- a/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
+++ b/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
@@ -35,26 +35,26 @@ metadata {
 
 		for (int i = 0; i <= 100; i += 20) {
 			status "temperature ${i}F": new physicalgraph.zwave.Zwave().sensorMultilevelV2.sensorMultilevelReport(
-					scaledSensorValue: i, precision: 1, sensorType: 1, scale: 1).incomingMessage()
+				scaledSensorValue: i, precision: 1, sensorType: 1, scale: 1).incomingMessage()
 		}
 
 		for (int i = 0; i <= 100; i += 20) {
 			status "humidity ${i}%": new physicalgraph.zwave.Zwave().sensorMultilevelV2.sensorMultilevelReport(
-					scaledSensorValue: i, precision: 0, sensorType: 5).incomingMessage()
+				scaledSensorValue: i, precision: 0, sensorType: 5).incomingMessage()
 		}
 
 		for (int i = 0; i <= 100; i += 20) {
 			status "luminance ${i} lux": new physicalgraph.zwave.Zwave().sensorMultilevelV2.sensorMultilevelReport(
-					scaledSensorValue: i, precision: 0, sensorType: 3).incomingMessage()
+				scaledSensorValue: i, precision: 0, sensorType: 3).incomingMessage()
 		}
 		for (int i = 200; i <= 1000; i += 200) {
 			status "luminance ${i} lux": new physicalgraph.zwave.Zwave().sensorMultilevelV2.sensorMultilevelReport(
-					scaledSensorValue: i, precision: 0, sensorType: 3).incomingMessage()
+				scaledSensorValue: i, precision: 0, sensorType: 3).incomingMessage()
 		}
 
 		for (int i = 0; i <= 100; i += 20) {
 			status "battery ${i}%": new physicalgraph.zwave.Zwave().batteryV1.batteryReport(
-					batteryLevel: i).incomingMessage()
+				batteryLevel: i).incomingMessage()
 		}
 	}
 
@@ -67,15 +67,15 @@ metadata {
 		}
 		valueTile("temperature", "device.temperature", inactiveLabel: false, width: 2, height: 2) {
 			state "temperature", label:'${currentValue}°',
-					backgroundColors:[
-							[value: 31, color: "#153591"],
-							[value: 44, color: "#1e9cbb"],
-							[value: 59, color: "#90d2a7"],
-							[value: 74, color: "#44b621"],
-							[value: 84, color: "#f1d801"],
-							[value: 95, color: "#d04e00"],
-							[value: 96, color: "#bc2323"]
-					]
+			backgroundColors:[
+				[value: 31, color: "#153591"],
+				[value: 44, color: "#1e9cbb"],
+				[value: 59, color: "#90d2a7"],
+				[value: 74, color: "#44b621"],
+				[value: 84, color: "#f1d801"],
+				[value: 95, color: "#d04e00"],
+				[value: 96, color: "#bc2323"]
+			]
 		}
 		valueTile("humidity", "device.humidity", inactiveLabel: false, width: 2, height: 2) {
 			state "humidity", label:'${currentValue}% humidity', unit:""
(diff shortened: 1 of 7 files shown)


devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -107,20 +110,30 @@ def updated(){
...
 
 // Event Generation
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd)
 {
-	[descriptionText: "${device.displayName} woke up", isStateChange: false]
+	def results = [createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)]
+	// try to configure until we've received a temperature update
+	if (!device.currentValue("temperature")) {
+		log.debug "temp is null, configuring"
+		results += configure()
```
mckeed: You'll also need to send at least some of configure() here in the case that preferences (timeout or period) have been changed since the last wakeup. See aeon-multisensor-6 for an example, where it does `setConfigured("false")` in `updated()`. `configure()` is only automatically called when the device joins, `updated()` is called whenever the preferences are changed.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -85,12 +86,14 @@ metadata {
...
-		}
 
 		main(["motion", "temperature", "humidity", "illuminance"])
-		details(["motion", "temperature", "humidity", "illuminance", "battery", "configure"])
+		details(["motion", "temperature", "humidity", "illuminance", "battery"])
+	}
+
+	preferences {
+		input name: "timeout", type: "number", title: "Motion timeout", description: "Enter guard interval before receiving another motion event, in seconds", range: "20..60", defaultValue: 20
+		input name: "period", type: "number", title: "Reporting period", description: "Enter reporting period, in minutes", range: "4..15", defaultValue: 5
```
mckeed: You should mention the range in the description text, like "Enter guard interval before receiving another motion event, in seconds (20-60)"

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
greens: Ah, okay. I wasn't sure whether/where there was some function that was fired on preference update or not.

greens:
@mckeed Didn't mean to have all of these other commits in this pull request as well. I'll remember in the future to do separate issue work on branches. I can go do the revert runaround, but they're here in the meantime if you want to take a look.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -31,7 +31,7 @@ metadata {
 		status "no motion (basic)"  : "command: 2001, payload: 00"
 		status "motion (binary)"    : "command: 3003, payload: FF"
 		status "no motion (binary)" : "command: 3003, payload: 00"
-		status "wakeup" 			: "command: 8407, payload: 00"
+		status "wakeup" 			: "command: 8407, payload: "
```
greens: Just thought I'd point out that this was never showing up for me when I tested on the simulator so I would do a raw command of `8407`, but that was giving me outofboundsexceptions so I went with `840700` and that seemed to do what I wanted.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -31,6 +31,7 @@ metadata {
 		status "no motion (basic)"  : "command: 2001, payload: 00"
 		status "motion (binary)"    : "command: 3003, payload: FF"
 		status "no motion (binary)" : "command: 3003, payload: 00"
+		status "wakeup" 			: "command: 8407, payload: "
```
greens, line 34: Just thought I'd point out that this was never showing up for me when I tested on the simulator so I would do a raw command of `8407`, but that was giving me outofboundsexceptions so I went with `840700` and that seemed to do what I wanted.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -85,12 +86,18 @@ metadata {
...
-		standardTile("configure", "device.configure", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
-			state "configure", label:'', action:"configuration.configure", icon:"st.secondary.configure"
-		}
 
 		main(["motion", "temperature", "humidity", "illuminance"])
-		details(["motion", "temperature", "humidity", "illuminance", "battery", "configure"])
+		details(["motion", "temperature", "humidity", "illuminance", "battery"])
+	}
+
+	preferences {
```
dsainteclaire, line 94: Because this will be visible to the user when they install it, I wonder if there's a better way to explain it to the user?  Thoughts @marstorp

greens: `		input description: "Please consult AEOTEC MULTISENSOR 6 operating manual for advanced setting options. You can skip this configuration to use default settings",
				title: "Advanced Configuration", displayDuringSetup: true, type: "paragraph", element: "paragraph"`

This explanation makes an appearance in the multisensor 6 DTH that I cribbed this from.

dsainteclaire:
Just the one comment about the preferences change.  Otherwise I don't see anything glaring.  Good job!

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -102,25 +109,39 @@ def installed(){
 def updated(){
```
marstorp, line 111: just want to point out an issue with our platform, sometimes the devices are added silently, i.e. the updated() method is not called from the mobile app.
If this is required to be executed on install then this will have to be triggered from the install method too :(

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
marstorp: Although simulator is still somewhat used today it is not up to date with the current platform which may explain the issues you are seeing.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
marstorp: Sounds like a reasonable description to me :)
However with default value set to "${motionDelayTime}" where motionDelayTime isn't defined can the user really skip this step?

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
greens: I think as written it's only expected to be called on a preference change.

munds: This was an issue I saw in other DTHs. Are we retroactively adding that as "best practice" now? @marstorp

marstorp: if it's only expected to be called on preference change I'm good with this, although I wonder if the user is presented with the choice to change the presences at install if it's a silent install.
But I'm fine with this as is :)

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
greens: The code in configure where the variables are used checks to see if they exist first and uses defaults if not.

marstorp: yes, true but when you enter the preference page can you skip it or do you have to enter a value?

greens: Just checked on my site. It looks like the field is blank. If I want the defaults to be 20s and 5 minutes, should I use 

`defaultValue: "20 seconds"`

or 

`defaultValue: 0`

greens: From the [docs](http://docs.smartthings.com/en/latest/device-type-developers-guide/device-preferences.html?highlight=defaultValue):

> Setting a default value (defaultValue: "foobar") for an input may render that selection in the mobile app, but the user still needs to enter data in that field. It’s recommended to not use defaultValue to avoid confusion.

So the defaultValue field seems to be useless?

dkirker: Yeah... defaultValue is not really used properly (I need to see if there is a mobile ticket open for this). It really gets bad in Android if required:true is set. I think it might work in iOS (it did for me). Check out PROB-1673. If you look at the Aeon Siren code, it has a good use of enforcing defaultValue configuration and making sure that updated values are sent to the device.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
dkirker: @marstorp Shouldn't matter either way. Now the logic is: configure() should get called by the platform (which is perfect for when the device is powered via the USB port), this device will also receive the configure() during the Z-Wave WakeUpNotification if the device is battery powered.

workingmonk:
can you add a short description to the PR title after the ticket? makes it easier for the deploy blog

PR #2490, opened 2017-11-06
[DVCSMP-3377] Schlage Locks - For some locks DTH is assigning default

vaish-rahul:
code name instead of code name provided by user.


vaish-rahul:
@tpmanley Please review the changes.

tpmanley (approved):

PR #2500, opened 2017-11-09
ICP-2151 ICP-2236 ICP-2237 Add Samsung Lock fingerprint to DTH

dkirker:


workingmonk:
certification complete?

workingmonk:
and what is the expected release date?

dkirker:
@workingmonk  Trying to get that from the cert team, that way we aren't putting the cart before the donkey. I'm not aware of any urgency. I'll merge after a "+1" and the green-light from cert.

PR #2503, opened 2017-11-10
DVCSMP-3374 Zen Thermostat has no settings for "Heat only" or "Cool only"

marstorp:
Customer reported on PR 2296;
"My Zen can only Heat (10A setting I think). There is no option for "Off, Heat" in the thermostat modes settings UI."
Adding ["off", "heat"] and ["off", "cool"] to settings.

### DIFF
--- a/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
+++ b/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
@@ -4,7 +4,7 @@
  *  Author: Zen Within
  *  Date: 2015-02-21
  *  Updated by SmartThings
- *  Date: 2017-10-12
+ *  Date: 2017-11-12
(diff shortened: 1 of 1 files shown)


greens (approved):
Zigbee doesn't have a way of querying the device for supported modes?

marstorp:
@greens if I understand Zigbee spec there is. However when I tried using those commands the thermostat always returned the same thing. I did leave the code, commented out as reference though. It's possible I didn't do the right thing :P

pdonovan:
I'm the original customer mentioned in the first comment.. I made a change similar to this myself in my own DTH just to see what would happen.
After I set the preference to [off, heat] I noticed that the logging for what the Zen was sending back in supportedModes (which previously always said off, heat, cool) started reporting [off,heat]. I'm not sure why that happened.

PR #2510, opened 2017-11-13
DVCSMP-1524 Remove polling capability from hub-polled devices

greens:
Removes the deprecated polling capability from these devices which are polled by the hub directly.


greens:
Is this still a change that should be done? Did the DTH control for hub polling mentioned in the jira issue ever materialize?

mckeed (approved):
Yeah, this is still needed. It shouldn't have any negative effect, but we should let support know when it goes out so they can keep an eye out just in case.

greens:
@mckeed How do we notify support about this, and what would this look like for users?

PR #2511, opened 2017-11-13
DVCSMP-2220 Aeon Siren not sending alert on secure inclusion failure

greens:
This just cribs the way other secure devices handle this failure. In the future we'd like to have this standardized or done automatically.

### DIFF
--- a/devicetypes/smartthings/aeon-siren.src/aeon-siren.groovy
+++ b/devicetypes/smartthings/aeon-siren.src/aeon-siren.groovy
@@ -97,9 +97,23 @@ def updated() {
 def parse(String description) {
 	log.debug "parse($description)"
 	def result = null
-	def cmd = zwave.parse(description, [0x98: 1, 0x20: 1, 0x70: 1])
-	if (cmd) {
-		result = zwaveEvent(cmd)
+	if (description.startsWith("Err")) {
+		if (state.sec) {
+			result = createEvent(descriptionText:description, displayed:false)
+		} else {
+			result = createEvent(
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-siren.src/aeon-siren.groovy
```diff
@@ -97,9 +97,23 @@ def updated() {
 def parse(String description) {
 	log.debug "parse($description)"
 	def result = null
-	def cmd = zwave.parse(description, [0x98: 1, 0x20: 1, 0x70: 1])
-	if (cmd) {
-		result = zwaveEvent(cmd)
+	if (description.startsWith("Err")) {
```
kianooshkarami, line 100: This reminds me, we should probably add something at the hub firmware to allow for the user to attempt secure inclusion without having to exclude and include to attempt again. I will make a note of it for future work, and we can discuss how/steps to bring this for users.

mckeed: The spec for S0 requires exclude on failed secure inclusion, but we should look into it for S2

mckeed (approved):

devicetypes/smartthings/aeon-siren.src/aeon-siren.groovy
kianooshkarami: Oh yeah, good point. It be nice to show an icon next to devices that failed secure inclusion to remind the user to do that.

PR #2512, opened 2017-11-14
[ICP-2233] During scanning of Lock Codes, Recently tab shows multiple

vaish-rahul:
events for added lock codes.

### DIFF
--- a/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
+++ b/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
@@ -243,9 +243,9 @@ def reloadAllCodes() {
 	sendEvent(lockCodesEvent(lockCodes))
 	def cmds = validateAttributes()
 	if (isYaleLock()) {
-		state.checkCode = 1
+		state.checkCode = state.checkCode ?: 1
 	} else {
-		state.checkCode = 0
+		state.checkCode = state.checkCode ?: 0
 	}
 	cmds += requestCode(state.checkCode)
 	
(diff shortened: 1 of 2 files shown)


vaish-rahul:
@tpmanley Please review this PR. The changes against the JIRA ticket are already verified by SRI-B QA team.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1195,7 +1195,7 @@ def reloadAllCodes() {
 		cmds << secure(zwave.userCodeV1.usersNumberGet())
 	} else {
 		sendEvent(name: "maxCodes", value: state.codes, displayed: false)
-		cmds << requestCode(1)
+		cmds << requestCode(state.checkCode)
```
tpmanley, line 1198: Do you know that `state.checkCode` will always be valid here? You have null checks in other places so should you use `requestCode(state.checkCode ?: 1)` here?

vaish-rahul: At line 1186 state.checkCode will become 1 if it's null. So it is fine to use state.checkCode here.

vaish-rahul: @jackchi Please review the changes in PR. The changes are tested by SRI-B QA team.

vaish-rahul:
@jackchi Please review this PR. The changes are validated by SRI-B QA team against the JIRA ticket.

jackchi:
👍  Thank you for fixing this issue. LGTM.

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -243,9 +243,9 @@ def reloadAllCodes() {
 	sendEvent(lockCodesEvent(lockCodes))
 	def cmds = validateAttributes()
 	if (isYaleLock()) {
-		state.checkCode = 1
+		state.checkCode = state.checkCode ?: 1
```
workingmonk, line 246: is this supposed to be a null check? Elvis operators treat 0 as false too. And if the value is anything other than 0/null/false, it will keep that value instead of setting the value to 1. is that the intended behavior?

vaish-rahul: Yes this is the intended behavior. 0 slot is not valid for Yale locks hence 1 should be picked up in that case. For Kwikset locks 0 is a valid slot however even if elvis operator considers it as false we are picking up 0 by default at line 248.

PR #2513, opened 2017-11-14
Don't suppress code update/unset notifications

rboy1:
Required for operation for SmartApps to confirm that all transactions send to lock have been processed by lock, don't suppress the notifications, pass them onto the SmartApps

@tylerlange can this please be reviewed and merged

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -773,7 +773,7 @@ def zwaveEvent(UserCodeReport cmd) {
 				result << codeSetEvent(lockCodes, codeID, codeName)
 			} else {
 				map.value = "$codeID changed"
-				map.isStateChange = false
+				map.isStateChange = true
 			}
 			map.descriptionText = "${getStatusForDescription('set')} \"$codeName\""
 			map.data = [ codeName: codeName, lockName: deviceName ]
@@ -805,7 +805,7 @@ def zwaveEvent(UserCodeReport cmd) {
 				result << codeDeletedEvent(lockCodes, codeID)
 			} else {
 				map.value = "$codeID unset"
-				map.isStateChange = false
+				map.isStateChange = true
 				map.data = [ lockName: deviceName ]
 			}
 		}


tylerlange:
@vaish-rahul  @jackchi  please review, suggested to continue to support community lock solutions.

jackchi:
Will do. Can @rboy1 provide a link to the Community Supported Lock Solutions so we can test it out?

rboy1:
@jackchi if you send me a PM with your email I can get you access to the server or we can test it out and let you know the results.

rboy1:
Also related similar issue on ZigBee DTH

https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2514

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -805,7 +805,7 @@ def zwaveEvent(UserCodeReport cmd) {
 				result << codeDeletedEvent(lockCodes, codeID)
 			} else {
 				map.value = "$codeID unset"
-				map.isStateChange = false
+				map.isStateChange = true
```
vaish-rahul, line 808: During scanning if a slot is found to be empty it is not a good experience to display `unset` event as nothing has been `unset` actually. If there are lock codes in `lockCodes` attribute but during scanning a slot is found to be empty then DTH will generate `codeChanged` event with value `$codeID deleted`. SmartApps can use this event for updating their state.

Still if your SmartApp is dependent on `$codeID unset` event then we can handle this by keeping `isStateChange` to true and `displayed `to false. Your SmartApp's subscription handler should still get called.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -773,7 +773,7 @@ def zwaveEvent(UserCodeReport cmd) {
 				result << codeSetEvent(lockCodes, codeID, codeName)
 			} else {
 				map.value = "$codeID changed"
-				map.isStateChange = false
+				map.isStateChange = true
```
vaish-rahul, line 776: While scanning let's say 8 slots are found to be filled. `lockCodes` attribute is updated with this information. Now user initiates scanning once again. Now DTH should not generate `$codeID changed` event because the code has not changed actually. The DTH will generate `$codeID set` event if a slot is found occupied during scanning and the information about the same is not available in `lockCodes` attribute. SmartApp can read the lock codes related information from `lockCodes` attribute. Also we have another attribute `scanCodes` whose value will change from scanning to complete. SmartApps can subscribe to `scanCodes` and read updated `lockCodes` attribute after scanning.

Still if your SmartApp is dependent on `$codeID changed` event then we can handle this by keeping `isStateChange` to true and `displayed `to false. Your SmartApp's subscription handler should still get called.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
rboy1: Perfect Rahul, that's a great solution.

Existing apps manage the users in the SmartApp, they keep track of the lock database through their own mechanisms, so they rely heavily on getting *every* notification from the lock to manage those databases. Without getting into the history of why SmartApps were designed this way, sometimes when the slot is supposed to be empty, they aren't! So there are safety mechanisms built in to handle clear codes, sometimes they *may* be empty and in those cases these notifications are critical for apps to track which slots were infact empty and which weren't.

Also how would you handle locks that don't report codes? How would the DTH reliably track codes without knowing which code was sent on the lock (you may be surprised but users do set codes on locks *while* using a SmartApp which leads to total havoc when locks don't report codes).

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
rboy1: Correct, see above

greens (changes requested):
Hi, just getting up to speed with everything here.

Seems like the solution we've landed on would be to have `isStateChange` be `true` and `displayed` be `false` in these specific cases.

rboy1:
Correct, in general the DTH should pass *ALL* events up to the SmartApps and if it's believes it should not be displayed to the user the `displayed` should be marked as `false`, and let SmartApps decide how to handle the events.

vaish-rahul:
@jackchi @greens: This request is being worked upon as part of DVCSMP-3396. You should go through the ticket's description to understand how this will affect Smart Locks SmartApp. Anyway, we will work on this for both Zigbee and Z-Wave locks and get the changes validated by QA team. We'll raise a single PR against [DVCSMP-3396](https://smartthings.atlassian.net/browse/DVCSMP-3396) for both Z-Wave and Zigbee locks. Since DVCSMP-3396 causes DVCSMP-3397 we'll have to make sure that Smart Locks changes are merged along with the DTH changes.

vaish-rahul:
@jack, @greens, @tylerlange, @rboy1: We don't need to merge this PR. I think https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2520 should be good enough to resolve the problems @rboy1 is facing. @rboy1 you can also have a look and probably test your use cases with DTHs from https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2520.

PR #2514, opened 2017-11-14
Don't suppress code update/unset notifications

rboy1:
SmartApps require all notifications from lock

Related to https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2513

@jackchi @tylerlange


vaish-rahul:
@jackchi @greens: This request is being worked upon as part of DVCSMP-3396. You should go through the ticket's description to understand how this will affect Smart Locks SmartApp. Anyway, we will work on this for both Zigbee and Z-Wave locks and get the changes validated by QA team. We'll raise a single PR against DVCSMP-3396 for both Z-Wave and Zigbee locks. Since DVCSMP-3396 causes DVCSMP-3397 we'll have to make sure that Smart Locks changes are merged along with the DTH changes.

vaish-rahul:
@jack, @greens, @tylerlange, @rboy1: We don't need to merge this PR. I think #2520 should be good enough to resolve the problems @rboy1 is facing. @rboy1 you can also have a look and probably test your use cases with DTHs from #2520.

PR #2520, opened 2017-11-16
[DVCSMP-3396] Locks DTH should generate event for changed and unset

vaish-rahul:
during scanning.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -768,14 +768,14 @@ def zwaveEvent(UserCodeReport cmd) {
 		} else {
 			// We'll land here during scanning of codes
 			codeName = getCodeName(lockCodes, codeID)
+			def changeType = getChangeType(lockCodes, codeID)
 			if (!lockCodes[codeID]) {
-				map.value = "$codeID set"
 				result << codeSetEvent(lockCodes, codeID, codeName)
 			} else {
-				map.value = "$codeID changed"
-				map.isStateChange = false
+				map.displayed = false
 			}
-			map.descriptionText = "${getStatusForDescription('set')} \"$codeName\""
+			map.value = "$codeID $changeType"
+			map.descriptionText = "${getStatusForDescription(changeType)} \"$codeName\""
 			map.data = [ codeName: codeName, lockName: deviceName ]
 		}
 	} else if(userIdStatus == 254 && isSchlageLock()) {
@@ -805,7 +805,7 @@ def zwaveEvent(UserCodeReport cmd) {
 				result << codeDeletedEvent(lockCodes, codeID)
 			} else {
 				map.value = "$codeID unset"
-				map.isStateChange = false
+				map.displayed = false
 				map.data = [ lockName: deviceName ]
 			}
 		}
--- a/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
+++ b/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
@@ -736,7 +736,7 @@ private def parseCommandResponse(String description) {
 				if ("set" == changeType) {
 					result << codeSetEvent(lockCodes, codeID, codeName)
 				} else {
-					responseMap.isStateChange = false
+					responseMap.displayed = false
 				}
 			}
 		} else {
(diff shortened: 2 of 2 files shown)


vaish-rahul:
@jackchi This PR is for accommodating the changes suggested by RBoy in Locks DTH. Please review the changes. However, this should be merged after https://github.com/SmartThingsCommunity/SmartThingsPrivate/pull/2788

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -768,14 +768,14 @@ def zwaveEvent(UserCodeReport cmd) {
 		} else {
 			// We'll land here during scanning of codes
 			codeName = getCodeName(lockCodes, codeID)
+			def changeType = getChangeType(lockCodes, codeID)
 			if (!lockCodes[codeID]) {
-				map.value = "$codeID set"
```
greens, line 772: My apologies for being a bit slow to pick up what's going on here, but why are we removing this value from these events?

Also what's the difference between the two events created in this case? (one from createSetEvent and one from the map) Are they each used differently?

vaish-rahul: Hi Steven. We are not removing the value from events. The value will be added at line number 777. We have just shifted it down because the logic is common for both the `if` and `else` loops.

- `map` will create an event with name `codeChanged`. This event is handled by Smart Locks SmartApp in case of code creation, code deletion, code update etc. This event is also handled by existing 3rd party SmartApps.
- `codeSetEvent` will create two events:
   1. An event that updates `lockCodes` attribute.
   2. `codeReport` event. This event is legacy event and is needed for supporting existing SmartApps of 3rd party developers. There is a plan to deprecate this event but not in near future.

jackchi (approved):

rboy1:
@vaish-Rahul this may not work always because if the value of the event doesn't change it won't be reported to the SmartApp. So the apps send a delete command and the lock responds but the value of the event doesn't change it won't be reported to the app and will cause the same issue. 

The ideal solution is what you proposed originally. Keep `isStateChanged` as true and `displayed` as false. That will ensure the events reach the SmartApp as the locks sends them in.

vaish-rahul:
@rboy1 In this case event's value does change. The values will be: 

- `1 set`, `2 set` etc. when code is `created` or when slot is found `occupied` during scanning.
- `1 deleted`, `2 deleted` etc. when code is `deleted`.
- `1 unset`, `2 unset` etc. when slot is found `empty` during scanning.
- `1 changed`, `2 changed` etc. when code is `updated` or when a slot is found `occupied` during scanning but `lockCodes` map already contains the slot.

Ideally we should rely on platform to set the `isStateChanged` flag. In this case the flag will be set to `true` because value will change as it contains the slot number.

When a app sends a delete command the code will be deleted from a slot. The slot number would be 1, 2, 3, etc and correspondingly the event's value will be `1 deleted`, `2 deleted`, `3 deleted`.

Can you let me know the use case in which DTH is failing to provide the event to your SmartApp?

rboy1:
Yes that would work most of the times, had a similar issue with the original DTH because of which the same patch had to be applied there.

Here's the scenario where this fails, if the SmartApp sends a delete code request to a slot that's empty (say 1) the DH will return `1 unset`, if the SmartApp sends another delete code request (it's very rare but it does happen due to race conditions of when methods are called) now the DH sends `1 unset` it doesn't reach the SmartApp because `isStateChanged` wasn't set and it assumes the lock didn't respond which them leads to more issues.

Hence the same patch was applied to the original code, to be safe the DH should pass all responses back to the SmartApps and let the SmartApps decide whether to ignore the response or not. So settings the `isStateChanged` to `true` for each response from the lock is the likely compatible way forward as you had originally proposed with the `displayed` set to `false`

vaish-rahul:
@rboy1 Please try executing the use case you mentioned. The event should be generated each time you call `deleteCode`.

In Z-Wave Lock DTH `isStateChange` flag is set to `true` at line no. 739 and it will remain `true` even in case of `<slot> unset` and `<slot> changed`.

In Zigbee Lock DTH `isStateChange` flag is set to `true` at line no. 700 and it will remain `true` even in case of `<slot> unset` and `<slot> changed`.

So we don't need to change the DTH code. Kindly check and let me know if something doesn't work for you.

rboy1:
I'll double check and run some tests. Just being a little careful on this one because it was very problematic the last time we tried to replicate the issue and debug it (previous handler)

rboy1:
It's looking good so far, will keep testing it. Thanks

rboy1:
Quick question - when does this change reflect in the v1 and v2 hubs?

vaish-rahul:
I guess on 5th December 2017. @jackchi Correct me if I am wrong.

PR #2521, opened 2017-11-16
DEVC-554 DTH implementation for Cooper/Eaton 5 button Scene Controller

ALamchaS:
Initial implementation of Cooper 5-scene controller DTH


devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
```diff
@@ -0,0 +1,268 @@
...
+	cmds << new physicalgraph.device.HubAction("delay 500")
+	cmds << new physicalgraph.device.HubAction(zwave.indicatorV1.indicatorGet().format())
+	//Indicator returns number which is a bit representation of current state of switches
+	sendHubCommand(cmds, 500)
+}
+
+def parse(String description) {
+	if (description.startsWith("Err")) {
+		if (description.startsWith("Err 106") && !state.sec) {
+			state.sec = 0
```
greens: Can this device ever add securely? My guess is not based on none of the other commands being encapsulated. Since that's the case this check seems unnecessary.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
```diff
@@ -0,0 +1,268 @@
...
+		def cmds = []
+		//Set Association for this group
+		cmds << new physicalgraph.device.HubAction(zwave.associationV1.associationSet(groupingIdentifier:cmd.groupingIdentifier, nodeId:[zwaveHubNodeId]).format())
+		cmds << new physicalgraph.device.HubAction(zwave.associationV1.associationSet(groupingIdentifier:cmd.groupingIdentifier, nodeId:[zwaveHubNodeId]).format())
+		sendHubCommand(cmds, 1500)
+	}
+}
+
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
```
greens: Seems strange to me that the device sends a basic set but never a basic report, but I don't have the docs to verify.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
```diff
@@ -0,0 +1,268 @@
...
+	sendHubCommand new physicalgraph.device.HubAction(zwave.indicatorV1.indicatorSet(value: binarySwitchState).format())
+}
+
+private addChildSwitches() {
+	for (i in 1..5) {
+		String childDni = "${device.deviceNetworkId}/${i}"
+		def child = childDevices.find{it.deviceNetworkId == childDni}
+		addChildDevice("Child Switch", "${childDni}", null,
+				[completedSetup: true, label: "${device.displayName} switch ${i}",
+				 isComponent: true, componentName: "switch$i", componentLabel: "Switch $i"])
```
greens: Not a big deal, but there's a lot of unnecessary closures here.

`"${childDni}"` is already a string and doesn't need to be substituted

`${i}` doesn't need a closure

greens (commented):
There are a couple of things that aren't a big deal that I think are worth a look, and there are a couple of places I'm asking for some clarification.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
```diff
@@ -0,0 +1,268 @@
...
+}
+
+private updateLocalSwitchState() {
+	def binarySwitchState = 0;
+	def multiplier = 1;
+	for (int i = 1; i <= 5; ++i) {
+		String childDni = "${device.deviceNetworkId}/${i}"
+		def child = childDevices.find{it.deviceNetworkId == childDni}
+		if (child?.device.currentState("switch")?.value == "on") {
+			binarySwitchState += multiplier
```
greens: Switch 1 should be bit 0, Switch 2 should be bit 1, etc

I think this will be reversed from how you'd want it to be, no?

greens (commented):
@mckeed @bflorian Thoughts on using the "Button" capability instead of "Switch"?

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
ALamchaS: I am not quite certain what you mean. Could you please elaborate?

If you mean binarySwitchState value, that is how it works. For button 1 we get multiplier == 1, so will bit 0 (least significant bit, which in IndicatorReport I called first bit, which may be confusing :/ ) will be 1 if switch is "on" and 0 if it is "off" (and it is not quite our design, that is how device handles value for IndicationSet command).

If you mean ${i} index in childDni, we kept approach from existing Aeon Minimote, where switches are numbered from 1 to 4 in UI, and keep the same values in childDni (probably to avoid confusion when debugging). Of course we can change that if you want.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
ALamchaS: No, it can not.

I missed that. This was left from template. Thanks for the catch.

Will remove this.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
ALamchaS: Thanks, I will fix those.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
ALamchaS: Device can control scene capable devices.
In that case it sends SceneActivationSet with appropriate sceneId when switch is switched on, and BasicSet(value: 0) if switch is switched off. In this mode, device does not send any BasicReport nor does it respond to BasicGet commands.

For non-scene capable devices, it is possible to configure level for each association group for each switch. In that case we get BasicSet with value as configured when scene is activated, and BasicSet(value: 0) when scene is deactivated. In this mode, device does not send any BasicReport nor does it respond to BasicGet commands.

ALamchaS: We currently work on acccount st.dth.dev@gmail.com, location name US_LAB

We will leave device plugged in overnight. You can impersonate to check it if you want.

If you prefer we can collect telnet logs  I don't have any other ideas.

devicetypes/smartthings/cooper-scene-controller.src/cooper-scene-controller.groovy
greens: Oh nope, you're right.

I read it wrong.

mckeed:
Yeah, the other certified devices like this are using _Button_ capability. _Switch_ is for devices that you turn on and off, not that turn other things on and off.

ALamchaS:
What about blue indicator LED on each button then?

It shows which scenes are currently active.

With buttons we would be forced to ignore this functionality.

greens:
Map "on" and "off" to "pushed" and "held".

Change the child device's tiles from actions to indicators.

See the "aeon-key-fob" and "child-button" DTHs for examples. I think the indicator light methods should fit well on top of them.

ALamchaS:
I have made suggested on/off -> pushed/held changes locally. I used Child Button DTH from Aeon minimote. It does not seem to show button state in UI. It makes set up a bit more complicated.

Tested this with Smart Lighting automation.
Results:
- Toggle on & off option becomes confusing and should always be disabled, as it will switch light on/off every second button press.
- To operate in synch with 5 stage keypad blue led indicator states, user needs to add 2 automations for each button - 1st for turning device on when button is pushed, 2nd for turning device off when button is held.

Please let me know if this behaviour is intended. If it is, I will submit my changes.

greens:
I may have been fundamentally misunderstanding what this device does. I think I have a better handle on it now.

Your first implementation seems to be correct (switches). Just make another commit with the style fixes and it should be good.

greens (approved):

greens (changes requested):
It looks like the manufacturer would prefer the name "Eaton" instead of "Cooper".

ALamchaS:
I managed to convert 1 sendEvent to createEvent without breaking anything.

- name change
- code refactored a little bit

greens (dismissed):

greens (dismissed):

ALamchaS:
Hi. I updated this Pull Request as there was a problem with Connect App. As we can not run update() on demand there, initial set up needs to be 100% successful. I extended set up time by about 50% to improve success rate of initial set up. I also removed All Off from supported functionalities list, as it fails a lot today.

greens (dismissed):
I like the refactoring too.

We seem to be having a lot of issues over there around devices' initial polls. Any insights on your side?

greens:
Hey @ALamchaS, just got off the phone with the product team. In light of not being able to have a no-functionality device in OneApp, can we use the same solution as the Neo dual switch? Have the parent device also expose the functionality of the first "child device"?

greens (dismissed):
I like this, but I think I'll wait to put it in until I have that talk with Bob and Vinay.

greens:
@workingmonk Would this device also be a good candidate for the generic metadata?

workingmonk:
Yes. it can use the generic-switch

greens:
@ALamchaS Can you add the references to the generic ui-metadata in both of these DTHs?

greens (approved):

PR #2523, opened 2017-11-17
DEVC-555 Create DTH for Eaton/Cooper Battery operated switch (RF9575)

ALamchaS:


greens (changes requested):
Is there intended to be any actual functionality with this device? Correct me if I'm wrong, but the `on` and `off` methods just update the state and don't actually send ZWave commands.

If this device is actually a switch, you need to send commands.

If this device isn't actually a switch we can control, there shouldn't be any control tiles.

Also the issue mentions that the device is battery-operated, but I don't see any reference to that here at all.

ALamchaS:
Device is not very intuitively designed. I think it is intended to be used exclusively as a remote for actual switch or dimmer.

- It does not support Wakeup commands.
- It wakes up ONLY when user presses it. (User needs to press buttons during pairing to allow hub to retrieve manufacturer specific and version reports)
- It does not respond to BasicGet and SwitchMultilevelGet commands

To act as switch it sends BasicGet to associated device and responds to BasicReport(0) with BasicSet(255) and to BasicReport(255) with BasicSet(0).

As long as it thinks it is on (which is crazy, as it does not behave as if it stored state), it enables dimmer up and dimmer down buttons.



Our approach was to make DTH act as virtual switch/dimmer, to make device behave as advertised by manufacturer.

Alternatives would be:
- handle the device as a button (which it is)
- ignore dimmer buttons
- treat dimmer buttons as buttons (which they are; still requires tricking the device to always think it is on)

greens:
I think my big issue is that you've presented controls when there actually aren't any. We can't control the device from the app at all (in fact, using the controls will make the app go out of sync with the actual state of the device), so I think these tiles should just be changed to indicators of the current level or on/off state.

I might replace the "Switch" capability with the "Button" capability. For convenience you should leave "Switch level", but replace the actual control with a value tile.

Also, does the device not have any **battery** functionality at all?

ALamchaS:
No, it does not. We placed z-wave string is in comments before footprint.

zw:S type:0000 mfr:001A prod:4243 model:0000 ver:3.01 zwv:3.67 lib:01 cc:72,77,86,85 ccOut:26

Supported command classes:
Manufacturer Specific | 0x72
Node Naming | 0x77
Version | 0x86
Association | 0x85

cc out:
Switch Multilevel | 0x26

It supports Basic command class too.

I will change DTH to disable controls.

greens (approved):

greens:
I'd misunderstood how this device worked. The button presses (virtual and otherwise) are meant to generate smart things events to be consumed by smartapps and automations. In light of me now understanding how this device was always intended to work, I approve of the changes.

greens (changes requested):
It looks like the manufacturer would prefer the name "Eaton" instead of "Cooper".

ALamchaS:
Changes:

- name changed from Cooper to Eaton
- implemented proper createEvent, response usage
- control tiles were changed to value tiles
- DTH no longer handles on/off/setLevel events from SmartApps at all to not desynchronize DTH state with device state
- device is aware of current state, so it will behave as designed by manufacturer when paired with more devices than just SmartThings Hub
- device is aware of current state, so LED indicator will use blinking signals as designed by manufacturer
- removed properties that are not required for basic device functionality

greens (dismissed):

ALamchaS:
Ok. So to make this process a bit faster, I prepared 2 versions of DTH for Eaton Anyplace Switch:

As we discussed, this device can work by using normal SmartThings automations, or direct z-wave association.

Button variant:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2523/commits/27eb3fc6494a2840908c291c9eed29359cd79760

Button variant makes it more similar to how ST usually handles similar "remote" devices, so it is more ST-centered approach. There is no way to synchronize devices associated with this switch, so it is possible first button press will sych button's state with other devices instead of changing their on/off state.

Switch variant:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2523/commits/bb85652d994b986c174215f910a1ebd6f8ac83bc

Switch variant will be different than usual implementation of similar devices. It may make blinking led indications a bit more meaningful. When using direct z-wave association, user can create automation that will use SmartThings to update cloud state of Anyplace Switch to keep it synchronized with z-wave associated devices.

Which version is better?
Button variant is my first button implementation, so please comment everything that can be improved.

greens:
Couple of questions:

For the switch implementation, is it possible that our representation of the switch gets out of sync with the state of the device? Is there any indicator on the device what the state is? Because I might just lean towards instead turning all events into toggles? i.e. if our representation is 'on', treat any event as an 'off'.

For the button implementation, why did you include a custom switch attribute?

I'm leaning towards switch, because that would give us the most functionality in OneApp.

ALamchaS:
When device's button it pressed, it sends BasicGet, and waits for BasicReport.
If BasicReport holds 0 value, device sends BasicSet(255), otherwise it sends BasicSet(0).

That is why we hold switch attribute in button representation - we need to know device's current state to report it.

In switch representation:
- Device always syncs switch state with our cloud state using BasicGet, it is impossible to get out of sync in this implementation. Sync problems encountered earlier affected dim up and dim down buttons, which we fortunately do not support anymore.
- State indicator is visible only after button press to conserve battery. If device  turned off - led blinks. If device turned on - led turns on for about 2 seconds.
- Device toggles itself using state we send it. If we used BasicGet as a trigger for toggle, we would still need to send something to device, otherwise it will not send BasicSet and led indicator will not work.

greens:
Which one do you like better?

ALamchaS:
Switch variant.

greens:
Alright. Works for me. Revert the button commit and I'll pass it along to get the ui metadata started.

greens (approved):
This looks good, but let's wait until Tim's got some ui-metadata done before we merge.

PR #2534, opened 2017-11-20
MSA-2379: FIBARO Motion Sensor ZW5 **Official Request by Fibaro"

SmartThings:
Submitting 'FIBARO Motion Sensor ZW5 **Official Request by Fibaro"' for publication


sagar-trility:
Just had one question. Wanted to understand more as to why the following line was removed from the new DTH file:

cmds += zwave.wakeUpV2.wakeUpIntervalSet(seconds: 7200, nodeid: zwaveHubNodeId)//FGMS' default wake up interval

I am asking because during testing I am not seeing the device wake up.

PR #2549, opened 2017-11-21
DVCSMP-690 Aeon multisensor improvements

greens:
Fixes major oversight where commands would not be handled.

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
+++ b/devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
@@ -127,7 +127,7 @@ def parse(String description)
 		results = createEvent(descriptionText:description, displayed:true)
 	} else {
 		def cmd = zwave.parse(description, [0x31: 2, 0x30: 1, 0x84: 1])
-		if(cmd) results += zwaveEvent(cmd)
+		if(cmd) results = zwaveEvent(cmd)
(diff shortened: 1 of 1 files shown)


greens:
@workingmonk , my last work on this device definitely broke it, so we might want to roll back those commits at least if we're doing a deploy soon.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -137,12 +137,13 @@ def parse(String description)
 // Event Generation
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd)
 {
-	def results = [createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)]
+	def results = []
+	results << createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)
 	// If we haven't configured yet, then do so now
 	if (!isConfigured()) {
-		results += configure()
+		results << configure()
```
mckeed: Since `configure()` returns commands, it has to be in the `response()` call to be bundled together as a HubMultiAction with the wakeUpNMI. See how Aeon Multisensor 6 does it with `result` vs `cmds`

mckeed (changes requested):

mckeed (approved):
Looks good. You should retarget this PR to the staging branch.

PR #2555, opened 2017-11-23
DEVC-557 DTH implementation for Eaton/Cooper Z-Wave outlet RFTR9505-TAW

ALamchaS:


devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
```diff
@@ -0,0 +1,172 @@
...
+				   isStateChange: true])
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.Command cmd) {
+	// Handles all Z-Wave commands we aren't interested in
+	log.debug "Unexpected zwave command $cmd"
+}
+
+private hubAction(formattedCommand) {
```
greens: If you're gonna do this already, I'd just pass in the unformatted command and call `format()` inside this method.

devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
```diff
@@ -0,0 +1,172 @@
...
+	new physicalgraph.device.HubAction(formattedCommand)
+}
+
+private sendCommand(formattedCommand) {
```
greens: You're not using this method.

devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
```diff
@@ -0,0 +1,172 @@
...
+		updateDataValue("productTypeId", cmd.productTypeId.toString())
+	}
+	// productId 0 is a valid value
+	if (cmd.productId != null) {
+		updateDataValue("productId", cmd.productId.toString())
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	//settings are read-only, so send event if settings are out of sync
```
greens: For here and below, you can do

`device.updateSetting("powerUpState", cmd.configurationValue[0])`

It's not documented, but I've tested that it works.

greens (changes requested):
A couple of things to fix, also I missed this in previous PRs, but as a general style and consistency suggestion, we should leave the responsibility of sending events and commands to the caller of `parse`. That is, `zwaveEvent` calls should return their events, and `parse` should handle the bundling and formatting of those and then return them instead of an empty list and instead of calling `sendEvent` in the `zwaveEvent` methods.

devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
ALamchaS: Thank You very much. That's huge help.

ALamchaS:
Thank you for suggestions. Could you please dismiss reviews on other 3 pull requests?

Ingvar decided we need to change name of all those DTHs from Cooper * to Eaton *

devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
ALamchaS: I tried this in dimmer implementation in zwave command handler method: 

`device.updateSetting("delayedOffTime", 15)`

This results in:

>  c24c73e7-270b-4629-a1ed-3fc5818128d4 2:12:01 PM: error org.hibernate.LazyInitializationException: failed to lazily initialize a collection of role: physicalgraph.device.Device.preferences, no session or session was closed 

Same method call works perfectly in refresh()

What am I doing wrong?

greens: I guess this is why it's undocumented. 

As an alternate solution, try doing what is done here: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/ge-link-bulb.src/ge-link-bulb.groovy

Specifically look at the relationship between `dimRate` and `state.rate`.

Essentially, the DTH has a duplication of the preference in `state.____`. Updates from the user change this value as well as updates from the device, and it is treated as the canonical value for logic purposes.

ALamchaS: Does zigbee handle state.rate updates from device internally? This DTH seems to handle only commands with resultMap.name other than "level".

We can use mechanism you describe like this:
- Add value tile for each property, so that user can go to preferences and change them if he does not like those values.
- Hold received configuration value in state and update preferences in next refresh call (or any other capability method if it works there; we would have to check if it works when called by SmartApp too)
- Combination of the above, which could be confusing as values in preferences will or will not match those in value tile depending on other user and/or SmartApp actions.

MAblewicz:
According to Ingvar Marstorp training here in SRPOL he told us that using sendHubCommand(cmds)  and sendEvent() is  better because have better control when they will be send. He also pointed out that if we start using those method we need to be consistent in our DTH with it. I also talked with David Sainte-Claire about using those methods and he suggested that we should follow Ingvar advices. So if there is no other reason than styling changes to avoid using sendHubCommand(cmds) and sendEvent() we will continue using them.

devicetypes/smartthings/cooper-receptacle.src/cooper-receptacle.groovy
greens: I think option 2 is the best, but I would still use the state value as the authoritative value instead of the preference. So in that case the three main points are:

* `update` gets called, so set the state value and send the zwave set.

* zwave report received so update the state value

* capability method called so set the preference to the state value if they're different (`updatePreference` does seem to work in any capability method)

ALamchaS: For now, I added attributes and standard tiles to display current configuration.

I tested option 2. I added SmartLighting automation that turned the receptacle on/off when remote switch is turned on/off. Unfortunately this resulted in org.hibernate.LazyInitializationException each time on() or off() method was called by SmartLighting SmartApp (by switching remote switch on/off).

ALamchaS:
I changed DTH to use current event/command approach, changed name from Cooper to Eaton and refactored the code a bit.

If sendEvent/sendHubCommand approach will be decided as better, we can change this DTH with no problems.

greens:
In light of `updateDeviceSetting` not working, I'd rather you just go back to the original code and remove the tiles. I don't think those two preferences should be crucial to most users.

greens (approved):

PR #2569, opened 2017-11-28
PROB-1904 Adding changeable icons back to zwave-relay and momentary-button-tile 

mikesip:
In a recent update to change these DTHs to use multi-attribute tiles the option to change the device icon for them was lost. This has caused a few users to not be able to change icons on devices that they could before (PROB-1904). This change adds back the ability to change the icons for these two device types.


workingmonk (approved):

workingmonk:
@mikesip if you have a ticket number can you add it to the PR title?

PR #2570, opened 2017-11-29
SAPLAT-81 Additional fingerprint for Inovelli 2-channel smart plug

bflorian:
Supports and correctly identifies new model of the device that is not functional with the default multi-endpoint DTH

### DIFF
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
@@ -28,7 +28,12 @@ metadata {
 		capability "Health Check"
 
 		fingerprint manufacturer: "015D", prod: "0221", model: "251C", deviceJoinName: "Show Home 2-Channel Smart Plug"
+		fingerprint manufacturer: "0312", prod: "0221", model: "251C", deviceJoinName: "Inovelli 2-Channel Smart Plug"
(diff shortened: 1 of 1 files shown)


devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
```diff
@@ -29,6 +29,7 @@ metadata {
 
 		fingerprint manufacturer: "015D", prod: "0221", model: "251C", deviceJoinName: "Show Home 2-Channel Smart Plug"
 		fingerprint manufacturer: "0312", prod: "B221", model: "251C", deviceJoinName: "Inovelli 2-Channel Smart Plug"
+		fingerprint manufacturer: "015D", prod: "2500", model: "2500", deviceJoinName: "Inovelli 2-Channel Smart Plug"
```
workingmonk: the manufacturer matches "Show Home" smart plug instead of the "Inovelli" Smart plug, is that right?

bflorian: Yes, I got that fingerprint from an actual device. Both are actually manufactured by Inovelli. They just brand the one model "Show Home".

We should check that outdoor device on my desk as well before we deploy this PR

workingmonk: cool. let me know once you add that PR

PR #2597, opened 2017-12-06
DEVC-558 DTH implementation for Eaton/Cooper Z-Wave Master AL series Dimmer(RF9540-N)

ALamchaS:


greens (changes requested):
I'd prefer if we just used one DTH with multiple fingerprints for this and the device in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2612 because of their similarity.

I'm inclined to go with the other DTH because it exposes the dimmer ramp time preference.

ALamchaS:
Ok. I have two questions: 
1. How would you like us to address configuration parameters that are handled by master dimmer and are not handled by accessory dimmer? The two ways I can come up with are:
- send those commands when handling manufacturer specific report response if this is first initialization, or initialize() if manufacturer data is already set.
- send those commands anyway (even though accessory dimmer will not process it - additional comments required not to confuse anyone reading the code).
2. About handling ramp time - do you prefer one level get command after dimmer ramp time, as is implemented in "Z-Wave Dimmer Switch Generic", or multiple level checks during ramp time as presented in the Eaton accessory dimmer DTH.

greens:
1. I'd just send them anyway and make sure they're commented.

2. I'd follow the behavior in the generic DTH (like you have here). Sending 10 gets in 10 seconds is unnecessarily chatty. Good catch.

ALamchaS:
Closed as this DTH was merged with Eaton Accessory Dimmer DTH (https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2612)

PR #2601, opened 2017-12-06
ICP-2152: LIFX - Devices are pairing double

munds:
This is a fix for parent app not uninstalling when there are no child devices present.

### DIFF
--- a/smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
+++ b/smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
@@ -276,7 +276,7 @@ def countChildren() {
 }
 
 def attemptUninstall() {
-    log.debug "Will uninstall if no child devices"
+	log.debug "Will uninstall if no child devices"
 	if (!getChildDevices()) {
 		try {
             log.debug "No children, performing uninstall"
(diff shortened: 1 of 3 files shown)


smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -269,9 +269,31 @@ def uninstalled() {
...
+	log.debug "Counting child devices"
+	state.attemptUninstall = true
+	runIn(11, attemptUninstall)
+}
+
+def attemptUninstall() {
+	log.debug "Will uninstall if no child devices"
+	if (!getChildDevices()) {
+		try {
+            log.debug "No children, performing uninstall"
```
larsfinander, line 282: whitespace

smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -269,9 +269,31 @@ def uninstalled() {
 	cirrus.unregisterServiceManager()
 }
 
+def countChildren() {
+	log.debug "Counting child devices"
+	state.attemptUninstall = true
+	runIn(11, attemptUninstall)
```
workingmonk, line 275: only do this if current count is <=1

bflorian:
This used to happen automatically. Do we understand why that’s not happening anymore?

Also, is this problem unique to LIFX or is it happening to all service manager devices?

workingmonk:
if this is a platform issue (which i guess it is), this is affecting other service managers too. this should be fixed in DM and not patched in here.

munds:
Closing this PR because the root cause can be solved by a DM fix.

PR #2603, opened 2017-12-06
DVCSMP-3349 Use old reporting logic for older firmware Centralite sensors

dkirker:
Centralite devices with firmware equal to and newer than 1.15.7 have different logic for battery reporting. For the devices older than 1.15.7 we should use the old logic that was removed in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2158 and https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2196

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -171,8 +171,9 @@ private Map getBatteryResult(rawValue) {
 			def pct = batteryMap[volts]
 			result.value = pct
 		} else {
-			def minVolts = 2.4
-			def maxVolts = 2.7
+			def useOldBatt = shouldUseOldBatteryReporting()
+			def minVolts = useOldBatt ? 2.1 : 2.4
+			def maxVolts = useOldBatt ? 3.0 : 2.7
 			// Get the current battery percentage as a multiplier 0 - 1
 			def curValVolts = Integer.parseInt(device.currentState("battery")?.value ?: "100") / 100.0
 			// Find the corresponding voltage from our range
@@ -183,15 +184,16 @@ private Map getBatteryResult(rawValue) {
 			// OR we have received the same reading twice in a row
 			// OR we don't currently have a battery reading
 			// OR the value we just received is at least 2 steps off from the last reported value
-			if(state?.lastVolts == null || state?.lastVolts == volts || device.currentState("battery")?.value == null || Math.abs(curValVolts - volts) > 0.1) {
+			// OR the device's firmware is older than 1.15.7
+			if(useOldBatt || state?.lastVolts == null || state?.lastVolts == volts || device.currentState("battery")?.value == null || Math.abs(curValVolts - volts) > 0.1) {
 				def pct = (volts - minVolts) / (maxVolts - minVolts)
 				def roundedPct = Math.round(pct * 100)
 				if (roundedPct <= 0)
 					roundedPct = 1
 				result.value = Math.min(100, roundedPct)
 			} else {
-				// Don't update as we want to smooth the battery values
-				result = null
+				// Don't update as we want to smooth the battery values, but do report the last battery state for record keeping purposes
+				result.value = device.currentState("battery").value
 			}
 			state.lastVolts = volts
 		}
@@ -236,4 +238,23 @@ def configure() {
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
 	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
+}
+
+private shouldUseOldBatteryReporting() {
+	def isFwVersionLess = true // By default use the old battery reporting
+	def deviceFwVer = "${device.getFirmwareVersion()}"
+	def deviceVersion = deviceFwVer.tokenize('.')  // We expect the format ###.###.### where ### is some integer
+
+	if (deviceVersion.size() == 3) {
+		def targetVersion = [1, 15, 7] // Centralite Firmware 1.15.7 contains battery smoothing fixes, so versions before that should NOT be smoothed
+		def devMajor = deviceVersion[0] as int
+		def devMinor = deviceVersion[1] as int
+		def devBuild = deviceVersion[2] as int
+
+		isFwVersionLess = ((devMajor < targetVersion[0]) ||
+			(devMajor == targetVersion[0] && devMinor < targetVersion[1]) ||
+			(devMajor == targetVersion[0] && devMinor == targetVersion[1] && devBuild < targetVersion[2]))
+	}
+
+	return isFwVersionLess // If f/w version is less than 1.15.7 then do NOT smooth battery reports and use the old reporting
 }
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -183,7 +184,8 @@ private Map getBatteryResult(rawValue) {
 			// OR we have received the same reading twice in a row
 			// OR we don't currently have a battery reading
 			// OR the value we just received is at least 2 steps off from the last reported value
-			if(state?.lastVolts == null || state?.lastVolts == volts || device.currentState("battery")?.value == null || Math.abs(curValVolts - volts) > 0.1) {
+			// OR the device's firmware is older than 1.15.7
+			if(state?.lastVolts == null || state?.lastVolts == volts || device.currentState("battery")?.value == null || Math.abs(curValVolts - volts) > 0.1 || isFirmwareVerLessThan(1, 15, 7)) {
```
tpmanley: Rather than setting `result = null` on line 196 I think we should generate the event with the current value (`device.currentState("battery").value`). This way we'll know that we're still getting battery reports even if the value isn't changing.

dkirker: Yeah, I can add that @tpmanley . I believe if the value is the same as the last event value the platform won't create/propagate it, so I will need to add result.isStateChange = true to make sure the event isn't dropped. It seems this has been affecting device health for some devices, as well.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -148,6 +148,7 @@ private Map translateZoneStatus(ZoneStatus zs) {
 private Map getBatteryResult(rawValue) {
 	log.debug "Battery rawValue = ${rawValue}"
 	def linkText = getLinkText(device)
+	//log.debug "${device.getFirmwareVersion()} isFirmwareVerLessThan(1, 15, 7) = ${isFirmwareVerLessThan(1, 15, 7)}"
```
tpmanley: The commented out log messages should be removed before merging

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -171,8 +172,8 @@ private Map getBatteryResult(rawValue) {
 			def pct = batteryMap[volts]
 			result.value = pct
 		} else {
-			def minVolts = 2.4
-			def maxVolts = 2.7
+			def minVolts = (isFirmwareVerLessThan(1, 15, 7)) ? 2.1 : 2.4
```
tpmanley: Rather than calling `isFirmwareVerLessThan` multiple times you could call it once and save to a variable.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -237,3 +239,16 @@ def configure() {
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
 	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
 }
+
+private isFirmwareVerLessThan(major, minor, build) {
+	def deviceFwVer = "${device.getFirmwareVersion()}"
```
tpmanley: I wish there was a better interface to get at the firmware version but it looks like this is the best we can do for now because we can't call functions or access members on the object returned by `getFirmwareVersion()`

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -237,3 +239,16 @@ def configure() {
...
+	def tempTok = deviceFwVer.tokenize('.') // We expect the format ###.###.### where ### is some integer
+	def devMajor = tempTok[0] as int
+	def devMinor = tempTok[1] as int
+	def devBuild = tempTok[2] as int
+	//def (devMajor as int, devMinor as int, devBuild as int) = deviceFwVer.tokenize('.') // Would be cleaner, but we don't seem to support multi-assignments
```
tpmanley: This should also be removed before merging

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
tpmanley: I suggest putting the firmware version check first since that will be the most common case

tpmanley (approved):
Couple of minor things but overall looks good

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: Ha! In my effort to be clean and resourceful, I missed one of the obvious actions! Good catch! Thanks

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: I agree. I doubt the DeviceFirmware toString will change, but this should be "ok" for now.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: Will do. If you know why multiple assignment throws an exception for me (if it is supposed to be supported) then I'll use that.

dkirker:
@tpmanley  Cool! Made the updates.

varzac (approved):

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -237,3 +239,15 @@ def configure() {
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
 	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
 }
+
+private isFirmwareVersionLessThan(major, minor, build) {
+	def deviceFwVer = "${device.getFirmwareVersion()}"
```
varzac: Actually just realized that it is possible for this to be `null`.  There can be times when we don't know about the devices firmware.  I think what we want to do here is if the value is `null` return `true` but then the naming of the function is kind of weird.  We could rename it something like `useOldBatteryReporting` instead

varzac (changes requested):
See my line comment

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: Ah right! Thanks for catching that null case. I had included that when I was designing it in my head.

As for the function naming.... hmmm...

posborne (approved):
Beyond the comments from Zach and Tom I have nothing additional to add.

dkirker:
@tpmanley  @varzac @posborne I think the revised changes should be satisfactory. @varzac Let me know what you think!

varzac (approved):

PR #2604, opened 2017-12-06
DVCSMP-2874 Aeon multisensor 6 missing method exception

greens:
The `secure` method was never defined in this DTH and also the device is only conditionally a listening device, so I've reflected that in the `poll` method.


dkirker (commented):
@jackchi Does the fixed ping() command look ok? If it is running off battery there is nothing we can do until it sends a wakeup notification.

However, one thought:
Would device health accept this as an "it is online" event?

createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)

If yes (would it even get it because isStateChange is false?), then I think this looks good. If no, then we should probably add a state variable that is a flag indicating that we should request battery state in the next wakeup.

@jackchi @mckeed Thoughts?

mckeed:
Wakeup events do work for device health, but this device sends multisensor events every 8 minutes, so those are what we're relying on to keep the device online.

It doesn't make a lot of sense to send a batteryGet as the ping when the batteries aren't being used. Maybe requesting the temperature would be better, since we shouldn't get to the point of pinging the device unless the multisensor events aren't getting received.

greens:
That's a good point. I'll make some changes to use temperature instead.

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -142,10 +142,15 @@ def installed(){
...
 def updated() {
 // Device-Watch simply pings if no device events received for 122min(checkInterval)
-	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	log.debug "Updated with settings: ${settings}"
 	log.debug "${device.displayName} is now ${device.latestValue("powerSource")}"
 
+	checkInterval = 2 * 60 * 60 + 2 * 60
+	if (reportInterval) {
+		checkInterval = 2*(timeOptionValueMap[reportInterval]/60) + (2 * 60)
+	}
```
greens: check my logic here, esp. re: accessing preference values.

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -142,10 +142,15 @@ def installed(){
 
 def updated() {
 // Device-Watch simply pings if no device events received for 122min(checkInterval)
-	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	log.debug "Updated with settings: ${settings}"
 	log.debug "${device.displayName} is now ${device.latestValue("powerSource")}"
 
+	checkInterval = 2 * 60 * 60 + 2 * 60
```
mckeed: missing `def`

mckeed:
I tried this out and ran into an issue: when the device is battery powered the check interval is changed right away but the reporting interval isn't changed until the next wakeup, which makes the device go offline when switching to a shorter interval

greens:
Ahh, okay. Missed that. So we should update the checkInterval inside `configure`.

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -399,6 +402,14 @@ def configure() {
 
 	setConfigured("true")
 
+	// set the check interval based on the report interval preference. (default 122 minutes)
+	// we do this here in case the device is in wakeup mode
+	def checkInterval = 2 * 60 * 60 + 2 * 60
+	if (reportInterval) {
+		checkInterval = 2*(timeOptionValueMap[reportInterval]/60) + (2 * 60)
```
tpmanley: The value for `"18 hours"` in `getTimeOptionValueMap` is actually only 6 hours

greens: That's a good catch. Looks like it's wrong in the aeon multisensor (original) code too. I'll go ahead and fix both under this commit.

mckeed: Math is wrong here. You're dividing the seconds by 60 but checkInterval is in seconds too.

mckeed (changes requested):

greens:
Can I get a second look at this?

greens:
@mckeed Incorporated your suggestions.

dkirker (approved):

mckeed (approved):

PR #2605, opened 2017-12-07
[ICP-2366] [ICP-2367] Yale Keyless Connected Smart Door Lock - Named

vaish-rahul:
code is not recognizable.
Yale Keyless Connected Smart Door Lock - no auto locked/auto unlocked
event in IDE.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -371,10 +371,16 @@ private def handleAccessAlarmReport(cmd) {
 	switch(cmd.zwaveAlarmEvent) {
 		case 1: // Manually locked
 			map.descriptionText = "Locked manually"
+			if(generatesDoorLockOperationReportBeforeAlarmReport()) {
+				map.isStateChange = true
+			}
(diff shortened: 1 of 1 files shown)


greens (commented):
What does this look like when tested with a lock that sends these event in the "expected" order? Do you then see two events?

greens (commented):
Does this still create two events in the case of these specific Yale Locks?

vaish-rahul:
Yes two events will be created in case of this specific Yale Locks. When a lock is sending report in expected order, which is AlarmReport followed by DoorLockOperationReport, then two events will not be generated because flag is turned on while handling AlarmReport only. While handling DoorLockOperationReport the system will set the flag to false automatically since there is no change in lock's state. Only the case where lock is sending reports in unexpected order, two events will be generated. What we have done now is that we have restricted the changes to specific model which is generating reports in unexpected order.

greens:
I'd like you to try implementing the solution Duncan mentioned in Slack:

>I think a good solution would be in the lock() command, send just the SET and schedule a follow-up method 5 or 6 seconds later. In that method, if we didn't get the notification we can send a GET, otherwise do nothing because we already have the event

We really want to avoid multiple events.

vaish-rahul:
Closing as @greens is working on a better fix to this problem.

PR #2524, opened 2017-11-17
DEVC-556 DTH implementation for Eaton/Cooper Accessory switch (RF9517)

ALamchaS:

PR #2606, opened 2017-12-08
DTH implementation for Eaton/Cooper Z-Wave Master Switch (RF9501)

MAblewicz:
DTH implementation for Eaton/Cooper Z-Wave Master Switch (RF9501):
http://www.cooperindustries.com/content/public/en/wiring_devices/products/lighting_controls/aspire_rf_wireless/switches/aspire_rf_15a_wireless_switch_rf9501.html

### DIFF
--- a/devicetypes/smartthings/eaton-master-switch.src/README.md
+++ b/devicetypes/smartthings/eaton-master-switch.src/README.md
@@ -1,0 +1,45 @@
+# Eaton Master Switch
+
+Cloud Execution
+
+Works with:
+
+* [Eaton Master Switch](http://www.cooperindustries.com/content/public/en/wiring_devices/products/lighting_controls/aspire_rf_wireless/switches/aspire_rf_15a_wireless_switch_rf9501.html)
+
+## Table of contents
+
+* [Capabilities](#capabilities)
+* [Installation](#installation)
+* [Supported Functionality](#supported-functionality)
+* [Deinstallation](#deinstallation)
+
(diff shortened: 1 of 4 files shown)


greens (dismissed):

ALamchaS:
In light of your comment in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2597 would you like us to merge this DTH with https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2524 ?

greens:
It looks like that DTH doesn't support protection mode? If there's specific functionality (that we're exposing in a tile), it makes using a single DTH difficult.

ALamchaS:
Accessory switch device does handle protection mode too. It was my first DTH written when Ingvar was teaching us, and I misunderstood what protection command really does, so I did not implement it.

Only real difference would be Relay Switch capability. Master switch has built-in relay switch, and accessory switch does not. We would have to remove this capability. Functionality will remain the same.

ALamchaS:
To speed this process up a bit: this is how resulting DTH will look like. I tested it with both Accessory and Master switch. Both handle Protection (Eaton/Cooper calls it Child Lockout feature) commands properly.

If you decide Relay Switch capability is important, I will revert those changes.

greens (approved):
Looks good.

dkirker (approved):

MAblewicz:
Device support added to generic Z-Wave switch DTH

PR #2626, opened 2017-12-13
Dvcsmp-1437 GE Fan switches don't support configuration

greens:
Upon investigation, these GE fan switches should never have used the GE-specific DTH because one model doesn't even support the configuration command class and the new model I added does, but not for indicator control.


mckeed (approved):
At some point it would be nice to get a specific DTH for these devices

greens:
Strong agree. 

We could easily do something like a toggle button that iterates through off/low/medium/high, but if we wanted any sort of actual stepped level controls it'd take some doing on the UI side I think?

PR #2627, opened 2017-12-13
ICP-2367 Yale Locks send lock operations before notification events

greens:
We need to enforce the order of notification and lock in order to get the most information possible for the lock/unlock events. This attempts to delay the lock command event until after the notification event. Because of isStateChange, the lock event will actually essentially get ignored as long as the notification event comes.


devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -313,6 +313,17 @@ def zwaveEvent(DoorLockOperationReport cmd) {
 		}
 	}
 	result ? [createEvent(map), *result] : createEvent(map)
+	if (generatesDoorLockOperationReportBeforeAlarmReport()) {
+		// we're expecting lock events to come after notification events, but for specific yale locks they come out of order
+		runIn(3, "delayLockEvent", [data: [event: result]])
```
vaish-rahul: This won't work. I see a problem.

`result` will not have any data for locked/unlocked event as you are not putting that into the `result` list. So the changes will not work when expected to work. Besides this, the changes will also break the existing functionality of `refresh` in DTH.

vaish-rahul (changes requested):

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
greens: Ah thanks, didn't notice there wasn't actually an assignment on the line above.

greens:
@vaish-rahul This should fix it. Thanks for catching that. Refresh should still work, there will just be a bit more of a delay for Yale locks.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -312,7 +312,18 @@ def zwaveEvent(DoorLockOperationReport cmd) {
 			result << response(secure(zwave.associationV1.associationGet(groupingIdentifier:1)))
 		}
 	}
-	result ? [createEvent(map), *result] : createEvent(map)
+	result = result ? [createEvent(map), *result] : createEvent(map)
```
mckeed: `createEvent` sets `isStateChange`. It would be better to move this line to the `else` block below and just send `map` to `delayLockEvent`.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -312,18 +312,17 @@ def zwaveEvent(DoorLockOperationReport cmd) {
...
         return [:]
 	} else {
-		return result
+		return result ? [createEvent(map), *result] : createEvent(map)
 	}
 }
 
 def delayLockEvent(data) {
-	sendEvent(data.event)
+	sendEvent(data.result ? [createEvent(data.map), *data.result] : createEvent(data.map))
```
vaish-rahul: Now theoretically this looks correct. But today I tried this out by self publishing the DTH and `delayLockEvent` doesn't get called. I made `generatesDoorLockOperationReportBeforeAlarmReport` return true explicitly and `delayLockEvent` gets scheduled to be called in 3 seconds but it doesn't get called and as a result event doesn't get generated. May be I am missing something or something is broken in production server because of which `delayLockEvent` is not getting called. Either way, @greens can you try executing this in your account by self publishing and check whether `delayLockEvent` gets called?

vaish-rahul (changes requested):

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
vaish-rahul: Good catch @mckeed

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
greens: I tested on my account with logging set up with the correct device and delayLockEvent was being called for me. 

Try testing it in the IDE with some additional logging statements. Because the event _shouldn't_ show up.

vaish-rahul: I added a log in `delayLockEvent` and it didn't print. Anyway, if it's working for you then it's Ok.

vaish-rahul (approved):

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -314,15 +314,16 @@ def zwaveEvent(DoorLockOperationReport cmd) {
...
         return [:]
 	} else {
 		return result ? [createEvent(map), *result] : createEvent(map)
 	}
 }
 
 def delayLockEvent(data) {
-	sendEvent(data.result ? [createEvent(data.map), *data.result] : createEvent(data.map))
+	log.debug "Sending cached lock operation: $data.map"
+	sendEvent(data.map)
```
vaish-rahul: Cool. I missed to check that `sendEvent` takes a Map and not a list. However, with this we will miss to send the `HubActions` added to `result` list in `zwaveEvent(DoorLockOperationReport cmd)`. Are we sure that this will not break anything?

mckeed: Those hubActions are primarily for fixing an association issue that came up with first-generation Schlage locks. I expect it won't be a problem to skip them for recent Yale models.

mckeed (dismissed):

mckeed (approved):

PR #2633, opened 2017-12-16
PROB-1940 Ecobee Thermostat Automation Setpoint off by 1 Degree

marstorp:
When an automation sends a setpoint to an ecobee thermostat that matches
the current setpoint, the setpoint changes by 1 degree.

This issue was caused by the need of forcing the new setpoint +-1° when rounding
issues between °F and °C caused the setpoint to remain the same when
changing it manually one step at a time.
Added fix to check if setpoints are set manually or by automation before applying
the +-1° adjustment.

Also added possibility to change the deadband used in device settings.


rohandesai (approved):
LGTM :shipit:

workingmonk (approved):

PR #2643, opened 2017-12-22
WWST-159 WWST-155 WWST-164 WWST-157 WWST-166 WWST-158 WWST-156 WWST-165 WWST-160 WWST-167 Move Honeywell Switch fingerprints

greens:
…65 WWST-160 WWST-167

Honeywell binary switches were put in the wrong DTH. Should go in the generic one instead of the one meant for GE devices.


dkirker (approved):

dkirker:
@greens Looks good. Forced the last two ticket numbers into the title. Not sure if it makes a difference for The Vinay's scripts, but it is there. :P

PR #2658, opened 2018-01-01
Fix for handler sending incorrect basic set values

erocm123:
According to Z-Wave documentation Basic Set and MultiLevel set accept values with the range of 0,1..99 (0x01..0x63), & 255. The slider in this handler can be set to 100 and there was no input checking in the setLevel method. This is causing the handler to send an invalid value of 100 to the device and several devices are not responding to the command.

This change is code taken from the SmartThings "Dimmer Switch" device handler. It merely checks the setLevel input value to make sure it is a valid value. If it is less than 0 it sends 0 if it is greater than 99 it sends 99.


mckeed (approved):
Thanks!

greens (approved):

PR #2665, opened 2018-01-03
[CHAD-1649] Temporary Fix to clear tamper

adityata:
* Tamper is cleared after 10s of it being detected
  This is because tamper is not handled locally

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
@@ -157,7 +157,7 @@ def updated() {
 			sendEvent(name: "powerSource", value: powerSource, displayed: false)
 		}
 	}
-	
+
 	if (powerSource == "battery") {
 		setConfigured("false") //wait until the next time device wakeup to send configure command after user change preference
 	} else { // We haven't identified the power supply, or the power supply is USB, so configure
@@ -299,6 +299,10 @@ def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
 	motionEvent(cmd.value)
 }
 
+def clearTamper() {
+	sendEvent(name: "tamper", value: "clear")
+}
+
 def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cmd) {
 	def result = []
 	if (cmd.notificationType == 7) {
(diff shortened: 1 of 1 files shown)


adityata:
@tpmanley

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -309,6 +313,8 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 				break
 			case 3:
 				result << createEvent(name: "tamper", value: "detected", descriptionText: "$device.displayName was tampered")
+				// Clear the tamper alert after 10s. This is a temporary fix for the tamper attribute until local execution handles it
+				runIn(10, clearTamper)
```
tpmanley: Should we also remove the clear in `case 0`?

adityata: No, the `case 0` is the proper way for the device to clear reports. Also, there would be no use of any changes as that part of code will not run (as `case 0` would have been handled locally)

kianooshkarami:
Just so I understand the reasoning behind the change, Would you please document the reasoning behind it?

adityata:
* Handling Tamper Attribute locally is not fully possible as we need the extra bit of information as to why the tamper happened (i.e. `descriptionText`). We do not yet have the functionality in Local Execution to set the `descriptionText` for events raised from hub.
  https://github.com/PhysicalGraph/hub-core/pull/972: Tried to address the tamper handling locally, but the `descrptionText` support is not still available. Hence this local change was put on hold.
* By default if `case 3` is not handled locally, it is handled in cloud, thus `tamper=detected` event with `descriptionText` is raised.
  But since `case 0` is handled locally, we will not be able send `tamper=clear` (if tamper alert is not handled locally) and this event will not raised.
* To provide a temoprary fix to this problem, we can send a `tamper=clear` event after some set time (say 10s).

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
mckeed: We should `unschedule(clearTamper)` before this, in case there are multiple alerts in the time frame.

mckeed (dismissed):

varzac (dismissed):

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
adityata: `groovy.lang.MissingMethodException: No signature of method: java.util.ArrayList.unschedule() is applicable for argument types: (java.lang.String) values: [clearTamper]`

Getting this exception

adityata:
Can someone please review and merge this? I don't have write access.

varzac (approved):

PR #2671, opened 2018-01-04
PROB-1869 Zen thermostat connecting with no battery state

marstorp:
Changed algorithm for calculating battery % so it no longer need the check
for upper/lower limit of the received voltage.


greens (approved):

workingmonk:
cc @tpmanley for battery related changes

tpmanley (approved):

marstorp:
Added a fix for "java.lang.NullPointerException: Cannot invoke method minus() on null object"

greens (approved):

PR #2682, opened 2018-01-08
ICP-2750 - State after cutting off power

juano2310:


larsfinander (approved):

devicetypes/smartthings/lifx-white-bulb.src/lifx-white-bulb.groovy
```diff
@@ -110,7 +110,8 @@ def setColorTemperature(kelvin) {
 def on() {
 	log.debug "Device setOn"
 	parent.logErrors() {
-		if (parent.apiPUT("/lights/${selector()}/state", [power: "on"]) != null) {
+    	def value = parent.apiPUT("/lights/${selector()}/state", [power: "on"])
+		if (value.data.results.status[0] == "ok") {
```
bflorian: What happens if the PUT request fails with a non 2XX error, like a 401 or 404? Will the `value.data.results.status[0]` path still be valid? Maybe we should check for a 2XX status code before trying to index into a data structure that might not be returned in all cases?

bflorian (approved):

PR #2689, opened 2018-01-10
[LSD-204] Adding fields to SmartApp definitions

gausnes:
\cc @rappleg @workingmonk @SpencerH @ajguyot @aaron-miller @Dianoga

### DIFF
--- a/smartapps/com-andrewreitz/jenkins-notifier.src/jenkins-notifier.groovy
+++ b/smartapps/com-andrewreitz/jenkins-notifier.src/jenkins-notifier.groovy
@@ -30,7 +30,8 @@ definition(
     description: "Turn off and on devices based on the state that your Jenkins Build is in.",
     category: "Fun & Social",
     iconUrl: "http://i.imgur.com/tyIp8wQ.jpg",
-    iconX2Url: "http://i.imgur.com/tyIp8wQ.jpg"
+    iconX2Url: "http://i.imgur.com/tyIp8wQ.jpg",
+    pausable: true
 )
 
 preferences {
--- a/smartapps/com-obycode/obything-music-connect.src/obything-music-connect.groovy
+++ b/smartapps/com-obycode/obything-music-connect.src/obything-music-connect.groovy
@@ -21,7 +21,10 @@ definition(
     category: "SmartThings Labs",
     iconUrl: "http://obycode.com/obything/ObyThingSTLogo.png",
     iconX2Url: "http://obycode.com/obything/ObyThingSTLogo@2x.png",
-    singleInstance: true)
+    singleInstance: true,
+    usesThirdPartyAuthentication: true,
+    pausable: false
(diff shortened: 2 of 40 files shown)


aaron-miller (approved):

workingmonk:
cc @bflorian @rajkaramchedu

rajkaramchedu:
@gausnes Didn't want to start a new email/Slack thread, so asking here. For docs purposes, a couple of questions:
- What are the definitions for `pausable` and `usesThirdPartyAuthentication`? 
- These new fields are limited only to Groovy SmartApps? (and not the new endpoint apps?)
- Any contextual info you, or anyone tagged in this PR, can provide that helps for documenting these new fields? Thanks.

gausnes:
>What are the definitions for pausable and usesThirdPartyAuthentication?
 * `pausable` - indicates that the respective SmartApp is able to be paused. The paused state stops the SmartApp from executing for Subscriptions and Schedules.
    * The default behavior is to allow the SmartApp to be paused.
 * `usesThirdPartyAuthentication` - This flag will be consumed by OneApp to display SmartApp that exist to handle integrations with third parties differently than the normal SmartApp.
    * The default behavior is to have this flag be false and appear as a normal SmartApp.
    * SmartApp marked with this flag will appear in a section of the App settings instead of the list of SmartApps
    * I think we may want to change the name of this flag or add another flavor that makes more sense. Perhaps `displayAsThirdParty`

>These new fields are limited only to Groovy SmartApps? (and not the new endpoint apps?)

I believe similar concepts are going to applied to Endpoint Apps, however not for the initial launch of OneApp.

>Any contextual info you, or anyone tagged in this PR, can provide that helps for documenting these new fields? Thanks.

Currently you need elevated privileges to set these fields, I would like to change this in the near future to allow Community Developed apps to use these flags as well. Something to note for the time being though.

rajkaramchedu:
@jimmyjames

rajkaramchedu:
@gausnes Thanks! It sounds like `usesThirdPartyAuthentication` is an umbrella term that consolidates terms like "service manager SmartApp," and "SmartThings Connector" (for C2C integration). Is that the intention/target?

gausnes:
@rajkaramchedu more or less. That's how I'm currently viewing this flag however that flag is mostly concerned with mobile UI so classification might get kind of weird.

E.g. we currently have a SmartApp that can utilize an OAuth client for possible notifications that we aren't classifying as `usesThirdPartyAuthentication` because the primary use case of the SmartApp is to function as an automation.

smartapps/dianoga/co2-vent/src/co2-vent.groovy
```diff
@@ -0,0 +1,69 @@
+/**
```
workingmonk: this is a new file. is the intent to make it part of this PR?

PR #2697, opened 2018-01-15
[DVCSMP-3413] Implemented Lock Code Scheduling functionality.

vaish-rahul:
info so that the Zigbee Locks get adapted to Smart Locks Version 2
features.

### DIFF
--- a/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
+++ b/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
@@ -13,6 +13,8 @@
  *  for the specific language governing permissions and limitations under the License.
  *
  */
+import java.util.Map
+
 import physicalgraph.zigbee.zcl.DataType
 
 metadata {
@@ -26,7 +28,7 @@ metadata {
 		capability "Battery"
 		capability "Configuration"
 		capability "Health Check"
-
+		
 		fingerprint profileId: "0104", inClusters: "0000,0001,0009,000A,0101,0020", outClusters: "000A,0019", manufacturer: "Yale", model: "YRD220/240 TSDB", deviceJoinName: "Yale Touch Screen Deadbolt Lock"
 		fingerprint profileId: "0104", inClusters: "0000,0001,0003,0009,000A,0101,0020", outClusters: "000A,0019", manufacturer: "Yale", model: "YRL220 TS LL", deviceJoinName: "Yale Touch Screen Lever Lock"
 		fingerprint profileId: "0104", inClusters: "0000,0001,0003,0009,000A,0101,0020", outClusters: "000A,0019", manufacturer: "Yale", model: "YRD210 PB DB", deviceJoinName: "Yale Push Button Deadbolt Lock"
@@ -481,6 +483,7 @@ private def parseAttributeResponse(String description) {
 		} else if (value == 2) {
 			responseMap.value = "unlocked"
 			responseMap.descriptionText = "Unlocked"
+			responseMap.data = [ notify: true, notificationText: "$deviceName at ${location.name} was unlocked" ]
 		} else {
 			responseMap.value = "unknown"
 			responseMap.descriptionText = "Unknown state"
@@ -575,10 +578,17 @@ private def parseCommandResponse(String description) {
 				break
 			case 2:
 				responseMap.value = "unlocked"
+				if (responseMap.data) {
+					responseMap.data.notify = true
+				} else {
+					responseMap.data = [notify: true]
+				}
 				if(codeName) {
 					responseMap.descriptionText = "Unlocked by \"$codeName\""
+					responseMap.data.notificationText = "$deviceName at ${location.name} was unlocked by \"$codeName\""
 				} else {
 					responseMap.descriptionText = "Unlocked ${desc}"
+					responseMap.data.notificationText = "$deviceName at ${location.name} was unlocked ${desc}"
 				}
 				break
 			case 3: //Lock Failure Invalid Pin
@@ -598,7 +608,13 @@ private def parseCommandResponse(String description) {
 			case 9: // unlocked using the key
 			case 14: // unlocked using the Thumbturn
 				responseMap.value = "unlocked"
+				if (responseMap.data) {
+					responseMap.data.notify = true
+				} else {
+					responseMap.data = [notify: true]
+				}
 				responseMap.descriptionText = "Unlocked ${desc}"
+				responseMap.data.notificationText = "$deviceName at ${location.name} was unlocked ${desc}"
 				break
 			case 10: //Auto lock
 				responseMap.value = "locked"
@@ -657,6 +673,7 @@ private def parseCommandResponse(String description) {
 					responseMap.descriptionText = "Deleted all user codes"
 					responseMap.data = [notify: true, notificationText: "Deleted all user codes in $deviceName at ${location.name}"]
 					result << createEvent(name: "lockCodes", value: util.toJson([:]), displayed: false, descriptionText: "'lockCodes' attribute updated")
+					result << createEvent(name: "lockCodesData", value: util.toJson([:]), displayed: false, descriptionText: "'lockCodesData' attribute updated")
 				} else {
 					if (lockCodes[codeID.toString()]) {
 						codeName = getCodeName(lockCodes, codeID)
@@ -664,6 +681,7 @@ private def parseCommandResponse(String description) {
 						responseMap.descriptionText = "Deleted \"$codeName\""
 						responseMap.data = [ codeName: codeName, notify: true, notificationText: "Deleted \"$codeName\" in $deviceName at ${location.name}" ]
 						result << codeDeletedEvent(lockCodes, codeID)
+						result << clearUserInfo(codeID)
 					}
 				}
 				break
@@ -735,6 +753,8 @@ private def parseCommandResponse(String description) {
 				responseMap.data = [ codeName: codeName ]
 				if ("set" == changeType) {
 					result << codeSetEvent(lockCodes, codeID, codeName)
+					// setting user info with always schedule type
+					setUserInfo([codeID: "" + codeID, phoneNum: "", accessType: "always", lastUpdated: new Date().getTime()])
 				} else {
 					responseMap.displayed = false
 				}
@@ -760,6 +780,7 @@ private def parseCommandResponse(String description) {
 				responseMap.descriptionText = "Deleted \"$codeName\""
 				responseMap.data = [ codeName: codeName, notify: true, notificationText: "Deleted \"$codeName\" in $deviceName at ${location.name}" ]
 				result << codeDeletedEvent(lockCodes, codeID)
+				result << clearUserInfo(codeID)
 			} else {
 				// Code slot is empty - can happen when a slot is found empty while scanning the lock
 				responseMap.value = "$codeID unset"
@@ -1175,4 +1196,73 @@ private boolean isMasterCode(codeID) {
 		codeID = codeID.toInteger()
 	}
 	(codeID == 0) ? true : false
+}
+
+/**
+ * API end-point for updating the 'lockCodesData' attribute with the user info
+ *
+ * @param userInfo: Map consisting values for code id, phone number, access type, and last usage
+ */
+def setUserInfo(userInfo) {
+	log.trace "[DTH] Executing 'setUserInfo()' by ${device.displayName} with userInfo := ${userInfo}"
+	def codeID = userInfo.codeID
+	if (codeID) {
+		codeID = codeID.toString()
+		def lockCodesData = loadlockCodesData(codeID)
+		lockCodesData[codeID].lastUpdated = userInfo.lastUpdated
+		lockCodesData[codeID].userInfo.phoneNum = userInfo.phoneNum ?: ""
+		def accessType = userInfo.accessType
+		if (accessType) {
+			if (accessType == "always") {
+				lockCodesData[codeID].scheduleInfo = [:]
+				lockCodesData[codeID].scheduleInfo.scheduleData = [:]
+			}
+			lockCodesData[codeID].scheduleInfo.scheduleType = accessType
+		}
+		sendEvent(lockCodesDataEvent(lockCodesData))
+	}
+}
+
+/**
+ * Removes the specified code if from lockCodesData and updates the attribute
+ *
+ * @param codeID: The code slot number
+ */
+def clearUserInfo(codeID) {
+	log.trace "[DTH] Executing 'clearUserInfo()' by ${device.displayName} with codeID := $codeID"
+	def lockCodesData = parseJson(device.currentValue("lockCodesData") ?: "{}") ?: [:]
+	codeID = codeID.toString()
+	lockCodesData.remove(codeID)
+	lockCodesDataEvent(lockCodesData)
+}
+
+/**
+ * Reads the 'lockCodesData' attribute and parses the same
+ *
+ * @returns Map: The lockCodesData map
+ */
+private Map loadlockCodesData(codeID) {
+	def lockCodesData = parseJson(device.currentValue("lockCodesData") ?: "{}") ?: [:]
+	if (!lockCodesData[codeID]) {
+		lockCodesData[codeID] = [:]
+	}
+	if (!lockCodesData[codeID].userInfo) {
+		lockCodesData[codeID].userInfo = [:]
+	}
+	if (!lockCodesData[codeID].scheduleInfo) {
+		lockCodesData[codeID].scheduleInfo = [:]
+	}
+	if (!lockCodesData[codeID].scheduleInfo.scheduleData) {
+		lockCodesData[codeID].scheduleInfo.scheduleData = [:]
+	}
+	lockCodesData
+}
+
+/**
+ * Populates the 'lockCodesData' attribute by calling create event
+ *
+ * @param lockCodesData The week day schedule info in a lock
+ */
+private Map lockCodesDataEvent(lockCodesData) {
+	createEvent(name: "lockCodesData", value: util.toJson(lockCodesData), displayed: false, descriptionText: "'lockCodesData' attribute updated")
 }
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -24,7 +24,7 @@ metadata {
 		capability "Battery"
 		capability "Health Check"
 		capability "Configuration"
-
+		
 		// Generic
 		fingerprint deviceId: "0x4003", inClusters: "0x98"
 		fingerprint deviceId: "0x4004", inClusters: "0x98"
@@ -92,8 +92,19 @@ metadata {
 	}
 }
 
+import java.text.SimpleDateFormat
+
 import physicalgraph.zwave.commands.doorlockv1.*
 import physicalgraph.zwave.commands.usercodev1.*
+
+private def getSCHEDULE_TYPE_DAILY_REPEATING() { "dailyRepeating" }
+private def getSCHEDULE_TYPE_WEEK_DAY() { "weekDay" }
+private def getSCHEDULE_TYPE_YEAR_DAY() { "yearDay" }
+private def getSCHEDULE_ENTRY_LOCK_CLASS_ID() { "4E" }
+private def getUSER_TYPE_OWNER() { 1 }
+private def getUSER_TYPE_WEEK_DAY() { 4 }
+private def getUSER_TYPE_YEAR_DAY() { 3 }
+private def getMAX_CODES_TO_SCAN() { 8 }
 
 /**
  * Called on app installed
@@ -151,7 +162,16 @@ def updated() {
  */
 def configure() {
 	log.trace "[DTH] Executing 'configure()' for device ${device.displayName}"
-	def cmds = doConfigure()
+	def cmds = []
+	if (!state.configured) {
+		// configureScheduling will be called in doConfigure so this will take care of scenario when lock is paired freshly
+		cmds = doConfigure()
+	} else {
+		// In case a lock is already paired then it needs to be configured for scheduling
+		cmds << setClock()
+		cmds << configureScheduling()
+		cmds = delayBetween(cmds, 200)
+	}
 	log.debug "Configure returning with commands := $cmds"
 	cmds
 }
@@ -169,8 +189,34 @@ def doConfigure() {
 	if (isSchlageLock()) {
 		cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
 	}
-	cmds = delayBetween(cmds, 4200)
+	cmds << setClock()
+	cmds << configureScheduling()
+	// delayBetween does the flattening
+	cmds = delayBetween(cmds, 200)
 	log.debug "Do configure returning with commands := $cmds"
+	cmds
+}
+
+/**
+ * Returns commands to be send to lock for scheduling related parameters
+ * @return - Commands to be send to lock if lock supports scheduling, empty list otherwise
+ */
+def configureScheduling() {
+	def cmds = []
+	if(device.currentValue("numberOfSlotsDailyRepeating") == null ||
+	device.currentValue("numberOfSlotsWeekDay") == null ||
+	device.currentValue("numberOfSlotsYearDay") == null) {
+		// Some schedule slots are never read before
+		if (SCHEDULE_ENTRY_LOCK_CLASS_ID in zwaveInfo.sec) {
+			log.debug "[DTH] supports ScheduleEntryLock command class"
+			cmds << secure(zwave.scheduleEntryLockV3.scheduleEntryTypeSupportedGet())
+		} else {
+			log.debug "[DTH] doesn't support ScheduleEntryLock command class"
+			sendEvent(name: "numberOfSlotsDailyRepeating", value: 0, displayed: false, descriptionText: "No. of daily repeating slots supported is 0")
+			sendEvent(name: "numberOfSlotsWeekDay", value: 0, displayed: false, descriptionText: "No. of week day slots supported is 0")
+			sendEvent(name: "numberOfSlotsYearDay", value: 0, displayed: false, descriptionText: "No. of year day slots supported is 0")
+		}
+	}
 	cmds
 }
 
@@ -214,7 +260,6 @@ def parse(String description) {
  * @param cmd: The ConfigurationReport command to be parsed
  *
  * @return The event(s) to be sent out
- *
  */
 def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
 	log.trace "[DTH] Executing 'zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd)' with cmd = $cmd"
@@ -231,8 +276,60 @@ def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport 
 			isStateChange: true, data: [lockName: deviceName, notify: true,
 				notificationText: "Deleted all user codes in $deviceName at ${location.name}"])
 			result << createEvent(name: "lockCodes", value: util.toJson([:]), displayed: false, descriptionText: "'lockCodes' attribute updated")
+			result << createEvent(name: "lockCodesData", value: util.toJson([:]), displayed: false, descriptionText: "'lockCodesData' attribute updated")
 		}
 		result << createEvent(name:"codeLength", value: length, descriptionText: "Code length is $length", displayed: false)
+		return result
+	} else if (isKwiksetLock()) {
+		def result = []
+		def codeID = cmd.parameterNumber
+		def scheduleType = getScheduleType(codeID)
+		def lockCodes = loadLockCodes()
+		def codeName = getCodeName(lockCodes, codeID)
+		def userType = cmd.scaledConfigurationValue
+		log.trace "[DTH] Executing 'ConfigurationReport()' with codeID := $codeID, scheduleType := $scheduleType, and userType := $userType"
+		if (isScheduleScanningInProgress(codeID) || scheduleType == "always" || scheduleType == "unknown") {
+			// We will land here when the schedules are being scanned
+			if (userType == USER_TYPE_OWNER) {
+				// no schedules exists for this code
+				setUserInfo([codeID: "" + codeID, phoneNum: "", accessType: "always", lastUpdated: new Date().getTime()])
+				log.trace "[DTH] ConfigurationReport() -  Schedule scanning complete for codeID := $codeID"
+				checkAndScanNextSchedule(codeID)
+			} else if (userType == USER_TYPE_YEAR_DAY) {
+				// year day schedule exists for this code
+				log.trace "[DTH] ConfigurationReport() -  Requesting year day schedule for codeID := $codeID"
+				result << response(requestYearDaySchedule(codeID))
+			} else if (userType == USER_TYPE_WEEK_DAY) {
+				// week day schedule exists for this code
+				log.trace "[DTH] ConfigurationReport() -  Requesting week day schedule for codeID := $codeID"
+				clearWeekdayScanningDataFromState(codeID)
+				result << response(requestAllWeekDaySchedule(codeID))
+			}
+		} else {
+			// We will land here when a schedule is being cleared by changing the user type
+			if (userType == USER_TYPE_OWNER) {
+				log.trace "[DTH] ConfigurationReport() -  $scheduleType deleted successfully for codeID := $codeID by changing the user type"
+				result << scheduleDeleteEvent(codeID)
+				def map = null
+				def deviceName = device.displayName
+				if (scheduleType == SCHEDULE_TYPE_YEAR_DAY) {
+					map = [name: "scheduleChanged", value: "$codeID $SCHEDULE_TYPE_YEAR_DAY deleted", descriptionText: "Cleared limited schedule for \"$codeName\"", isStateChange: true, displayed: true]
+					map.data = [lockName: deviceName, notify: true, notificationText: "Cleared limited schedule for \"$codeName\" in $deviceName at ${location.name}"]
+					result << createEvent(map)
+				} else if (scheduleType == SCHEDULE_TYPE_WEEK_DAY) {
+					map = [name: "scheduleChanged", value: "$codeID $SCHEDULE_TYPE_WEEK_DAY deleted", descriptionText: "Cleared recurring schedule for \"$codeName\"", isStateChange: true, displayed: true]
+					map.data = [lockName: deviceName, notify: true, notificationText: "Cleared recurring schedule for \"$codeName\" in $deviceName at ${location.name}"]
+					result << createEvent(map)
+				}
+			} else {
+				log.trace "[DTH] ConfigurationReport() -  $scheduleType deleted failed for codeID := $codeID by changing the user type"
+				if (scheduleType == SCHEDULE_TYPE_YEAR_DAY) {
+					result << createEvent(name: "scheduleChanged", value: "$codeID $SCHEDULE_TYPE_YEAR_DAY delete failed", descriptionText: "Limited schedule delete failed for \"$codeName\"", isStateChange: true, displayed: false)
+				} else if (scheduleType == SCHEDULE_TYPE_WEEK_DAY) {
+					result << createEvent(name: "scheduleChanged", value: "$codeID $SCHEDULE_TYPE_WEEK_DAY delete failed", descriptionText: "Recurring schedule delete failed for \"$codeName\"", isStateChange: true, displayed: false)
+				}
+			}
+		}
 		return result
 	}
 	return null
@@ -311,9 +408,13 @@ def zwaveEvent(DoorLockOperationReport cmd) {
 	} else if (cmd.doorLockMode == 0x01) {
 		map.value = "unlocked with timeout"
 		map.descriptionText = "Unlocked with timeout"
+		map.data.notify = true
+		map.data.notificationText = "${device.displayName} at ${location.name} was unlocked with timeout"
 	}  else {
 		map.value = "unlocked"
 		map.descriptionText = "Unlocked"
+		map.data.notify = true
+		map.data.notificationText = "${device.displayName} at ${location.name} was unlocked"
 		if (state.assoc != zwaveHubNodeId) {
 			result << response(secure(zwave.associationV1.associationSet(groupingIdentifier:1, nodeId:zwaveHubNodeId)))
 			result << response(zwave.associationV1.associationSet(groupingIdentifier:2, nodeId:zwaveHubNodeId))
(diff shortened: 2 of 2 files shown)


vaish-rahul:
@greens, @mckeed Please review the changes.

vaish-rahul:
This PR is dependent on following PR: https://github.com/SmartThingsCommunity/SmartThingsCapabilities/pull/248

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -575,10 +578,17 @@ private def parseCommandResponse(String description) {
 				break
 			case 2:
 				responseMap.value = "unlocked"
+				if (responseMap.data) {
```
greens, line 581: Why not just one line here?

`responseMap.data = [notify : true]`

Is there a subtlety I'm missing?

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -598,7 +608,13 @@ private def parseCommandResponse(String description) {
 			case 9: // unlocked using the key
 			case 14: // unlocked using the Thumbturn
 				responseMap.value = "unlocked"
+				if (responseMap.data) {
```
greens, line 611: same question as above

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
vaish-rahul: It's possible that responseMap.data already has some values in it by the time it reaches this line. For e.g. at line number 560 the responseMap.data can get assigned with a map and if that happens then doing `responseMap.data = [notify : true]` will overwrite the existing map with new one. To avoid this problem `responseMap.data.notify = true` is done which will just add another key-value pair to the existing map.

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
vaish-rahul: same answer as above.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1874,568 @@ def deleteCode(codeID) {
...
+}
+
+/**
+ * Returns the command(s) for deleting week day schedule
+ * 
+ * @param codeID: The code slot number
+ *
+ * @return The command(s) for deleting week day schedule
+ */
+private def getWeekDayDeleteScheduleCmd(schSettings) {
```
greens: Does this still need to be defined?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,374 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+				def startMonth = (cmd.startMonth < 10) ? "0${cmd.startMonth}" : "${cmd.startMonth}"
+				def startDay = (cmd.startDay < 10) ? "0${cmd.startDay}" : "${cmd.startDay}"
+				def startHour = (cmd.startHour < 10) ? "0${cmd.startHour}" : "${cmd.startHour}"
+				def startMinute = (cmd.startMinute < 10) ? "0${cmd.startMinute}" : "${cmd.startMinute}"
+				def stopYear = (cmd.stopYear < 10) ? "0${cmd.stopYear}" : "${cmd.stopYear}"
+				def stopMonth = (cmd.stopMonth < 10) ? "0${cmd.stopMonth}" : "${cmd.stopMonth}"
+				def stopDay = (cmd.stopDay < 10) ? "0${cmd.stopDay}" : "${cmd.stopDay}"
+				def stopHour = (cmd.stopHour < 10) ? "0${cmd.stopHour}" : "${cmd.stopHour}"
+				def stopMinute = (cmd.stopMinute < 10) ? "0${cmd.stopMinute}" : "${cmd.stopMinute}"
+				state["setYearDayStartDate$codeID"] = "20${startYear}-${startMonth}-${startDay}"
```
greens: You could save a lot of lines using:

`sprintf("%02d", variableName)`

greens: or in this case specifically:

`sprintf("20%02d-%02d-%02d", startYear, startMonth, startDay)`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1874,568 @@ def deleteCode(codeID) {
...
+			def keySet = lockCodes.keySet()
+			codesToScanSchedules = []
+			// Saving the list of lock codes for which the schedule scanning should be executed
+			// In case the user creates schedule while the scanning is in progress, those codes need not be scanned
+			for (def i = 1; i <= MAX_CODES_TO_SCAN; i++) {
+				if (i.toString() in keySet) {
+					codesToScanSchedules << i
+				}
+			}
+			state.codesToScanSchedules = codesToScanSchedules
```
greens, line 1916: In what cases would this not be equivalent to `lockCodes.keySet()`?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1874,568 @@ def deleteCode(codeID) {
...
+ * Checks if there are more schedules to be scanned
+ *
+ * @returns true if more schedules are to be scanned, else false
+ */
+private def hasSchedulesToScan() {
+	def codesToScanSchedules = state.codesToScanSchedules
+	if (codesToScanSchedules && codesToScanSchedules.size() > 0) {
+		return true
+	}
+	return false
```
greens: `return state.codesToScanSchedules && state.codesToScanSchedules.size() > 0` ?

or

`return states.codesToScanSchedules?.size() > 0`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
vaish-rahul: We scan only the first 8 slots when the lock is paired freshly. Among the first 8 slots it is possible that only few slots have lock code for e.g. only slot 1, 3, 4, 8. In this case the `codesToScanSchedules` list will contain 1, 3 ,4, 8 and for these slots schedules will be scanned. Meanwhile a user can create lock code in other available slots, either from smartapp or directly on the lock. In this case the new slots will get added to `lockCodes` map but we shouldn't be scanning those slots.

vaish-rahul:
@greens I have made changes according to your review comments. Please check.

vaish-rahul:
@greens, @jodyalbritton Any other review comments on this PR?

greens:
Lots to get through. Still working on it in-between other things.

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -1176,3 +1197,72 @@ private boolean isMasterCode(codeID) {
...
+ * @param userInfo: Map consisting values for code id, phone number, access type, and last usage
+ */
+def setUserInfo(userInfo) {
+	log.trace "[DTH] Executing 'setUserInfo()' by ${device.displayName} with userInfo := ${userInfo}"
+	def codeID = userInfo.codeID
+	if (codeID) {
+		codeID = codeID.toString()
+		def lockCodesData = loadlockCodesData(codeID)
+		lockCodesData[codeID].lastUsage = userInfo.lastUsage
+		lockCodesData[codeID].userInfo.phoneNum = userInfo.phoneNum ?: ""
```
mckeed: What is phoneNum used for? This doesn't seem like the right place to be storing phone numbers.

vaish-rahul: In DTH we store code name for e.g. "Plumber" etc., in lockCodes attribute, which is actually the guest name. This was implemented in Smart Locks V1. The reason was that if user uninstalls the smartapp and installs it again then lockCodes data shouldn't get erased and hence it was stored in DTH.

In Smart Locks V2 this is extended to storing guest's phone number also. If we shift the phone number to SmartApp then the solution will break if user uninstalls and installs the SmartApp. In this case the code name will be picked up from DTH but phone number will be missing.

devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -1176,3 +1197,72 @@ private boolean isMasterCode(codeID) {
...
+ *
+ * @param userInfo: Map consisting values for code id, phone number, access type, and last usage
+ */
+def setUserInfo(userInfo) {
+	log.trace "[DTH] Executing 'setUserInfo()' by ${device.displayName} with userInfo := ${userInfo}"
+	def codeID = userInfo.codeID
+	if (codeID) {
+		codeID = codeID.toString()
+		def lockCodesData = loadlockCodesData(codeID)
+		lockCodesData[codeID].lastUpdated = userInfo.lastUpdated
```
greens, line 1212: Is the `lastUpdated` field being used for anything?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -24,7 +24,7 @@ metadata {
 		capability "Battery"
 		capability "Health Check"
 		capability "Configuration"
```
greens, line 26: Once we decide the name of the capability, it should be added here, right?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -169,11 +189,37 @@ def doConfigure() {
 	if (isSchlageLock()) {
 		cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
 	}
-	cmds = delayBetween(cmds, 4200)
+	cmds << setClock()
+	cmds << configureScheduling()
+	// delayBetween does the flattening
+	cmds = delayBetween(cmds, 200)
```
greens, line 195: Just pointing out that we've had issues elsewhere with short delays between lock commands. I'd prefer that this value at least remained `4200`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+ *
+ * @param cmd: The ScheduleEntryTypeSupportedReport command to be parsed
+ *
+ * @return The event(s) to be sent out
+ *
+ */
+def zwaveEvent(physicalgraph.zwave.commands.scheduleentrylockv3.ScheduleEntryTypeSupportedReport cmd) {
+	log.trace "[DTH] Executing 'zwaveEvent(physicalgraph.zwave.commands.scheduleentrylockv3.ScheduleEntryTypeSupportedReport cmd)' with cmd = $cmd"
+	def result = []
+	def numberOfSlotsDailyRepeating = cmd.numberOfSlotsDailyRepeating
```
greens, line 1177: Can we do:

`def numberOfSlotsDailyRepeating = cmd.numberOfSlotsDailyRepeating ?: 0`

here?

greens: Or even just put that statement in the `descriptionText` directly?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+			}
+		} else {
+			def savedStartTime = state["setDailyRepeatingStartTime$codeID"]
+			def savedEndTime = state["setDailyRepeatingEndTime$codeID"]
+			def savedWeekDayBitmask = state["setDailyRepeatingBitMask$codeID"]
+			
+			def changeType = getScheduleChangeType(codeID, SCHEDULE_TYPE_DAILY_REPEATING)
+			
+			if (savedStartTime && savedEndTime && savedWeekDayBitmask) {
+				def startTime = getWeekDayHoursAndMinutes(savedStartTime)
```
greens, line 1248: Every time you call `getWeekDayHoursAndMinutes` you're immediately breaking it up into these separate variables. You should either change the signature to just return a list and then do:

`def (startHour, startMin) = getWeekDayHoursAndMinutes(savedStartTime)`

or just reference startTime's individual attributes every time.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+				def endTime = getWeekDayHoursAndMinutes(savedEndTime)
+				
+				def startHour = startTime.hour
+				def startMin = startTime.minute
+				def endHour = endTime.hour
+				def endMin = endTime.minute
+		
+				def durMap = getDurationInHoursAndMinutes(startHour, startMin, endHour, endMin)
```
greens, line 1256: same rationale for `getDurationInHoursAndMinutes`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+def zwaveEvent(physicalgraph.zwave.commands.scheduleentrylockv3.ScheduleEntryLockWeekDayReport cmd) {
+	log.trace "[DTH] Executing 'zwaveEvent(physicalgraph.zwave.commands.scheduleentrylockv3.ScheduleEntryLockWeekDayReport cmd)' with cmd = $cmd"
+	if (device.currentValue("numberOfSlotsWeekDay")) {
+		def result = []
+		def codeID = cmd.userIdentifier
+		def lockCodes = loadLockCodes()
+		def codeName = getCodeName(lockCodes, codeID)
+		
+		def daysOfWeek = state["setWeekDayWeekDays$codeID"]
+		if (daysOfWeek && daysOfWeek.size() > 0) {
```
greens, line 1323: `daysOfWeek?.size() > 0`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+			if (dayOfWeek in daysOfWeek) {
+				def startTime = getWeekDayHoursAndMinutes(state["setWeekDayStartTime$codeID"])
+				def startHour = startTime.hour
+				def startMin = startTime.minute
+				
+				def endTime = getWeekDayHoursAndMinutes(state["setWeekDayEndTime$codeID"])
+				def endHour = endTime.hour
+				def endMin = endTime.minute
+				
+				if (cmd.startHour == startHour && cmd.startMinute == startMin && cmd.stopHour == endHour && cmd.stopMinute == endMin) {
```
greens, line 1334: As a general comment, can this be commented a bit more? It's not entirely clear at a glance what is happening in this code.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+					}
+				} else {
+					log.trace "[DTH] ScheduleEntryLockWeekDayReport() -  Schedule set failed for codeID := $codeID"
+					result << createEvent(name: "scheduleChanged", value: "$codeID $SCHEDULE_TYPE_WEEK_DAY failed", descriptionText: "Recurring schedule set failed for \"$codeName\"", isStateChange: true, displayed: false)
+					clearWeekdayScheduleDataFromState(codeID)
+				}
+			}
+		} else {
+			def scheduleSlotId = cmd.scheduleSlotId
+			scheduleSlotId = scheduleSlotId - 1
```
greens, line 1355: just do this in one line

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1035,6 +1163,366 @@ def zwaveEvent(physicalgraph.zwave.commands.applicationstatusv1.ApplicationRejec
...
+		} else {
+			def startDate = state["setYearDayStartDate$codeID"]
+			def startTime = state["setYearDayStartTime$codeID"]
+			def endDate = state["setYearDayStopDate$codeID"]
+			def endTime = state["setYearDayStopTime$codeID"]
+			
+			def changeType = getScheduleChangeType(codeID, SCHEDULE_TYPE_YEAR_DAY)
+			
+			if (startDate && startTime && endDate && endTime) {
+				def startDateTime = getYearDayTimeDetails(startDate, startTime)
```
greens, line 1467: same rationale as `getDurationInHoursAndMinutes` changes.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+}
+
+/**
+ * Starts the scanning of schedule for the first code present in the 'lockCodes' map
+ */
+def startScanningSchedule() {
+	log.trace "[DTH] Executing 'startScanningSchedule()' by ${device.displayName}"
+	def codesToScanSchedules = state.codesToScanSchedules
+	def codeID = -1
+	if (codesToScanSchedules && codesToScanSchedules.size() > 0) {
```
greens, line 1901: `codesToScanSchedules?.size() > 0`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		// Schedule scanning did not complete the last time. Scan the remaining schedules.
+		codeID = codesToScanSchedules[0]
+	} else {
+		def lockCodes = loadLockCodes()
+		if (lockCodes && lockCodes.size() > 0) {
+			def keySet = lockCodes.keySet()
+			codesToScanSchedules = []
```
greens, line 1908: In the creation/assignment above, you could just do

`def codesToScanSchedules = state.codesToScanSchedules ?: []`

and avoid this line.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+/**
+ * Scans the schedule for the specified code slot number
+ * 
+ * @param codeID: The code slot number
+ */
+private def scanSchedule(codeID) {
+	if (isKwiksetLock()) {
+		// For Kwikset locks we can find out the user type and then issue the appropriate schedule get command
+		log.trace "[DTH] scanSchedule() -  Calling configurationGet() for codeID := $codeID"
+		sendHubCommand(new physicalgraph.device.HubAction(secure(zwave.configurationV2.configurationGet(parameterNumber: codeID))))
```
greens, line 1938: I added some code recently so that you shouldn't have to cast Zwave commands to HubActions anymore for `sendHubCommand`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+
+/**
+ * API end-point for changing the user type of a code slot to type owner. This is applicable only for Kwikset locks.
+ *
+ * @param codeID: The code slot number
+ *
+ * @returns cmds: The commands fired for changing the user type
+ */
+def changeUserTypeToOwner(codeID) {
+	if (isKwiksetLock()) {
```
greens, line 2114: `if (isKwiksetLock() && codeID)`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		
+		def durMap = getDurationInHoursAndMinutes(startHour, startMin, endHour, endMin)
+		def durMin = durMap.durMin
+		def durHour = durMap.durHour
+		
+		// For weekDayBitMask, the order of days is Sat, Fri, Thu, Wed, Tue, Mon, Sun with Sat as the most significant bit
+		short weekDayBitMask = schSettings.weekDayBitMask
+		
+		state["setDailyRepeatingStartTime$codeID"] = schSettings.startTime
+		state["setDailyRepeatingEndTime$codeID"] = schSettings.endTime
```
greens: This is the only time you're using `endTime` in this scope. It doesn't need to be split into constituent fields.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		log.trace "[DTH] Executing 'getDailyRepeatingSetScheduleCmd()' by ${device.displayName} with schSettings := $schSettings"
+		def cmds = []
+		
+		def codeID = Integer.parseInt(schSettings.codeId, 10)
+		
+		def startTime = getWeekDayHoursAndMinutes(schSettings.startTime)
+		def startHour = startTime.hour
+		def startMin = startTime.minute
+		
+		def endTime = getWeekDayHoursAndMinutes(schSettings.endTime)
```
greens, line 2143: not used in this scope

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		
+		def setCmds = []
+		def getCmds = []
+		def daysOfWeek = []
+		// dayOfWeek - Sunday = 0, Monday = 1, Tuesday = 2, Wednesday = 3, Thursday = 4, Friday = 5, Saturday = 6
+		for (def i = 0; i <= 6; i++) {
+			if (isBitSet(weekDayBitMask, i)) {
+				daysOfWeek << i
+				setCmds << secure(zwave.scheduleEntryLockV3.scheduleEntryLockWeekDaySet(startHour: startHour, startMinute: startMin,
+					setAction: 1, scheduleSlotId: (i + 1), stopHour: endHour, stopMinute: endMin, userIdentifier: codeID, dayOfWeek: i))
```
greens, line 2204: what's the relation between the `scheduleSlotId` and `dayOfWeek`?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		state["setYearDayStopDate$codeID"] = endDate
+		state["setYearDayStopTime$codeID"] = endTime
+		
+		def startDateTime = getYearDayTimeDetails(startDate, startTime)
+		def startHour = startDateTime.hour
+		def startMinute = startDateTime.minute
+		def startDay = startDateTime.day
+		def startMonth = startDateTime.month
+		def startYear = startDateTime.year.toString()
+		startYear = startYear.substring(2, startYear.length())
```
greens, line 2247: If you're just trying to get the last 2 digits of the year and `startDateTime` is already an Integer:

`def startYear = startDateTime.year % 100`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+		def endDateTime = getYearDayTimeDetails(endDate, endTime)
+		def stopHour = endDateTime.hour
+		def stopMinute = endDateTime.minute
+		def stopDay = endDateTime.day
+		def stopMonth = endDateTime.month
+		def stopYear = endDateTime.year.toString()
+		stopYear = stopYear.substring(2, stopYear.length())
+		stopYear = Integer.parseInt(stopYear, 10)
+		
+		cmds << secure(zwave.scheduleEntryLockV3.scheduleEntryLockYearDaySet(scheduleSlotId: 1, setAction: 1, startDay: startDay, startHour: startHour, startMinute: startMinute,
```
greens, line 2259: As a general comment, you have a bunch of these functions where you break out all your arguments into separate assignments and then just use them once as arguments to a function call.

I'd recommend just referencing them directly instead of having the extra assignment.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
mckeed: Unfortunately the destructuring assignment syntax `def (startHour, startMin)` is broken by the sandbox parser, or at least was last time I used it. I agree just using startTime.hour is stylistically better than startHour, though.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+ */
+private def getWeekDayDeleteScheduleCmd(codeID) {
+	def cmds = []
+	def setCmds = []
+	def getCmds = []
+	for (def i = 0; i <= 6; i++) {
+		setCmds << secure(zwave.scheduleEntryLockV3.scheduleEntryLockWeekDaySet(setAction: 0, scheduleSlotId: (i + 1), userIdentifier: codeID))
+		getCmds << requestWeekDaySchedule((i + 1), codeID)
+	}
+	if (setCmds && getCmds) {
```
greens, line 2304: Is there any way both of these won't be populated?

greens: Also doesn't the `requestAllWeekDaySchedule` function below generate `getCmds` exactly?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+ *
+ * @param weekDayBitmask: The week day bit mask
+ *
+ * @param bitNum: The bit number to check
+ *
+ * @return - true if the bit is set, false otherwise
+ */
+def isBitSet(weekDayBitmask, bitNum) {
+	def mask = 1 << bitNum
+	if ((weekDayBitmask & mask) != 0) {
```
greens, line 2393: `return (weekDayBitmask & (1 << num)) != 0`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+private Map getDurationInHoursAndMinutes(startHour, startMin, endHour, endMin) {
+	def startDate = new Date()
+	startDate.set(hourOfDay: startHour, minute: startMin, second: 0, year: 2000, month: 1, date: 1)
+
+	def endDate = new Date()
+	endDate.set(hourOfDay: endHour, minute: endMin, second: 0, year: 2000, month: 1, date: 1)
+
+	long difference = endDate.getTime() - startDate.getTime();
+	difference = difference / 1000
+	def durSeconds = difference % 60
```
greens, line 2421: Check my math, but I'm pretty sure `durSeconds` is always 0.

greens: I think the better approach here, since you're always dealing in full minutes and hours, is to compute using minutes:

`def difInMinutes = ((endHour * 60) + endMin) - ((startHour * 60) + startMin)`
`def durMin = difInMinutes % 60`
`def durHour = (difInMinutes / 60) % 24)`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+
+/**
+ * Removes the specified code if from lockCodesData and updates the attribute
+ *
+ * @param codeID: The code slot number
+ */
+def clearUserInfo(codeID) {
+	log.trace "[DTH] Executing 'clearUserInfo()' by ${device.displayName} with codeID := $codeID"
+	def lockCodesData = parseJson(device.currentValue("lockCodesData") ?: "{}") ?: [:]
+	codeID = codeID.toString()
```
greens, line 2437: as a general style comment, very simple statements like this don't need to be broken out into one line

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1365,6 +1866,579 @@ def deleteCode(codeID) {
...
+}
+
+/**
+ * Removes the specified code if from lockCodesData and updates the attribute
+ *
+ * @param codeID: The code slot number
+ */
+def clearUserInfo(codeID) {
+	log.trace "[DTH] Executing 'clearUserInfo()' by ${device.displayName} with codeID := $codeID"
+	def lockCodesData = parseJson(device.currentValue("lockCodesData") ?: "{}") ?: [:]
```
greens, line 2436: near-duplication of `loadLockCodesData()`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1526,6 +2631,54 @@ private def codeSetEvent(lockCodes, codeID, codeName) {
...
+ *
+ * @return The list of events to be sent out
+ */
+private def scheduleSetEvent(codeID, scheduleType) {
+	// codeID seems to be an int primitive type
+	codeID = codeID.toString()
+	def lockCodesData = loadlockCodesData(codeID)
+	lockCodesData[codeID].lastUpdated = new Date().getTime()
+	lockCodesData[codeID].scheduleInfo = [:]
+	lockCodesData[codeID].scheduleInfo.scheduleData = [:]
```
greens, line 2649: Don't bother clearing this if you're just going to re-assign.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1588,7 +2741,73 @@ private def getChangeType(lockCodes, codeID) {
...
+ *
+ * @param codeID The code slot number
+ *
+ * @param scheduleType The schedule type to check for
+ *
+ * @return "set" or "update" basis the presence of the code id in the lockCodesData map
+ */
+private def getScheduleChangeType(codeID, scheduleType) {
+	def changeType = "set"
+	def lockCodesData = device.currentValue("lockCodesData")
```
greens, line 2754: Some of these lines can be replaced with a `loadLockCodesData()` call.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1588,7 +2741,73 @@ private def getChangeType(lockCodes, codeID) {
...
+	if (lockCodesData) {
+		lockCodesData = parseJson(lockCodesData)
+		def data = lockCodesData[codeID.toString()]
+		if (data) {
```
greens, line 2758: this should really be

`if (data?.scheduleInfo?.scheduleData)`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1588,7 +2741,73 @@ private def getChangeType(lockCodes, codeID) {
...
+/**
+ * Checks if schedule scanning is in progress
+ *
+ * @param codeID The code slot number
+ *
+ * @return true if scanning is in progress, else false
+ */
+private def isScheduleScanningInProgress(codeID) {
+	def codesToScanSchedules = state.codesToScanSchedules
+	if (codesToScanSchedules && codeID.toInteger() in codesToScanSchedules) {
```
greens, line 2783: `return (state.codesToScanSchedules && (codeID.toInteger() in state.codesToScanSchedules))`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1588,7 +2741,73 @@ private def getChangeType(lockCodes, codeID) {
...
+/**
+ * Returns the schedule type basis the code slot number
+ *
+ * @param codeID The code slot number
+ *
+ * @return The schedule type
+ */
+private def getScheduleType(codeID) {
+	def scheduleType = "always"
+	def lockCodesData = device.currentValue("lockCodesData")
```
greens, line 2798: `loadLockCodesData()`

greens: `scheduleType = lockCodesData[codeID.toString()]?.scheduleInfo?.scheduleType ?: "always"`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1602,6 +2821,40 @@ private def getStatusForDescription(changeType) {
...
 }
 
+/**
+ * Parse the time using SimpleDateFormat
+ *
+ * @param time: The time in HH:mm format
+ *
+ * @return The map with hour and minute
+ */
+private def getWeekDayHoursAndMinutes(time) {
```
greens, line 2831: if the time string is formatted as "HH:mm", it's much simpler to just do:

`time = time.split(':')`
`return [hour: time[0] as Integer , minute: time[1] as Integer]`

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1602,6 +2821,40 @@ private def getStatusForDescription(changeType) {
...
+ * Parse the date and time using SimpleDateFormat
+ *
+ * @param date: The date in yyyy-MM-dd format
+ *
+ * @param time: The time in HH:mm format
+ *
+ * @return The map with year, month, day, hour and minute
+ */
+private def getYearDayTimeDetails(date, time) {
+	def sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm")
```
greens, line 2849: I might recommend doing the same splitting as above here as well on the date (using a hyphen), since you're just putting them into the `Calendar` and the immediately pulling them back out.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1695,3 +3006,53 @@ def readCodeSlotId(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd) {
...
+			def dstOffset = location.timeZone.dstSavings / 1000
+			def dstFlag = 0
+			if(dstOffset < 0) {
+				dstFlag = 1
+				dstOffset = dstOffset.abs()
+			}
+			def minutesDst = dstOffset  / 60
+			log.debug "Setting offset on lock hourTzo: ${hours} minuteOffsetDst: ${minutesDst} minuteTzo: ${minutes} signOffsetDst: ${dstFlag} signTzo: ${timeZoneFlag}"
+			cmds << secure(zwave.scheduleEntryLockV3.scheduleEntryLockTimeOffsetSet(hourTzo: hours, minuteOffsetDst: minutesDst, minuteTzo: minutes, signOffsetDst: dstFlag, signTzo: timeZoneFlag))
+			}
```
greens, line 3044: fix indentation

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1695,3 +3006,53 @@ def readCodeSlotId(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd) {
...
+			timeZoneOffset = timeZoneOffset.abs()
+		}
+
+		def timeobt = timeZoneOffset.longValue()
+		def minutes = (timeobt % 3600) / 60
+		def hours = (timeobt - minutes * 60) / 3600
+
+		if(!location.timeZone.useDaylightTime() || !location.timeZone.inDaylightTime(new Date())) {
+			log.debug "Setting offset on lock hourTzo: ${hours} minuteOffsetDst: 0 minuteTzo: ${minutes} signOffsetDst: 0 signTzo: ${timeZoneFlag}"
+			cmds << secure(zwave.scheduleEntryLockV3.scheduleEntryLockTimeOffsetSet(hourTzo:hours, minuteOffsetDst:0, minuteTzo:minutes, signOffsetDst:0, signTzo: timeZoneFlag))
```
greens, line 3033: default values (`minuteOffsetDst` and `signOffsetDst`) can be left out here.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1695,3 +3006,53 @@ def readCodeSlotId(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd) {
...
+		def now = new Date().toCalendar()
+		short hour = now.get(Calendar.HOUR_OF_DAY)
+		short min = now.get(Calendar.MINUTE)
+		short sec = now.get(Calendar.SECOND)
+		short day = now.get(Calendar.DAY_OF_MONTH)
+		short month = now.get(Calendar.MONTH) + 1
+		int year = now.get(Calendar.YEAR)
+
+		log.debug "Setting time on lock - $year-$month-$day $hour:$min:$sec (YYYY-MM-DD hh:mm:ss)"
+		cmds << secure(zwave.timeParametersV1.timeParametersSet(hourUtc: hour, minuteUtc: min, secondUtc: sec, day: day , month: month, year: year))
```
greens, line 3055: just reference the values of the calendar object directly inline

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1612,6 +2865,64 @@ def clearStateForSlot(codeID) {
 	state["setname$codeID"] = null
 }
 
+/**
+ * Clears the daily repeating start time, end time and weekday bit mask from the state basis the code slot number
+ *
+ * @param codeID: The code slot number
+ */
+def clearDailyRepeatingScheduleDataFromState(codeID) {
+	state["setDailyRepeatingStartTime$codeID"] = null
```
greens, line 2874: A couple of things about this style of state storage:

Why not just store `state["code$codeID"]` as a dict?

I think that would make adding/removing it a lot simpler.

Also I'd check to see that both of these steps are necessary. It would be my instinct that just calling `state.remove` should get rid of the data stored in the field without you having to specifically null it out.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
greens: how about 

`codesToScanSchedules = keySet.findall {it <= MAX_CODES_TO_SCAN}`

greens (changes requested):
I think you've got enough here for a first pass.

Keeping in mind that this is largely just a style review.

Do you think you could go through for me the flow from adding a device to setting a code through some app, and then deleting that code?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -386,15 +487,15 @@ private def handleAccessAlarmReport(cmd) {
 			break
 		case 2: // Manually unlocked
 			map.descriptionText = "Unlocked manually"
-			map.data = [ method: "manual" ]
+			map.data = [ method: "manual", notify: true, notificationText: "$deviceName at ${location.name} was unlocked manually" ]
```
mckeed, line 490: What is the purpose of `notify` in the event data? Can't that be handled by the SmartApp?

PR #2707, opened 2018-01-17
ICP-2954 Fingerprints for Yale Conexis

greens:
Looks like we forgot to add the fingerprint for the Conexis lock when we added the Keyless.


dkirker (approved):
Not sure what happened around the other commits in this, but looks good. (Yay whitespace mess!)

PR #2721, opened 2018-01-19
ICP-2939 Initial battery poll being ignored

greens:
This is purely a timing issue caused by secure commands creating a bit of a logjam.

If anyone has a better solution I'd love to try it.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -170,7 +170,7 @@ def doConfigure() {
 	if (isSchlageLock()) {
 		cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
 	}
-	cmds = delayBetween(cmds, 4200)
+	cmds = delayBetween(cmds, 10*1000)
 	log.debug "Do configure returning with commands := $cmds"
 	cmds
 }
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -170,7 +170,7 @@ def doConfigure() {
 	if (isSchlageLock()) {
 		cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
 	}
-	cmds = delayBetween(cmds, 10*1000)
+	cmds = delayBetween(cmds, 30*1000)
 	log.debug "Do configure returning with commands := $cmds"
 	cmds
 }


dkirker (dismissed):
Oh man. The timing with Z-Wave locks is crazy. This will probably be fine. Unless @mckeed has any complaints.

greens:
@dkirker need another okay on this one now that SR-Pol has confirmed what time works for them.

dkirker:
It looks *ok*, however I would like @mckeed to comment as well. Especially on updated() and the use of doConfigure() there. I have left a more detailed comment in the ticket.

mckeed (approved):
This is something we probably want to look at moving to the local engine.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -170,7 +170,7 @@ def doConfigure() {
 	if (isSchlageLock()) {
 		cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
 	}
-	cmds = delayBetween(cmds, 10*1000)
+	cmds = delayBetween(cmds, 30*1000)
```
vaish-rahul: It would have been better to increase the delay just for the battery get command. As of now, with this change even codeLength for Schlage locks will be received after 30 seconds of delay.

CC: @greens, @dkirker, @mckeed

mckeed: The same issue could cause that to be ignored, too. If the delay is causing a problem we should probably go back to the original delay but also schedule a follow-up callback with runIn() to check that battery and code length were received and if not request them again.

greens: I agree that the `runIn()` call would have been the better solution. I think the ideal solution we should work towards is to somehow guarantee that a delay between two commands is preserved respective to the two commands.

vaish-rahul: As of now we can change in following manner:

cmds << secure(zwave.doorLockV1.doorLockOperationGet())
if (isSchlageLock()) {
	cmds << secure(zwave.configurationV2.configurationGet(parameterNumber: getSchlageLockParam().codeLength.number))
}
cmds = delayBetween(cmds, 4200)
cmds << "delay 30000" << secure(zwave.batteryV1.batteryGet())

At least for Schlage locks the codeLength will be read without a big delay. The codeLength field for Schlage locks is highly important from lock codes point of view. Smart Locks smartapp displays the codeLength field as hint text to the user for Schlage locks.

greens: But we're adding the extra delay because the commands pile up and then get sent immediately one after the other. The lock can't handle this speed and starts dropping commands and not responding to them. There's no guarantee that the lock code length request would make it through and be answered under the same circumstances.

vaish-rahul: We don't see the problem of lock not responding in SRI-B. The QA team tested the functionality on Yale Z-Wave, Kwikset Z-Wave and Schlage Z-Wave locks. The lock is reporting even if we change the delay to 200ms. It may be possible that the lock in which the issue was observed had very less battery power remaining.

greens: I think it's a function of how many Z-Wave devices you have on your system? I was able to reproduce here in MV as well, but needed less of a buffer than SR-Pol did. The problem is that SR-Pol essentially gets final say about whether a device "works" or not, so at least for this issue we had to solve it in a way that worked for them.

PR #2729, opened 2018-01-22
ICP-2850 Setting color temperature causes offline device to come online

greens:
The event generated by setting the color name is causing devices to be marked online.

### DIFF
--- a/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
@@ -153,6 +153,7 @@ def installed() {
 }
 
 def setColorTemperature(value) {
+    // TODO: Use this method for all zigbee white lights, but also change logic in appengine-zigbee to round
     zigbee.setColorTemperature(value) + ["delay 1500"] + zigbee.colorTemperatureRefresh()
 }
 
--- a/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
@@ -153,8 +153,13 @@ def installed() {
 }
 
 def setColorTemperature(value) {
-    // TODO: Use this method for all zigbee white lights, but also change logic in appengine-zigbee to round
-    zigbee.setColorTemperature(value) + ["delay 1500"] + zigbee.colorTemperatureRefresh()
+    value = value as Integer
+    def tempInMired = Math.round(1000000 / value)
+    def finalHex = zigbee.swapEndianHex(zigbee.convertToHexString(tempInMired, 4))
+
+    zigbee.command(COLOR_CONTROL_CLUSTER, MOVE_TO_COLOR_TEMPERATURE_COMMAND, "$finalHex 0000") +
+    ["delay 1500"] +
+    zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_COLOR_TEMPERATURE)
 }
 
 def setLevel(value) {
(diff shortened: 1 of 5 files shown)


greens:
So a couple of questions about this.

This fix is an option, where we won't set the color temperature unless we get an update from the device.

We could also change the logic to only change the color temperature if the device is online, but that may have side effects.

Or we could get the device health team to do extra checking around which events cause devices to come back online

@tpmanley Would this change in the DTH also fix the issue for local execution lights?

devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
```diff
@@ -153,6 +153,7 @@ def installed() {
 }
 
 def setColorTemperature(value) {
+    // TODO: Use this method for all zigbee white lights, but also change logic in appengine-zigbee to round
```
greens: Noticed this is a copy of the truncating logic I just fixed in the DTHs. Only used in this one place, but if we fix it in appengine it could be used for all of these.

workingmonk (dismissed):

devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
tpmanley: Since `zigbee.setColorTemperature` truncates instead of rounds do you want to change it to round here and then update all the DTHs to call `zigbee.setColorTemperature()` once it's fixed there?

tpmanley:
I believe you can remove `command "setGenericName"` from all the DTHs since I don't think any SmartApps can or should be calling it.

tpmanley:
Consider adding the `setGenericName` stuff to zll-rgbw-bulb.groovy to make it consistent with the other RGBW/TW DTHs.

devicetypes/smartthings/zll-white-color-temperature-bulb.src/zll-white-color-temperature-bulb.groovy
```diff
@@ -79,6 +79,9 @@ def parse(String description) {
     log.debug "description is $description"
     def event = zigbee.getEvent(description)
     if (event) {
+        if (event.name=="colorTemperature") {
```
tpmanley: nit: spaces around `==`

tpmanley (dismissed):

devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
greens: Yeah. Seems like the right thing to do.

greens:
@tpmanley Incorporated some of your comments. I didn't add the color temperature attribute to the rgbw bulb since my guess is that it was left off due to confusion it might cause with the color vs color temp.

workingmonk:
setGenericName was added as a command because the first iteration had a tile to quickly switch between the value color temp values. over time that tile action was taken out, but the command persisted

tpmanley:
@greens unless I'm missing something it's inconsistent between the Zigbee RGBW Bulb and the ZLL RGBW Bulb DTH and I don't think there is any reason it shouldn't be the same

greens:
@tpmanley Done.

tpmanley (approved):
Looks good, just squash the commits before the merge

berkens:
I noticed that the color temperature can be changed by the slider from 2700K to 6500K.

However 2700K is not always the lower limit of the bulb / controller. 

It would be nice if the lower limit was 2000K. 

I returned an Osram bulb because it was not so warm white. 

Later on I experienced that a led strip controller with the same DTH the warm white leds were illuminated, but the cool white leds were also still on (dim). 

I changed the DTH for my led controller so the color temperature range is now 2000K - 6500K. The cool white leds are now off at 2000K.

PR #2737, opened 2018-01-24
PROB-1941 Smart Care apps missing Save button on iOS

marstorp:
The definition for the page in the preferences section did not specify
'install: true', instead this was specified in the declaration for the
dynamicPage. However, it seems changing the definition in the
dynamic page declaration doesn't matter as in iOS it always uses
the first declaration found.

Fixing this issue by adding 'install: true' to the preference declaration
and removing the page definition from dynamicPage.


rohandesai (approved):
a simple change to show "Save" button for iOS. 👍

PR #2747, opened 2018-01-30
ICP-3135 Wrong icon type for smart plug in SC

greens:
the switch device type gets a switch icon, but this is a plug so it needs the plug icon


dkirker (approved):
@greens Looks good! And I tested and yes, this does in fact work. :)

@workingmonk Really stupid question... But since this change literally only seems to affect the icon in Connect, is it possible that we could get this to tail my commit in staging? Or would you rather it wait a week? (I honestly have no preference either way. If you know that the ocfDeviceType property affects other things then let's wait for ST QA to have a look.)

PR #2764, opened 2018-02-05
Implementation of Eaton RF Master Switch, Eaton RF Accessory Switch, …

ALamchaS:
…and Eaton RF Receptacle using generic switch DTH

### DIFF
--- a/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
+++ b/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
@@ -147,7 +147,6 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 }
 
 def on() {
-	def supportsSwitchBinary = !device.getDataValue("MSR").equals("001A-5352-0000")
 	delayBetween([
 		zwave.basicV1.basicSet(value: 0xFF).format(),
 		getSwitchStateZwaveCommand().format()
--- a/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
+++ b/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
@@ -15,33 +15,36 @@ metadata {
 	definition (name: "Z-Wave Switch Generic", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.switch", runLocally: true, minHubCoreVersion: '000.019.00012', executeCommandsLocally: true) {
 		capability "Actuator"
 		capability "Health Check"
- 		capability "Switch"
+		capability "Switch"
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
 		capability "Light"
 
 		fingerprint inClusters: "0x25", deviceJoinName: "Z-Wave Switch"
-		fingerprint mfr:"001D", prod:"1A02", model:"0334", deviceJoinName: "Leviton Appliance Module"
-		fingerprint mfr:"0063", prod:"4F50", model:"3031", deviceJoinName: "GE Plug-in Outdoor Switch"
-		fingerprint mfr:"001D", prod:"1D04", model:"0334", deviceJoinName: "Leviton Outlet"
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
```diff
@@ -151,22 +147,23 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 }
 
 def on() {
+	def supportsSwitchBinary = !device.getDataValue("MSR").equals("001A-5352-0000")
```
greens: I think you forgot to delete this line.

greens (changes requested):
How does the accessory switch respond to the binary switch command class? It's listed as supported on the Sigma device database.

Does basic CC control the actual switch's on/off state and switch CC control associated devices?

devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
ALamchaS: Yes. Thanks for catching this.

ALamchaS:
Accessory switch does not respond to binary switch command class at all. I attached pdf "documentation" we received in related Jira ticket.

Raw description comparison:

Accessory switch:
Raw Description | zw:L type:1201 mfr:001A prod:5352 model:0000 ver:3.13 zwv:3.52 lib:03 cc:27,75,86,70,85,77,2B,2C,72,73,87

Master Switch:
Raw Description | zw:L type:1003 mfr:001A prod:534C model:0000 ver:3.21 zwv:3.67 lib:03 cc:25,27,75,86,70,71,85,77,2B,2C,72,73,82,87

COMMAND_CLASS_SWITCH_BINARY id is 0x25 - Master switch reports it as supported, accessory switch does not.

Other difference: 
- Accessory switch reports type as 1201, so GENERIC_TYPE_SWITCH_REMOTE and SPECIFIC_TYPE_SWITCH_REMOTE_BINARY.
- Master switch reports type as 1003 (same as receptacle), so GENERIC_TYPE_SWITCH_BINARY and SPECIFIC_TYPE_SCENE_SWITCH_BINARY

Basic set does control Accessory switch's state.

After associating Master Switch with Accessory Switch (using AssociationSet commands from DTH, as those devices can not set up their own z-wave network), physically pushing switches is handled properly, but BasicSet commands originating from Hub do not get propagated by any of the devices. To operate 100% correctly, automation would be required. I found topic where people discuss this problem using other controller, link: http://forum.micasaverde.com/index.php?topic=13664.0

@MAblewiczS mentioned there was discussion regarding Accessory Switch on Slack. @mckeed told z-wave only switches not connected to load will probably get separate generic DTH. For now I will leave Accessory switch implementation as it is, and wait for your opinion.

Other important subject - we removed methods that were some kind of implementation for inverting led indicator, as this does not seem to be used by any capability and if this code was executed, Eaton devices would interpret configurationSet calls used there differently.

Can you please confirm if this is OK with other devices grouped in this generic DTH?

devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
```diff
@@ -113,13 +116,6 @@ def zwaveEvent(physicalgraph.zwave.commands.switchbinaryv1.SwitchBinaryReport cm
 	[name: "switch", value: cmd.value ? "on" : "off", type: "digital"]
 }
 
-def zwaveEvent(physicalgraph.zwave.commands.configurationv1.ConfigurationReport cmd) {
```
greens, line 116: This was for either the GE or the Leviton devices. I'm okay with losing it.

greens:
@ALamchaS My binary switch comments were based on the official Sigma documentation for this device:

https://products.z-wavealliance.org/products/742

I'm not surprised that it doesn't match your actual experience with the device.

Your changes look good, but we'll see where we end up after the discussion about what to do with the accessory devices.

greens (approved):

PR #2766, opened 2018-02-05
Revert 2585 dvcsmp 3412

nathankooistra:
Revert the SmartApp change

PR #2767, opened 2018-02-05
Revert "DVCSMP-3412 Utilize sunrise/sunsetTime instead of external API"

nathankooistra:
Revert SmartApp


workingmonk:
👍 based on the approved PR https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2766

PR #2775, opened 2018-02-07
Added fingerprint for Eaton RF 5-Scene Keypad

MMateusiakS:
Added fingerprint for Eaton RF 5-Scene Keypad


ALamchaS:
Please note that for this device to work via z-wave network directly, additional configuration is required. This configuration can be done only using other z-wave controller. http://www.cooperindustries.com/content/dam/public/wiringdevices/products/documents/instruction_sheets/rfwcddsc_rev._a_.pdf

PR #2777, opened 2018-02-07
Add modification for generic DTH for Z-Wave controller. Added fingerp…

MMateusiakS:
Add modification for generic DTH for Z-Wave controller. 
Added fingerprint for Eaton RF Anyplace Switch.


ALamchaS:
Please note this PR removes on/off custom commands and buttons, as those will not work with this sleepy device.

mckeed (changes requested):
If a device is sleepy, it should be added to Z-Wave Remote, not this one.

Those on/off commands are required for Z-Wave certification of the hub.

PR #2781, opened 2018-02-08
Report correct status when locked via keypad

rboy1:
Schlage and Yale, with One Touch locking enabled are locked manually from the keypad, report the status correct to differentiate manual locking and keypad locking

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -406,6 +406,7 @@ private def handleAccessAlarmReport(cmd) {
 			} else {
 				// locked by pressing the Schlage button
 				map.descriptionText = "Locked manually"
+				map.data = [ method: "keypad" ]
 			}
 			break
 		case 6: // Unlocked with keypad


rboy1:
@vaish-rahul @jackchi 

Folks can you review this change and accept it, the lock isn't reporting the source of an external keypad lock. Creating an issue to correctly identify the intent, internal manual lock is different from external keypad lock.

vaish-rahul (approved):

vaish-rahul:
@greens Please review.

greens (approved):

rboy1:
Guys do we have to wait for a new firmware build for the changes to be pushed to the hub or with new architecture can incremental changes to DTH's get pushed separately?

greens:
For DTH changes, they'll first go to staging next Tuesday, and then if nothing bad happens production the Tuesday after. They don't require hub updates.

workingmonk:
@tpmanley @mckeed  does this code block run locally? if yes, I am guessing that the cloud change would not help non-v1 users, right? we would need this change to go out to protocol library?

greens:
@workingmonk Lock DTH isn't marked to run locally, so none of this does.

tpmanley:
Z-Wave Lock is one of the DTHs that ran locally in AppEngine and it still does run locally post-AppEngine even though it doesn't have the metadata reflecting that. However this particular code block does not run locally today.

rboy1:
@tpmanley so does that mean the changes get pushed to the v2 hubs after 2 weeks once it hits production or are the DTH's split between running locally for some parts and in the DTH for other parts on the cloud with the post-AppEngine?

tpmanley:
@rboy1 the code in the hub is only updated with a full firmware release. Most of the lock code management stuff runs in the cloud so that the lock code team can iterate on it faster. Only the basic lock control (lock/unlock) runs locally right now so that it can still be controlled by locally running smartapps like Lighting Automation and Smart Home Monitor.

rboy1:
So you're saying part of the DH runs in the cloud and other parts locally? E.g. lock() and unlock() run the hub, the rest of the DH runs in the cloud?

tpmanley:
@rboy1 that's correct

PR #2787, opened 2018-02-12
DVCSMP-3567 Fingerprints for directly connected Hue lights

tpmanley:
Although not officially certified or supported many users connect Hue lights
directly to the SmartThings hub and most of the type they join as Things.
This adds fingerprints for most of the models listed in:
https://developers.meethue.com/documentation/supported-lights.

https://smartthings.atlassian.net/browse/DVCSMP-3567


dkirker (approved):
As long as these all work and there are no serious concerns, this looks good!

PR #2788, opened 2018-02-12
Updated 4CH Sonoff, works in ST App

GeorgeIoak:


GeorgeIoak:
I don't think that came over well so sorry about that if it's not usable. I made some changes which now allows the ST app to control the Sonoff and display ON/OFF states correctly for all channels. Unfortunately the ST app does not change when the Sonoff initiates the changes.

PR #2789, opened 2018-02-12
DVCSMP-3566 Remove blacklisted Kwikset Smartcode lock

dkirker:


tpmanley (changes requested):
Need to also add this to the zigbee lock dth

dkirker:
@tpmanley D'oh!! Fixed. 😛

tpmanley (approved):

PR #2808, opened 2018-02-17
SAPLAT-96 Virtual Device Manager to create virtual devices from the app

workingmonk:


workingmonk:
Made changes to the virtual devices to set the initial state so that there are no issues in one app

workingmonk:
screenshot_20180216-165720

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,78 @@
...
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+definition(
+    name: "Virtual Device Manager",
+    namespace: "smartthings",
+    author: "SmartThings",
+    description: "Tests child device creation and update",
```
tpmanley: Copy /paste? Maybe "Create virtual devices"

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,78 @@
...
+preferences {
+    page name: "mainPage", install: true, uninstall: true
+
+}
+
+def mainPage() {
+    dynamicPage(name: "mainPage") {
+        section {
+            input "switchCount", "number", title: "Select number of Virtual Switches", defaultValue: 1, required: true
+            input "dimmerCount", "number", title: "Select number of Virtual Dimmers", defaultValue: 1, required: true
```
tpmanley: My personal experience is making one virtual device at a time.  What do you think about changing this to focus on that use case and prompt for the device name, switch or dimmer and hub?

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,78 @@
...
+            input "theHub", "hub", title: "Select hub for local execution", multiple: false, required: false
+        }
+    }
+}
+
+def installed() {
+    log.debug "Installed with settings: ${settings}"
+    state.lastSwitchDni = 0
+    state.lastDimmerDni = 0
```
tpmanley: I think we should use a single counter and name for these devices (ex: virtual-$num) because they may get retyped in the future which will lead to confusion if the DNI doesn't match the type.

workingmonk:
@tpmanley fixed based on your comments

workingmonk:
@bflorian can you confirm if updated is always called after installation? I can see that happen which causes 2 virtual devices to be created with the new logic

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,57 @@
...
+        section {
+            input "virtualDeviceType", "enum", title: "Which type of virtual device do you want to create?", multiple: false, required: true, options: ["Virtual Switch", "Virtual Dimmer Switch"]
+            input "theHub", "hub", title: "Select hub for local execution (Optional)", multiple: false, required: false
+        }
+    }
+}
+
+def installed() {
+    log.debug "Installed with settings: ${settings}"
+    state.lastDni = 1
```
tpmanley: Should be renamed to nextDni based on how it's used

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,57 @@
...
+}
+
+def updated() {
+    log.debug "Updated with settings: ${settings}"
+    initialize()
+}
+
+def initialize() {
+    def latestDni = state.lastDni
+    def d = addChildDevice("smartthings", virtualDeviceType, "virtual-$latestDni", theHub?.id, [completedSetup: true, label: "$virtualDeviceType $latestDni"])
```
tpmanley: Does virtualDeviceType have a value when this is called from installed? If not you could make sure  that is set before making the child device.

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,57 @@
...
+preferences {
+    page name: "mainPage", install: true, uninstall: true
+
+}
+
+def mainPage() {
+    dynamicPage(name: "mainPage") {
+        section {
+            input "virtualDeviceType", "enum", title: "Which type of virtual device do you want to create?", multiple: false, required: true, options: ["Virtual Switch", "Virtual Dimmer Switch"]
+            input "theHub", "hub", title: "Select hub for local execution (Optional)", multiple: false, required: false
```
tpmanley: Might be more clear to say "Select the hub (required for local execution)"

workingmonk:
@tpmanley updated

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,62 @@
...
+preferences {
+    page name: "mainPage", install: true, uninstall: true
+
+}
+
+def mainPage() {
+    dynamicPage(name: "mainPage") {
+        section {
+            input "virtualDeviceType", "enum", title: "Which type of virtual device do you want to create?", multiple: false, required: true, options: ["Virtual Switch", "Virtual Dimmer Switch"]
+            input "theHub", "hub", title: "Select the hub (required for local execution) (Optional)", multiple: false, required: false
```
tpmanley, line 36: Do you think we could remove the `(Optional)` now because it's implied by the `required for local execution`?

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,62 @@
...
+    initialize()
+}
+
+def initialize() {
+    def latestDni = state.nextDni
+    if (virtualDeviceType) {
+        def d = addChildDevice("smartthings", virtualDeviceType, "virtual-$latestDni", theHub?.id, [completedSetup: true, label: "$virtualDeviceType $latestDni"])
+        latestDni++
+        state.nextDni = latestDni
+    }
```
tpmanley: nit: `} else {`

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,62 @@
...
+    else {
+        log.error "Failed creating Virtual Device because the device type was missing"
```
tpmanley: Do we get this error when the app is installed?

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,62 @@
...
+    description: "Creates virtual devices",
+    iconUrl: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience.png",
+    iconX2Url: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience@2x.png",
+    iconX3Url: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience@2x.png"
+)
+
+
+preferences {
+    page name: "mainPage", install: true, uninstall: true
+
```
tpmanley: nit: extra newline

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -0,0 +1,62 @@
+/**
+ *  Virtual Device Manager
+ *
+ *  Copyright 2015 Bob Florian/SmartThings
```
tpmanley, line 4: Should update copyright date/author since it's a new file

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
workingmonk: Yeah, but for Virtual locations with no hubs, the `(Optional)` makes it clear that this is work for them too

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
workingmonk: Bob said the `updated()` isn't supposed to be called at install, but it is getting called. this error is to track that during our testing and QA

smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
workingmonk: Copyright should be the oldest based on when the file was used, right? The code is based on Bob's code written in 2015, so copyright is 2015.
older is better

tpmanley (approved):

PR #2827, opened 2018-02-23
DVCSMP-3412 Restore the code change to Sunrise/Sunset

nathankooistra:
Restoring the PR that was later reverted, but started out as https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2585 with some additional logging.


rappleg:
I know that @nathankooistra did a lot of testing with this locally and seemed to be working as expected. Would like to get it out to dev/staging so we could get some more eyes on it since DST is coming up soon.

rappleg (approved):

PR #2828, opened 2018-02-23
ICP-3774 Update pingable contact sensors to poll status

greens:
As a stop-gap before a more robust solution to offline state change mismatches, we'll now ping contact sensors for their contact state instead of for battery or temperature.

These were all of the "Contact Sensor" and "Water Sensor" devices I could find that also supported "Health Check" and were not sleepy.

Let me know if I missed any.


tpmanley:
I think we should do motion as well

greens:
But motion is instantaneous. Motion would only matter in this situation if someone's dancing in front of your sensor as the power comes back on (+ ping time)

greens:
Added the only health-checked motion sensor I could find.

tpmanley (approved):

PR #2829, opened 2018-02-24
DVCSMP-3535 Add battery metric logging

dkirker:


devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -99,6 +99,11 @@ def parse(String description) {
 		} else {
 			Map descMap = zigbee.parseDescriptionAsMap(description)
 			if (descMap?.clusterInt == 0x0001 && descMap.commandInt != 0x07 && descMap?.value) {
+				if (descMap?.attrInt == 0x003E) {
```
tpmanley: you could instead log the `attrId` and then you wouldn't need the if statement. It would also be helpful in case there is an unexpected attribute being reported.

ex: `log.info "BATT METRICS - 0x${descMap?.attrId} attr; value: 0x${descMap?.value}, curr percent: ${device.currentState("battery")?.value}, device: ${device.getDataValue("manufacturer")} ${device.getDataValue("model")}"`

dkirker: @tpmanley That's actually not a bad idea. I added that, and the code has been updated.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -99,6 +99,7 @@ def parse(String description) {
 		} else {
 			Map descMap = zigbee.parseDescriptionAsMap(description)
 			if (descMap?.clusterInt == 0x0001 && descMap.commandInt != 0x07 && descMap?.value) {
+				log.info "BATT METRICS - ${descMap?.attrInt} attr; value: ${descMap?.value}, dec value: ${Integer.parseInt(descMap.value, 16)}, curr percent: ${device.currentState("battery")?.value}, device: ${device.getDataValue("manufacturer")} ${device.getDataValue("model")}"
```
tpmanley: I think it would be clearer to put `attr:` before the attribute value

tpmanley (dismissed):

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: @tpmanley I tweaked it. And apparently it means I dismissed your approval. Anyway, once I get it again I'll merge it.

tpmanley (approved):

PR #2835, opened 2018-02-27
Make generic switch cloud DTH match LEE version

greens:
If we genericize the generic switch DTH (and make it match local execution implementation) we should be able to get the Eaton accessory dimmer to fit without any changes.


ALamchaS:
Looks good to me.

Just tested this on Eaton Accessory Switch device, and local operation works correctly.

ALamchaS (approved):

mckeed:
The idea behind making a separate DTH for accessory switches is that we are going to want a way to distinguish them in the future. We're going to be adding device categories that we'll want to be different between real switches and accessory/remote switches. We also might decide that these should have the Remote Control capability since they are controlling devices (vs controlled).

mckeed (dismissed):
After talking with Steven I agree this is the best way to handle accessory switches. Switch mirroring isn't going anywhere, so switch types can be seen as controlling devices in general.

greens:
@mckeed double-check my merge?

mckeed (dismissed):
The indentation is off on the fingerprint

greens:
should be fixed now

mckeed:
Sorry I missed this earlier – it looks like your change to refresh got clobbered in the merge

greens:
Thanks for catching that.

greens:
@dkirker  adding you since Duncan is on vacation

dkirker (approved):

PR #2845, opened 2018-02-28
ICP-3834 Add executeCommandsLocally flag to legacy locally executing DTHs

tpmanley:
Just default the value to `false` to match the current behavior.


rappleg (approved):

varzac:
I don't know how to approve via mobile, but approve from me

PR #2849, opened 2018-03-01
ICP-3812 Update Door Knocker to prepare for Pabal support

ajguyot:
With this setup Door Knocker will send the "{{kSensor}} detected a knock." notification through Pabal on accounts in which the service is enabled. Once we have a valid code and param name for the Door Knocker notification string we can swap those in and the app will begin sending with Pabal translations immediately with no extra changes necessary.

Pabal ignores invalid codes and defaults back to the plain Message so this SmartApp will work perfectly fine as is. If we'd rather wait for the code before merging then I can update this PR to swap the code in as soon as we have it.

/cc @bflorian

### DIFF
--- a/smartapps/imbrianj/door-knocker.src/door-knocker.groovy
+++ b/smartapps/imbrianj/door-knocker.src/door-knocker.groovy
@@ -73,9 +73,9 @@ def doorClosed(evt) {
 def doorKnock() {
   if((openSensor.latestValue("contact") == "closed") &&
     (now() - (60 * 1000) > state.lastClosed)) {
-	def kSensor = knockSensor.label ?: knockSensor.name
+    def kSensor = knockSensor.label ?: knockSensor.name
     log.debug("${kSensor} detected a knock.")
-    send(localization.translate("{{kSensor}} detected a knock.", [kSensor: kSensor]))
+    send(kSensor)
   }
 
   else {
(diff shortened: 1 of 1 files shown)


ajguyot:
To confirm, this update will result in identical functionality to the existing version for now, including sending SMS notifications for users who previously had it set up with a `phone` param (but not allowing any new users to send SMS). I removed the direct call to translate the message that was in this SmartApp because that translation occurs automatically within Data Management, so it's unnecessary to do it here first.

ajguyot:
Updated to include the Pabal codes proposed in ICP-3812, so this SmartApp should now be complete. We're still waiting on getting those codes actually added to Pabal though so currently if installed it would still send untranslated, but as soon as the codes are added it will translate.

ajguyot:
As per [ICP-3812](https://smartthings.atlassian.net/browse/ICP-3812) the new Door Knocker codes have finally been added to Pabal. Reopening this PR, if I can get an approval on it then the Door Knocker changes for Pabal support will be complete.

lilianaST (dismissed):

smartapps/imbrianj/door-knocker.src/door-knocker.groovy
```diff
@@ -16,8 +16,7 @@ definition(
     description: "Alert if door is knocked, but not opened.",
     category: "Convenience",
     iconUrl: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience.png",
-    iconX2Url: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience%402x.png",
-    pausable: true
+    iconX2Url: "https://s3.amazonaws.com/smartapp-icons/Convenience/Cat-Convenience%402x.png"
```
workingmonk: any reason we took out pausable?

ajguyot: Oh yeah I can add that back in now. Back in March there was a bug where pausable apps were failing to compile so I pulled it since we wanted this out right away (lol), but we've fixed that so I'll update

ajguyot:
@workingmonk Updated to add pausable back

lilianaST (approved):

bflorian (approved):

erodewald:
Is this related? https://www.reddit.com/r/SmartThings/comments/9p0vrr/any_idea_why_the_someone_is_knocking_at_the_door/

I'm seeing widespread confirmation that string interpolation isn't working for the push notification. I personally see it in both classic and oneapp.

ajguyot:
🤔 Well the phrase `knockSensor.name` is never used in quotes so it doesn't make much sense to me that the code here would be resulting in this. Also this code was tested and I can't see how that extremely clear error would not have been caught. I'd make a ticket with the Pabal team and have them verify. Feel free to tag me and I can double check my test installation of this app tomorrow and see if I can pin down the source as here or Pabal.

erodewald:
@ajguyot That was my assessment too, thanks for the suggestion, I'll tag you on the issue in Pabal tomorrow.

PR #2850, opened 2018-03-02
VOICE-217 Prevent spurious device delete & recreate

surfous:
If LIFX API doesn’t return a 200 response when querying devices in a location, we returned an empty list which led us to delete all devices. Those that were part of a SmartApp could not be removed, and did not churn as such.

Now we:
* return null if not a 200 response - causes SA to error out instead of deleting devices
* do not force delete devices when using Cirrus - this would have exacerbated the problem

As this is being fast-tracked, changes are intentionally minimal.

### DIFF
--- a/smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
+++ b/smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
@@ -54,37 +54,37 @@ def getClientId()                { return appSettings.clientId }
 private getVendorName() { "LIFX" }
 
 def authPage() {
-	if (state.lifxAccessToken) {
-		def validateToken = locationOptions() ?: []
-	}
+    if (state.lifxAccessToken) {
+        def validateToken = locationOptions() ?: []
+    }
 
-	if (!state.lifxAccessToken) {
-		log.debug "no LIFX access token"
-		// This is the SmartThings access token
-		if (!state.accessToken) {
-			log.debug "no access token, create access token"
-			state.accessToken = createAccessToken() // predefined method
-		}
-		def description = "Tap to enter LIFX credentials"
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -431,7 +431,7 @@ void updateDevicesFromResponse(devices) {
 
     children.findAll { !deviceIds.contains(it.deviceNetworkId) }.each {
         log.trace "deleting child device $it.label"
-        deleteChildDevice(it.deviceNetworkId, true)
+        deleteChildDevice(it.deviceNetworkId, false)
```
workingmonk: can you change it to `deleteChildDevice(it.deviceNetworkId)`
default is false.

smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -377,7 +377,7 @@ def devicesList(selector = '') {
 			state.lifxAccessToken = null
 		} else {
 			log.debug("No response from device list call. ${resp.status} ${resp.data}")
-			return []
+			return null
```
workingmonk: instead of returning null and failing later with NPE, can we throw an exception here with the right message? I am worried that this might result in a few thousand NPE on the dashboard and might send us down another rabbit hole

surfous:
@workingmonk hokay, I've made the mods you've requested - I'm throwing a RuntimeException because unchecked seemed to fit the circumstance better, and no other standard exception looked like a good match. I'm hoping I'm allowed to throw this from the sandbox. 

Also, other changes are all whitespace. I saw the indents were mixed and I switched them to spaces.

PR #2853, opened 2018-03-02
Update the Inovelli 2 Channel Plug with changes from the Inovelli website

mjfakler:
The Inovelli 2 channel smart plug has `isComponent: true` which should be false. This is an update from the [Inovelli website](https://inovelli.com/smartthings-device-handlers/#NZW37Alt) for their device handler.

### DIFF
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
@@ -222,7 +222,7 @@ def updated() {
 		}
 		state.oldLabel = device.label
 	}
-	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	sendEvent(name: "needUpdate", value: device.currentValue("needUpdate"), displayed: false, isStateChange: true)
 }
 def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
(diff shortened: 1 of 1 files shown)


devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
```diff
@@ -222,7 +222,7 @@ def updated() {
 		}
 		state.oldLabel = device.label
 	}
-	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
tpmanley, line 225: I don't think we want to remove `offlinePingable`. Without that it won't be able to automatically come back online if it happens to go offline.

devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
```diff
@@ -252,13 +252,19 @@ private channelNumber(String dni) {
...
+			addChildDevice("Switch Child Device", "${device.deviceNetworkId}-ep${i}", null, [completedSetup: true, label: "${device.displayName} (CH${i})",
+				isComponent: false, componentName: "ep$i", componentLabel: "Channel $i"
+			])
+		}
+	} catch (e) {
+		runIn(2, "sendAlert")
 	}
 }
-
+private sendAlert() {
```
tpmanley, line 265: Why is this needed? The `Switch Child Device` is published by us so I'm not sure if this error would ever be hit.

devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
```diff
@@ -252,13 +252,19 @@ private channelNumber(String dni) {
...
 private void createChildDevices() {
 	state.oldLabel = device.label
-	for (i in 1..2) {
-		addChildDevice("Switch Child Device", "${device.deviceNetworkId}-ep${i}", null, [completedSetup: true, label: "${device.displayName} (CH${i})",
-																						 isComponent: true, componentName: "ep$i", componentLabel: "Channel $i"
-		])
+	try {
+		for (i in 1..2) {
+			addChildDevice("Switch Child Device", "${device.deviceNetworkId}-ep${i}", null, [completedSetup: true, label: "${device.displayName} (CH${i})",
+				isComponent: false, componentName: "ep$i", componentLabel: "Channel $i"
```
bflorian, line 258: We don't want to do this. The motivation comes from wanting to be able to re-label the device, but not treating it as a component has other implications for One App and the API. There is this pull request that would allow components to be re-labeled: https://github.com/PhysicalGraph/data-management/pull/4770

bflorian (commented):
I'd rather push on getting the referenced DM change done that mark these child devices as not being components.

mjfakler:
Closing, there is a better solution not involving changing the DTH.

PR #2855, opened 2018-03-03
DVCSMP-3639 Fix ZLL RGBW Bulb Poll for Health Check

workingmonk:


devicetypes/smartthings/zll-dimmer-bulb.src/zll-dimmer-bulb.groovy
```diff
@@ -113,6 +112,8 @@ def poll() {
  * PING is used by Device-Watch in attempt to reach the Device
  * */
 def ping() {
+	unschedule()
```
tpmanley: `configureHealthCheck` will call `unschedule` so we don't want to also call it here. If we do and the health check was already configured then it would be left unscheduled.

devicetypes/smartthings/zll-rgb-bulb.src/zll-rgb-bulb.groovy
```diff
@@ -110,9 +109,23 @@ def configure() {
...
+	configureHealthCheck()
 	refreshAttributes()
 }
 
+def configureHealthCheck() {
+	Integer hcIntervalMinutes = 12
+	if (!hasConfiguredHealthCheck) {
+		log.debug "Configuring Health Check, Reporting"
+		unschedule("healthPoll")
+		runEvery5Minutes("healthPoll")
```
tpmanley, line 122: Need to add the `healthPoll` function to this and the other DTHs that don't have it yet

devicetypes/smartthings/zll-white-color-temperature-bulb-5000k.src/zll-white-color-temperature-bulb-5000k.groovy
```diff
@@ -117,7 +116,9 @@ def poll() {
 /**
  * PING is used by Device-Watch in attempt to reach the Device
  * */
-def ping() {
+ def ping() {
```
tpmanley, line 119: nit: extra space before `ping` here and in the next one

devicetypes/smartthings/zll-rgb-bulb.src/zll-rgb-bulb.groovy
```diff
@@ -110,9 +109,29 @@ def configure() {
...
 
 def ping() {
+	unschedule()
+	configureHealthCheck()
 	refreshAttributes()
 }
 
+def configureHealthCheck() {
+	Integer hcIntervalMinutes = 12
+	if (!hasConfiguredHealthCheck) {
```
tpmanley, line 119: Need to add this to the top of each file: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zll-white-color-temperature-bulb.src/zll-white-color-temperature-bulb.groovy#L14

devicetypes/smartthings/zll-dimmer-bulb.src/zll-dimmer-bulb.groovy
```diff
@@ -13,13 +13,12 @@
...
 import groovy.transform.Field
 
-@Field Boolean hasConfiguredHealthCheck = false             
+@Field Boolean hasConfiguredHealthCheck = false
 
 metadata {
 	definition(name: "ZLL Dimmer Bulb", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", runLocally: true, minHubCoreVersion: '000.021.00001', executeCommandsLocally: true) {
 		capability "Actuator"
 		capability "Configuration"
-		capability "Polling"
```
tpmanley, line 22: You should be able to remove the `poll` command from all these DTHs too. I'm checking on the details of how to remove the capability and ensure that `poll` stops being called.

gausnes:
I think it might be worth trying to fix the issues surrounding polling. This is the first time that polling issues have been brought to my attention at the least, not sure how widely known it is that this is a problem. This approach is going to cause considerably more pressure on ticker and doesn't scale as well. Also schedules can be a PITA to deal with from my experience.

Any additional thoughts @rappleg @Bijnagte?

PR #2862, opened 2018-03-06
ICP-3846 Ecobee Sensor's temperature is missing "units"

marstorp:
Adding unit to temperature event and changing invalid value "--" for
"unknown" to 0 as temperature is of type NUMBER.


rohandesai (approved):
👍

PR #2873, opened 2018-03-07
Add usage information for virtual device creator

varzac:
This adds some additional information to the preferences page for the
virtual device manager to explain what the different actions will do.


smartapps/smartthings/virtual-device-manager.src/virtual-device-manager.groovy
```diff
@@ -15,7 +15,7 @@
  */
 
 definition(
-    name: "Virtual Device Manager",
+    name: "Virtual Device Creator",
```
tpmanley, line 18: @workingmonk Is it possible to change the filename so that it matches the new name?

PR #2874, opened 2018-03-07
DVCSMP-3277 Set Aeon Minimote Child buttons hub

varzac:
Previously these child devices were not associated with a hub.  As a
result the information about the child devices weren't synced locally.
This prevented us from being able to generate events for those child
devices.

This is a part of: https://smartthings.atlassian.net/browse/DVCSMP-3277


tpmanley (dismissed):
Would be nice if this could go against staging so it gets released next week

varzac:
Done

tpmanley (approved):

varzac:
I don't have permissions to merge this.  @workingmonk Are you OK with this?

PR #2875, opened 2018-03-07
[WWST-172] ecobee-add switch 20180307

twack:


devicetypes/smartthings/ecobee-switch.src/ecobee-switch.groovy
```diff
@@ -0,0 +1,95 @@
...
+	log.trace "[DTH] Executing 'on' for ${this.device.displayName}"
+    boolean desiredState = true
+	parent.controlSwitch( this.device.deviceNetworkId, desiredState )
+}
+
+def off() {
+	log.trace "[DTH] Executing 'on' for ${this.device.displayName}"
+    boolean desiredState = false
+	parent.controlSwitch( this.device.deviceNetworkId, desiredState )
+}
```
bflorian, line 90: I hate to be picky but let's not have messed up whitespace in brand new code. We officially are supposed to use tabs but I won't complain about four spaces.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -317,57 +409,75 @@ def updated() {
...
-		if (!ecobeesensors) {
-			delete = getChildDevices().findAll { !thermostats.contains(it.deviceNetworkId) }
-		} else {
-			delete = getChildDevices().findAll { !thermostats.contains(it.deviceNetworkId) && !ecobeesensors.contains(it.deviceNetworkId)}
-		}
-	}
-	log.warn "delete: ${delete}, deleting ${delete.size()} thermostats"
-	delete.each { deleteChildDevice(it.deviceNetworkId) } //inherits from SmartApp (data-management)
+    log.debug "[SM] removeChildDevice() - Delete this: ${delete}"
+    delete.each {
```
bflorian: I'm not seeing where `delete` is set any longer, and for that matter, why we need to do this at all. Isn't the only reason for having this method to update the settings when a device is deleted

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -317,57 +409,75 @@ def updated() {
...
+        deleteChildDevice(childDevice.device.deviceNetworkId)
+    }
+    if (!settings['selectedDevices']) //empty list, bail, service manager will be deleted with no devices
+    	return
```
bflorian: Rather than assume the platform behavior of deleting the service manager when the last device is deleted I think it would be safer in this case to just update the settings to an empty list and let the platform do (or not do) what it wants with the service manager.

bflorian (changes requested):
Besides the specific comments, I'm really confused as to how polling is working here, with poll, pollChild, pollChildren, and pollHandler. We may want to make it a separate PR but I think we should refactor this to not use the built in device poll() capability (which is being deprecated anyway) and have the service manager control polling of all the devices.

devicetypes/smartthings/ecobee-switch.src/ecobee-switch.groovy
twack: My editors show correctly. I will make sure that I change editor defaults. Updated file to tab correctly.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
twack: Updated to suggested method

twack:
ticket for refactor of polling method:
https://smartthings.atlassian.net/browse/WWST-202
Ecobee will be updating their service capabilities to publish events. That will be a good time to implement the preferred polling methodology.

bflorian (approved):

PR #2885, opened 2018-03-09
Add more features to the virtual device creator

varzac:
This adds the ability to name the device being created, and also adds a
list of the devices created with the app.

I have this pointed at production because that is where the changes to this are current.  After next weeks roll down to staging, we can point this at staging with the aim to get it out the week after that.  I don't think there is any hurry to get this out, and we may get some beta feedback between now, and the week after next that we want to incorporate.

### DIFF
--- a/smartapps/smartthings/virtual-device-creator.src/virtual-device-creator.groovy
+++ b/smartapps/smartthings/virtual-device-creator.src/virtual-device-creator.groovy
@@ -1,5 +1,5 @@
 /**
- *  Virtual Device Manager
+ *  Virtual Device Creator
  *
  *  Copyright 2015 Bob Florian/SmartThings
  *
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/virtual-device-creator.src/virtual-device-creator.groovy
```diff
@@ -33,20 +33,35 @@ preferences {
...
 def mainPage() {
     dynamicPage(name: "mainPage") {
-        section {
-            paragraph "You have created ${state.nextDni? state.nextDni - 1 : 0} virtual devices through this app"
+        section ("New Device") {
             input "virtualDeviceType", "enum", title: "Which type of virtual device do you want to create?", multiple: false, required: true, options: ["Virtual Switch", "Virtual Dimmer Switch"]
             input "theHub", "hub", title: "Select the hub (required for local execution) (Optional)", multiple: false, required: false
         }
-        remove("Remove All Virtual Devices", "This will remove all virtual devices created through this app.")
+        section("Device name") {
```
tpmanley: Maybe capitalize Name

tpmanley (approved):

PR #2889, opened 2018-03-09
DVCSMP-3684 Fix simulated RGB/RGBW DTHs to avoid iOS ST Classic crash

surfous:
Sending colorControl.color attribute as a hex RGB string rather than a COLOR_MAP eliminates this crash.

Note that the capability docs specify this attribute’s data type as COLOR_MAP:
* [Color Control capability](https://smartthings.developer.samsung.com/develop/api-ref/capabilities.html#Color-Control)
* [COLOR_MAP data type](https://smartthings.developer.samsung.com/develop/api-ref/capabilities.html#Data-types)


rohandesai (approved):
LGTM code wise.

PR #2895, opened 2018-03-11
Update fibaro-rgbw-controller.groovy

vseven:
There is a error in the hueToRgb function.  See https://community.smartthings.com/t/device-handler-for-nodemcu-using-mqtt/121076/43 for more information.   The removal of the $ seems to fix it.


greens (approved):
Good catch!

PR #2906, opened 2018-03-13
Add more features to the virtual device creator

varzac:
This adds the ability to name the device being created, and also adds a
list of the devices created with the app.


tpmanley (approved):

varzac:
@workingmonk are you OK to merge this?

workingmonk:
@varzac can you add a ticket number so that Steven can verify this

PR #2909, opened 2018-03-14
[DEVC-594] Implement new DTH for Coolcam Neo Smart Wall Light Switch NAS-SC02ZU (contains endpoint child device DTH).

ALamchaS:
I could not use Switch Endpoind DTH, as NEO light switch endpoints do not support basic command class.

I added capablility: "Health Check" to endpoint DTH, as this child device was reported as ACTIVE even when parent device went offline.
Side effect: Now each switch will have separate status. So after going offline, pressing 1 switch will make just this one switch go back online. It will not change state of second switch.


devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
```diff
@@ -0,0 +1,164 @@
...
+	[:]
+}
+
+def zwaveEvent(physicalgraph.zwave.Command cmd) {
+	[descriptionText: "$device.displayName: $cmd", isStateChange: true]
+}
+
+def on() {
+	delayBetween([
+		zwave.switchBinaryV1.switchBinarySet(switchValue: 0xFF).format(),
```
greens: Does a switch binary set turn both lights off or just the first endpoint?

I would have expected you to need to encapsulate these commands.

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
```diff
@@ -0,0 +1,164 @@
...
+def ping() {
+	refresh()
+}
+
+def refresh() {
+	zwave.switchBinaryV1.switchBinaryGet().format()
+}
+
+// sendCommand is called by endpoint 2 child device handler
+def sendCommand(endpointDevice, commands) {
```
greens: you're not using this `endpointDevice` argument.

greens (changes requested):
a couple thoughts

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
ALamchaS: This is what configuration parameter does.

It decides which endpoints are controlled by binary switch set command sent from association group 1 (designed for controllers).

Configuration parameter 4 value:
0 - binary switch set commands from association group 1 will not control any endpoint. We would need to encapsulate to control each switch.
1 - binary switch set commands from association group 1 will control endpoint 1 (this is default setting)
2 - binary switch set commands from association group 1 will control endpoint 2
3 - binary switch set commands from association group 1 will control both endpoint 1 and endpoint 2

I decided to use default setting because it is default and it makes on() and off() commands for one endpoint faster.

I will add comments to explain this.

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
ALamchaS: Yes. I wanted binary switch endpoint to be reusable. This argument may be required for any multichannel device with more than one child device to identify calling endpoint.

I can try to add some code to make endpoint aware of its channel number, so It could pass number as argument. 

This will add some copy-pasting to future parent devices or a verbose comment explaining what needs to be done in parent dth for this to work.

ALamchaS: Last option will require type checking in any future parent device that uses more endpoint types (which is unlikely but possible).

ALamchaS: Last options would be to leave it be and add comments, add some redundant endpoint identification logic, or remove this argument, making endpoint dth dedicated to this device in the process.

greens: The dth is called `neo-coolcam-light-switch.groovy`. If you want it to be reused, you would want to change the name.

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
greens: Okay. Thanks.

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
greens: Never mind. I get it. Looks good.

greens (dismissed):

devicetypes/smartthings/neo-coolcam-light-switch.src/neo-coolcam-light-switch.groovy
```diff
@@ -0,0 +1,164 @@
...
+			state "default", label: '', action: "refresh.refresh", icon: "st.secondary.refresh"
+		}
+
+		main "switch"
+		details(["switch", "refresh"])
+	}
+}
+
+def installed() {
+	def componentLabel = "NEO Coolcam Light Switch ep2"
```
mckeed: "ep2" isn't a very user-friendly term for the label. How about just "NEO Coolcam Light Switch 2"? And maybe use ".. Switch 1" for the deviceJoinName of the main device?

mckeed:
Could we change the name of this DTH to "Z-Wave Dual Switch"? There are several devices that work like this one and I could see wanting to reuse this handler.

Also we'd want the configurationSet behind a `zwaveInfo.mfr ==` check

ALamchaS:
I can make it a bit more generic-ish.

Potential differences:

- handling of non-multichannel binary switch commands

Requires changes.

- NEO uses Switch Binary commands exclusively, other vendor may use Basic or both

If other vendor's endpoints do not support Switch Binary, adding if-else in on(), off() and refresh() in both parent and child DTH, or separate DTH would be required.

Required changes:
- Change DTH name to "Z-Wave Dual Switch"
- Change NEO Coolcam Light Switch deviceJoinName to "NEO Coolcam Light Switch 1"
- Generate child device name by replacing parent's last character ('1') with '2' (eg. "NEO Coolcam Light Switch 1" and "NEO Coolcam Light Switch 2")
- For this to work deviceJoinName for additional devices will have to be similar for additional devices, to generate 2nd switch  - some kind of comment should be enough
- Change parent's on(), off(), refresh() to use multichannel encapsulation - i imagine this will work correctly with most devices with similar design
- Remove code that sets configuration parameter 4 for NEO Light Switch, or set it to 0 (no endpoint is mapped to devce in association group 1)

greens:
I think Duncan's suggestion for the last point was to still set the configuration parameter, but check the manufacturer info first.

But the other point would be that if you just use multichannel encap for the "parent" device this would be a more portable DTH.

ALamchaS:
Yes. It will be more likely to work the same for any other similar device when we use multichannel encapsulation  for endpoint 1 (a.k.a parent).

The reason why I say we may remove configuration was not Duncan's comment regarding configuration changes. It's just that if we use multichannel encryption for parent, we do not use feature configured by parameter 4.

Just to clarify: Configuration parameter 4 decides how device handles non-encapsulated binary switch commands from hub.
And since we will encapsulate all of them, we may:
- ignore this parameter to remove redundant code,
- or set it to 0 to indicate the fact we always send multichannel encrypted Binary Switch commands.

ALamchaS:
Ok. First of the two above commits contains code changes mentioned earlier.

I tested this, and it works with the device.

All suggestions are welcome.

devicetypes/smartthings/zwave-dual-switch.src/zwace-dual-switch.groovy
```diff
@@ -0,0 +1,182 @@
...
+
+		main "switch"
+		details(["switch", "refresh"])
+	}
+}
+
+def installed() {
+	def componentLabel
+	if (device.displayName.endsWith('1')) {
+		componentLabel = device.displayName.substring(0, device.displayName.length() - 1) + '2'
```
greens: If you wanted to be particularly Groovy you could do

`"${device.displayName[0..-2]}2"`

😛

greens (dismissed):
I think this looks pretty good. @mckeed ?

ALamchaS:
I fixed my critical spelling mistake (zwace-dual-switch.groovy => zwave-dual-switch.groovy).

I also made those String operations more Groovy, as you suggested. Thanks.

greens (approved):

greens:
Uploaded UI Metadata for this to staging if you want to try it out.

I'll give @mckeed one more day and then I'll merge.

PR #2927, opened 2018-03-19
ICP-3923 GoControl Integrations

greens:


devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -30,6 +30,7 @@ metadata {
 		fingerprint mfr: "0060", prod: "0001", model: "0003", deviceJoinName: "Everspring Motion Sensor"  // Everspring HSP02
 		fingerprint mfr: "011A", prod: "0601", model: "0901", deviceJoinName: "Enerwave Motion Sensor"  // Enerwave ZWN-BPC
 		fingerprint mfr: "0063", prod: "4953", model: "3133", deviceJoinName: "GE Portable Smart Motion Sensor"
+		fingerprint mfr: "014F", prod: "2002", model: "0203", deviceJoinName: "Linear PIR Motion Sensor"
```
dkirker: So... The device supports temperature. I noticed that it works fine with the Z-Wave Plus Motion/Temp Sensor, if state.sec gets set to 0 (it defaults to secure if state.sec is undefined -- I tested with an installed() that set state.sec-0). Maybe we could not use state.sec and just read the z-wave raw data for 's' and use this fingerprint with that DTH?

cc: @mckeed

greens: Yeah that's the old method of checking security I think. It's what got us in trouble with the Fibaro Smoke Sensor.

greens:
Moved the PIR sensor and updated the security handling.

greens:
@dkirker ping

dkirker (approved):

workingmonk:
the non-fingerprint change is in zwave-plus-motion-temp-sensor.groovy

PR #2938, opened 2018-03-22
ICP-4326 Add Health Check to Z-Wave Relay

greens:
Doesn't seem to be any particular reason not to have this, other than the age of the DTH


greens:
Actually while I'm here it might be a good time to modernize the fingerprints here a little bit.

dkirker (approved):

PR #2944, opened 2018-03-23
DVCSMP-3572 Z-Wave Multilevel devices cannot express 100%

greens:
This just parses a value of 99 as 100. Because of how the Z-Wave spec is written 99=100%, 1<1%, but we were treating it as 1-to-1. This is the least painful fix. Exact mapping would with rounding would mean setting any value over 50 would map to that level + 1. Here that only happens at 99.

This will need to be accompanied by a hubcore fix for LE devices.


ALamchaS (approved):

greens:
Added a fix due to the bug in iOS that @BroderickCarlin found.

greens:
pinging @tpmanley or @mckeed

tpmanley (approved):

BroderickCarlin (approved):

varzac (approved):

PR #2946, opened 2018-03-23
ICP-4373 - Update location

juano2310:


smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -375,6 +375,14 @@ def devicesList(selector = '') {
         } else if (resp.status == 401) {
             log.warn "Access token is not valid"
             state.lifxAccessToken = null
+        } else if (resp.status == 404 && resp.data.contains('Could not find location_id')) {
+            log.warn "Location is not valid"
+            devices = devicesList()
+            devices.each { device ->
+                if (device.location.id != settings.selectedLocationId && getChildDevice(device.id)) {
+                    app.updateSetting("selectedLocationId", device.location.id)
```
bflorian: I think it would be a good idea to also modify the `settings` map here directly, i.e. `settings.selectedLocationId=device.location.id`,  because I don't think `app.updateSetting()` does that, and we'd wind up updating the setting multiple times for a location with many lights.

smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -375,6 +375,14 @@ def devicesList(selector = '') {
         } else if (resp.status == 401) {
             log.warn "Access token is not valid"
             state.lifxAccessToken = null
+        } else if (resp.status == 404 && resp.data.contains('Could not find location_id')) {
+            log.warn "Location is not valid"
+            devices = devicesList()
```
bflorian: Is there any possibility of a recursive loop here since we are calling `devicesList()` from inside `devicesList()`? Maybe we should check explicitly on the selector not being empty before making the recursive call. As it is we are depending on the response to prevent the recursive loop.

smartapps/smartthings/lifx-connect.src/lifx-connect.groovy
```diff
@@ -375,6 +375,15 @@ def devicesList(selector = '') {
         } else if (resp.status == 401) {
             log.warn "Access token is not valid"
             state.lifxAccessToken = null
+        } else if (resp.status == 404 && "$resp.data".startsWith('[error:Could not find location_id') && selector != '') {
```
bflorian: So `resp.data` in this case is a JSON object, right? Meaning that `"$resp.data"` is the same as calling `toString()` on that object? That means that if there are other fields in the response and the order of those fields ever changes the test might break. Wouldn't it be better to say `resp.data?.error == "Could not find location_id"`? It would be more resilient.

bflorian (approved):

PR #2948, opened 2018-03-24
Cycle a Switch and a DirecTV Controller

LennyZ71:
A smart app that allows a switch to be cycled on and/off based on set amount of time.
A controller that allows DirecTV genies to be controlled, also provides a remote UI.
A EndPoint for Plex_Servers.


LennyZ71:
Didn't realize pushing from IDE would create pull request. Removing this one.

PR #2963, opened 2018-03-27
WWST-202, DVCSMP-3666 Refactor Ecobee service manger

marstorp:
WWST-202 - Refactor Ecobee service manger to not use DTH polling (being depreciated)
Move device polling to be controlled by the Service Manager (Connect) app and implement
device health in Switch, Thermostat and Sensor.
Addition of event subscriptions can't be done until Ecobee releases an API that allows for that.

DVCSMP-3666 - Ecobee (Connect) not able to re-authenticate ST subscription
If, for some reason the authorization grant, token or credentials are invalid,
expired or revoked, Ecobee (connect) may still try to poll the cloud and keep trying to refresh authToken.
For a user to resolve this the ST Ecobee integration needs to re-authenticate the ST subscription in the
Ecobee cloud, however as the connect app skips the login page if authToken exits, there's no check to
see if the token is valid, there's no way for the user to re-authenticate.

Also, changed use of atomicState to state to reduce possible data corruption and added a mechanism
to stop and restart polling when user launches the SM to add/remove devices to avoid data corruption

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -15,6 +15,7 @@
  *	Author: SmartThings
  *	Date: 2013-06-13
  */
+import groovy.json.JsonOutput
 metadata {
 	definition (name: "Ecobee Thermostat", namespace: "smartthings", author: "SmartThings") {
 		capability "Actuator"
@@ -135,16 +136,29 @@ metadata {
 void installed() {
     // The device refreshes every 5 minutes by default so if we miss 2 refreshes we can consider it offline
     // Using 12 minutes because in testing, device health team found that there could be "jitter"
-    sendEvent(name: "checkInterval", value: 60 * 12, data: [protocol: "cloud"], displayed: false)
+	initialize()
+}
+def initialize() {
+	sendEvent(name: "DeviceWatch-Enroll", value: JsonOutput.toJson([protocol: "cloud", scheme:"untracked"]), displayed: false)
+	updateDataValue("EnrolledUTDH", "true")
 }
 
-// Device Watch will ping the device to proactively determine if the device has gone offline
-// If the device was online the last time we refreshed, trigger another refresh as part of the ping.
+def updated() {
+	log.debug "updated()"
+	parent.setName(device.label, device.deviceNetworkId)
+	initialize()
+}
+
+// Called when the DTH is uninstalled, is this true for cirrus/gadfly integrations?
+// Informs parent to purge its associated data
+def uninstalled() {
+    log.debug "uninstalled() parent.purgeChildDevice($device.deviceNetworkId)"
+    // purge DTH from parent
+    parent?.purgeChildDevice(this)
+}
+
 def ping() {
-    def isAlive = device.currentValue("deviceAlive") == "true" ? true : false
-    if (isAlive) {
-        refresh()
-    }
+	log.debug "ping() NOP"
 }
 
 // parse events into attributes
@@ -153,13 +167,12 @@ def parse(String description) {
 }
 
 def refresh() {
-	log.debug "refresh"
-	poll()
+	log.debug "refresh, calling parent poll"
+	parent.poll()
 }
 
 void poll() {
-	log.debug "Executing 'poll' using parent SmartApp"
-	parent.pollChild()
+	log.debug "poll not implemented as it is done by parent SmartApp every 5 minutes"
 }
 
 def generateEvent(Map results) {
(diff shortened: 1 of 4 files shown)


workingmonk:
ecobee switch needs `updated()`

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -135,16 +136,29 @@ metadata {
...
+// Informs parent to purge its associated data
+def uninstalled() {
+    log.debug "uninstalled() parent.purgeChildDevice($device.deviceNetworkId)"
+    // purge DTH from parent
+    parent?.purgeChildDevice(this)
 }
 
-// Device Watch will ping the device to proactively determine if the device has gone offline
-// If the device was online the last time we refreshed, trigger another refresh as part of the ping.
 def ping() {
```
workingmonk, line 160: why is ping() needed for a C2C device?

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
```diff
@@ -153,13 +167,12 @@ def parse(String description) {
...
-	log.debug "refresh"
-	poll()
+	log.debug "refresh, calling parent poll"
+	parent.poll()
 }
 
 void poll() {
-	log.debug "Executing 'poll' using parent SmartApp"
-	parent.pollChild()
+	log.debug "poll not implemented as it is done by parent SmartApp every 5 minutes"
```
rohandesai, line 175: should we remove this method since it is not being called anywhere?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -131,20 +188,46 @@ def callback() {
...
+			state.refreshToken = resp.data.refresh_token
+			state.authToken = resp.data.access_token
 		}
-
-		if (atomicState.authToken) {
+		if ( state.authToken ) {
+			// get jwt for switch+ devices
+			obtainJsonToken()
+		}
+		if (state.authToken) {
```
rohandesai, line 204: Is there a reason we are doing the two `state.authToken` checks separately?

rohandesai (commented):
Mostly LGTM except a couple minor comments.

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: it was when device was enrolled as tracked, I left it in as last I removed a command we ended up with missing method exceptions.

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: unfortunately this method may be called by SA and routines as once upon a time the polling capability was defined for this DTH. removing this command will result in missing method exceptions if there are existing SA and routines calling it.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: You'll have to ask Todd about that as he added this for the addition of the early access switch API.

rohandesai (approved):
LGTM 👍

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
workingmonk: can you call updated() from ping() that way these devices can get enrolled properly. also, update the log statement to log.info "Ecobee ping() incorrectly called" or something. log.info will put it to sumo

rohandesai:
New changes look good for app settings. 👍

devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
marstorp: only time ping should be called is if the device is enrolled as tracked and there hasn't been any events sent from the thermostat for 12 minutes. As the thermostat now enrolls as untracked on the first poll response I don't se why we should add another call to updated when ping is called.

rgangarapu1 (commented):
I have tested in Staging with Staging Client Id and Staging serverUrl and its working fine. nothing is broken

rgangarapu1 (commented):
I have tested with Na04 with production account and everything works fine. I can see th status change when the ecobee is disconnected.

rohandesai:
New changes look good for the ecobee-switch 👍

PR #2964, opened 2018-03-28
 [DEVC-597] Add fingerprint for Bosh Radion ZigBee Motion Detector RFDL-ZB-MS to generic Smart Sense Motion Detector DTH.

ALamchaS:
Device seems to work correctly using generic DTH.

According to printed instruction sheet, device reports movement using red LED only once every 3 minutes to preserve battery.

I was quite surprised to find out the same holds true for motion events sent via zigbee (this is mentioned in instruction sheet too).

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -175,7 +175,13 @@ private Map getBatteryResult(rawValue) {
 				volts = maxVolts
 			def pct = batteryMap[volts]
 			result.value = pct
-		} else {
+		} else if (device.getDataValue("manufacturer") == "Bosch") {
+			def minValue = 21
+			def maxValue = 30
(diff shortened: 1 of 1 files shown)


greens (changes requested):
Can you double-check the battery reporting code? I think it's pretty non-standard and I'd be surprised if it still worked as expected.

ALamchaS:
Ok. So it seems Bosch reports raw value 30 for 100%, so I should probably make a manufacturer check to use "old" battery reporting.

greens (approved):

greens:
@tpmanley: @ALamchaS 's changes in his last commit around battery reporting look correct to me, but can you double-check that they make sense zigbee-wise?

Olek's comment from the ticket:

> It uses 1 or 2 parallel CR123 batteries.
> That means they should have about 3V when fully charged, and 2V when empty. This seems consistent with raw value 30 report that we get for power configuration cluster attribute 0x0020 when using fresh batteries.

tpmanley (approved):

PR #2969, opened 2018-03-30
Initial heaths.Timer app

heaths:
Simple application to turn a light switch (or any switch) off after a certain amount of time has elapsed since it was turned on. You could use this for a porch light after normal hours, a scented wax "candle", and more.

Icon source is co-located here for ease (all related source), but the PNGs themselves are hosted on my own GitHub account.

### DIFF
--- a/smartapps/heaths/timer.src/timer.groovy
+++ b/smartapps/heaths/timer.src/timer.groovy
@@ -1,0 +1,63 @@
+/**
+ *  Timer
+ *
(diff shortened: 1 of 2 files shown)


heaths:
Closing since 1) this will never be merged since SmartThings v2 community support from Samsung seems to be dead (if it ever really existed), and 2) a recent update to SmartThings adds auto-off to most (all?) devices.

PR #2978, opened 2018-04-03
[DEVC-596] Add implementation for Bosh Radion ZigBee Multisensor RFMS-ZBMS-3 to generic ZigBee open close sensor DTH.

ALamchaS:
DTH works correctly for both contact and water detector modes. I tested it with both cloud and local execution.

There are problems with Samsung SmartThings apk for water detector mode only.

Are there any known issues with OneApp not supporting dynamic DTH changes or with SmartSense Moisture Sensor Metadata?

### DIFF
--- a/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
@@ -27,12 +27,12 @@ metadata {
 
 		command "enrollResponse"
 
-
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3300-S"
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3300"
 		fingerprint inClusters: "0000,0001,0003,0020,0402,0500,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3320-L", deviceJoinName: "Iris Contact Sensor"
 		fingerprint inClusters: "0000,0001,0003,0020,0402,0500,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3323-G", deviceJoinName: "Centralite Micro Door Sensor"
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "Visonic", model: "MCT-340 E", deviceJoinName: "Visonic Door/Window Sensor"
+		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "Bosch", model: "RFMS-ZBMS", deviceJoinName: "Bosch multi-sensor"
 	}
 
 	simulator {
@@ -97,6 +97,14 @@ def parse(String description) {
 					sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 				} else {
 					log.warn "TEMP REPORTING CONFIG FAILED- error code: ${descMap.data[0]}"
+				}
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0001 && device.getDataValue("model") == "RFMS-ZBMS") {
+				if (Integer.parseInt(descMap.value, 16) == 0x0015) {
+					//multi-sensor is in contact or tilt detector mode - both act as contact sensor so no action is necessary
+				} else if (Integer.parseInt(descMap.value, 16) == 0x002A) {
+					//multi-sensor is in water detector mode, DTH should be changed to water sensor DTH
+					log.debug "SmartSense Moisture Sensor"
+					setDeviceType("SmartSense Moisture Sensor")
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


greens:
@ALamchaS the one thing I can think of with switching the device type like this would be that `installed` might not get called, but this is uncharted territory for me.

I don't have any idea how well OneApp handles it. What was the problem you had in water detector mode?

ALamchaS:
OneApp "Devices" tab:
- Status is stuck at "Checking status..."
- Icon is the same as open/close sensor

Bosch multi-sensor device page:
- Main object is blue circle with measured temperature.
- Other available values are "Battery" and "Connected to $hubName Hub"
- There is no water sensor value at all on device page.

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -27,12 +27,12 @@ metadata {
 
 		command "enrollResponse"
 
-
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3300-S"
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3300"
 		fingerprint inClusters: "0000,0001,0003,0020,0402,0500,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3320-L", deviceJoinName: "Iris Contact Sensor"
 		fingerprint inClusters: "0000,0001,0003,0020,0402,0500,0B05", outClusters: "0019", manufacturer: "CentraLite", model: "3323-G", deviceJoinName: "Centralite Micro Door Sensor"
 		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "Visonic", model: "MCT-340 E", deviceJoinName: "Visonic Door/Window Sensor"
+		fingerprint inClusters: "0000,0001,0003,0402,0500,0020,0B05", outClusters: "0019", manufacturer: "Bosch", model: "RFMS-ZBMS", deviceJoinName: "Bosch multi-sensor"
```
tpmanley, line 35: Is `Bosch multi-sensor` the correct join name? Typically all the words are capitalized.

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -182,5 +190,10 @@ def configure() {
 
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
+	def commands = refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300)
+	// send refresh cmds as part of config
```
tpmanley, line 194: This comment should be above the line above since it's referring to the call to `refresh()`

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -98,6 +98,14 @@ def parse(String description) {
 				} else {
 					log.warn "TEMP REPORTING CONFIG FAILED- error code: ${descMap.data[0]}"
 				}
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0001 && device.getDataValue("model") == "RFMS-ZBMS") {
```
tpmanley, line 101: s/`0x0500`/`zigbee.IAS_ZONE_CLUSTER`/

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -98,6 +98,14 @@ def parse(String description) {
...
+				if (Integer.parseInt(descMap.value, 16) == 0x0015) {
+					//multi-sensor is in contact or tilt detector mode - both act as contact sensor so no action is necessary
+				} else if (Integer.parseInt(descMap.value, 16) == 0x002A) {
+					//multi-sensor is in water detector mode, DTH should be changed to water sensor DTH
+					log.debug "SmartSense Moisture Sensor"
+					setDeviceType("SmartSense Moisture Sensor")
```
tpmanley, line 107: My understanding is when the device type is changed only `updated` will be called and not `configure`. That looks like it should be OK because `configure` does the same thing in both DTHs.

greens: @ALamchaS This explains why you have "Checking Status" when you switch the device type. The initial poll for device state is in `configure`. If you switch the device type then that won't happen. Maybe add an `updated` method to both DTHs that calls `refresh` and take the `refresh` call out of `configure`?

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
ALamchaS: This is how it is written on the box. But it does not look like production-ready package, so I will change that.

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
ALamchaS: This state persists even after I trigger wet/dry events, so it is not the reason for this problem.

I will try making water sensor the main DTH today. I will see if this situation gets inversed.
If it does, this is caused by dynamic DTH change,
if not - something may be wrong with this water sensor UI metadata.

Thanks for making me think of that.

ALamchaS: Ok. So after changing DTH situation is: water sensor works OK. contact sensor does not display correctly in OneApp.

greens: @tpmanley Do you think there might be some wrinkles here due to local execution?

PR #2981, opened 2018-04-03
WWST-207 HOTFIX Ecobee JWT auth errors

marstorp:
sendCommandToEcobee() is calling poll and if poll causes child to call sendCommandToEcobee
this may lead to recursive calls.

Refactor sendCommandToEcobee, removing poll and instead try to send
the command one more time after refresh of authToken.

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -1072,7 +1072,7 @@ private boolean sendCommandToEcobee(Map bodyParams) {
 		body: toJson(bodyParams)
 	]
 	def keepTrying = true
-	cmdAttempt = 1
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -1071,34 +1071,44 @@ private boolean sendCommandToEcobee(Map bodyParams) {
 		headers: ["Content-Type": "application/json", "Authorization": "Bearer ${state.authToken}"],
 		body: toJson(bodyParams)
 	]
+	def keepTrying = true
+	cmdAttempt = 1
```
bflorian: Needs a `def`

rohandesai:
Code changes look good to me.

bflorian (approved):

PR #2984, opened 2018-04-03
ICP-4139 Aeon Multisensor not getting initial state when added on bat…

greens:
…tery

- Removes cruft around old 'powerSupply' implementation that should have been removed
- Updates 'batteryStatus' when power source is updated, since battery is parsed locally now
- Updates 'batteryStatus' whenever we receive a command and the device is on battery power. This should be done locally, but it's a weird hybrid attribute so that seems unlikely.
- Runs configure despite power supply on the initial 'updated()' call, since even sleepy devices will still be awake and we need to get that initial state

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
@@ -142,29 +142,27 @@ def installed(){
 
 def updated() {
 	log.debug "Updated with settings: ${settings}"
-	log.debug "${device.displayName} is now ${device.latestValue("powerSource")}"
 
-
-	def powerSource = device.latestValue("powerSource")
-
-	if (!powerSource) { // Check to see if we have updated to new powerSource attr
-		def powerSupply = device.latestValue("powerSupply")
-
-		if (powerSupply) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -142,29 +142,27 @@ def installed(){
 
 def updated() {
 	log.debug "Updated with settings: ${settings}"
-	log.debug "${device.displayName} is now ${device.latestValue("powerSource")}"
 
-
-	def powerSource = device.latestValue("powerSource")
-
-	if (!powerSource) { // Check to see if we have updated to new powerSource attr
```
greens, line 150: I know this was meant as migration code, but everyone's device has probably transitioned over by now.

dkirker: I'll defer to @mckeed  on this. It is probably fine now.

dkirker:
It looks good, but I am going to defer to @mckeed on the old powerSupply -> powerSource code, since I added it per his recommendation.

greens:
Deferring to the changes under https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3010

PR #2858, opened 2018-03-05
[DVCSMP-3542] Onboarding Aeotec Nano Dimmer & Nano Switch

sagar-trility:

PR #2985, opened 2018-04-04
 [ICP-4513] Add fingerprint for Aeotec Labs Nano Switch (ZW139-C) to Z-wave Switch Secure DTH.

ALamchaS:
Please note that this DTH has no metadata.

### DIFF
--- a/devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
+++ b/devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
@@ -12,7 +12,7 @@
  *
  */
 metadata {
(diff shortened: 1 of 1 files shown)


greens:
I think we probably just want this fingerprint to go in with https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2858/

ALamchaS:
This is different nano switch. It does not support Command Class Meter.

That is why I added it to secure switch and not metering switch.

My first impression was the same as yours, but different prod number (0003 instead of 0103) and the fact it paired as simple z-wave switch instead of metering switch (which contains fingerprint inClusters: "0x25,0x32") made me investigate this subject.

greens:
Good catch. Check out the device in ICP-4536 as well.

greens (approved):

PR #2992, opened 2018-04-10
PROB-2053 Ecobee Integration Failing for Some Users

marstorp:
When Ecobee (connect) receives HttpResponseException with status 400 a "java.io.IOException: Stream closed" is thrown while accessing the error object. As a workaround a new string object is created using toString() on the response data.

Also changing refresh of authToken to only request refresh of the current used and to handle the new token when the method returns rather than have authToken call the method again, which can cause recursion.

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -20,7 +20,7 @@
  *      JLH - 02-15-2014 - Fuller use of ecobee API
  *      10-28-2015 DVCSMP-604 - accessory sensor, DVCSMP-1174, DVCSMP-1111 - not respond to routines
  */
-
+import groovy.json.JsonSlurper
 include 'localization'
 
 definition(
@@ -80,6 +80,7 @@ def authPage() {
 
 	if(state.authToken) {
 		if(!state.jwt) {
+			state.jwt = true
 			refreshAuthToken()
 		}
 		description = "You are connected."
@@ -199,7 +200,8 @@ def callback() {
 		}
 		if ( state.authToken ) {
 			// get jwt for switch+ devices
-			obtainJsonToken()
+			state.jwt = true
+			refreshAuthToken()
 		}
 		if (state.authToken) {
 			success()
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -556,66 +546,87 @@ def poll() {
...
-						storeThermostatData(resp.data.thermostatList)
+				httpGet(pollParams) { resp ->
+					isThermostatPolled = true
+					if(resp.status == 200) {
+						if (thermostats || ecobeesensors) {
+							storeThermostatData(resp.data.thermostatList)
+						}
+						if (ecobeesensors) {
+							updateSensorData(resp.data.thermostatList.remoteSensors)
+						}
```
ron2911, line 588: should we add an `else` for other `resp.status` codes like we have for switches below?

marstorp: Other than 200 there shouldn't really be any other successful response statuses. If there wasn't any data updated during the poll, the health check coming in next PR will count this loss of data towards the 3 attempts before marking the device offline.
As log.warn only shows up in live logs they don't give that much information in prod.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -812,9 +825,10 @@ def toQueryString(Map m) {
 	return m.collect { k, v -> "${k}=${URLEncoder.encode(v.toString())}" }.sort().join("&")
 }
 
-private refreshAuthToken() {
+boolean refreshAuthToken() {
 	log.debug "refreshing auth token"
 	def notificationMessage = "is disconnected from SmartThings, because the access credential changed or was lost. Please go to the Ecobee (Connect) SmartApp and re-enter your account login credentials."
+	def status = false
```
ron2911: should we rename this to `isSuccess` or something more relevant?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -542,9 +527,14 @@ def purgeChildDevice(childDevice) {
 		log.info "No more thermostats to poll, unscheduling"
 		unschedule()
 		state.authToken = null
+		runIn(1, "killMe")
 	}
 }
 
+def killMe() {
+	app.delete()
```
ron2911: should we add `try {} catch {}` here just in case?

ron2911 (approved):
LGTM 👍

PR #2997, opened 2018-04-10
Fixing filename and vid

workingmonk:


greens (approved):
thanks

@ALamchaS Something we both need to be aware of in the future: The DTH name field and the filename need to match.

workingmonk:
namespace is the first level folder name. .src and .groovy name needs to match groovy the name in the definition

ALamchaS:
Good to know. Thanks.

PR #3002, opened 2018-04-11
PROB-2053 Ecobee Integration Failing for Some Users

marstorp:
Fixing null pointer exception that was introduced in PR #2992
in the purgeChildDevice() method.


ron2911 (approved):
👍

PR #3005, opened 2018-04-12
DVCSMP-3749 Ecobee device health does not work if poll fails

marstorp:
Also added cleanup of non installed device data
Changed update of thermostat tile to not have to poll the cloud

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -125,10 +125,24 @@ def restartPoll() {
 }
 
 def ecobeeDeviceList() {
-	def thermostatList = getEcobeeDevices()
+	getEcobeeDevices()
 
-	def p = dynamicPage(name: "deviceList", title: "Select Your ecobee Devices", uninstall: true) {
-		def numThermostats = thermostatList.size()
+	def thermostatList = thermostatsDiscovered()
+	def numThermostats = thermostatList.size()
+	def sensors = sensorsDiscovered()
+	def numSensors = sensors.size()
+	def switches = switchesDiscovered()
+	def numSwitches = switches.size()
+
+	if (!numThermostats && !numSensors && !numSwitches) {
+		return dynamicPage(name: "deviceList", title: "No devices found", uninstall: true) {
+			section ("") {
+				paragraph "Could not find any devices avilable for SmartThings to control. Please check your ecobee account and retry later."
+			}
+		}
+	}
+
+	return dynamicPage(name: "deviceList", title: "Select Your ecobee Devices", uninstall: true) {
 		if (numThermostats > 0)  {
 			def preselectedThermostats = thermostatList.collect{it.key}
 			section("") {
@@ -138,8 +152,6 @@ def ecobeeDeviceList() {
 						description: "Tap to choose", metadata:[values:thermostatList], defaultValue: preselectedThermostats)
 			}
 		}
-		def sensors = sensorsDiscovered()
-		def numSensors = sensors.size()
 		if (numSensors > 0)  {
 			def preselectedSensors = sensors.collect{it.key}
 			section("") {
@@ -148,8 +160,6 @@ def ecobeeDeviceList() {
 						type: "enum", required:false, description: "Tap to choose", multiple:true, options:sensors, defaultValue: preselectedSensors)
 			}
 		}
-		def switches = switchesDiscovered()
-		def numSwitches = switches.size()
 		if (numSwitches > 0)  {
 			def preselectedSwitches = switches.collect{it.key}
 			section("") {
@@ -159,7 +169,6 @@ def ecobeeDeviceList() {
 			}
 		}
 	}
-	return p
 }
 
 def oauthInitUrl() {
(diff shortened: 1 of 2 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -307,71 +316,101 @@ def getEcobeeDevices() {
...
-						includeRuntime: true,
-						includeSensors: true
+	def isThermostatPolled = false
+	def isSwitchesPolled = false
+	def pollAttempt = 1
+	// try obtain devices twice, in case authToken needs to be refreshed
+	while (!(isThermostatPolled && isSwitchesPolled) && (pollAttempt < 3)) {
+		try {
+			// First get thermostats their remote sensors
+			if (!isThermostatPolled) {
```
ron2911, line 326: do we need to check this boolean again here? we already have a check above in the while loop.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -307,71 +316,101 @@ def getEcobeeDevices() {
...
-				headers: ["Content-Type": "application/json;charset=UTF-8", "Authorization": "Bearer ${state.authToken}"],
-		]
-		httpGet(switchListParams) { resp ->
-			if (resp.status == 200) {
-				resp.data?.devices?.each {
-					if (it.type == "LIGHT_SWITCH") {
-						switchList[it?.identifier] = it
-						switchList[it?.identifier] << [deviceAlive: (it?.connected ?: false)]
+			// Now get light swiches
+			if (!isSwitchesPolled) {
```
ron2911, line 370: do we need to check this boolean again here? we already have a check above in the while loop.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -307,71 +316,101 @@ def getEcobeeDevices() {
...
+						path:    "/1/thermostat",
+						headers: ["Content-Type": "text/json", "Authorization": "Bearer ${state.authToken}"],
+						// TODO - the query string below is not consistent with the Ecobee docs:
+						// https://www.ecobee.com/home/developer/api/documentation/v1/operations/get-thermostats.shtml
+						query: [format: 'json', body: toJson(bodyParams)]
+				]
+				httpGet(deviceListParams) { resp ->
+					isThermostatPolled = true
+					if (resp.status == 200) {
+						resp.data.thermostatList.each { stat ->
```
ron2911, line 347: should we add the null safety operator? `resp.data?.thermostatList?.each`

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -422,10 +461,11 @@ def initialize() {
 	def childThermostats = thermostats.collect { dni ->
 		def d = getChildDevice(dni)
 		if(!d) {
-			d = addChildDevice(app.namespace, getChildName(), dni, null, ["label":"${thermostatList[dni]}" ?: getChildName()])
+			d = addChildDevice(app.namespace, getChildName(), dni, null, ["label":"${thermostatList[dni].data.name}" ?: getChildName()])
```
ron2911, line 464: should we add a null safety check for `thermostatList[dni]?.data?.name` just in case?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -494,6 +534,21 @@ def initialize() {
...
+	// purge state from devices that are not selected
+	def thermostatList = state.thermostats ?: [:]
+	def remoteSensors = state.remoteSensors2 ?: [:]
+	def switchList = state.switchList ?: [:]
+
+	// clean up device lists
+	thermostatList.keySet().removeAll(thermostatList.keySet() - thermostats)
+	remoteSensors.keySet().removeAll(remoteSensors.keySet() - ecobeesensors)
+	switchList.keySet().removeAll(switchList.keySet() - ecobeeswitches)
+	state.thermostats = thermostatList
```
ron2911, line 547: nit: Can we just do `state.thermostats = thermostatList.keySet().removeAll(thermostatList.keySet() - thermostats)` over here?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -554,6 +610,27 @@ def poll() {
...
+		stat.polled = false
+		stat.data = stat.data ? stat.data << [deviceAlive:false] : [deviceAlive:false]
+	}
+	state.thermostats = thermostatList
+	switchList.each { sdni, sw ->
+		sw.deviceAlive = false
+		sw.polled = false
+	}
+	state.remoteSensors2 = remoteSensors
+	state.thermostats = thermostatList
```
ron2911, line 630: we are setting `state.thermostats` two times here.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -629,15 +704,57 @@ def poll() {
...
+			}
+		}?.keySet()
+		def switchesOffline = switchList.findAll { sdni, sw ->
+			if ((sw.deviceAlive == false) && (sw.polled == false)) {
+				sw.pollAttempts = sw.pollAttempts ? sw.pollAttempts + 1 : 1
+				if (sw.pollAttempts > 2) {
+					return sdni
+				}
+			}
+		}?.keySet()
```
ron2911, line 746: It seems like all of the above three are doing the same thing. Should we move this out into a helper function?

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -671,7 +788,8 @@ void controlSwitch( dni, desiredState ) {
 			try {
 				httpPut(params) { resp ->
 					keepTrying = false
-					log.debug "RESPONSE CODE: ${resp.status}"
+					def rspDataString = "${e.resp?.data}".toString()
```
ron2911: do we have access to `e` here? also `rspDataString` is not being used anywhere in this success block.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: I do this in case this poll was ok but the switch poll fails in order to avoid polling thermostats a second time when we only need to poll the switches

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: If we get a 200 ok on the request then the response should have data with a thermostatList, if not there's something seriously wrong that we might want to discover sooner rather than later.

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: This check might be unecessary as the logic dictates we should at least have been able to poll thermostats before getting here, but it won't prevent the switches to be polled either...

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: name comes from getThermostatDisplayName(data) where data is resp.data.thermostatList, if name isn't present in that data it will default to (stat.modelNumber == "siSmart" ? "Smart Si" : "Smart") + " (${stat.identifier})").toString(), so there should always be a name present, if the name still is empty it will default to "Ecobee Thermostat"

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: Maybe, if you trust the state assignment, nothing I've tested though

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: two are doing pretty much the same thing, the third does not, I didn't feel like sorting out the helper method for two things when the third was different 😪 Leaving that for another day....

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
marstorp: copy/paste issue but the string is used in the log

smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
ron2911: 👍

ron2911 (approved):
LGTM 👍

PR #3009, opened 2018-04-13
[ICP-4628] Reduce ramp time for the device to work correctly using local execution.

ALamchaS:
@MAblewiczS @KKlimczukS @MMateusiakS


ALamchaS:
If this works, additional configurationGet calls need to be removed - those were added for debugging.

greens (approved):
In the future, I think you can use the `scaledConfigurationValue` with a decimal number to auto-pad your values to fit the size.

PR #3015, opened 2018-04-16
ICP-4668 Add join names for already certified Leviton devices

greens:
Adding join name for previously-certified Leviton Dimmer Switch


greens:
@workingmonk This adds fingerprints for previously-certified Leviton devices that were certified prior to the requirement of a join name for certification. The goal is to have these fingerprints in production tomorrow after being put in staging today.

PR #3016, opened 2018-04-16
ICP-4587 Change GE Plug-in Outdoor Switch fingerprint to production m…

greens:
…odel


greens:
@workingmonk same rationale as https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3015

PR #3020, opened 2018-04-18
[ICP-4701] Add initial state check.

ALamchaS:


devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -233,10 +233,17 @@ def zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerS
...
 	if (msr == "011A-0601-0901") {  // Enerwave motion doesn't always get the associationSet that the hub sends on join
 		result << response(zwave.associationV1.associationSet(groupingIdentifier: 1, nodeId: zwaveHubNodeId))
-	} else if (!device.currentState("battery")) {
-		if (msr == "0086-0102-0059") {
-			result << response(zwave.securityV1.securityMessageEncapsulation().encapsulate(zwave.batteryV1.batteryGet()).format())
-		} else {
+	} else if (msr == "0086-0102-0059") {
+		def cmds = []
+		if (!device.currentState("battery")) {
+			cmds << zwave.securityV1.securityMessageEncapsulation().encapsulate(zwave.batteryV1.batteryGet()).format()
```
dkirker: Would it be better to update the command method to use the new way to detect secure (see https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/b71b3fad4b8caff7841cbc4c379b74354edc1db0/devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy#L267) and use this here?

I would say to drop the specific check for the one device and just do the state checks for battery and contact. Might it be better to just do what we do in zwave-motion-sensor with initialPoll or also add a battery check in to configure()? I know there are some instances where sometimes the mobile client doesn't send the proper command to call updated() so I wonder if there are edge cases where this may never get executed?

dkirker:
Other than maybe doing some things a little differently, this does look good. Let me know if you have any thoughts to my ideas above.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
ALamchaS: I will try first two suggestions - it should be better to generalize this code a little. I am a bit afraid it can break this DTH for other devices that use it, as we do not have all the devices to check it.

Initial poll would not fix issue ICP-4701 by itself, as we already send something similar in configure(), and apparently do not get any response (although this may be caused by wakeUpNoMoreInformation from updated arriving before configure() is called), but I agree updated() seems a bit weird right now.

Now that I think about it, this code should not be in updated() in the first place, as those devices are most likely sleepy, so when we save preferences in Classic, those commands will most likely be sent while the device is asleep. So installed, which is called once during inclusion seems more logical choice.

I think adding battery check to configure could be good, if not for this:
[createEvent(map), response(zwave.wakeUpV1.wakeUpNoMoreInformation())]
in response to battery report. So this 20s delay before wakeUpNoMoreInformation should be better. I will have to check if this is used somewhere else to end communication.

Thanks for input.

ALamchaS: I have a question. This device seems to handle wakeUpNoMoreInformation command properly only if it was security encapsulated. Should I make a modification like this for all secure devices here, or should I make a check if cc 84 is mentioned in zwInfo.sec first?

dkirker: I think if you replace the `if (state.sec == 1)` with `if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo.zw?.endsWIth("s"))` in command() and use either command() or commands() that would probably be safe.

dkirker: @mckeed  Might also have some input.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -187,24 +172,26 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
...
-		cmds << command(zwave.batteryV1.batteryGet())
-	} else {
-		// If we check the battery state we will send NoMoreInfo in the handler for BatteryReport so that we definitely get the report
-		cmds << zwave.wakeUpV1.wakeUpNoMoreInformation().format()
+		request << zwave.batteryV1.batteryGet()
+	}
+
+	def response = response(command(zwave.wakeUpV1.wakeUpNoMoreInformation()))
+	if (request.size() > 0) {
+		response = new physicalgraph.device.HubMultiAction(commands(request, 1000) + ["delay 2000", command(zwave.wakeUpV1.wakeUpNoMoreInformation())])
```
ALamchaS: For some reason, using request() instead of new HubMultiAction causes errors like:
groovy.lang.MissingMethodException: No signature of method: physicalgraph.device.HubAction.call() is applicable for argument types: (java.util.ArrayList) values: [[9881007204, delay 2000, 9881008408]]
Possible solutions: wait(), any(), wait(long), any(groovy.lang.Closure), each(groovy.lang.Closure), collect()

Probably some problems with dynamic typing.

ALamchaS (commented):
I encountered one problem during development.

@mckeed @greens @dkirker , any ideas or changes required?

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
greens: Are you saying that:

`response( commands(request,1000) + ["delay 2000", command(zwave.wakeUpV1.wakeUpNoMoreInformation())] )`

was throwing that error?

I think it's because you took out the `command()` calls which outputs a string. Another subtlety in the original code is that `noMoreInformation` should always be unencapsulated. And your code relies on the device to respond with the battery status within two seconds or else it might have unexpected behavior with the noMoreInformation.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -76,7 +76,7 @@ private getCommandClassVersions() {
 def parse(String description) {
 	def result = null
 	if (description.startsWith("Err 106")) {
-		if (state.sec) {
+		if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo?.zw?.endWith("s")) {
```
dkirker: endWith should be endsWith

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
dkirker: response supports the following:

response(String action)
response(physicalgraph.zwave.Command cmd)
response(List actions)

Anyway, command will encapsulate any command if possible. If it is something that you don't want encapsulated, then you wouldn't wrap it with command (or commands), but just append .format(). A good example of the mixed use would be updated() and refresh() in zwave-lock.src/zwave-lock.groovy (replacing secure with command for use here). Then you should be able to take that array or cmds and pass it to response() and then just return that.

So maybe:

```
def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd) {
	def event = createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)
	def cmds = []
	if (!state.MSR) {
		cmds << zwave.manufacturerSpecificV2.manufacturerSpecificGet()
	}

	if (device.currentValue("contact") == null) {
		// In case our initial request didn't make it, initial state check no. 3
		cmds << zwave.sensorBinaryV2.sensorBinaryGet(sensorType: zwave.sensorBinaryV2.SENSOR_TYPE_DOOR_WINDOW)
	}

	if (!state.lastbat || now() - state.lastbat > 53 * 60 * 60 * 1000) {
		cmds << zwave.batteryV1.batteryGet()
	}

	def request = []
	if (cmds.size() > 0) {
		request = commands(cmds, 1000)
		request << "delay 2000"
	}
	request << zwave.wakeUpV1.wakeUpNoMoreInformation().format()

	[event, response(request)]
}
```

dkirker (changes requested):

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
ALamchaS: Thanks. Good catch.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
ALamchaS: @greens 
> Are you saying that:
>
> response( commands(request,1000) + ["delay 2000", command(zwave.wakeUpV1.wakeUpNoMoreInformation())] )
>
> was throwing that error?

Yes.

> I think it's because you took out the command() calls which outputs a string

They are still there. In commands(...) method.

> private commands(commands, delay = 200) {
> 	delayBetween(commands.collect { command(it) }, delay)
> }

> noMoreInformation should always be unencapsulated.

During development this device did not turn the red led off when it was not encapsulated. Manual states it will turn off 2 seconds after receiving noMoreInformation. I will test it without encapsulation 10-20 times more and put it behind mfr/prod/model check or remove it based on results.

> And your code relies on the device to respond with the battery status within two seconds or else it
> might have unexpected behaviour with the noMoreInformation.

I do not think it is a good idea to send it only after we get battery report. Given the amount of "no battery/state after pairing" issues, battery report seems far from guaranteed. And we should always put the device to sleep. Should I make it 20000ms like some other devices?

ALamchaS: You guys are right, non-encrypted noMoreInformation works well.

@dkirker Thanks for help, your code does not throw any errors.

greens: Reasonable point about the `nomoreinformation` after battery report. We'll see how it goes.

greens (approved):

dkirker (approved):

PR #3036, opened 2018-04-26
DVCSMP-3852 Ecobee NullPointerException prevents complete poll

marstorp:
Users with mismatching settings/application state data for sensors, settings have less
sensors than what application state data indicates and less sensors than installed
sensors DTHs are getting NullPointerException during poll preventing a complete
poll and refresh of devices.

Adding check before updating sensor data, if not selected update the child sensor
with data field "DeviceIssue" to request user to remove and re-add it.

Also replacing the check for sensor type `!= "thermostat" to `== "ecobee3_remote_sensor"`
to avoid adding non temperature sensors as remote sensors.

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -352,7 +352,7 @@ def getEcobeeDevices() {
 							thermostatList[dni].pollAttempts = 0
 							// compile all remote sensors conected to the thermostat
 							stat.remoteSensors.each { sensor ->
-								if (sensor.type != "thermostat") {
+								if (sensor.type == "ecobee3_remote_sensor") {
 									def rsDni = "ecobee_sensor-"+ sensor?.id + "-" + sensor?.code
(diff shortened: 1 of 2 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -907,7 +907,7 @@ def updateSensorData(sensorData) {
 						// is preserved and not changed to name from ecobee cloud as this is the first name change is allowed
 						child.setDisplayName(it.name)
 					}
-					if (!remoteSensors[dni] || remoteSensors[dni].deviceAlive) {
+					if (remoteSensors[dni] && remoteSensors[dni].deviceAlive) {
```
workingmonk, line 910: you inverted the logic here, right?
before it would have entered the block if it was a new device `!remoteSensor[dni]`, but now it won't

marstorp: Yes as we don't want to add the data for a non selected device even if there's a loitering child hanging around.

rohandesai (approved):
👍

PR #3045, opened 2018-04-27
PROB-1507 Fix bug in Color Coordinator with setting color

MichaelStruck:
Updated based on user feedback. Removed the temp light change in the HSL area.


vl4ds:
👍

MichaelStruck:
Updated with adding pause back in

tpmanley (approved):
I tested out the change and confirmed it fixed the issues where the slave lights would flash a little and not follow the color of the master light.

Issue #3046, opened 2018-04-28
Lock status wrong - Yale Assure Lock SL(YRD256-ZW2-619)


gonella:
Hi,

I have Yale Assure Lock SL(YRD256-ZW2-619) with zwave module installed, see link below. Lock is sending wrong status to smarthings, even it is physically connected it is saying unlocked. Looking the groovy file available, I can't see entry for my lock. Is there any workaround? 

The lock is installed by name "Yale Assure Lock", but it seems it supports only YRD416, YRD426, YRD446, my lock is YRD256.

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy

https://www.amazon.com/gp/product/B076WZT4VD/ref=oh_aui_detailpage_o04_s00?ie=UTF8&psc=1

workingmonk:
@gonella can you check the issue again EOD Tuesday? The PR that you referenced will be deployed then. In the short term, you can copy the DTH into your account and test it manually to see if it fixes your issue

gonella:
Hi @workingmonk I tried, but it didn't work. The problem still the same, even is physically locked, it is reporting unlocked. Every time I locked the door by the app, a few ms 1-2 seconds later start reporting unlocked. Any idea?

workingmonk:
cc @dkirker @mckeed

PR #3047, opened 2018-04-28
ICP-4838 Add a timed initial state check to Z-Wave locks

dkirker:
Verify that we have received the initial state of the lock 2 minutes after it has been installed and at 2 minute intervals until we have received the initial battery and lock states.

Sometimes the user can get in to a situation where a lock may not receive packets requesting the initial states, or we never send the request for the initial states.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -102,6 +102,34 @@ import physicalgraph.zwave.commands.usercodev1.*
 def installed() {
 	// Device-Watch pings if no device events received for 1 hour (checkInterval)
 	sendEvent(name: "checkInterval", value: 1 * 60 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+
+	scheduleInstalledCheck()
+}
+
+/**
+ * Verify that we have actually received the lock's initial states.
+ * If not, verify that we have at least requested them or request them,
+ * and check again.
+ */
+def scheduleInstalledCheck() {
+	runIn(120, installedCheck, [forceForLocallyExecuting: true])
+}
+
+def installedCheck() {
+	if (device.currentState("lock") && device.currentState("battery")) {
+		unschedule("installedCheck")
+	} else {
+		// We might have called updated() or configure() at some point but not have received a reply, so don't flood the network
+		if (!state.lastLockDetailsQuery || secondsPast(state.lastLockDetailsQuery, 2 * 60)) {
+			def actions = updated()
+
+			if (actions) {
+				sendHubCommand(actions.toHubAction())
+			}
+		}
+
+		scheduleInstalledCheck()
+	}
 }
 
 /**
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -102,6 +102,30 @@ import physicalgraph.zwave.commands.usercodev1.*
...
+		// We might have called updated() or configure() at some point but not have received a reply, so don't flood the network
+		if (!state.lastLockDetailsQuery || secondsPast(state.lastLockDetailsQuery, 2 * 60)) {
+			def actions = updated()
+
+			if (actions) {
+				sendHubCommand(actions.toHubAction())
+			}
+		}
+
+		runIn(120, installedCheck)
```
tpmanley: Did you test with the DTH running locally? I believe you need to add a third argument `[forceForLocallyExecuting: true]` to get this to run if the DTH is running locally.

dkirker: @tpmanley I will give this a try. Thanks for the heads up! The account I was using with the Connect Home I have doesn't have the account permissions to set the runLocally flag.

dkirker: @tpmanley Update incoming with that change!

gonella:
Hi @dkirker, do you know if this fix could help on this issue? 

https://github.com/SmartThingsCommunity/SmartThingsPublic/issues/3046

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -102,6 +102,34 @@ import physicalgraph.zwave.commands.usercodev1.*
...
+ * If not, verify that we have at least requested them or request them,
+ * and check again.
+ */
+def scheduleInstalledCheck() {
+	runIn(120, installedCheck, [forceForLocallyExecuting: true])
+}
+
+def installedCheck() {
+	if (device.currentState("lock") && device.currentState("battery")) {
+		unschedule("installedCheck")
```
tpmanley, line 120: Do you need to unschedule? It looks like the function is only called when the `runIn` fires so I don't think there will be any outstanding schedule that needs to be unscheduled.

tpmanley (approved):

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
dkirker: @tpmanley Probably not. I thought I had remembered hearing that was best practice, but I could be making it up. If anyone thinks it should be removed I can def do that. What I haven't tried is to see if it is possible to schedule multiple instances (I've seen a few duplicate log.debug from installedCheck), and if calling unscheduled once unschedules them all. Or if that is even possible.

kianooshkarami (approved):

dkirker:
@workingmonk Alright, so this is def ready to go to staging, and I will coordinate with Steven C. to test the rest of this week. I am confident the impact is low.

PR #3055, opened 2018-05-03
ICP-4613 Add GE polling setup in installed

dkirker:
There is a chance that when adding a device updated() isn't called. We can safely set up polling in configure() as well. This will ensure that it is called either way.

### DIFF
--- a/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
+++ b/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
@@ -84,14 +84,17 @@ metadata {
 def installed() {
 	// Device-Watch simply pings if no device events received for 32min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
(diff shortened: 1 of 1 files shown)


dkirker:
@mckeed @kianooshST Could I get a code review, please? I added a polling scheduling to configure() for GE devices on top of the one in updated(), but I am open to removing the one in updated(). There's a chance, at least for now, that updated() isn't called in the latest app.

cc @workingmonk I'll be working with Steven to QA so we can have this released with the latest deploy.

workingmonk:
Can you target this to master instead of staging?

kianooshkarami (dismissed):

mckeed:
Isn't this already being done in updated()? See https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3008

dkirker:
@mckeed  There are instances where updated doesn't get called, it seems. They come and go, but it is probably safe to assume it won't always be called.

dkirker:
@mckeed I added the polling setup to installed(). Originally I was going to remove it elsewhere, but it occurred to me that we need it in updated() in the event that someone re-types to us.

kianooshkarami:
@mckeed Would these devices not be polled by local polling?

mckeed (dismissed):

mckeed (approved):
Hub local polling is only for Basic, this change is for Meter.

PR #3056, opened 2018-05-04
ICP-4556 Call refresh in installed to get initial state

dkirker:
Refresh is called from installed() to ensure that it gets run

### DIFF
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -64,12 +64,49 @@ metadata {
 def installed() {
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	initialize()
 }
 
 def updated() {
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
-	response(refresh())
+
+	runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])
+}
+
+def initialize() {
+	def cmds = []
+
+	// Set a limit to the number of times that we run so that we don't run forever and ever
+	if (!state.initializeCount) {
+		state.initializeCount = 1
+	} else if (state.initializeCount <= 10) { // Keep checking for ~2 mins (10 * 12 sec intervals)
+		state.initializeCount = state.initializeCount + 1
+	} else {
+		state.initializeCount = 0
+		return // TODO: This might be a good opprotunity to mark the device unhealthy
+	}
+
+	if (!device.currentState("alarm")) {
+		cmds << zwave.basicV1.basicGet().format()
+		cmds << "delay 5"
+	}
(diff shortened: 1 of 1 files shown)


dkirker:
@mckeed  @kianooshST Could I get a code review, please? Like with the GE polling set PR, I am up for removing the refresh() in updated(). The benefit in both locations should guarantee that we get some sort of update (if installed is too early then as long as updated is called we should get it there; if updated isn't called then we should still have it in installed); the downside being that there will be 2  refresh() calls for most devices. I could add this to configure() and define that capability, but since no devices currently use configure() for what it is meant for, I am hesitant to add it there.

ALamchaS:
How about calling something like:

runIn(3, "initialize", [overwrite: true])

in both updated and installed? forceForLocallyExecuting: true may be required too, I do not know.

You would need to use sendHubCommand in initialize, but it should prevent calling refresh twice.

def initialize() {
    sendHubCommand(refresh())
}

dkirker:
@ALamchaS That's not a bad idea. I was concerned about the resolution of runIn. But on second thought, I think even the possibility of runIn not happening for 1 minute is acceptable.

dkirker:
@ALamchaS Already, I went ahead and updated. I do believe that forceForLocallyExecuting is required, so it is in there.

ALamchaS (approved):

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -64,12 +64,17 @@ metadata {
 def installed() {
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	runIn(10, "initialize", [overwrite: true, forceForLocallyExecuting: true])
```
kianooshkarami: Why not 0 delay? Unless I do not the devices behavior well, I think sending the commands to the hub is much better idea specially because after a device is added, the hub reserves a period of time where the messages for recently added node will have higher priority and will then be sent ahead of other messages. 

My idea would be to start a timer and check every 10 seconds for about a few minutes to see if we have received the data that we wanted or else issue new commands. I think this will prove reliability and certainty that we have what we wanted.

dkirker: @kianooshST That is an interesting idea. I can do this much like what I did for the lock. Actually, I am beginning to wonder if that might be better behavior to do in general. We kind of do it with WakeUpNotification for Z-Wave devices, but for devices that don't knock on the door we end up making a lot of assumptions.

mckeed:
For siren I think we can reasonably just create an "off" event from installed(). If the device even can join while its siren is sounding, I think the amount of users who would do that is sufficiently small.

dkirker:
@kianooshST @ALamchaS Alright, so I updated the logic a little so it isn't just a check once and forget. 

@mckeed We still have to check the battery. Either way, our QA, cert, and Samsung QA gets very picky if we don't have initial states in a timely fashion.

mckeed:
Why not just send the refresh directly from installed() instead of doing the firstCheck thing?

mckeed (dismissed):

mckeed (dismissed):

mckeed (dismissed):

dkirker:
Ok. I'm done with the commit spam (if Github is sending them). I've tested this pretty extensively. Unfortunately, this particular siren flavor (the Lowes model) does not support external power (the Everspring model does). @mckeed @ALamchaS

kianooshkarami (approved):

mckeed (approved):

PR #3076, opened 2018-05-10
DVCSMP-4222 Fix for kickstarter devices

workingmonk:


tpmanley:
`SmartSense Garage Door Multi` as well?

workingmonk:
added garage door multi too

tpmanley (approved):

PR #3077, opened 2018-05-11
DVCSMP-3952 Support battery percentage reporting

dkirker:


devicetypes/smartthings/smartsense-moisture-sensor.src/smartsense-moisture-sensor.groovy
```diff
@@ -204,20 +222,37 @@ def ping() {
...
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 
+	log.debug "Configuring Reporting"
+	def configCmds = []
+
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
+	if (device.getDataValue("manufacturer") == "Samjin") {
+		configCmds += zigbee.configureReporting(0x0001, 0x0021, DataType.UINT8, 30, 21600, 0x05)
```
varzac: The last argument is the reportable change.  In this case it will only inform us if the battery percentage changes by at least 2.5%  I don't have a problem with that necessarily, I just want to make sure that is the intent.  Also, you can replace the `0x0001` with `zigbee.POWER_CONFIGURATION_CLUSTER`

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -101,9 +101,14 @@ def parse(String description) {
 			map = parseIasMessage(description)
 		} else {
 			Map descMap = zigbee.parseDescriptionAsMap(description)
-			if (descMap?.clusterInt == 0x0001 && descMap.commandInt != 0x07 && descMap?.value) {
+
+			if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.commandInt != 0x07 && descMap.value) {
```
varzac, line 117: might as well change the `?.` to just `.` for the `clusterInt` as well to be consistent.  `descMap` can't be null anyways.  Same goes for the other DTHs

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -234,23 +252,39 @@ def ping() {
...
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 
+	log.debug "Configuring Reporting"
+	def configCmds = []
+
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
+	if (device.getDataValue("manufacturer") == "Samjin") {
+		configCmds += zigbee.configureReporting(0x0001, 0x0021, DataType.UINT8, 30, 21600, 0x05)
```
varzac: same as the previous comment.

varzac (dismissed):
Few minor comments, but otherwise looks good.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
dkirker: I'm not sure I understand what you mean. I am going to keep descMap?.clusterInt as the first item in the if (since we do it in every other if in these DTHes), just to at least be safe. Unless it adds some ungodly overhead. This way, if descMap for some reason is null we don't spit up exceptions. (I do see some other areas where it is used from previous commits that isn't needed. And now I am tempted to go on a cleaning spree... 😮 )

varzac: parseDescriptionAsMap will always return an empty map, unless it throws an exception itself.  But it's not a big deal.  It can stay as is.

tpmanley (approved):

PR #3081, opened 2018-05-11
ONEAPP-3651 - Translations for SmartApps

juano2310:


smartapps/smartthings/keep-me-cozy-ii.src/i18n/ar-AE.properties
```diff
@@ -20,3 +20,5 @@
 '''Next page'''=الصفحة التالية
 '''Text'''=النص
 '''Number'''=الرقم
+'''CHOOSE THERMOSTAT…'''=اختيار الثرموستات...
```
workingmonk, line 23: the `...` is the special unicode character rather than the 3 single periods

smartapps/smartthings/keep-me-cozy-ii.src/i18n/ar-AE.properties
```diff
@@ -20,3 +20,5 @@
...
+'''OPTIONALLY CHOOSE TEMPERATURE SENSOR TO USE INSTEAD OF THE THERMOSTAT'S…'''=اختيار مستشعر درجة الحرارة لاستخدامه بدلاً من الثرموستات بشكل اختياري...
```
workingmonk, line 24: same for this one

smartapps/smartthings/keep-me-cozy-ii.src/i18n/ar-AE.properties
workingmonk: also the text is all caps instead of how it is in the code, not sure if that will work

smartapps/egid/smart-windows.src/i18n/ar-AE.properties
```diff
@@ -32,3 +32,4 @@
 '''Next page'''=الصفحة التالية
 '''Text'''=النص
 '''Number'''=الرقم
+'''NOTIFICATIONS'''=الإشعارات
```
workingmonk, line 35: notification is not all caps in the code, will that work??

PR #3091, opened 2018-05-16
ICP-4638 Request current states for Leviton after setLevel

dkirker:


dkirker:
We are checking to see if this is just because of an old firmware.

PR #3094, opened 2018-05-17
Fixing bug ICP-4992. Fixing unsecure join

MMateusiakS:
Please review the PR. 
@greens @dkirker @ALamchaS @MAblewiczS @KKlimczukS

### DIFF
--- a/devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
+++ b/devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
@@ -117,7 +117,7 @@ def refresh() {
 }
 
 private command(physicalgraph.zwave.Command cmd) {
-	if (zwaveInfo && zwaveInfo.zw?.endsWith("s")) {
+	if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo?.zw?.contains("s")) {
 		zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
 	} else {
 		cmd.format()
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
```diff
@@ -119,7 +117,7 @@ def refresh() {
 }
 
 private command(physicalgraph.zwave.Command cmd) {
-	if (state.sec != 0) {
+	if (zwaveInfo && zwaveInfo.zw?.endsWith("s")) {
```
dkirker: We should probably just use:

`zwaveInfo?.zw?.contains("s")`

I am not 100% sure that we are guaranteed that "s" is on the end. (In fact, we should probably make sure all instances of the code like this is made uniform.) cc @kianooshST  @mckeed

mckeed: Some devices using this DTH were added before the new fingerprint type that works with zwaveInfo, so this should be `if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo.zw?.endsWIth("s"))`

mckeed (changes requested):

devicetypes/smartthings/zwave-switch-secure.src/zwave-switch-secure.groovy
ALamchaS: @mckeed 
I commited modifications requested by you and Donald. I have a question: Do the devices that use this DTH since before the new fingerprint type have zwaveInfo at all? Or should it be something like: 
if ((zwaveInfo == null || zwaveInfo.zw == null && state.sec !=0) || ...
?

greens: I'm pretty sure they'd have a zwaveInfo object, but I'll defer to @mckeed on this one.

Also I know there are DTHs that do `.contains("s")`, but I think we want to do `.endsWidth("s")` generally.

mckeed: The implementation of `zwaveInfo` returns an empty map as the failure case, so it should never be null.

mckeed (approved):

PR #3095, opened 2018-05-18
DVCSMP-3639 Implement same "polling" method across all ZLL devices

dkirker:


varzac (approved):

devicetypes/smartthings/zll-rgb-bulb.src/zll-rgb-bulb.groovy
```diff
@@ -101,6 +101,8 @@ def refresh() {
 }
 
 def poll() {
+	configureHealthCheck()
```
varzac, line 104: Do we need to reconfigure the schedule every time poll is called?

varzac:
So I'm wondering if we want to just remove the poll capability from all of these.  Our end goal is to have all of that polling handled locally, however, if they have the poll capability the cloud is going to continue to poll them even when they are marked to run locally.  @gausnes what are your thoughts on that?  We could always leave the poll capability for now, and remove when we are ready to go fully local (which might be as soon as a couple weeks depending on timings of the 22 release and the release to 21 on other platforms).

gausnes:
I think I'm fine with either since we will eventually need to do more cleanup of cloud polling when local polling is available.

It might actually be good to leave the `poll` capability so we remember to clean up this DTH.

devicetypes/smartthings/zll-rgb-bulb.src/zll-rgb-bulb.groovy
dkirker: It should only get called once if the "custom health check" hasn't been setup. I added it to poll to account for existing devices with the DTH already setup, since deploy (from what I recall) won't call update automagically.

dkirker:
I'm definitely all for yanking out the polling capability once we have supported it hub-side.

PR #3097, opened 2018-05-21
[CHAD-1623, CHAD-2144] Mark cree and zwave water valve to run locally

varzac:


tpmanley (dismissed):

varzac:
@workingmonk we'd like this to go to staging so it can go to prod tomorrow to prepare for the customer beta for hubcore 22 release.  Steven is looking at regression to make sure these changes don't break 21 and Nate is looking at the local execution changes.

tpmanley (approved):

PR #3103, opened 2018-05-22
CHAD-2137 Remove duplicate events being generated on some z-wave dimmers

BroderickCarlin:
There has been a complaint about some Z-Wave dimmers producing duplicate events. The root cause of this issue was creating an event on both the command in the cloud DTH and the report on the hub. These duplicate events were not being caught by the duplicate filter because one of the events contained a unit of `%` while the other was unit-less.

The solution to the problem shown here was to remove the unit of the event when the command is sent.


mckeed (dismissed):
Probably also need the same change to Z-Wave Dimmer Switch Generic.

This might cause issues with the client UI since the real event comes several seconds later than this fake event.

BroderickCarlin:
@mckeed  `Z-Wave Dimmer Switch Generic` does not suffer from the same issue because it has it's `executeCommandsLocally` flag set to `true`. It was because of this that I did not also make the change on that DTH, but that DTH would run into the same issues if the flag to run commands locally was set to `false` while leaving the `runLocally` flag set to `true`.

Have not personally witnessed any UI issues, but it is something to be aware of

mckeed:
Ah, interesting. I wonder if there's a reason this one shouldn't execute commands locally.

I didn't think about the fact that the UI is already handling the locally-executing setLevel commands that don't send the instant event, so not much reason to worry about this.

BroderickCarlin:
This DTH has support for LED indicators that I don't _believe_ we have any support for locally, which is what is causing this one to execute commands in the cloud

BroderickCarlin:
Just a follow up: removed the unit from the level event on the two cloud DTH's so that they are consistent with the rest of the level events we create across devices. We will now rely on the duplicate filter to make sure that only a single event is ever actually reported.

varzac (dismissed):

tpmanley (dismissed):

BroderickCarlin:
Realized after some quick testing that the previous change of removing units did not completely fix the issue at hand, as we also needed to make sure to handle the 0-99 to 0-100 mapping in the cloud. Without this an event with a value of 99 and an event with a value 100 will both be reported to the user.

tpmanley:
That will be covered by https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2944

BroderickCarlin:
Alright, undoing the changes that were just made in relation to the 0-99 to 0-100 mapping. This ticket is now being blocked by #2944

greens:
@mckeed `dimmer-switch-generic` executing locally is relatively recent. Not really any reason the GE dimmer couldn't also execute locally, we just haven't put it in yet.

The only issue we've run into is the polling of the level being done too quickly when executing locally so we get intermediate reports, but this is fixed in the latest hubcore version.

greens (approved):

PR #3104, opened 2018-05-22
CHAD-2137 Remove duplicate events being generated on some z-wave dimmers V2

BroderickCarlin:
Not pointing any fingers at @greens but the [previous PR](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3103) was merged before it was ready. (I should have thrown the `DO_NOT_MERGE` label on there) 👀 

It was determined that we can't reliably rely on filtering of duplicate events and so removing the creation of the event from the command creation function is going to be the most robust way to avoid these duplicate events. Really the only downside of this approach is going to be a slight delay between the command being sent to the device and the event being generated to the user, currently about 4 seconds.


greens:
Hm, everything seemed to make sense with those changes to me. 

But I don't think that this change is beneficial, as it essentially makes the cloud DTH non-functional.

Even if the code is never called I think it's a good idea to have all of our public DTHs work in the cloud.

varzac:
@greens the devices will still generate the event when the device reports the change (or we read after the send).  It just won't be instantly generated when we send the change level command.

BroderickCarlin:
I disagree that this change makes the cloud DTH non-functional. Currently this DTH is setup to run half locally and half in the cloud, denoted by having `runLocally: true` and `executeCommandsLocally: false`. If we look at just the cloud DTH prior to this change we are creating identical `level` events when we send a command and when we get a response. We would normally assume these events would be filtered by the duplicate event filter, but we are seeing that this isn't as reliable as we would like. With the changes shown in this PR you can have the DTH run completely in the cloud and you will still receive a single event, as expected. Hopefully this helps to clarify why we want/need this PR :)

greens:
Oh okay. I see it now. I'll let you merge this one when you're ready :)

greens (approved):

PR #3106, opened 2018-05-22
WWST-488 Add join name for danfoss thermostat

greens:


greens:
This is a device for M6 that is getting re-certified.

dkirker (approved):

PR #3117, opened 2018-05-25
ICP-3042 Fan Controller DTH

greens:
First draft of this DTH. We've been told this needs to happen since fan controllers have non-standard behavior when using the standard dimmer dth.


greens:
Topics for discussion:

* A lot of fan controllers have a sort of "burst" on mode where they go up to max (report max as well) and then go back down to the set value. Ideally we want that "burst" to be hidden.
* Is cycling from off->low->medium->high an acceptable main tile use, or should the main tile just have up/down arrow controls instead of the slider?
* These icons are not perfect
* The only other device that uses this capability is the Samsung AC, it's not really fully fleshed out

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -0,0 +1,145 @@
...
+
+def zwaveEvent(physicalgraph.zwave.commands.switchmultilevelv1.SwitchMultilevelSet cmd) {
+	fanEvents(cmd)
+}
+
+def fanEvents(physicalgraph.zwave.Command cmd) {
+	def value = (cmd.value ? "on" : "off")
+	def result = [createEvent(name: "switch", value: value)]
+    result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)
+    result << createEvent(name: "fanSpeed", value: cmd.value/33)
```
greens: I'll fix this

dkirker:
I think up/down might be better. I've never been a fan (haha, fan) of the cycle through. I would also add a small control tile next to refresh to turn the fan off; or even toggle on/off. I'm not sure how many controllers actually save their previous value, but having that tile would be nice.

dkirker:
An alternate would be to do what the community did for the King of Fans controller. I believe the main tile is on/off with showing the current state, and then there are individual tiles for each button below. Annoyingly, either implementation assumes the number of supported speeds.

ALamchaS:
I vote up/down. Cycling seems unnecessarily restrictive, and a bit strange if off is included.

Burst mode: Do you want to expose enable/disable for supporting devices? And will the delay be constant for simplicity or customized for each device?

Are there other devices other than Honeywell Speed Fan Controller (ICP-3042) and Samsung AC compatible with this implementation? 

Problem with custom number of speeds would be presentation.
- Is it possible to make something like Connect app's thermostat's setpoint temperature tile with Speed and speed mode number instead of "Set" and temperature in SmartThings Classic?
- Those dark/light rectangles from KoFans that present speed modes seem OK, but I do not know if it is possible to do that with custom number of rectangles and still look good.
- Names would be a pain to translate, especially if we want to create more than 4 modes. 
- In theory we could use some fan icon with different number of ~ symbolizing strength of air movement. Then number of ~ would indicate speed mode number, but would not be readable for more than 5~. Since we use +/- or cycling instead of slider, handling math of custom number of speeds should be easy. Handling icon selection would ugly if we wanted max speed to look the same for all devices.
- Normal slider could still be used by rounding slider value - this could be useful for 0-100 level speed controllers that could just use changing icons/texts to those matching fan controller better than light controller.

And one more problem is that I do not think we are guaranteed that plugin for Connect will use the same design principle.

greens:
Updated this a bit. The main control is now just a display, with the up/down controls like for a thermostat. Added an on/off tile.

Tried to get slider to work, but it was temperamental.

@ALamchaS : This DTH is just for the z-wave fan controllers. So far they all only have four states (0, 33, 66, 99 in switch multilevel).

ALamchaS (dismissed):
Looks good to me.

Will it get some kind of generic metadata?

greens:
@ALamchaS No. Probably not yet at least. Working on a custom one.

dkirker (dismissed):
Looks good. UI meta should also be on staging.

dkirker:
One question @greens If you tap the primary control, does it do anything? We might want to have it toggle on/off. If not, I think that can be done by giving each state a "action: "toggleFan" " value (or some command like that).

greens:
I'll add making the main tile toggle the device off and add in fingerprints

greens:
I think this is the version I'm ready to commit. Has fingerprints and the main tile toggles.

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		standardTile("refresh", "device.switch", width: 2, height: 2, inactiveLabel: false, decoration: "flat") {
+			state "default", label: '', action: "refresh.refresh", icon: "st.secondary.refresh"
+		}
+		main "fanSpeed"
+		details(["fanSpeed", "refresh"])
+	}
+
+	fingerprint mfr: "001D", prod: "1001", model: "0334", deviceJoinName: "Leviton 3-Speed Fan Controller"
```
dkirker: These will need to be moved up to the definition.

dkirker (dismissed):

greens:
Nice catch. Thanks. Fixed.

tpmanley (approved):
LGTM

PR #3131, opened 2018-06-01
ICP-4842 Keep sending association set until Enerwave motion indicates it is associated

dkirker:
Add battery event checking to hold over until CHAD-2158 is in a released firmware

### DIFF
--- a/devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
+++ b/devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
@@ -142,18 +142,40 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 	result
 }
 
+def zwaveEvent(physicalgraph.zwave.commands.associationv2.AssociationReport cmd) {
+	def result = []
+
+	if (cmd.groupingIdentifier == 1) {
+		if (cmd.nodeId.any { it == zwaveHubNodeId }) {
+			state.group1Assoc = true
+			result << createEvent(descriptionText: "$device.displayName is associated in group ${cmd.groupingIdentifier}", isStateChange: false)
+		} else {
+			result << createEvent(descriptionText: "Associating $device.displayName in group ${cmd.groupingIdentifier}", isStateChange: false)
+			result << response(zwave.associationV1.associationSet(groupingIdentifier:1, nodeId:zwaveHubNodeId))
+		}
+	}
+	result << decrWakeUpRequestRefCount()
+
+	result
+}
+
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd)
 {
 	def result = [createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)]
 
-	if (isEnerwave() && device.currentState('motion') == null) {  // Enerwave motion doesn't always get the associationSet that the hub sends on join
-		result << response(zwave.associationV1.associationSet(groupingIdentifier:1, nodeId:zwaveHubNodeId))
+	incrWakeUpRequestRefCount(true)
+
+	if (isEnerwave() && !state.group1Assoc) {  // Enerwave motion doesn't always get the associationSet that the hub sends on join, so check and re-send if needed
+		result << response(zwave.associationV1.associationGet())
+		incrWakeUpRequestRefCount()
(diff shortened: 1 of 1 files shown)


dkirker:
@mckeed  I can definitely remove the expanded battery checking code if you don't think it is needed. I have tested with it in my DTH as well.

greens:
If we need to re-send the association and poll the battery, does that result in any weird situations where the noMoreInfo is sent before we've received the battery report?

dkirker:
@greens That could possibly happen. From what I have seen when using a zniffer we'll more than likely get the battery report. That said, the device is configured to send the Wake Up every 30 minutes, so missing a battery report might be ok, since the checkInterval is set much much higher.

dkirker:
@greens  So, I went ahead and implemented a ref count for the number of requests we make in the wake up event. Since the cause of CHAD-2158 was only internal I axed the shouldCheckBattery stuff (I also realized there was a line of code I hadn't committed in it that would result in a bug with it always returning true).

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -142,18 +142,40 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 	result
 }
 
+def zwaveEvent(physicalgraph.zwave.commands.associationv2.AssociationReport cmd) {
+	def result = []
+
+	if (cmd.groupingIdentifier == 1) {
+		if (cmd.nodeId.any { it == zwaveHubNodeId }) {
+			state.group1Assoc = true
+			result << createEvent(descriptionText: "$device.displayName is associated in group ${cmd.groupingIdentifier}", isStateChange: false)
```
greens, line 151: should these be events or debug logs?

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -276,3 +298,29 @@ private command(physicalgraph.zwave.Command cmd) {
...
+private incrWakeUpRequestRefCount() {
+	if (state.wakeUpRequestRefCount == null) {
+		state.wakeUpRequestRefCount = 0
+	}
+
+	state.wakeUpRequestRefCount = state.wakeUpRequestRefCount + 1
+}
+
+private decrWakeUpRequestRefCount() {
+	def result = ""
```
greens: do empty event strings cause any trouble?

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
dkirker: I ended up going with events, like in the Wake Up Notification (`createEvent(descriptionText: "${device.displayName} woke up", isStateChange: false)`), so that they might be more readily accessible in the event list for troubleshooting a customer issue. Though, log.info could be equally as useful. Either shouldn't hurt the system (though, given some changes, it is possible that those events may stop working).

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
dkirker: So far in my testing: nope.

greens (dismissed):

dkirker:
@greens So... I had a chance to do a quick one over of this again (and also noticed an issue when I looked at some log output last night), and I found two small bugs which I just fixed (the first battery state request made outside of a WakeUpNotification would result an exception, and it was possible to get in a state where the ref counter was never going to reach 0 if two WakeUpNotifications were received before we sent the NoMoreInfo for the first one or if the device never responded to all of our requests in a wake up cycle). Could I get a new review, please? :)

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
dkirker: I went ahead and wrapped it in a response() call just so I could guarantee that the function was following its promise.

dkirker:
And squashed into 1 commit.

greens (approved):
It looks good but now I almost think it would be worth just sometimes sending two wakeups instead of having to worry about reference counting.

PR #3138, opened 2018-06-06
WWST-530 ICP-5169 Aeon Nano Non-metered switch fingerprint/Secure Switch health check

greens:
Adds the fingerprint for the US version. Identical to the EU version in all respects.

ICP-5169 Secure Switch does not support health check

Updating the DTH to add health check


csuk07:
@dkirker bump

dkirker (approved):

PR #3139, opened 2018-06-06
Temporarily revert "[ICP-4701] Add initial state check. (#3020)"

dkirker:
This reverts commit 55781c68e3e292295d332c7f87ff0c6808ae6a97.

Temporary revert to un-block deploy.


dkirker:
Not reverting as the code this would revert is deemed safe.

PR #3143, opened 2018-06-06
ICP-4701 Initial state for Aeotec door window

greens:
This is kind of the nuclear option, but in the case that the device goes to sleep before we can poll it, there's not much else we can do. The "bad" state only remains for 4 hours until the next wakeup.

### DIFF
--- a/devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
+++ b/devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
@@ -100,6 +100,9 @@ def parse(String description) {
 def installed() {
 	// Device-Watch simply pings if no device events received for 482min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 4 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	// this is the nuclear option because the device often goes to sleep before we can poll it
+	sendEvent(name: "contact", value: "open", descriptionText: "$device.displayName is open")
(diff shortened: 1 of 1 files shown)


ALamchaS:
I like it. I think Certification will most likely report bad battery level after pairing, so i am not 100% sure about this one.

dkirker:
I'm fine with the contact state. I am hesitant with the battery state. I know it will update later, I just feel odd about saying all is "well" when maybe it isn't.

greens:
@ALamchaS Yeah. I think cert will throw flags on this one, but they're also throwing flags on no initial state. I think this has a better chance of getting deferred or accepted because:

@dkirker I was trying to think of corner cases where this would be bad, considering this is only for initial install. Any time you change a battery while already paired I would expect an unsolicited report. So for initial install you'd have to take it out of the box, throw out the battery it came with and go get a garbage one that _only_ has enough charge to add and then dies in 4 hours without sending an unsolicited "low battery" report. Just seems like an extremely unlikely extreme corner case.

workingmonk:
LGTM, not sure if @mckeed is ok with this change

workingmonk (approved):

dkirker:
@greens  Sure, I was thinking more of getting a sensor from a friend/second hand, or one that was laying around for a while and being repaired or didn't come with a battery. I suspect "faking it until you make it" should be ok here since this device isn't something like a lock.

greens:
@mckeed Thoughts on the nuclear option?

greens:
@mckeed ping

mckeed (approved):
Sure, we don't really have a choice since the client requires initial state now.

One improvement might be in initialPoll to check whether the device actually uses the Binary Sensor CC, since I think it's deprecated in favor of Notification CC now.

PR #3169, opened 2018-06-18
ICP-5223 change ocf icon for zwave metering dimmer

greens:
All of these devices are switches, so they should probably use the switch icon instead of the bulb.


greens:
@dkirker ping

dkirker (approved):

PR #3179, opened 2018-06-20
DVCSMP-4223 NPE Check for zwave info

workingmonk:
Fix based on @tpmanley quick NPE analysis.

PR against production for hotfix release

Bug introduced in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2858


rappleg (approved):

tpmanley:
-It would be a good idea to add the null check on `zwaveInfo` in `configure` as well-

NM, not needed

mckeed:
We've been leaving this out because `getZwaveInfo` is guaranteed not to return null

tpmanley:
Good point. I guess the null reference was on `zw` and not `zwaveInfo`

mckeed (approved):

greens (approved):

PR #3186, opened 2018-06-21
DVCSMP-4224 Add "codeId" in parallel with "usedCode" to follow capability definition

dkirker:


greens (approved):

vaish-rahul:
[This](https://github.com/SmartThingsCommunity/SmartThingsCapabilities/blob/master/capabilities/lock.src/lock.yml) specifies `codeId `to be a string. What you are adding in the map is an integer. Will this create a problem?
CC: @Bijnagte, @niuzhibin

dkirker:
@vaish-rahul I made an update at https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3223 in the event that it does.

PR #3188, opened 2018-06-21
[PROB-2107] Fix SSL handshake Exception with Harmony and use TLSv1.2

rappleg:
Didn't add the tlsVersion to the async calls because it already uses TLS v1.2.

/cc @gausnes @workingmonk


gausnes (approved):

workingmonk:
closing this one for now as we have flipped the flag at the service level

PR #3190, opened 2018-06-21
ICP-5091 Enable Tamper detection by default for Fibaro CO Sensor

greens:


dkirker (dismissed):
It looks good. There's the chance that the initial syncStart might not run because it is called from updated(), however there is syncNext which gets called from the WakeUpNotification, so I think that will be ok.

greens:
updated

dkirker (approved):

PR #3201, opened 2018-06-28
Add initial state check for FireAngel smoke detector.

ALamchaS:


greens (approved):
Looks good. I'll merge if Rafal certifies it.

PR #3204, opened 2018-06-29
Adding generic ui-metadata to simulated switch and dimmers

workingmonk:
cc @tpmanley @bflorian


tpmanley:
👍

 ticket?

stevenchitov:
It was verified on Staging using OneApp. Implemented, works fine for Simulated Dimmer Bulb, Simulated Dimmer Switch, Simulated Switch, Virtual dimmer Switch, Virtual Switch

PR #3213, opened 2018-07-03
ICP-5324, ICP-5325 - Switching Leviton dimmer on after setting level > 0

KKlimczukS:
ICP-5324 - [US] WWST GAP - Incorrect status displayed on ST Android and iOS app - fix
ICP-5325 - [US] WWST GAP - Incorrect status displayed on ST CLASSIC Android and iOS app - fix

Switching Leviton dimmer on after setting level > 0
@greens @mckeed @MAblewiczS @ALamchaS @MMateusiakS @DCzarneckaS


greens (commented):
This only works if we turn off local execution. As it stands now, the MRVL devices are probably also broken.

MMateusiakS:
Hey @greens , I found PR by @KKlimczukS  who is on long sick leave. What should we do this this PR? If his change does not add any improvements should this be fixed in hubCore?

greens:
Leviton have confirmed this is a bug on their side that they will be addressing.

PR #3223, opened 2018-07-04
DVCSMP-4224 Change codeId value type from integer to string to match the spec

dkirker:
User string and not integer which was used in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3186

### DIFF
--- a/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
+++ b/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
@@ -558,7 +558,7 @@ private def parseCommandResponse(String description) {
 				return null
 			}
 			codeName = getCodeName(lockCodes, codeID)
-			responseMap.data = [ codeId: codeID, usedCode: codeID, codeName: codeName, method: "keypad" ]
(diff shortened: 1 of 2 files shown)


greens (dismissed):

workingmonk:
```
def a = 5
println a.getClass()

def b = a
println b.getClass()

def c = "${a}"
println c.getClass()

def d = a.toString()
println d.getClass()
```

output was
```
class java.lang.Integer
class java.lang.Integer
class org.codehaus.groovy.runtime.GStringImpl
class java.lang.String
```

dkirker:
I'm going to check on the string type to make sure it is handled right. I would *assume* that GStringImpl would be treated similarly, but I am not 100% sure.

vaish-rahul:
@niuzhibin, @dkirker When can this be resolved? @niuzhibin I can't initiate SQE testing unless all features are implemented.

dkirker:
@vaish-rahul I update the type. I haven't had a chance to test this with a lock, but you are welcome to get started with it and check. cc @niuzhibin

dkirker:
@greens or @workingmonk If you don't mind (re)reviewing this, please? Thank you.

workingmonk (approved):

PR #3224, opened 2018-07-04
WWST-1050 DTH for Zipato Siren

ALamchaS:
Includes minor refactoring and fix for issue ICP-5323

cc: @greens @dkirker  @MAblewiczS @MMateusiakS @KKlimczukS @DCzarneckaS

### DIFF
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -20,6 +20,7 @@ metadata {
 		capability "Actuator"
 		capability "Alarm"
 		capability "Battery"
+		capability "Configuration"
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
@@ -30,6 +31,8 @@ metadata {
 		fingerprint mfr: "0258", prod: "0003", model: "0088", deviceJoinName: "Neo Coolcam Siren Alarm"
 		fingerprint mfr: "021F", prod: "0003", model: "0088", deviceJoinName: "Dome Siren"
 		fingerprint mfr: "0060", prod: "000C", model: "0001", deviceJoinName: "Utilitech Siren"
+		//zw:F type:1005 mfr:0258 prod:0003 model:1088 ver:2.94 zwv:4.38 lib:06 cc:5E,86,72,5A,73,70,85,59,25,71,87,80 role:07 ff:8F00 ui:8F00
+		fingerprint mfr: "0258", prod: "0003", model: "1088", deviceJoinName: "Zipato Siren Alarm"
 	}
 
 	simulator {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -129,25 +136,21 @@ def poll() {
...
-	return (new Date().time - timestamp) > (seconds * 1000)
-}
-
 def on() {
 	log.debug "sending on"
-	[
-		zwave.basicV1.basicSet(value: 0xFF).format(),
-		zwave.basicV1.basicGet().format()
-	]
+	if (zwaveInfo.mfr == "0258" && zwaveInfo.model == "1088") {
```
greens: Can you add some comments here about why we're doing this for future devs?

ALamchaS:
Hi @greens . I added comment that you requested.

greens (approved):

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -129,25 +136,25 @@ def poll() {
...
-		zwave.basicV1.basicGet().format()
-	]
+	// ICP-5323: Zipato siren sometimes fails to make sound for full duration
+	// Those alarms do not end with Siren Notification Report.
+	// For those cases we add additional state check after alarm duration to
+	// synchronize cloud state with actual device state.
+	if (zwaveInfo.mfr == "0258" && zwaveInfo.model == "1088") {
+		[
+			zwave.basicV1.basicSet(value: 0xFF).format(),
+			zwave.basicV1.basicGet().format(),
```
kianooshkarami, line 148: Why is this doing double get? why not just do 2nd get?

dkirker: Actually, wouldn't a runIn be more appropriate for the later state check?

greens: My guess is that the device doesn't send the report in response to the set. The first get is to get the "on", the second get is to get the "off".

ALamchaS: It is as Steven said.

When we turn the Zipato Siren on it should sound the alarm for 60 seconds then turn off by itself. When this works correctly, we get notification that the siren is off and everything is OK.

Sometimes alarm fails to play for full 60s, and then we never get this notification. So we first check if the siren is on to update cloud state for the device (on). Then after 63 seconds we refresh to make sure the device is off.

runIn would be better in that we could make the second update conditional (if still on, update state), but the way I understand this method it may be less reliable as to when exactly it will be called. I would like this to be as precise as  reasonably applicable since this is a security-related device.

PR #3233, opened 2018-07-06
HMS-202 Add Health Check to simulated devices

surfous:
Replaces #3170 

### Update health check for _(commit 1)_:
* simulated-lock
* simulated-thermostat

### Added health check for _(commit 2)_:
* simulated-minimote
* simulated-presence-sensor
* simulated-refrigerator
* simulated-refrigerator-door
* simulated-refrigerator-temperature-control
* simulated-temperature-sensor

### Already had health check
* simulated-dimmer-switch

### Remove ping() methods and unnecessary merge conflict file _(commit 4)_


devicetypes/smartthings/testing/simulated-thermostat.src/simulated-thermostat.groovy
```diff
@@ -306,6 +314,8 @@ def parse(String description) {
 
 def ping() {
     log.trace "Executing ping"
+    sendEvent(name: "DeviceWatch-DeviceStatus", value: "online")
```
workingmonk: if this is untracked, i don't think we need the ping() method, but check with the DH team

munds (approved):
Looks good! I don't know why setLevel is calling setTemperature but after speaking with @surfous he said it's a relic of the past that is needed for proper operation (mainly to set the ambient temperature).

PR #3239, opened 2018-07-10
ICP-5167 Remove deprecated capability

greens:
"Garage Door Control" has been replaced be "Door Control"

### DIFF
--- a/devicetypes/smartthings/zwave-garage-door-opener.src/zwave-garage-door-opener.groovy
+++ b/devicetypes/smartthings/zwave-garage-door-opener.src/zwave-garage-door-opener.groovy
@@ -17,7 +17,6 @@ metadata {
 	definition (name: "Z-Wave Garage Door Opener", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.017.0012', executeCommandsLocally: false) {
 		capability "Actuator"
 		capability "Door Control"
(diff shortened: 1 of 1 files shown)


CosmicPuppy:
* When will all affected DTHs be updated?
* Is there a process to notify Community Developers to update their DTHs and affected SmartApps?

dkirker:
@CosmicPuppy  While the Garage Door Control capability has been flagged as deprecated (in favor of Door Control), we have not actually come up with a process yet for deprecation (or at least removal). There are still a few places (in solutions and SmartApps) internally where Garage Door Control is still in use at this time. This PR was looking to fix an issue with the ST client, but that was fixed through another means.

CosmicPuppy:
Thanks for the reply @dkirker.

I'm still a tad confused though. Will this PR be accepted / merged? If so, then any SmartApps which have an input filter of "Garage Door", are no longer going to function (i.e., no longer going to be able to select instances of this Device). This includes **ActionTiles** which has implemented a custom Garage Door Tile. Why not leave the extra Capability claim in place until deprecation notification process has been established?

Or is this actually PR being rejected?

dkirker:
@CosmicPuppy No changes are being made at this time. As for this PR, it is closed so I don't see a reason it would be coming back. Since it is closed, the devices that use it will remain unchanged. The change that was made in place of this won't affect SmartApps (for this very reason).

PR #3247, opened 2018-07-11
ICP-4820 - Remove polling after activity

juano2310:
By removing polling we prevent an immediate response with the wrong value.


munds (approved):
Complicated fix :+1:

PR #3250, opened 2018-07-12
ICP-5250 Fibaro Wall Plug US not updating power

greens:


dkirker (changes requested):
While your changes look fine, I'm hoping you don't mind making some formatting changes to that function. I'll leave them in-line on the main file.

dkirker:
Ha! I guess I can't, so I'll do my best here.

The two if blocks that act on ep should be an if and an else if.

There are some if and else statements that are one line that should be expanded a bit for readability.
This:

if (blah) {
some work
} else {
some other work
}

Instead of:

if (blah) {some work}
else {some other work}

And the same for the switch statement.

greens:
Updated.

dkirker (approved):

PR #3255, opened 2018-07-13
[WWST-889] Danalock V3 Z-Wave genericized DTH with no codes support.

ALamchaS:
This can work with generic zwave-lock DTH with some modification. It is significantly smaller - close to 1/3rd of generic zwave-lock DTH size.

Base DTH is huge, so I may have missed some useless code. Please feel free to comment on everything, @greens , @dkirker , @mckeed 

cc for SRPOL PE:
@MMateusiakS @DCzarneckaS  @MAblewiczS  @KKlimczukS

### DIFF
--- a/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
@@ -122,12 +122,6 @@ def updated() {
 				cmds << doConfigure()
 			}
 			cmds << refresh()
-			if (!state.MSR) {
-				cmds << zwave.manufacturerSpecificV1.manufacturerSpecificGet().format()
-			}
-			if (!state.fw) {
-				cmds << zwave.versionV1.versionGet().format()
-			}
 			hubAction = response(delayBetween(cmds, 30 * 1000))
 		}
 	} catch (e) {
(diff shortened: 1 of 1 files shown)


greens:
I think the problem with putting it in the generic lock DTH is that it would get picked up automatically by lock code manager.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -0,0 +1,593 @@
...
+	def hubAction = null
+	try {
+		def cmds = []
+		if (!device.currentState("lock") || !state.configured) {
+			log.debug "Returning commands for lock operation get and battery get"
+			if (!state.configured) {
+				cmds << doConfigure()
+			}
+			cmds << refresh()
+			if (!state.MSR) {
```
greens: I think this is a holdover from an older way of doing things. We can probably just use `zwaveInfo.mfr` for anything we would be using this for if we're starting from scratch.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -0,0 +1,593 @@
...
+				cmds << zwave.manufacturerSpecificV1.manufacturerSpecificGet().format()
+			}
+			if (!state.fw) {
```
greens: nothing is ever done with this so we can probably leave it out

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -0,0 +1,593 @@
...
+
+/**
+ * Returns the list of commands to be executed when the device is being configured/paired
+ */
+def doConfigure() {
+	log.trace "[DTH] Executing 'doConfigure()' for device ${device.displayName}"
+	state.configured = true
+	def cmds = []
+	cmds << secure(zwave.doorLockV1.doorLockOperationGet())
+	if (zwaveInfo.mfr != "010E") {
```
greens: Why restrict this based on mfr?

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
ALamchaS: Ok. Thanks. I removed this.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
ALamchaS: Removed. Thanks.

greens (approved):

PR #3258, opened 2018-07-16
[ICP-5298] Updated fingerprint for new hardware version of Zipato Siren Alarm.

ALamchaS:


greens (approved):
At least this fingerprint is using the correct manufacturer info.

PR #3263, opened 2018-07-17
[GWWST-115] Added generic button metadata for child button DTH.

ALamchaS:
Hi, @greens We are doing summer cleaning, and try to do something about ancient and forgotten Jira tickets.
This PR adds generic metadata to child button DTH.

SRPOL PE:
@MMateusiakS @DCzarneckaS @MAblewiczS @KKlimczukS

### DIFF
--- a/devicetypes/smartthings/child-button.src/child-button.groovy
+++ b/devicetypes/smartthings/child-button.src/child-button.groovy
@@ -14,7 +14,7 @@
  *
  */
 metadata {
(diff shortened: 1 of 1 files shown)


greens:
If it's already got a device-specific metadata I don't think we should convert it to generic until we start to have problems.

greens:
Problems other than that OneApp doesn't work with child devices.

PR #3265, opened 2018-07-17
[ICP-5100] Fix issue where device does not always respond to on/off commands

sagar-trility:


sagar-trility:
@greens Can you review this PR.

greens (approved):

PR #3269, opened 2018-07-18
[ICP-5298] DTH hotfix, update mfr/model checks.

ALamchaS:
@greens , could you please review this?

There will be some changes to this DTH after Vodafone finally says which melody and volume they want.


ALamchaS:
Ok. This bug where device does not update its state after 1 minute reoccured, so I updated this DTH to work with new mfr/model values.

greens (approved):

PR #3271, opened 2018-07-18
[ICP-5416, ICP-5419] Move ZLL 5000k bulb to cloud

varzac:
This move the "ZLL White Color Temperature Bulb 5000k" DTH to execute in
the cloud until 23.  This is because in the current local implementation
there isn't a read after a delay causing us to be out of sync
temporarily.

This addresses: https://smartthings.atlassian.net/browse/ICP-5419
This addresses: https://smartthings.atlassian.net/browse/ICP-5416

### DIFF
--- a/devicetypes/smartthings/zll-white-color-temperature-bulb-5000k.src/zll-white-color-temperature-bulb-5000k.groovy
+++ b/devicetypes/smartthings/zll-white-color-temperature-bulb-5000k.src/zll-white-color-temperature-bulb-5000k.groovy
@@ -13,7 +13,7 @@
  */
 
 metadata {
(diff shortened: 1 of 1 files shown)


tpmanley (approved):

greens:
Similar problem with ICP-4554 you think?

greens:
I think removing devices from LE might get some pushback.

varzac:
@greens with regards to ICP-4554, this issue is isolated to this particular DTH and that doesn't look to be using this DTH.  Although I didn't read through all the comments on that ticket so I might have missed something.

varzac:
@greens You are OK with this now correct?

greens:
Yeah.

greens (approved):

PR #3277, opened 2018-07-19
WWST-1093 Fibaro Wall Plug UK

greens:
The fibaro DTH for this device is not functional. Results are much better with our generic.

### DIFF
--- a/devicetypes/fibargroup/fibaro-wall-plug-eu-zw5.src/fibaro-wall-plug-eu-zw5.groovy
+++ b/devicetypes/fibargroup/fibaro-wall-plug-eu-zw5.src/fibaro-wall-plug-eu-zw5.groovy
@@ -12,8 +12,7 @@ metadata {
 
 		command "reset"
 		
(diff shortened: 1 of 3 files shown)


dkirker (approved):

ericangST:
@greens  I'm talking to Fibaro, and they tested the Fibaro Wall Plug US, but noticed that it was pairing with the wrong name.

I asked them to check the Fingerprint of their device and it's not the same as what we have in the DTH.

This is what they pulled from their device

type:1001 mfr:0072 prod:0501 model:0F14

I'm asking them why we have a totally different fingerprint so I'm having them get back to me

I do see that the device is certified, you can find it here
https://products.z-wavealliance.org/products/2622

You think we can get this pushed in? I don't think we need to deal with JIRA here

ericangST:
Just got a reply back from Fibaro team, they were wondering if it would be okay if we removed "Model" 

This way it's more generic and they're okay with that. Thoughts?

Something like 
type:1001 mfr:0072 prod:0501

Would be listed they saw the same thing in the fibaro flood sensor DTH
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/fibaro-flood-sensor.src/fibaro-flood-sensor.groovy

PR #3278, opened 2018-07-20
ICP-5475 [UK] Incorrect device states on ST Classic app

MAblewicz:
Fixed device states - it should be now only "HEAT" and "cleared"
added battery state initial check


devicetypes/smartthings/zwave-basic-heat-alarm.src/zwave-basic-heat-alarm.groovy
```diff
@@ -145,20 +148,39 @@ def createHeatEvents(name) {
 	switch (name) {
 		case "heat":
 			text = "$device.displayName heat was detected!"
-			result = createEvent(name: "heat", value: "detected", descriptionText: text)
+			result = createEvent(name: "temperatureAlarm", value: "heat", descriptionText: text)
 			break
 		case "tested":
```
greens: we should probably just get rid of the "tested" references here

dkirker: Removing it, or at least commenting it out, is probably a good idea. I was going to suggest for capabilities that don't have a "tested" attribute value that we could send one of the events followed by a "clear", but I think that would have bad consequences for automations.

I was thinking this morning of proposing a sort of "Alarm Test" capability that would remove the "tested" value from capabilities and better show a history of alarm tests (i.e. you could call in to getting the latest state and look at that event's timestamp).

MAblewicz: Sure i will remove tested. I think reporting this as HEAT is OK. That's the way certification was initially checking that device.

MAblewicz:
@dkirker @greens @mckeed  i removed unnecessary tested state. If code is ok now could you approve this PR

greens (approved):

PR #3289, opened 2018-07-23
ONEAPP-6300 Fix battery reporting for old Fibaro Flood Sensor

greens:


dkirker (approved):
Forgot about this one... I'm surprised no one was screaming about it.

PR #3300, opened 2018-07-25
ICP-5435 Add call to Dome motion/light to perform initial check for all supported attributes

dkirker:
Some minor formatting, too.


greens (changes requested):
Wouldn't it be better to put this in installed? Right now it looks like the device will get queried twice on install. 

Also we should be sending a wakeupnomoreinformation after with a significant delay.

dkirker:
@greens  Change made

greens (approved):

PR #3312, opened 2018-07-30
Rachio Integration

tonesto7:


juano2310:
@tonesto7 @workingmonk is this a new integration or just a relocation of the old one?

tonesto7:
@juano2310 
This is update and relocation of the original

tonesto7:
@juano2310 It applies critical fixes to many bugs and addresses changes to Rachio's API that has broken certain parts of the official integration. It also adds support for multiple controllers

juano2310:
I don't see any apparent issue but it would be interesting, if possible, to make the same commit against the old repo to see the diff.

workingmonk:
@juano2310 we can take these files and open the PR against the private repo to see the diff

devicetypes/rachio/rachio-iro2-controller.src/rachio-iro2-controller.groovy
```diff
@@ -0,0 +1,674 @@
...
+def devVer() { return "2.0.0" }
+
+metadata {
+	definition (name: "Rachio Sprinkler Controller", namespace: "rachio", author: "Rachio") {
+		capability "Refresh"
+        capability "Switch"
+        capability "Actuator"
+        capability "Valve"
+        capability "Sensor"
+        capability "Polling"
```
marstorp, line 30: Don't use capability "Polling" as it isn't reliable calling the DTH poll method. 
Also as the DTH is calling the connect apps poll() method in the its poll() and that the connect app is calling poll() every 15 minutes, which is polling every device, there's no need for the DTH to do any additional polling.
To many invocations of the app may cause it to be rate limited and not run at all so try to minimize the load as much as possible.

devicetypes/rachio/rachio-iro2-controller.src/rachio-iro2-controller.groovy
```diff
@@ -0,0 +1,674 @@
...
+void installed() {
+	initialize()
+	state.isInstalled = true
+}
+
+void updated() {
+	initialize()
+}
+
+def ping() {
```
marstorp: ping() is only called for tracked devices, this DTH is enrolled as "untracked" thus ping is not called and should be removed

workingmonk:
@juano2310 and @marstorp, based on your comments in the private repo, lets merge this change and we can fix the comments in staging before this goes to prod on 8/14

juano2310:
@workingmonk  I'm fine with that.

workingmonk:
r

PR #3315, opened 2018-07-30
ICP-3042 Fan Controller DTH (#3117)

workingmonk:
cc @greens when the UI is ready


greens:
This has been verified.

PR #3316, opened 2018-07-31
Add "Switch" capability to Springs Window Fashions Shade DH

ericvitale:
Added the "Switch" capability along with the required on() and off() methods. This in conjunction with the "Switch Level" capability allows SmartApps that control switches and dimmers to also control devices that use this DH.


CosmicPuppy (commented):
I heartily approve of this. It makes compatibility with many SmartApps much easier, including Smart Lighting and ActionTiles. ...Terry.

nmelanson75:
This fix works awesome except when using with homebridge it shows up as a light switch dimmer. Any ways I can change it to a shade/blind?

PR #3321, opened 2018-07-31
ICP-5567 - Change Schlage device join names to suit different models

f-yoneyama:
Change-Id: If805650c3561ffdaecfae61cb8edf7910b90f20c


MAblewicz (approved):
Looks good

PR #3328, opened 2018-08-03
[WWST-1150] DTH for 4 endpoint Ecolink Siren

MMateusiakS:
@greens @MAblewiczS  @DCzarneckaS DTH ready for review

### DIFF
--- a/devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
+++ b/devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
@@ -14,7 +14,7 @@
  *
  */
 metadata {
-	definition (name: "Ecolink Wireless Siren", namespace: "SmartThings", author: "SmartThings", mnmn: "SmartThings", vid: "generic-switch") {
+	definition (name: "Ecolink Wireless Siren", namespace: "SmartThings", author: "SmartThings", mnmn: "SmartThings", vid: "generic-siren") {
 		capability "Actuator"
 		capability "Health Check"
 		capability "Switch"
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
```diff
@@ -0,0 +1,184 @@
...
+	} else {
+		String childDni = "${device.deviceNetworkId}-ep$srcEndpoint"
+		def child = childDevices.find { it.deviceNetworkId == childDni }
+
+		child?.handleZWave(encapsulatedCommand)
+	}
+}
+
+def on() {
+	sendEvent(name: "switch", value: "on")
```
greens: We shouldn't need to synthesize this event here. The `basicGet` should take care of that for us.

devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
```diff
@@ -0,0 +1,184 @@
...
+	
+	endPoints.each {
+		cmds << basicSetCmd(0x00, it)
+		cmds << basicGetCmd(it)
+	}
+	delayBetween(cmds, 50)
+}
+
+def addChildSwitches() {
+	for (i in 2..4) {
```
greens: you could use `endPoints.each` here

devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
```diff
@@ -0,0 +1,184 @@
...
+
+def basicSetCmd(value, endPoint) {
+	multiChannelCmdEncapCmd(zwave.basicV1.basicSet(value: value), endPoint)
+}
+
+def basicGetCmd(endPoint) {
+	multiChannelCmdEncapCmd(zwave.basicV1.basicGet(), endPoint)
+}
+
+def multiChannelCmdEncapCmd(cmd, endPoint) {
```
greens, line 176: this and `encap` appear to be the same function

devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
```diff
@@ -0,0 +1,184 @@
...
+def basicGetCmd(endPoint) {
+	multiChannelCmdEncapCmd(zwave.basicV1.basicGet(), endPoint)
+}
+
+def multiChannelCmdEncapCmd(cmd, endPoint) {	
+	def cmds = []
+	cmds << zwave.multiChannelV3.multiChannelCmdEncap(destinationEndPoint: endPoint).encapsulate(cmd).format()
+}
+
+def sendCommand(device, commands) {
```
greens: why not just have the child devices send their endpoint number instead of parsing it out of the DNI?

greens (changes requested):
I think you'll want to look at the zwave-siren DTH. It's important that we use the "Alarm" capability. It will essentially be redundant with "Switch", but it's important for how the device is presented and used. Speaking of which, using the generic-switch ui-metadata won't work, because that will cause the device to show up as a switch instead of an alarm/siren. Use the `generic-siren` metadata instead.

MMateusiakS:
Hey @greens , I have made changes according to your comments. Thanks for advice about Alarm capability! Waiting for second review.

devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
```diff
@@ -0,0 +1,206 @@
...
+
+def zwaveEvent(physicalgraph.zwave.Command cmd) {
+	[:]
+}
+
+def createEvents(value) {
+	
+	def val = "off"
+	if(value == 0) {
+		val = "off"
```
greens: you've already defined `val` as `"off"` here, so you can just check for `!= 0`

greens: you could use another ternary operator here like you do for `value1`

greens (approved):

PR #3333, opened 2018-08-06
CHAD-2368 Secure handling check

dkirker:


devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -533,7 +533,12 @@ def refresh() {
  * @returns ret: The encapsulated command
  */
 private secure(physicalgraph.zwave.Command cmd) {
-	zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
+	// Default to secure encapsulation, unless: zwaveInfo is defined and the device is secure and the command is one we should not encapsulte
+	if (!(zwaveInfo?.zw?.contains("s") && zwaveInfo.sec?.contains(String.format("%02X", cmd.commandClassId)))) {
```
kianooshST: You have two different implementation of the check, maybe I am missing something, but why not go with `f (!(zwaveInfo?.zw?.contains("s") && zwaveInfo.sec?.contains(String.format("%02X", cmd.commandClassId))))` across all DTHs?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -1413,7 +1413,12 @@ def deleteCode(codeID) {
  * @returns ret: The encapsulated command
  */
 private secure(physicalgraph.zwave.Command cmd) {
-	zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
+	// Default to secure encapsulation, unless: zwaveInfo is defined and the device is secure and the command is one we should not encapsulte
+	if (!(zwaveInfo?.zw?.contains("s") && zwaveInfo.sec?.contains(String.format("%02X", cmd.commandClassId)))) {
```
kianooshkarami: Why are we not checking for `state.sec` here and we are else where? Asking more out of curiosity.

greens: Older DTHs would set the `sec` variable as a flag when they successfully parsed an encrypted packet. I believe this was before we added the 's' to the end of a secure device's zwaveInfo object.

It's a deprecated practice that we still have to support for legacy reasons.

devicetypes/smartthings/zwave-metering-switch-secure.src/zwave-metering-switch-secure.groovy
```diff
@@ -296,19 +296,19 @@ private encapSequence(commands, delay=200) {
 private encap(physicalgraph.zwave.Command cmd) {
 	def secureClasses = [0x25, 0x5A, 0x70, 0x85, 0x8E]
```
greens: You can remove this.

Also, maybe there's a way to merge these two methods together now?

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
dkirker: At least in this case, or any case where I don't check it, it is probably because "state.sec" wasn't used anywhere in the code.

greens:
Are we going to get flak for not doing "answered as asked" for certain things (specifically wakeup?)

I can't think of a way to do it that isn't gross.

dkirker:
@greens  If it is a requirement I was going on the assumption that if a device indicated a command class as supported in the Secure Command Supported Report that it would only ever send data relating to it securely. It seems silly that it wouldn't... Given the "All non-securely supported command classes MUST also be supported securely. All non-securely controlled command classes MUST also be controlled securely." requirements, I am tempted to just use the logic:

`if device is secure
secure()
else
insecure()`

instead of checking zwaveInfo.sec[].

greens (dismissed):
You're probably good with it as is.

greens (dismissed):

greens (approved):
I think we can all agree that this is a mess and that automatic encapsulation determination can't come soon enough.

dkirker:
@greens Agreed...

PR #3334, opened 2018-08-06
ICP-5264 Fibaro Smoke Sensor not updating to custom ui-metadata

greens:


greens:
Kind of the nuclear option here. Reaching out to Doug to see if he can determine why the vid generation wouldn't be happening correctly.

dkirker:
That is weird...

dkirker (approved):
If the problem is super complex, then this should be fine. I don't think we have an equivalent generic either.

PR #3349, opened 2018-08-11
DVCSMP-4434 Fix SmartSense Temp/Humidity Sensor joining as a Thing

tpmanley:


varzac (approved):
Did something change?

varzac:
OK, so we still aren't sure why adding the orvibo fingerprint caused the generic one to stop working, but this should have the details correct for the old devices.

tpmanley:
@varzac it's due to the way fingerprinting works today. If any fingerprint in the DTH has manufacturer/model then they all need it.

PR #3351, opened 2018-08-13
DVCSMP-4418 Add generic metadata reference and icons for Inovelli sma…

greens:
…rt plug


tpmanley (approved):
I tested and the devices now show up in OneApp. They don't show up with the parent child relationship like they do in Classic but we know that's an issue that's being worked by another team.

tpmanley:
FYI @erocm123

PR #3360, opened 2018-08-14
WWST-1345 Z-Wave Battery Thermostat

greens:
New DTH for z-wave battery thermostats, specifically the TBZ48. Based almost entirely on the z-wave thermostat DTH, but with a longer checkinterval and no polling. The other DTH we have for these thermostats (ct100) supports humidity measurement and is therefore unsuitable.

### DIFF
--- a/devicetypes/smartthings/zwave-battery-thermostat.src/zwave-battery-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-battery-thermostat.src/zwave-battery-thermostat.groovy
@@ -1,0 +1,669 @@
+/**
+ *  Copyright 2018 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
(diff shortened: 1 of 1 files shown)


greens:
Happy to un-cruft this a little bit, but worried that too much un-crufting might break whatever hacks they've done in OneApp to get current z-wave thermostat DTH to work.

shrakuma:
can this be reviewed ASAP please

devicetypes/smartthings/zwave-battery-thermostat.src/zwave-battery-thermostat.groovy
```diff
@@ -0,0 +1,669 @@
...
+		attribute "thermostatFanState", "string"
+
+		command "switchMode"
+		command "switchFanMode"
+		command "lowerHeatingSetpoint"
+		command "raiseHeatingSetpoint"
+		command "lowerCoolSetpoint"
+		command "raiseCoolSetpoint"
+
+		fingerprint deviceId: "0x08"
```
mckeed: Should take this one out as it will put it in contention with zwave-thermostat as the fallback generic thermostat handler.

mckeed:
I don't see any periodic batteryGet. Probably want to add that in case the device doesn't send regular battery reports on its own. The CT-100 handler does it from parse which is pretty gross but probably has the benefit of piggybacking on the FLiRS stay-awake period.

greens:
I was hoping it just sent them periodically, but this is probably safer.

mckeed (approved):

dkirker (approved):

greens:
@workingmonk targeted this for staging since it's a new DTH for one device.

PR #3371, opened 2018-08-17
[CHAD-2217] Mark Everspring Flood Sensor for local execution

varzac:
This adds local execution to fibaro motion zw5 and everspring flood
sensor

### DIFF
--- a/devicetypes/smartthings/everspring-flood-sensor.src/everspring-flood-sensor.groovy
+++ b/devicetypes/smartthings/everspring-flood-sensor.src/everspring-flood-sensor.groovy
@@ -12,7 +12,7 @@
  *
  */
 metadata {
-	definition (name: "Everspring Flood Sensor", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.sensor.moisture") {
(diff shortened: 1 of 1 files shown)


tpmanley (dismissed):
We usually use 4 digit build numbers for minHubCoreVersion

greens (approved):

greens (changes requested):
Why target for staging? Probably better to aim for master unless this is time-sensitive?

Also this will break some functionality with the Fibaro due to their "multiStatus" tile.

varzac:
We want to it out next week

varzac:
Hmm, that's a good point.  @tpmanley do you think we should just pull the fibaro water out of 24?  So what is the behavior @greens, I'm not sure how the multiStatus thing works.

tpmanley:
@varzac yes let's pull it out if we're not confident it will work correctly.

varzac:
Updated

tpmanley (approved):

nathnelst:
The "Everspring Flood Sensor" has been tested with a self-published DTH and 23.4 build of HubCore and Water status was reported as expected, could be used in a locally executing smartapp, and a cloud executing smartapp.

varzac:
@tpmanley @workingmonk could one of you merge this?

greens (approved):

PR #3373, opened 2018-08-17
WWST-1402: Update zen-thermostat.groovy by splitting Thermostat capability.

munds:
This is a fix for WWST-1402, where the UI wasn't being updated properly due to mismatched href.

cc @tpmanley @workingmonk

### DIFF
--- a/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
+++ b/devicetypes/zenwithin/zen-thermostat.src/zen-thermostat.groovy
@@ -14,7 +14,11 @@ metadata {
         capability "Refresh"
         capability "Sensor"
         capability "Temperature Measurement"
-        capability "Thermostat"
+        capability "Thermostat Heating Setpoint"
+        capability "Thermostat Cooling Setpoint"
+        capability "Thermostat Fan Mode"
+        capability "Thermostat Mode"
+        capability "Thermostat Operating State"
 
         command "setpointUp"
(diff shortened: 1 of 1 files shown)


greens (approved):
Interested to see if this actually works fully with OneApp.

dkirker:
ui meta might need to be updated :/

dkirker (approved):

munds:
@dkirker - the UI metadata is already updated 👍 

@greens - agreed, I'm interested in that as well.

workingmonk:
the ticket does NOT mention why the capabilities are being split into individual pieces. You could have both the main capability and the split up one. 

cc @marstorp do we have any other DTH will both the main thermostat capability and the split up individual thermostat capabilities?

greens:
I think the short answer is that the monolithic capability is deprecated and ui-metadata uses the split ones.

workingmonk:
I am ok with the new additions, but is keeping the Thermostat capability hurting anything? Removing it will make it unusable in some smartapps that rely on that capability to search for devices.

greens:
Are any of those smartapps ours? Because if so, we should really update them to use non-deprecated capabilities.

workingmonk:
there are quite a few in our public and private repo, if you search for `capability.thermostat`

k-morawski:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3556

PR #3375, opened 2018-08-20
DVCSMP-4451 Fix for Schlage FE599 and similar locks

rboy1:
@greens @workingmonk 
Submitting a patch to fix an issue with non motorized locks which don't report re-locked events.

Can be unsecured multiple times using they keypad, each time it's unsecured using a code a notification should be sent (some locks don't sent re-secured notifications so this unlock event is not passed to SmartApps without this fix)


mckeed (approved):

dkirker (approved):

workingmonk:
@tpmanley @dkirker can you add a ticket number for QA to verify these changes?

tpmanley:
Done: DVCSMP-4451

PR #3398, opened 2018-08-23
[ICP-5791] Setting initial state for Ecolink Siren

MMateusiakS:
@greens @MAblewiczS  @DCzarneckaS @PKacprowiczS 

I missed setting initial state for One App. Sorry!

### DIFF
--- a/devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
+++ b/devicetypes/smartthings/ecolink-wireless-siren.src/ecolink-wireless-siren.groovy
@@ -14,7 +14,7 @@
  *
  */
 metadata {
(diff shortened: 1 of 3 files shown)


greens:
Shouldn't the `refresh()` call in `initialize()` be doing this?

I think you're probably just missing a `response()` call around the `refresh`.

MMateusiakS:
@greens  response() worked fine. New Vid with new icon make fixes to the reported issues. But still there is no switch  in OneApp in the device tile comparing to ST classic. Is this case of metadata to be changed a bit? Also child DTH was moved to new directory.

greens (approved):
The icons seem to be a cloud issue based on what I've been told. It may or may not get fixed.

PR #3399, opened 2018-08-23
WWST-1214 ICP-5800 Ecolink ZigBee firefighter

greens:
Fixes typo that was preventing ui metadata from being loaded


workingmonk:
This might create a new file, but since it is NOT in prod, I am ok with this change. But, zigbee did change their name from ZigBee to Zigbee, if that matters :)

greens:
Zig🅱️ee

workingmonk (approved):

PR #3413, opened 2018-08-28
[ICP-5888][ICP-5889] Fix for wrong Ecolink Firefighter's status after Hub being offline.

PKacprowiczS:
I've added one more attribute read in _refresh()_  to ensure that device will go back to proper status.
@greens @MAblewiczS please review.


greens:
You'll also need to set `offlinePingable` to `"1"` when you set the checkInterval.

Device Health only pings devices (which we decided were critical) where that is true.

greens:
Looks good, but I'll wait until we've verified 5888 and 5889 are fixed.

greens (approved):
Approved. Just waiting on verification of ICP-5888.

greens (approved):

PR #3428, opened 2018-08-31
WWST-1263: ZW121-A Aeotec LED Strip - US

joshk-trility:
Updated generic RGBW led to support the Aeotec LED strip.
Added multi-attribute tiles on details screen. Fixed some logging,
cleaned up unused tiles.
Also fixed occasional error when setting the color (WWST-1418).
Fixed color temperature calculations and controls.


joshk-trility:
@greens This should be final. Could you please review?

greens (approved):
This looks good. People might get a bit tripped up after the first conversion, but after that we should be okay.

PR #3431, opened 2018-09-03
set data value to notify hub use different message style 

FenMei007S:
@tpmanley @varzac   add updateDataValue


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -58,6 +58,7 @@ metadata {
 }
 
 def installed() {
+    updateDataValue("updatedOnOff", "true")
```
tpmanley: We decided to change the key to `"onOff"` and the value to `"catchall"`. Also put this in `updated` instead of installed. Finally the PR should be against staging since that's where the DTH was merged.

varzac:
This looks good, but I'm not sure exactly what's going on with it being 6 commits? I see the multi switch merged into staging with only 1 commit, so we should make sure that this is only 1 commit.  Sometimes github gets confused with the commit list though.

varzac:
You should squash commits, and you should also make sure to set this value in installed as well as updated like we discussed on the ticket.

varzac:
I created a PR and merged this in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3459

PR #3433, opened 2018-09-03
[ICP-5948] Updates DTH for Eaton accessory dimmer , add vid and mnmm

DCzarneckaS:
@MMateusiakS @greens @PKacprowiczS @MAblewiczS @MGoralczykS


greens:
This already has device-specific metadata from before generic metadata existed, but it looks like it never got pushed to prod. This will be a better long-term solution anyway.

greens (approved):

PR #3437, opened 2018-09-04
[WWST-1279] ZWA001-A Aeotec LED Bulb 6 Multi-white

joshk-trility:
Added new DTH for this device.

### DIFF
--- a/devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
+++ b/devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
@@ -17,7 +17,9 @@
  */
 
 metadata {
-	definition (name: "Aeon LED Bulb 6 Multi-White", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb") {
+	definition (name: "Aeon LED Bulb 6 Multi-White", namespace: "smartthings", author: "SmartThings",
+				ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb",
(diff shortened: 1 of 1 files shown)


joshk-trility:
@greens This is ready to go. Would you kindly review it?

joshk-trility:
@greens, I've updated this DTH as well to make it similar to the multi-color bulb DTH. Ready for review.

devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
```diff
@@ -0,0 +1,223 @@
...
+}
+
+def refresh() {
+	commands([zwave.switchMultilevelV3.switchMultilevelGet()] + queryAllColors(), 500)
+}
+
+def ping() {
+	log.debug "ping().."
+	unschedule(offlinePing)
+	runEvery5Minutes(offlinePing)
```
mckeed: Is there a requirement behind pinging this device every 5 minutes while offline? Pinging an offline Z-Wave device can tie up the entire Z-Wave system for up to several seconds so we try not to do it too much.

joshk-trility: This is a good question and it was something that @greens and I discussed. This bulb doesn't Hail or notify in any way when it comes back online (e.g. power is restored). Since the device health ping has an increasing backoff (up to a few hours if left overnight), then it can take a really long time for this bulb to be considered online again. This offline ping is a bandaid to deal with that.

greens: I think the 5 minutes is @mckeed 's concern. I probably should have considered this too. Maybe 30m?

joshk-trility: Done.

mckeed (approved):

PR #3438, opened 2018-09-04
ICP-5963 Add "Configuration" capability to zwave motion/light

greens:
Having this missing was preventing the checkInterval from being set in instances where updated() was not called. Updated was not being called in OneApp unless the "rename" button was pressed.


tpmanley (approved):

MAblewicz (commented):
Looks good

PR #3465, opened 2018-09-11
WWST-1595 Zigbee Accessory Dimmer

greens:
Primary intention is to work with the Aurora Wireless Wall Remote

### DIFF
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -1,5 +1,5 @@
 /**
- *  Copyright 2015 SmartThings
+ *  Copyright 2018 SmartThings
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  *  in compliance with the License. You may obtain a copy of the License at:
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -15,7 +15,7 @@
 import physicalgraph.zigbee.zcl.DataType
 
 metadata {
-	definition (name: "ZigBee Accessory Dimmer", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller", runLocally: false, minHubCoreVersion: '000.019.00012', executeCommandsLocally: false, mnmn: "SmartThings", vid: "generic-dimmer") {
+	definition (name: "ZigBee Accessory Dimmer", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller", mnmn: "SmartThings", vid: "generic-dimmer") {
 		capability "Actuator"
 		capability "Switch"
 		capability "Button"
@@ -113,6 +113,7 @@ def on() {
 
 def setLevel(value) {
 	if (value != 0) sendEvent(name: "switch", value: "on")
+	else if (value == 0) sendEvent(name: "switch", value: "off")
 	sendEvent(name: "level", value: value)
 }
 
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -21,7 +21,6 @@ metadata {
 		capability "Switch"
 		capability "Button"
 		capability "Switch Level"
-		capability "Health Check"
 
 		fingerprint profileId: "0104", inClusters: "0000,0003", outClusters: "0000,0004,0003,0006,0008,0005", manufacturer: "Aurora", model: "Remote50AU", deviceJoinName: "Aurora Wireless Wall Remote"
 	}
@@ -123,22 +122,11 @@ def setLevel(value) {
 	if (value != 0) sendEvent(name: "switch", value: "on")
 	sendEvent(name: "level", value: value)
 }
-/**
- * PING is used by Device-Watch in attempt to reach the Device
- * */
-def ping() {
-	return refresh()
-}
-
-def refresh() {
-	zigbee.readAttribute(0x0000, 0x0000)
-}
 
 def installed() {
 	sendEvent(name: "switch", value: "off", isStateChange: false, displayed: false)
 	sendEvent(name: "level", value: 0, isStateChange: false, displayed: false)
 	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
-	sendEvent(name: "checkInterval", value: 24 * 60 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 }
 
 def configure() {
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -62,8 +62,6 @@ def parse(String description) {
 					sendEvent(name: "level", value: STEP)
 				}
 				sendEvent(name: "switch", value: device.currentValue("switch") == "on" ? "off" : "on")
-			} else {
-				log.warn "ON/OFF REPORTING CONFIG FAILED- error code:${cluster.data[0]}"
 			}
 		} else if (descMap && descMap.clusterInt == 0x0008) {
 			def currentLevel = device.currentValue("level") as Integer ?: 0
@@ -113,18 +111,20 @@ def on() {
 }
 
 def setLevel(value) {
-	if (value != 0) sendEvent(name: "switch", value: "on")
-	else if (value == 0) sendEvent(name: "switch", value: "off")
-	sendEvent(name: "level", value: value)
+	if (value == 0) sendEvent(name: "switch", value: "off")
+	else sendEvent(name: "switch", value: "on") {
+		sendEvent(name: "level", value: value)
+	}
 }
 
 def installed() {
-	sendEvent(name: "switch", value: "off", isStateChange: false, displayed: false)
-	sendEvent(name: "level", value: 0, isStateChange: false, displayed: false)
+	sendEvent(name: "switch", value: "on", isStateChange: false, displayed: false)
+	sendEvent(name: "level", value: 100, isStateChange: false, displayed: false)
 	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
 }
 
 def configure() {
     // strangely, these are necessary to have the device report when its buttons are pressed
-	zigbee.onOffConfig() + zigbee.levelConfig() + zigbee.configureReporting(0x0005, 0x0001, DataType.UINT8, 1, 3600, null)
+	zigbee.addBinding(0x0006) + zigbee.addBinding(0x0008) + zigbee.addBinding(0x0005)
+//	zigbee.onOffConfig() + zigbee.levelConfig() + zigbee.configureReporting(0x0005, 0x0001, DataType.UINT8, 1, 3600, null)
 }
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -1,0 +1,154 @@
+/**
+ *  Copyright 2015 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition (name: "ZigBee Accessory Dimmer", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller", runLocally: false, minHubCoreVersion: '000.019.00012', executeCommandsLocally: false, mnmn: "SmartThings", vid: "generic-dimmer") {
+		capability "Actuator"
+		capability "Configuration"
+		capability "Switch"
+		capability "Button"
+		capability "Switch Level"
+		capability "Health Check"
+
+		fingerprint profileId: "0104", inClusters: "0000,0003", outClusters: "0000,0004,0003,0006,0008,0005", manufacturer: "Aurora", model: "Remote50AU", deviceJoinName: "Aurora Wireless Wall Remote"
+	}
+
+	tiles(scale: 2) {
+		multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
+			tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
+				attributeState "on", label:'${name}', action:"switch.off", icon:"st.switches.light.on", backgroundColor:"#00A0DC", nextState:"turningOff"
+				attributeState "off", label:'${name}', action:"switch.on", icon:"st.switches.light.off", backgroundColor:"#ffffff", nextState:"turningOn"
+				attributeState "turningOn", label:'${name}', action:"switch.off", icon:"st.switches.light.on", backgroundColor:"#00A0DC", nextState:"turningOff"
+				attributeState "turningOff", label:'${name}', action:"switch.on", icon:"st.switches.light.off", backgroundColor:"#ffffff", nextState:"turningOn"
+			}
+			tileAttribute ("device.level", key: "SLIDER_CONTROL") {
+				attributeState "level", action:"switch level.setLevel"
+			}
+		}
+		main "switch"
+		details(["switch"])
+	}
+}
+
+def getSTEP() {10}
+
+// Parse incoming device messages to generate events
+def parse(String description) {
+	log.debug "description is $description"
+
+	def event = zigbee.getEvent(description)
+	if (event) {
+		if (event.name=="level" && event.value==0) {}
+		else {
+			sendEvent(event)
+		}
+	} else {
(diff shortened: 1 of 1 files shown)


greens:
Here's the first draft.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,154 @@
...
+					}
+				}
+			} else if (descMap.commandInt == 0x01) {
+//				def step = descMap.data[0] == "00" ? 10 : -10
+				sendEvent(name: "level", value: descMap.data[0] == "00" ? 100 : STEP)
+				sendEvent(name: "switch", value: "on" )
+//				def events = []
+//                for (currentLevel..target).step(step).each{
+//					events << createEvent(name: "level", value: currentLevel + it)
+//				}
```
greens: aborted attempts at the faked intermediate step events

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,154 @@
...
+					sendEvent(name: "level", value: value)
+				} else if (descMap.data[0] == "01") {
+					log.debug "move down"
+					value = Math.max(currentLevel - STEP, 0)
+					sendEvent(name: "level", value: value)
+					if (value == 0) {
+						sendEvent(name: "switch", value: "off")
+					}
+				}
+			} else if (descMap.commandInt == 0x01) {
```
greens, line 85: we're essentially turning the middle buttons (when held) into "go to max" and "go to min"

greens: with nothing in-between

fine tuning can be handled by the single press events

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,142 @@
...
+			sendEvent(event)
+		}
+	} else {
+		def descMap = zigbee.parseDescriptionAsMap(description)
+		if (descMap && descMap.clusterInt == 0x0006) {
+			if (descMap.commandInt == 0x01 || descMap.commandInt == 0x00) {
+				if (device.currentValue("level") == 0) {
+					sendEvent(name: "level", value: STEP)
+				}
+				sendEvent(name: "switch", value: device.currentValue("switch") == "on" ? "off" : "on")
```
tpmanley, line 64: This device has explicit On and Off buttons, correct? If so I think it may be better to create the `on` event on receipt of the `On` command rather than toggling. Should also force each event to be a state change event.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,142 @@
...
+				def value = Math.min(currentLevel + STEP, 100)
+				log.debug "move to ${descMap.data}"
+				if (descMap.data[0] == "00") {
+					log.debug "move up"
+					sendEvent(name: "switch", value: "on")
+					sendEvent(name: "level", value: value)
+				} else if (descMap.data[0] == "01") {
+					log.debug "move down"
+					value = Math.max(currentLevel - STEP, 0)
+					sendEvent(name: "level", value: value)
```
tpmanley: We usually don't create level events for a value of 0 because when they are turned on it is still 0.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,142 @@
...
+			}
+		} else {
+			log.warn "DID NOT PARSE MESSAGE for description : $description"
+			log.debug "${descMap}"
+		}
+	}
+}
+
+def off() {
+	sendEvent(name: "switch", value: "off")
```
tpmanley: Should add `isStateChange: true` to all these events

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,142 @@
...
+def installed() {
+	sendEvent(name: "switch", value: "off", isStateChange: false, displayed: false)
+	sendEvent(name: "level", value: 0, isStateChange: false, displayed: false)
+	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
+}
+
+def configure() {
+	log.debug "Configuring Reporting, IAS CIE, and Bindings."
+	def cmds = []
+	return zigbee.onOffConfig() +
```
tpmanley: I don't think these configs will do anything since the device doesn't have On/Off or Level cluster servers, just clients.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,142 @@
...
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition (name: "ZigBee Accessory Dimmer", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller", runLocally: false, minHubCoreVersion: '000.019.00012', executeCommandsLocally: false, mnmn: "SmartThings", vid: "generic-dimmer") {
+		capability "Actuator"
+		capability "Configuration"
+		capability "Switch"
+		capability "Button"
+		capability "Switch Level"
+
```
tpmanley, line 24: If we want to add Health Check might be able to setup attribute reporting on any implemented attribute in the Basic cluster

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: Only problem is that then we have to maintain the state and there's no way to get that from the device. (and it's not indicated on the device anywhere)

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: I did a little futzing about to make sure that if you turn it on from level = 0, level goes up to 10, but avoiding that situation in the first place is probably the better choice.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: I tried doing read attributes of basic cluster, but wasn't getting a response.

greens:
Updated. Still can't get the device to report any attributes in the basic cluster, so I don't think this device can support health check.

greens:
Waiting for a final review from somebody.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,123 @@
...
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition (name: "ZigBee Accessory Dimmer", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller", runLocally: false, minHubCoreVersion: '000.019.00012', executeCommandsLocally: false, mnmn: "SmartThings", vid: "generic-dimmer") {
```
tpmanley: Can remove `minHubCoreVersion` and `executeCommandsLocally` since it doesn't run locally.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
tpmanley: Does this device have separate on and off buttons or is it just a single button that toggles between on/off?

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,123 @@
...
+def off() {
+	sendEvent(name: "switch", value: "off", isStateChange: true)
+}
+
+def on() {
+	sendEvent(name: "switch", value: "on", isStateChange: true)
+}
+
+def setLevel(value) {
+	if (value != 0) sendEvent(name: "switch", value: "on")
```
tpmanley: Need to set `switch` to `off` if level is 0

greens:
Done.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: a single button

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,124 @@
...
+		}
+	} else {
+		def descMap = zigbee.parseDescriptionAsMap(description)
+		if (descMap && descMap.clusterInt == 0x0006) {
+			if (descMap.commandInt == 0x01 || descMap.commandInt == 0x00) {
+				if (device.currentValue("level") == 0) {
+					sendEvent(name: "level", value: STEP)
+				}
+				sendEvent(name: "switch", value: device.currentValue("switch") == "on" ? "off" : "on")
+			} else {
```
tpmanley: I think this `else` can be removed

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,124 @@
...
+	sendEvent(name: "switch", value: "off", isStateChange: true)
+}
+
+def on() {
+	sendEvent(name: "switch", value: "on", isStateChange: true)
+}
+
+def setLevel(value) {
+	if (value != 0) sendEvent(name: "switch", value: "on")
+	else if (value == 0) sendEvent(name: "switch", value: "off")
```
tpmanley: Could simplify to just `else`

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,124 @@
...
+	sendEvent(name: "level", value: value)
```
tpmanley: We usually don't set level to 0

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,124 @@
...
+}
+
+def installed() {
+	sendEvent(name: "switch", value: "off", isStateChange: false, displayed: false)
+	sendEvent(name: "level", value: 0, isStateChange: false, displayed: false)
```
tpmanley: Usually the default level is 100

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,124 @@
...
+	def event = zigbee.getEvent(description)
+	if (event) {
+		if (event.name=="level" && event.value==0) {}
+		else {
+			sendEvent(event)
+		}
+	} else {
+		def descMap = zigbee.parseDescriptionAsMap(description)
+		if (descMap && descMap.clusterInt == 0x0006) {
+			if (descMap.commandInt == 0x01 || descMap.commandInt == 0x00) {
```
tpmanley, line 60: Does it use both of these commands? Seems odd if it has one button that can generate two different commands.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -122,3 +123,8 @@ def installed() {
 	sendEvent(name: "level", value: 0, isStateChange: false, displayed: false)
 	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
 }
+
+def configure() {
+    // strangely, these are necessary to have the device report when its buttons are pressed
+	zigbee.onOffConfig() + zigbee.levelConfig() + zigbee.configureReporting(0x0005, 0x0001, DataType.UINT8, 1, 3600, null)
```
tpmanley: Guessing you could call `zigbee. addBinding(0x0006)`, etc.... Since this device doesn't have reportable attributes it's only the binding that should matter.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: It does, but I think it's for talking to actual devices that are keeping on/off so it keeps track of state internally

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
greens: yep, that was it

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,132 @@
...
+def installed() {
+	sendEvent(name: "switch", value: "on", isStateChange: false, displayed: false)
+	sendEvent(name: "level", value: 100, isStateChange: false, displayed: false)
+	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
+}
+
+def configure() {
+    // strangely, these are necessary to have the device report when its buttons are pressed
+	zigbee.addBinding(0x0006) + zigbee.addBinding(0x0008) + zigbee.addBinding(0x0005)
+//	zigbee.onOffConfig() + zigbee.levelConfig() + zigbee.configureReporting(0x0005, 0x0001, DataType.UINT8, 1, 3600, null)
```
varzac: You can remove this commented line.  You could also change the above to use `zigbee.ONOFF_CLUSTER`, `zigbee.LEVEL_CONTROL_CLUSTER` for the first 2.  We don't have a constant currently for the scenes cluster though.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
varzac: I'm wondering if we want to track state separately or just use what the device gives us.  If they have other bulbs that they are controlling directly from the remote it would be weird for us to be generating an "off" event when the bulbs are actually getting an on.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
varzac: As I commented above, we can use the command it sends us to know.  If the command ID is `0x01` it's an on and `0x00` is an off.  I think we could consider the argument that we should explicitly mark the event with `isStateChange: true` as well, but I don't feel too strongly on that.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
tpmanley: I agree, if it's tracking state on it's own I think we should use that.

greens: The intended operation with Aurora's hub is to control other lights, I assume.

I don't know any way they'd be able to do this on our system, though. And there's no outward indication on the device or any way for us to determine it's internal state.

Tracking the device's on/off state, but just having a purely invented level value seems counterintuitive to me and also removes the ability to use the app as a "virtual" on/off switch.

tpmanley: 👍  I see your point

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,131 @@
...
+	} else {
+		sendEvent(name: "switch", value: "on")
+        sendEvent(name: "level", value: value)
+	}
+}
+
+def installed() {
+	sendEvent(name: "switch", value: "on", isStateChange: false, displayed: false)
+	sendEvent(name: "level", value: 100, isStateChange: false, displayed: false)
+	sendEvent(name: "button", value: "pressed", isStateChange: false, displayed: false)
```
tpmanley, line 125: I think you need to create a `numberOfButtons` event too.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,131 @@
...
+}
+
+def configure() {
+    // strangely, these are necessary to have the device report when its buttons are pressed
```
tpmanley: nit: indentation looks off with the line below

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -0,0 +1,131 @@
...
+
+def on() {
+	sendEvent(name: "switch", value: "on", isStateChange: true)
+}
+
+def setLevel(value) {
+	if (value == 0) {
+		sendEvent(name: "switch", value: "off")
+	} else {
+		sendEvent(name: "switch", value: "on")
```
tpmanley, line 117: nit: indentation looks off

tpmanley (dismissed):
Just a couple of final minor things and then it looks good

tpmanley (approved):

PR #3467, opened 2018-09-12
ONEAPP-7061 Add Light capability to the GE Link Bulb DTH

tpmanley:
This is being done so that GE Link bulbs are available alert targets for SHM.

https://smartthings.atlassian.net/browse/ONEAPP-7061


tpmanley:
I still need to verify this before we merge.

tpmanley:
I confirmed this works in SHM.

varzac (approved):

PR #3472, opened 2018-09-12
WWST-1258: ZW140-A Aeotec Dual Nano Switch - US

joshk-trility:
Added Dual Nano Switch support to the generic zwave dual switch DTH.
Updated said DTH to support security encapsulation. Fixed issue where
parent device would show "on" when child device was activated.
Fixed an issue where the child device would not correctly report as
offline.

### DIFF
--- a/devicetypes/smartthings/zwave-binary-switch-endpoint.src/zwave-binary-switch-endpoint.groovy
+++ b/devicetypes/smartthings/zwave-binary-switch-endpoint.src/zwave-binary-switch-endpoint.groovy
@@ -51,7 +51,7 @@ def updated() {
 
 def configure() {
 	// Device-Watch simply pings if no device events received for checkInterval duration of 32min
-	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: parent.hubID])
 	refresh()
 }
 
--- a/devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
+++ b/devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
@@ -143,20 +143,20 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 }
 
 def on() {
-	// parent DTH conrols endpoint 1
+	// parent DTH controls endpoint 1
 	def endpointNumber = 1
 	delayBetween([
-		command(encap(endpointNumber, zwave.switchBinaryV1.switchBinarySet(switchValue: 0xFF))),
-		command(encap(endpointNumber, zwave.switchBinaryV1.switchBinaryGet()))
+		encap(endpointNumber, zwave.switchBinaryV1.switchBinarySet(switchValue: 0xFF)),
+		encap(endpointNumber, zwave.switchBinaryV1.switchBinaryGet())
 	])
 }
 
 def off() {
-	// parent DTH conrols endpoint 1
+	// parent DTH controls endpoint 1
 	def endpointNumber = 1
 	delayBetween([
-		command(encap(endpointNumber, zwave.switchBinaryV1.switchBinarySet(switchValue: 0x00))),
-		command(encap(endpointNumber, zwave.switchBinaryV1.switchBinaryGet()))
+		encap(endpointNumber, zwave.switchBinaryV1.switchBinarySet(switchValue: 0x00)),
+		encap(endpointNumber, zwave.switchBinaryV1.switchBinaryGet())
 	])
 }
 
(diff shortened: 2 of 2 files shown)


devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -88,22 +95,32 @@ def parse(String description) {
...
+		log.warn "Unable to extract encapsulated cmd from $cmd"
+		createEvent(descriptionText: cmd.toString())
+	}
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.multichannelv3.MultiChannelCmdEncap cmd) {
 	def encapsulatedCommand = cmd.encapsulatedCommand([0x32: 3, 0x25: 1, 0x20: 1])
 	if (cmd.sourceEndPoint == 1) {
-		zwaveEvent(encapsulatedCommand)
+		zwaveEvent(encapsulatedCommand, 1)
```
greens, line 123: Trying to understand why this (and all the signatures) was changed?

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -164,13 +181,13 @@ def sendCommand(endpointDevice, commands) {
...
 	}
-	result = commands.collect { encap(endpointNumber, it) }
+	result = commands.collect { command(encap(endpointNumber, it)) }
 	sendHubCommand(result, 100)
 }
 
 def encap(endpointNumber, cmd) {
 	if (cmd instanceof physicalgraph.zwave.Command) {
-		zwave.multiChannelV3.multiChannelCmdEncap(destinationEndPoint: endpointNumber).encapsulate(cmd).format()
+		zwave.multiChannelV3.multiChannelCmdEncap(destinationEndPoint: endpointNumber).encapsulate(cmd)
```
greens: at this point you might be better served just wrapping the result of this function in a `command()` call since it's used every place `encap()` is.

devicetypes/smartthings/zwave-binary-switch-endpoint.src/zwave-binary-switch-endpoint.groovy
```diff
@@ -51,7 +51,7 @@ def updated() {
 
 def configure() {
 	// Device-Watch simply pings if no device events received for checkInterval duration of 32min
-	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: parent.hubID])
```
greens, line 54: this works?

neat

joshk-trility: I was referred to the Fibaro dual switch implementation for this solution.

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
joshk-trility: Each time the child device was turned on/off, the switch would generate a BasicReport that wasn't associated with either endpoint. This caused, for example, endpoint 1 to flip state when endpoint 2 was actuated. Since we're only interested in events from the endpoints, we filter out the non-associated ones.
There are probably a few ways to go about solving that problem, but this was the quickest. It also mimics how it's done in the Innovelli dual switch DTH.

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
joshk-trility: Yes, that would be a bit cleaner. I'll fix that up.

joshk-trility:
Addressed comment about `command()` calls.

joshk-trility:
BTW, this is still blocked by WWST-1660.

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -88,22 +95,32 @@ def parse(String description) {
...
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd, endpoint=null) {
+	(endpoint == 1) ? [name: "switch", value: cmd.value ? "on" : "off"] : [:]
 }
 
-def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
-	[name: "switch", value: cmd.value ? "on" : "off"]
+def zwaveEvent(physicalgraph.zwave.commands.switchbinaryv1.SwitchBinaryReport cmd, endpoint=null) {
+	(endpoint == 1) ? [name: "switch", value: cmd.value ? "on" : "off"] : [:]
```
greens, line 107: I'll need to verify that the Coolcam switch wasn't reporting on/off for the main device without encapsulation. If so, this might break the implementation for that device.

greens (approved):

PR #3476, opened 2018-09-14
Making all runIn callbacks use public methods in the Fibaro DTHs

tpmanley:
The recent Grails/Groovy update requires runIn callbacks to be public or else they won't run.


greens (approved):

tpmanley:
In my testing the `runIn` callbacks actually were running even without these changes which is unexpected. That being said this change _should_ be required and it doesn't hurt anything to make it.

PR #3477, opened 2018-09-14
DVCSMP-4472 Aeon Multisensor 6 not finishing configure if reportInterval is not set

tpmanley:
Previously an exception was being thrown if the reportInterval setting was not set. Apparently it's not enough to check for "truthiness" of a setting so instead check if it's in the timeOptionValueMap and default to 1 hour if it is not.

https://smartthings.atlassian.net/browse/DVCSMP-4472

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
@@ -442,10 +442,7 @@ def configure() {
 
 	// set the check interval based on the report interval preference. (default 122 minutes)
 	// we do this here in case the device is in wakeup mode
-	def checkInterval = 2 * 60 * 60 + 2 * 60
-	if (reportInterval) {
-		checkInterval = 2 * timeOptionValueMap[reportInterval] + (2 * 60)
-	}
+	def checkInterval = 2 * (timeOptionValueMap[reportInterval] ?: 60 * 60) + 2 * 60
 	sendEvent(name: "checkInterval", value: checkInterval, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -442,10 +442,7 @@ def configure() {
 
 	// set the check interval based on the report interval preference. (default 122 minutes)
 	// we do this here in case the device is in wakeup mode
-	def checkInterval = 2 * 60 * 60 + 2 * 60
-	if (reportInterval) {
-		checkInterval = 2 * timeOptionValueMap[reportInterval] + (2 * 60)
-	}
+	def checkInterval = 2 * (timeOptionValueMap[reportInterval] ?: 60 * 60) + 2 * 60
```
tpmanley, line 445: @greens any idea why the default previously used here isn't the same as what's used here: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/732a0287331fa208854a195357c77e6d436827f2/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy#L412?

greens: no reason other than 8 minutes is a pretty short duration to fallback to

greens (approved):

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
tpmanley: I feel like these values should probably be the same but for now I'm going to leave it as is and at least get this fix out.

varzac (approved):

PR #3432, opened 2018-09-03
[ICP-5919] Edited configuration for Ecolink Door/Window Sensor

PKacprowiczS:
@greens @MAblewiczS @MMateusiakS @DCzarneckaS @MGoralczykS 
Please have a look and tell me what you think.

PR #3487, opened 2018-09-17
ICP-5919 Make temperature report less frequently than IAS so that ope…

greens:
…n/closed can bring device online

### DIFF
--- a/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
@@ -183,5 +183,5 @@ def configure() {
 
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -183,5 +183,5 @@ def configure() {
 
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	return refresh() + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 300) // send refresh cmds as part of config
+	return refresh() + zigbee.iasZoneConfig(60, 300) + zigbee.batteryConfig() + zigbee.temperatureConfig(120, 300) // send refresh cmds as part of config
```
varzac, line 186: The minReportingInterval can stay the same for temp, it's the max that matters.  I would leave the min at 30, but max up to 600.  @tpmanley that work for you?

varzac:
Any reason not to do this as a part of https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3432/files?  Or is the intent to replace that PR?

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
tpmanley: I think you could go even higher with the max for temp like 3600 (an hour) seems good to me. It's going to get pushed immediately if it changes and I don't see much benefit of sending it more frequently.

greens:
Just slipped my mind that there was already a PR

PR #3491, opened 2018-09-18
WWST-1670: ZW095-A Home Energy Meter Gen5 2P, 200A

joshk-trility:
Updated the aeon-home-energy-meter DTH to support the Aeotec Home
Energy Meter Gen 5.

### DIFF
--- a/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
+++ b/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
@@ -72,6 +72,12 @@ metadata {
 def installed() {
 	log.debug "installed()..."
 	sendEvent(name: "checkInterval", value: 1860, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "0"])
+	response(refresh())
+}
+
+def updated() {
+	log.debug "updated()..."
+	response(refresh())
 }
 
 def ping() {
@@ -89,14 +95,35 @@ def parse(String description) {
 	return result
 }
 
-def zwaveEvent(physicalgraph.zwave.commands.meterv1.MeterReport cmd) {
-	if (cmd.scale == 0) {
-		[name: "energy", value: cmd.scaledMeterValue, unit: "kWh"]
-	} else if (cmd.scale == 1) {
-		[name: "energy", value: cmd.scaledMeterValue, unit: "kVAh"]
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
+	def encapsulatedCommand = cmd.encapsulatedCommand()
+	if (encapsulatedCommand) {
+		zwaveEvent(encapsulatedCommand)
+	} else {
+		log.warn "Unable to extract encapsulated cmd from $cmd"
+		createEvent(descriptionText: cmd.toString())
 	}
-	else {
-		[name: "power", value: Math.round(cmd.scaledMeterValue), unit: "W"]
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.meterv1.MeterReport cmd) {
+	meterReport(cmd.scale, cmd.scaledMeterValue)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.meterv2.MeterReport cmd) {
+	meterReport(cmd.scale, cmd.scaledMeterValue)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.meterv3.MeterReport cmd) {
+	meterReport(cmd.scale, cmd.scaledMeterValue)
+}
+
+private meterReport(scale, value) {
+	if (scale == 0) {
+		[name: "energy", value: value, unit: "kWh"]
+	} else if (scale == 1) {
+		[name: "energy", value: value, unit: "kVAh"]
+	} else {
+		[name: "power", value: Math.round(value), unit: "W"]
 	}
 }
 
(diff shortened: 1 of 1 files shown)


joshk-trility:
@greens Another DTH ready for review!

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -92,29 +133,58 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
...
-		zwave.configurationV1.configurationSet(parameterNumber: 113, size: 4, scaledConfigurationValue: 300).format() // every 5 min
-	])
-	log.debug cmd
-	cmd
+	log.debug "configure()..."
+	if (zwaveInfo.model.equals("005F"))
+		delayBetween([
+			// Send combined power in watts to report group 1 every 5 seconds (if difference exceeds 10% or 50W)
+			encap(zwave.configurationV1.configurationSet(parameterNumber: 101, size: 4, scaledConfigurationValue: 2)),
+			encap(zwave.configurationV1.configurationSet(parameterNumber: 111, size: 4, scaledConfigurationValue: 5)),
```
greens: Report only happens if the change is more than 10%, right? Worried about some case where we're getting a report every 5s.

greens (approved):

greens (dismissed):

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
PKacprowiczS: Actually it is causing report every 5s with no 10% change.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -75,14 +95,35 @@ def parse(String description) {
 	return result
 }
 
-def zwaveEvent(physicalgraph.zwave.commands.meterv1.MeterReport cmd) {
-	if (cmd.scale == 0) {
-		[name: "energy", value: cmd.scaledMeterValue, unit: "kWh"]
-	} else if (cmd.scale == 1) {
-		[name: "energy", value: cmd.scaledMeterValue, unit: "kVAh"]
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
```
PKacprowiczS, line 98: Incoming Crc16Encap commands aren't being handled. According to http://manuals-backend.z-wave.info/make.php?lang=en&sku=ZW095-A&cert=ZC10-15040002 they're disabled by default, but in my case I've received these kind of commands. So, we may add handler for them, manually disable it (parameter 13) or make sure device is in factory default settings (parameter 255) just after pairing and before our configuration.
Personally, I'd go for 1st option.

PKacprowiczS (changes requested):
Generally it is working, but there's some issues that need to be addressed.
@MAblewiczS

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
joshk-trility: I was seeing something similar with the energy value being reported too often, but not for the power. If we back off the reporting period for the power, we don't get any MeterReports until that period is met. For example, In our testing, when we turned off our test load, the app would not reflect the power change until the next reporting period (~5 minutes).
I'll look into this some more.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
joshk-trility: Based on this comment and the above one about too many MeterReports, I'm wondering if we really should just set this device to factory settings when installed. It sounds like CRC-16 was enabled for whatever reason on your device and maybe the MeterReport threshold was also reduced to a very noisy level.
I'll take a look into this a bit more.

joshk-trility:
By restoring the device to the default factory settings, we should avoid any unexpected configuration which would result in a game of whack-a-mole with fixes. I reviewed the default settings from the manual and deemed them sufficient for our purposes. I'm willing, of course, to discuss the defaults further if we believe they will not do what we want.
@PKacprowiczS  This should address the excessive power reports you're seeing, as I saw that it fixed the frequent energy reports as well in my test setup. Additionally, we don't need to encapsulate CRC-16 commands since it won't be enabled.
@greens Ready for another review.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
greens: Yeah. Go ahead and add a CRC unencapsulation method just in case.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -92,29 +133,54 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
...
-		zwave.configurationV1.configurationSet(parameterNumber: 103, size: 4, scaledConfigurationValue: 0).format(),    // no third report
-		zwave.configurationV1.configurationSet(parameterNumber: 113, size: 4, scaledConfigurationValue: 300).format() // every 5 min
-	])
-	log.debug cmd
-	cmd
+	log.debug "configure()..."
+	if (zwaveInfo.model.equals("005F"))
+		delayBetween([
+			// Reset the device to the default settings
+			encap(zwave.configurationV1.configurationSet(parameterNumber: 255, size: 4, scaledConfigurationValue: 1))
```
greens, line 174: Looking at the manual it's not entirely clear what the default reporting interval is for the first two groups. Is it 5 minutes?

joshk-trility: Group 1 defaults to 1 minute, Group 2 is 2 minutes.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
joshk-trility: Done

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
greens: can we set them to 5 so we maintain the same reporting interval across devices?

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -75,14 +95,52 @@ def parse(String description) {
...
+	]
+	def version = versions[cmd.commandClass as Integer]
+	def ccObj = version ? zwave.commandClass(cmd.commandClass, version) : zwave.commandClass(cmd.commandClass)
+	def encapsulatedCommand = ccObj?.command(cmd.command)?.parse(cmd.data)
+	if (encapsulatedCommand) {
+		zwaveEvent(encapsulatedCommand)
 	}
-	else {
-		[name: "power", value: Math.round(cmd.scaledMeterValue), unit: "W"]
+	[:]
```
PKacprowiczS, line 122: Leaving it here like that causes every Crc16Encap command return empty map instead of correct report.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
PKacprowiczS: Resetting device to default configuration seems to not work, or "default" is different from what is stated in the manual. I'm getting reports every 5 second, so I think we should do all of important configurations anyway and don't look at the 'defaults'.

PKacprowiczS (changes requested):
CRC 16 handler needs to be fixed.

k-morawski:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3525

PR #3497, opened 2018-09-20
ICP-6184: Fix NullPointerException when handling color events

joshk-trility:
Added check for uninitialized state for existing devices.


joshk-trility:
Wrong base branch.

PR #3500, opened 2018-09-21
WWST-1485: Modify multi switch fingerprint

FenMei007S:
@tpmanley @varzac  Dear Tom and Zach, the cooperate switch be changed to other model(it is also a 2 gang switch), so update the fingerprint. plz merage to china STG server


varzac (approved):

tpmanley:
Using https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3502 instead so it goes right to staging.

PR #3501, opened 2018-09-21
[WWST-1653; WWST-1654] New DTH - Z-Wave Panic Button

PKacprowiczS:
@greens @MAblewiczS @MMateusiakS @DCzarneckaS @MGoralczykS 
Right now I'd rather treat this as a sketch. I'm not sure if I should be using 'Child Button' DTH, because device's buttons do not support "Held" state.
Anyway, please tell what you think about it.

### DIFF
--- a/devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
+++ b/devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
@@ -45,16 +45,17 @@ metadata {
 
 def installed() {
 	log.debug "Installed $device.displayName"
-	sendEvent(name: "DeviceWatch-Enroll", value: [protocol: "zwave", scheme:"untracked"].encodeAsJson(), displayed: false)
+	//sendEvent(name: "DeviceWatch-Enroll", value: [protocol: "zwave", scheme:"untracked"].encodeAsJson(), displayed: false)
 	if(hasMultipleButtons()) {
 		addChildButtons()
 	}
-	setPanicAlarm("clear")
 	def cmds = [
 			zwave.batteryV1.batteryGet().format(),
 			zwave.notificationV3.notificationGet().format(),
+			"delay 1000",
 			zwave.wakeUpV1.wakeUpNoMoreInformation().format()
 	]
+	sendEvent(name: "panicAlarm", value: "clear", descriptionText: "Panic Alarm status: clear", displayed: true, isStateChange: true)
 	response(cmds)
 }
 
@@ -88,11 +89,13 @@ def zwaveEvent(physicalgraph.zwave.commands.sensorbinaryv2.SensorBinaryReport cm
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
 	def results = [createEvent(descriptionText: "$device.displayName woke up", isStateChange: false)]
 
-	def prevBattery = device.currentState("battery")
-	if (!prevBattery || (new Date().time - prevBattery.date.time)/60000 >= 60 * 53) {
+	if (!state.lastbatt || (now() - state.lastbatt) >= 56*60*60*1000) {
 		results += response(zwave.batteryV1.batteryGet().format())
 	}
-	results += response(zwave.wakeUpV1.wakeUpNoMoreInformation().format())
+	results += response([
+			"delay 1000",
+			zwave.wakeUpV1.wakeUpNoMoreInformation().format()
+	])
 	results
 }
 
@@ -118,7 +121,7 @@ private setPanicAlarm(status) {
 
 private setButtonState(buttonId, state) {
 	def event
-	String childDni = "${device.deviceNetworkId}/$buttonId"
+	String childDni = "${device.deviceNetworkId}:$buttonId"
 	def child = childDevices.find { it.deviceNetworkId == childDni }
 	if (!child) {
 		log.error "Child device $childDni not found"
@@ -132,7 +135,7 @@ private setButtonState(buttonId, state) {
 private addChildButtons() {
 	log.debug "Creating child devices"
 	for (i in 1..3) {
-		String childDni = "${device.deviceNetworkId}/$i"
+		String childDni = "${device.deviceNetworkId}:$i"
 		def child = childDevices.find { it.deviceNetworkId == childDni }
 		addChildDevice("Child Button", childDni, null, [
 				completedSetup: true,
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -0,0 +1,149 @@
...
+		child?.sendEvent(name: "button", value: "$state")
+		event = createEvent(descriptionText: "Button $buttonId was $state", isStateChange: true)
+	}
+	event
+}
+
+private addChildButtons() {
+	log.debug "Creating child devices"
+	for (i in 1..3) {
+		String childDni = "${device.deviceNetworkId}/$i"
```
varzac: We've been trying to be a bit more consistent with how we handle child DNIs and on the zigbee side we have settled on `<deviceDNI>:<endpoint>` so using the colon as a separator. Could you change this to be the same so we can try to have consistency across devices.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -0,0 +1,149 @@
...
+
+		main(["panic"])
+
+		details(["panic", "buttons", "battery"])
+	}
+}
+
+def installed() {
+	log.debug "Installed $device.displayName"
+	sendEvent(name: "DeviceWatch-Enroll", value: [protocol: "zwave", scheme:"untracked"].encodeAsJson(), displayed: false)
```
greens: this is deprecated

You'll want to set a checkInterval of (presumably) 8 hours and 2 minutes.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -0,0 +1,149 @@
...
+	if(hasMultipleButtons()) {
+		addChildButtons()
+	}
+	setPanicAlarm("clear")
+	def cmds = [
+			zwave.batteryV1.batteryGet().format(),
+			zwave.notificationV3.notificationGet().format(),
+			zwave.wakeUpV1.wakeUpNoMoreInformation().format()
```
greens, line 53: You should put a delay in before the `noMoreInformation` or the device might go to sleep before it reports the batteries and notification.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -0,0 +1,149 @@
...
+def zwaveEvent(physicalgraph.zwave.commands.sensorbinaryv2.SensorBinaryReport cmd) {
+	def status = cmd.sensorValue ? "panic" : "clear"
+	setPanicAlarm(status)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
+	def results = [createEvent(descriptionText: "$device.displayName woke up", isStateChange: false)]
+
+	def prevBattery = device.currentState("battery")
+	if (!prevBattery || (new Date().time - prevBattery.date.time)/60000 >= 60 * 53) {
```
greens: you can just use `now()`

Can you also move the `60*1000` to the other side of the comparison?

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -0,0 +1,149 @@
...
+		results += response(zwave.batteryV1.batteryGet().format())
+	}
+	results += response(zwave.wakeUpV1.wakeUpNoMoreInformation().format())
```
greens: you'll also want a delay before this one as well

greens:
The pushed/held/double thing in Child Button will create problems since the ui-metadata has to include all possible options even if they'll never occur, and I think those get passed to the automation tools. :/ 

worst case scenario we have to create a separate ui-metadata and cloned DTH for this, but I'll see if we can figure something else out.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
PKacprowiczS: Thing is, I don't know what wakeUpInterval this device has. When asked with *Wake Up Interval Get* it reports 0 seconds.

PKacprowiczS:
@greens further changes. Please have a glimpse when you can.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
PKacprowiczS: In 75983b0 commit I've just commented this line, as I'm not sure how to handle this "not-so-waking-up" case. 
I mean this device is waking up, but not in constant interval. I think it is happening when there wasn't any interaction with device for a long time, button is pushed and then device is sending wakeUpNotification. Although I'm not 100% sure about that, I'll need to investigate it further.

greens: We automatically set a wakeup interval of 4 hours to all devices that claim they support it during join. If it's not durable that makes me nervous. Watch hub logs during join and you should see the device's minimum and maximum wakeup interval values. It might be that it doesn't allow a value <= 4 hours.

devicetypes/smartthings/zwave-panic-button.src/zwave-panic-button.groovy
```diff
@@ -45,16 +45,17 @@ metadata {
...
 		addChildButtons()
 	}
-	setPanicAlarm("clear")
 	def cmds = [
 			zwave.batteryV1.batteryGet().format(),
 			zwave.notificationV3.notificationGet().format(),
+			"delay 1000",
 			zwave.wakeUpV1.wakeUpNoMoreInformation().format()
 	]
+	sendEvent(name: "panicAlarm", value: "clear", descriptionText: "Panic Alarm status: clear", displayed: true, isStateChange: true)
```
greens, line 55: was the notification report not taking care of this?

PKacprowiczS: nope, device doesn't answer to notificationGet at all. I should probably delete this command

greens: please do. other than that it looks fine

greens:
If the device is sending wakeup notifications every 4 hours like it should, you should set the checkInterval to 8 hours + 2 minutes.

PKacprowiczS:
Closing as this probably won't be certified anytime soo... never.

PR #3506, opened 2018-09-24
Wwst 1735 Zipato Siren loudness Change

MAblewicz:
@greens @mckeed @tpmanley please review this change. It is high priority request from Vodafone. Changing zipato siren loudness parameter from 2 to 3. It's MFR checked so it should only impact zipato siren.

@workingmonk Zipato requested this as priority change. Will it be possible to make this change to Prod this tuesday>


greens:
Moved to cleaner PR targeted at production here:

https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3507

PR #3517, opened 2018-09-25
WWST-1565 - August Smart Lock Pro fingerprint

MAblewicz:
@greens @tpmanley @dkirker could you review this. Only fingerprint change required.


dkirker (approved):

MAblewicz:
We will create new DTH for that lock because its has additional contact sensor

MAblewicz:
We will create new DTH for that lock because its has additional contact sensor

ericangST:
Due to the nature of Open/Close (working via Bluetooth), let's re-use existing DTH, it'll minimized and shorten development for this device. Main Functionality should be Open/Close and Battery Reporting

PR #3533, opened 2018-09-27
ICP-6274: SmartSense Motion Sensor incorrectly reporting inactive after join

varzac:
This changes the way configuration works to only send the read of the
zone status attribute until after the enroll.  Reading before the enroll
was causing the device to erroneously generate a no motion event and
then after enroll immediately creating a motion event.

This resolves: https://smartthings.atlassian.net/browse/ICP-6274

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -297,8 +297,12 @@ def configure() {
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 
 	log.debug "Configuring Reporting"
-	def configCmds = []
+	def configCmds = [zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000)]
+	def batteryAttr = device.getDataValue("manufacturer") == "Samjin" ? 0x0021 : 0x0020
 
+	configCmds += zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryAttr)
+
+	configCmds += zigbee.enrollResponse()
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
 	if (device.getDataValue("manufacturer") == "Samjin") {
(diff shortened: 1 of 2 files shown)


varzac:
I targeted staging because I didn't know what the timeline is on these, but we can switch if we wanted.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -307,8 +314,9 @@ def configure() {
 		configCmds += zigbee.batteryConfig()
 	}
 	configCmds += zigbee.temperatureConfig(30, 300)
+	configCmds += zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS)
```
tpmanley, line 314: We need to add a second read of the battery here. That was added recently (via the call to refresh) to work around an issue in a particular version of hubCore.

varzac: updated

BroderickCarlin (dismissed):

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -297,8 +297,15 @@ def configure() {
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 
 	log.debug "Configuring Reporting"
-	def configCmds = []
+	def configCmds = [zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000)]
+	def batteryAttr = 0x0020
```
tpmanley: Could simplify to `def batteryAttr = device.getDataValue("manufacturer") == "Samjin" ? 0x0021 : 0x0020`

tpmanley (dismissed):

BroderickCarlin (dismissed):
I have tested this with the hub and device in question and have confirmed functionality

varzac:
Added a fix for the multi sensor as well.

tpmanley (dismissed):

varzac:
If we want to go straight into staging I don't have permissions to merge so @tpmanley you'll have to do it.  If not, I can change the target to master.

tpmanley:
Let's go to `master` with this change to make sure we have enough time to sufficiently test before it is released.

varzac:
Sigh, with the rebase your approval was dismissed...

tpmanley (approved):

PR #3563, opened 2018-10-08
[ICP-6462] KeyWe lock status fix

PKacprowiczS:
@greens @MAblewiczS @KKlimczukS @MMateusiakS @DCzarneckaS @MGoralczykS 
This lock treats DoorLockOperation different than others, details in task.


greens (approved):
Approving this, since it should only affect "refresh" and I think Duncan put in a LE fix for this before he left.

PR #3570, opened 2018-10-09
[ICP-6467] Fixing bug with no added child devices

PKacprowiczS:
@greens @MAblewiczS @MMateusiakS @KKlimczukS @DCzarneckaS @MGoralczykS 
This is also for ICP-6476


greens:
Why does sending a second manufacturer get solve the problem? Does this still work if you take the manufacturer get out of configure?

PKacprowiczS:
This happens when I can't replicate caching issue... I've just copied wrong command, there should be second endPoint get. Sorry.

greens:
So is the problem that the original endpoint get wasn't being received? If you were able to fix the issue with the wrong command I'm very suspect about this fix.

PKacprowiczS:
There was some discussion in ST community about 'no child created' issue, and someone came up with 'fix' which tried to create child devices when updated() is called and Shravan asked me to implement it. And as this DTH is trying to be as generic as possible, the only way to trigger creating child devices is to send getEndpoints command once again.
That previous commit with wrong command hasn't fix anything at all. This issue is so random, we've just got lucky that it didn't occurred when retested.

greens (approved):

greens:
@tpmanley and @workingmonk these changes have been tested and verified at SR-Pol and I've been asked to push them to staging to be tested on Monday so they can go to prod sooner.

PR #3571, opened 2018-10-09
[ICP-6497] Updates DTH for NoeLight Switch

DCzarneckaS:
@greens  @MMateusiakS


greens:
Does `refresh()` only update the state for the parent device or the child device?

Which device does this line https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/a53ef059e60eb23653d196cfb87a3ae8e9e38ffd/devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy#L78 update?

DCzarneckaS:
@greens  I will update this refresh also for child device to be sure about this.

PR #3579, opened 2018-10-11
[ICP-6497] Updates DTH parent and Child for Aeotec Dual Nano Switch

DCzarneckaS:
@MAblewiczS  @greens


devicetypes/smartthings/zwave-binary-switch-endpoint.src/zwave-binary-switch-endpoint.groovy
```diff
@@ -43,6 +43,7 @@ metadata {
 
 def installed() {
 	configure()
+	response(refresh())
```
tpmanley: You need to return the combined results of both configure and refresh.

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -66,6 +66,7 @@ def installed() {
 		log.warn "Failed to add endpoint 2 ($desc) as Z-Wave Binary Switch Endpoint - $e"
 	}
 	configure()
+	response(refresh())
```
MAblewicz: Move response(refresh) to configure() where You are already calling refresh or remove refresh there and combine like Tom said. First way looks easier

MAblewicz (changes requested):

DCzarneckaS:
@MAblewiczS  Please check new commits.

PR #3581, opened 2018-10-15
Make ecobee act as on/off switch, fix cycle thru supported modes

jchobantonov:
Make ecobee act as on/off switch, so in case of leak sensor put below A/C triggers it could turn off the A/C.

Fix available modes for Ecobee 3 Lite thermostat. The thermostat should be able to cycle thru different modes (from within smartthings classic app) like off, cool, heat, auto, but it didn't because of the logic in ecobee-connect.groovy (only off and cool are available):
Ecobee 3 Lite reports autoHeatCoolFeatureEnabled as false, and heatStages as 0, although when invoking to set heat, cool, or auto mode the thermostat is able to do so.

### DIFF
--- a/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
+++ b/devicetypes/smartthings/ecobee-thermostat.src/ecobee-thermostat.groovy
@@ -25,6 +25,7 @@ metadata {
 		capability "Refresh"
 		capability "Relative Humidity Measurement"
 		capability "Health Check"
+		capability "Switch"
 
 		command "generateEvent"
 		command "resumeProgram"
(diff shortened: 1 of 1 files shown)


jchobantonov:
@tpmanley could you give this pull request traction? I'm sorry I don't know who could approve/merge this pull request so I have added you as I saw you do a merge against master

tpmanley:
@workingmonk , can you take a look? I think Ecobee is more your team than mine

workingmonk:
We are working with Ecobee to move this integration over to ST Schema and better feature support. We don't want to add custom logic with Switch capability. I would leave that decision to the partner on what the experience for 'off' and 'on' is, and if they want to support that or not.

Adding Switch capability to a device has other implications. With voice agents you could ask for "Turn off the lights" and this device may be perceived as a light/outlet because of the Switch capability.

PR #3582, opened 2018-10-15
ICP-6497 added refresh support to second endpoint

DCzarneckaS:
ICP-6497 added refresh support to second endpoint

@greens @tpmanley @dkirker @mckeed @MAblewiczS Please review my changes


ericangST:
If we get this merged into STG, we'll have this in PROD on the 30th, so we'll get this hopefully certified on the 31st to meet our M10 delivery

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -86,6 +86,7 @@ def configure() {
 	}
 	commands << command(zwave.basicV1.basicGet())
 	response(commands)
+	refresh()
```
greens: you'd want a `response` around this too.

or `response(commands + refresh())` I think would work

MAblewicz (approved):
Looks good now

greens (approved):

PR #3583, opened 2018-10-15
ICP-6205 Fix RGB<->HSV conversion for Z-Wave color bulbs

greens:
This has been broken. Seemingly due to assuming that computer modulo and mathematical modulo are the same wrt negative numbers. They're not. Also we get to use some nice utility functions that someone wrote but never used.

### DIFF
--- a/devicetypes/smartthings/aeon-led-bulb-6.src/aeon-led-bulb-6.groovy
+++ b/devicetypes/smartthings/aeon-led-bulb-6.src/aeon-led-bulb-6.groovy
@@ -16,6 +16,8 @@
  *  Date: 2018-8-31
  */
 
+import physicalgraph.developer.ColorUtilities
+
 metadata {
 	definition (name: "Aeon LED Bulb 6 Multi-Color", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb") {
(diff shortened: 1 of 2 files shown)


varzac (approved):
This seems fine to me.

mckeed (approved):
Can you give an example of how this conversion was wrong? From my spot checking it seems like they're giving the same values up to rounding differences. So I'm wondering if the problem is somewhere else.

greens:
@mckeed from hub logs attached to the Jira ticket:

```
90b41fe1-7f6e-4124-a5d7-cc1780277261 08:53:37: debug 'zw device: 0A, command: 9881, payload: 00 33 04 03 0F ' parsed to [[name:color, value:#2b0f1c, isStateChange:true, displayed:true, linkText:Aeon LED Bulb 6 Multi-Color, descriptionText:Aeon LED Bulb 6 Multi-Color color is #2b0f1c], [name:hue, value:-7.738094753427404, isStateChange:true, displayed:true, linkText:Aeon LED Bulb 6 Multi-Color, descriptionText:Aeon LED Bulb 6 Multi-Color hue is -7.738094753427404], [name:saturation, value:65.11628215234937, isStateChange:true, displayed:true, linkText:Aeon LED Bulb 6 Multi-Color, descriptionText:Aeon LED Bulb 6 Multi-Color saturation is 65.11628215234937]]
```

The negative hue value was the issue.

PR #3584, opened 2018-10-16
[WWST-1611] New fingerprint for Zigbee Thermostat - KONOz Lux Thermostat.

PKacprowiczS:
@greens @MAblewiczS @MMateusiakS @KKlimczukS @DCzarneckaS @MGoralczykS 
Please have a look at this.

### DIFF
--- a/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
+++ b/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
@@ -1,0 +1,337 @@
+/**
+ *  Copyright 2018 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ *	CentraLite Thermostat
+ *
+ *	Author: SRPOL
+ *	Date: 2018-10-15
+ */
+
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition (name: "Zigbee Thermostat", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "SmartThings-smartthings-Z-Wave_Thermostat") {
+		capability "Actuator"
+		capability "Temperature Measurement"
+		capability "Thermostat Mode"
+		capability "Thermostat Fan Mode"
+		capability "Thermostat Cooling Setpoint"
+		capability "Thermostat Heating Setpoint"
+		capability "Thermostat Operating State"
+		capability "Configuration"
+		capability "Health Check"
+		capability "Refresh"
+		capability "Sensor"
+
+		command "lowerHeatingSetpoint"
+		command "raiseHeatingSetpoint"
+		command "lowerCoolSetpoint"
+		command "raiseCoolSetpoint"
+
+		fingerprint profileId: "0104", inClusters: "0000,0001,0003,0004,0005,0020,0201,0202,0204,0B05", outClusters: "000A, 0019",  manufacturer: "LUX", model: "KONOZ", deviceJoinName: "LUX KONOz Thermostat"
+	}
+
+	tiles {
+		multiAttributeTile(name:"temperature", type:"generic", width:3, height:2, canChangeIcon: true) {
+			tileAttribute("device.temperature", key: "PRIMARY_CONTROL") {
+				attributeState("temperature", label:'${currentValue}°', icon: "st.alarm.temperature.normal",
+						backgroundColors:[
+								// Celsius
+								[value: 0, color: "#153591"],
+								[value: 7, color: "#1e9cbb"],
+								[value: 15, color: "#90d2a7"],
+								[value: 23, color: "#44b621"],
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -0,0 +1,337 @@
...
+	}
+}
+
+def parse(String description) {
+	def map = zigbee.getEvent(description)
+	def result
+	if(!map) {
+		result = parseAttrMessage(description)
+	} else {
+		log.warn "Some event is not parsed: ${map}"
```
varzac: This is odd, because you are writing printing this warning when the library successfully parses an event.  I think a message like "Unexpected event: " would make more sense.

PKacprowiczS: Mostly it was for debugging purposes, I wanted to check if there's any command from the device which is returned by zigbee.getEvent() (plot twist, there isn't). But yeah, I can agree that this is odd.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -0,0 +1,337 @@
...
+			zigbee.readAttribute(FAN_CONTROL_CLUSTER, FAN_MODE)
+}
+
+def ping() {
+	refresh()
+}
+
+def configure() {
+	def binding = zigbee.addBinding(THERMOSTAT_CLUSTER) + zigbee.addBinding(FAN_CONTROL_CLUSTER)
+	def startValues = zigbee.writeAttribute(THERMOSTAT_CLUSTER, HEATING_SETPOINT, DataType.INT16, 0x0A28) +
```
varzac: Is there any reason these values were chosen?  The heating set point seems a little high.

varzac (dismissed):
Couple of minor questions, but overall looks good.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
PKacprowiczS: They're taken from ZCL reference but... It appears I've accidentally switched heating and cooling setpoints default values...

varzac: Ah, I see now, they are the default values for those attributes.  If they are actually the defaults I wouldn't think we would have to set them on join, but I don't have any problem with it.  Swap them so they are the correct defaults and this PR looks good to me.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -0,0 +1,337 @@
...
+ *	Author: SRPOL
+ *	Date: 2018-10-15
+ */
+
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition (name: "Zigbee Thermostat", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "SmartThings-smartthings-Z-Wave_Thermostat") {
+		capability "Actuator"
+		capability "Temperature Measurement"
```
greens, line 24:
```suggestion
		capability "Temperature Measurement"
		capability "Thermostat"
```

greens: In order to work with SmartApps, it still needs the old deprecated capability.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -0,0 +1,337 @@
...
+	if (degrees != null) {
+		def degreesInteger = Math.round(degrees)
+
+		def celsius = (temperatureScale == "C") ? degreesInteger : (fahrenheitToCelsius(degreesInteger) as Double).round(2)
+		return zigbee.writeAttribute(THERMOSTAT_CLUSTER, HEATING_SETPOINT, DataType.INT16, hex(celsius * 100)) +
+				zigbee.readAttribute(THERMOSTAT_CLUSTER, HEATING_SETPOINT)
+	}
+}
+
+def raiseHeatingSetpoint() {
```
greens, line 319: Some of our other thermostat DTHs have a delay here before sending the actual command in order to collect multiple arrow presses into just one set instead of several "up by 1"s.

I know OneApp doesn't work like this, but I think it's worth doing for the classic DTH for a better user experience.

If it's too complicated and error-prone that you don't think it's worth it, I'd understand that too.

greens:
Based on https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3054/files this device should support "battery"

greens (dismissed):
This looks good to me. @varzac want to take another look?

varzac (dismissed):

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -0,0 +1,372 @@
...
+	def startValues = zigbee.writeAttribute(THERMOSTAT_CLUSTER, HEATING_SETPOINT, DataType.INT16, 0x07D0) +
+			zigbee.writeAttribute(THERMOSTAT_CLUSTER, COOLING_SETPOINT, DataType.INT16, 0x0A28)
+
+	return binding + startValues + zigbee.batteryConfig() + refresh()
+}
+
+def getBatteryPercentage(rawValue) {
+	def result = [:]
+	result.name = "battery"
+	if(rawValue == 0) {
```
greens, line 217: if the device supports having batteries installed while it's plugged in, I'd manually change this to 100% just to make sure users aren't constantly changing batteries in a plugged-in thermostat.

greens (approved):

PR #3586, opened 2018-10-16
WWST-1644 Zigbee Lock: Removed extra cluster from Schlage BE468 fingerprint

tpmanley:
The FC57 cluster is not in the production firmware.

https://smartthings.atlassian.net/browse/WWST-1644


tpmanley:
@MGoralczykS @greens

greens (approved):

PR #3594, opened 2018-10-18
WWST-1633 - Yale Siren (with Tamper Alert capability) in generic zwave-siren DTH

KKlimczukS:
@greens @MAblewiczS @MMateusiakS @DCzarneckaS @MGoralczykS @PKacprowiczS 
Please, take a look at this DTH.


greens:
I would instead just recommend adding this to https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy

Since OneApp still doesn't support preferences, it's not worth the extra DTH. That was the only difference I noticed. Let me know if there are other ones that necessitate a new DTH.

The only changes you _should_ need for the above DTH is adding support for security encapsulation.

KKlimczukS:
Yale External Siren has an important "Tamper alert" feature and it should be exposed in preferences (even if they are not supported in OneApp yet - it will be ready for future upgrade of OneApp and still it can be fully used in Smartthings Classic) - otherwise (in generic DTH) it will have to be completely disabled and user will not have any option to arm/disarm the siren - and that feature in case of this external device is critical.
Of course, tamper alert can be armed by default in generic DTH, but that will only cause problems for a user while taking the device off the wall or removing the device cover.

greens:
If that's the case then you need to include the "Tamper Alert" capability and actually generate "tamper" events. You're not doing that right now. If you add that, I'd rename it `zwave-siren-tamper`. Also this _should_ be able to run locally.

devicetypes/smartthings/zwave-siren-tamper.src/zwave-siren-tamper.groovy
```diff
@@ -0,0 +1,283 @@
...
+}
+
+
+def zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
+	log.debug "BatteryReport:  $cmd"
+
+	def batteryLevel = cmd.batteryLevel != 255 ? cmd.batteryLevel : 0
+	def map = [name: "battery", unit: "%", value: batteryLevel]
+	createEvent(map)
+}
```
greens:
```
def zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
	def map = [name: "battery", unit: "%"]
	if (cmd.batteryLevel == 0xFF) {
		map.value = 1
		map.descriptionText = "$device.displayName has a low battery"
	} else {
		map.value = cmd.batteryLevel
	}
	state.lastbatt = new Date().time
	createEvent(map)
}
```

This is the battery handling code from zwave-siren. It would be good to be consistent. Specifically we turn 255 into 1%, not 0.

greens: you can ignore the 'lastbatt' stuff

devicetypes/smartthings/zwave-siren-tamper.src/zwave-siren-tamper.groovy
```diff
@@ -0,0 +1,283 @@
...
+		secure(zwave.basicV1.basicSet(value: 0xFF)),
+		"delay 3000",
+		secure(zwave.basicV1.basicSet(value: 0x00)),
+		secure(zwave.basicV1.basicGet()),
+		secure(zwave.alarmV2.alarmGet(zwaveAlarmType: 0x07))
+	]
+}
+
+
+private secure(physicalgraph.zwave.Command cmd) {
```
greens: This should always be conditional just in case.

devicetypes/smartthings/zwave-siren-tamper.src/zwave-siren-tamper.groovy
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		main "alarm"
+		details(["alarm", "test", "off", "refresh", "battery", "tamper"])
+	}
+}
+
+def installed() {
+	// Device-Watch simply pings if no device events received for 32min(checkInterval)
+	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
greens: For a battery-powered device, this should be longer. `zwave-siren` uses 122 minutes

devicetypes/smartthings/zwave-siren-tamper.src/zwave-siren-tamper.groovy
```diff
@@ -0,0 +1,283 @@
...
+		if (cmd) {
+			result = zwaveEvent(cmd)
+		}
+	}
+	log.debug "Parse returned ${result?.inspect()}"
+	return result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
+	def encapsulatedCommand = cmd.encapsulatedCommand([0x59: 1, 0x2B: 1, 0x71: 2, 0x85: 2, 0x25: 2, 0x7A: 1])
```
greens: Can we use notification v3 (`0x71: 3`) instead of v2?

greens: They are backwards compatible with each other, but you may need to change how some things are named.

devicetypes/smartthings/zwave-siren-tamper.src/zwave-siren-tamper.groovy
```diff
@@ -0,0 +1,283 @@
...
+		Short comfortLED = (settings.comfortLED as Short) ?: 0
+		Boolean tamper = (settings.tamper as Boolean) ?: 0
+
+		if (alarmLength != state.alarmLength || alarmLEDflash != state.alarmLEDflash || comfortLED != state.comfortLED || tamper != state.tamper) {
+			state.alarmLength = alarmLength
+			state.alarmLEDflash = alarmLEDflash
+			state.comfortLED = comfortLED
+			state.tamper = tamper
+
+			commands << new physicalgraph.device.HubAction(zwave.configurationV1.configurationSet(parameterNumber: 1, size: 1, configurationValue: [alarmLength]).format())
```
greens: these don't need to be hubactions, as demonstrated by the basicSet you did below

Also they should be conditionally secured

devicetypes/smartthings/zwave-siren-tamper.src/README.md
```diff
@@ -0,0 +1,35 @@
+# Yale External Siren
```
greens: You'll probably want to update all the instances of the word "Yale"

greens:
Changed my mind about this after looking at ICP-6798. Please see my comments on Jira. Sorry.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -126,10 +160,37 @@ def configure() {
...
+		if (alarmLength != state.alarmLength || alarmLEDflash != state.alarmLEDflash || comfortLED != state.comfortLED || tamper != state.tamper) {
+			state.alarmLength = alarmLength
+			state.alarmLEDflash = alarmLEDflash
+			state.comfortLED = comfortLED
+			state.tamper = tamper
+
+			cmds << zwave.configurationV1.configurationSet(parameterNumber: 1, size: 1, configurationValue: [alarmLength]).format()
+			cmds << zwave.configurationV1.configurationSet(parameterNumber: 2, size: 1, configurationValue: [alarmLEDflash ? 1 : 0]).format()
+			cmds << zwave.configurationV1.configurationSet(parameterNumber: 3, size: 1, configurationValue: [comfortLED]).format()
+			cmds << zwave.configurationV1.configurationSet(parameterNumber: 4, size: 1, configurationValue: [tamper ? 1 : 0]).format()
```
greens: For parallel structure's sake, you should call `format()` on all the commands or none of them.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -161,18 +228,22 @@ def on() {
...
+			secure(zwave.basicV1.basicSet(value: 0x00)),
+			secure(zwave.basicV1.basicGet()),
+			secure(zwave.alarmV2.alarmGet(zwaveAlarmType: 0x07))
+		]
+	} else {
+		[
+			zwave.basicV1.basicSet(value: 0x00).format(),
+			zwave.basicV1.basicGet().format()
+		]
+	}
```
greens, line 227: I'd structure this to just append the `alarmGet` for Yale devices instead of all the duplication.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -189,23 +260,61 @@ def ping() {
...
+		return delayBetween([
+			secure(zwave.basicV1.basicGet()),
+			secure(zwave.alarmV2.alarmGet(zwaveAlarmType: 0x07)),
+			zwave.batteryV1.batteryGet().format()
+		], 2000)
+	} else {
+		return delayBetween([
+			zwave.basicV1.basicGet().format(),
+			zwave.batteryV1.batteryGet().format()
+		], 2000)
```
greens: same here

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -189,23 +260,61 @@ def ping() {
...
+		def cmd = zwave.parse(description, [0x20: 1])
+		if (cmd) {
+			result = zwaveEvent(cmd)
+		}
 	}
 	log.debug "Parse returned ${result?.descriptionText}"
 	return result
 }
 
+private secure(physicalgraph.zwave.Command cmd) {
```
greens, line 284: make this conditional on the `zwaveInfo` like it is in other DTHs.

greens: Doing that lets you avoid some of the code duplication above

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -189,23 +249,60 @@ def ping() {
...
+	if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo?.zw?.contains("s")) {
```
greens: you don't need the `state.sec` check because that's a legacy structure we have kept for backwards compatibility on old DTHs where security was handled differently.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -93,21 +118,25 @@ def initialize() {
...
 	if (!device.currentState("alarm")) {
-		cmds << zwave.basicV1.basicGet().format()
-		cmds << "delay 5"
+		if (isYale()) {
+			cmds << secure(zwave.basicV1.basicGet())
+			cmds << secure(zwave.alarmV2.alarmGet(zwaveAlarmType: 0x07))
+		} else {
+			cmds << secure(zwave.basicV1.basicGet())
+			cmds << "delay 5"
+		}
```
greens, line 124: small nitpick. You could write this to only have one `basicGet` statement.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -238,7 +332,32 @@ def zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
...
 	def isActive = false
-	if (cmd.notificationType == 0x0E) { //Siren notification
+	def result = []
+
+	if (cmd.notificationType == 0x07) { // Burglar notification
+		def map = [name: "tamper", value: "detected"]
+		map.data = [sirenName: deviceName]
+		switch (cmd.event) {
+			case 0:
+				map.value = "clear"
```
greens, line 342: Forgot that tamper clear doesn't work when the device is locally executing.

See Tom's comment here:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3619

greens: You'll have to automatically clear it. It's not ideal, but it's the best workaround we have atm.

greens (approved):

PR #3597, opened 2018-10-19
Fix Pabal parameter

ajguyot:
Pabal requires a key of `n` and a value of `${knockSensor.name}` (brackets must be included) to successfully fill the device name.


erodewald (approved):
Simple fix, looks good -- matches what Jihee described.

PR #3599, opened 2018-10-20
Add fingerprint for RGBgenie ZW-1001 Z-Wave Dimmer controller

vseven:
@greens  - Adding a fingerprint for a RGBgenie ZW-1001 Z-Wave Dimmer controller.   http://www.rgbgenie.com/ . 
By default the device was recognized as a "Secure Dimmer" but it didn't work under that device type.  Here is the info from the devices tab after inclusion:  
Raw Description 	zw:Ls type:1101 mfr:0330 prod:0201 model:D002 ver:1.03 zwv:4.61 lib:03 cc:5E,55,6C,98,9F role:05 ff:8600 ui:8600

Let me know if you need anything else from me.  Also it seems this device type doesn't work in the new app?

### DIFF
--- a/devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
+++ b/devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
@@ -40,6 +40,7 @@ metadata {
 		fingerprint mfr: "001A", prod: "4449", model: "0003", deviceJoinName: "Eaton RF Dimming Plug-In Module"
 		fingerprint mfr: "014F", prod: "5744", model: "3530", deviceJoinName: "GoControl In-Wall Dimmer"
 		fingerprint mfr: "0307", prod: "4447", model: "3034", deviceJoinName: "Satco In-Wall Dimmer"
+		fingerprint mfr: "0330", prod: "0201", model: "D002", deviceJoinName: "RGBgenie ZW-1001 Z-Wave Dimmer"
 	}
 
 	simulator {


greens (approved):

greens:
I couldn't find the certification for this device on the z-wave alliance site: https://products.z-wavealliance.org/ (which might just mean it's a white labeled re-packaging but also might mean it's not certified and hence your issues). 

Just so I can look into it. Can you elaborate on what didn't work when the device was added using the Secure Dimmer DTH?

And what doesn't work in the new App?

vseven:
The underlying controller is made by Sunricher.  Its advertised as a single color LED dimmer, 4 channels (all identical levels).  The "secure dimmer" device does nothing, no commands work.  Product info:  https://rgbgenie.com/?product=rgbgenie-single-color-led-controller-z-wave .  I've been working with the RGBgenie people to try and get them working without user intervention and can get more info if I know the questions to ask.  I also physically have one of everything RGBgenie sells so I can test or try whatever or attempt to read numbers off the chips.

On a related note they also have a ZW-1002 RGBW controller which comes in as a Aeon LED Bulb and works by default with that DTH.  They also have a ZigBee 3.0 version of both controllers but I've had 0 luck getting those to even discover....not even a catchall event although the devices think they are paired.  Tom Manley said it was something to do with the device using the "ZLL profile on the Basic cluster" and that a future hub firmware update will hopefully help that. 

As for "doesn't work" I mean it just shows connected with no controls.  Unless its a bug from it being pulled in as a secure dimmer and once you look at it in the new app it doesn't update when you change device types,

greens:
>Unless its a bug from it being pulled in as a secure dimmer and once you look at it in the new app it doesn't update when you change device types,

Bingo. New App doesn't do great when you switch the device type after adding. That's something I've made the powers that be aware of, but I don't have a timeline for it to be fixed.

Nothing's immediately jumping out to me in the secure dimmer DTH, but it hasn't seen a lot of attention in a while so it's possible there's something going on there. As a heads up: we've seen some conflict issues around device classification when the old DTH has a fingerprint with the `deviceId` field, so your added fingerprint still might not cause it to get picked up and we might need to refactor the generic fingerprint in `secure-dimmer`.

vseven:
So related but unrelated.  As I mentioned their ZW-1002 RGBW device comes in naturally as a Aeon RGBW Bulb device type but also works with the RGBW Light device type.  For that should I put in a pull for adding the fingerprint to the generic RGBW Light or put in a entire new DTH based on the RGBW DTH but called "RGBgenie RGBW Controller" or something with its fingerprint?  Technically its not a bulb but not sure if you guys would rather have it detect on its own or reuse when possible.

greens:
We typically aim for re-use, so the correct thing is to just add a fingerprint with a join name. There's nothing inherent to the aeon DTH that necessitates it being called "aeon". Why there are two DTHs that seem near-identical predates me, and they could both use a little love and updating. We're only just now looking harder at Z-Wave rgbw bulbs.

That said, I did just put in a fix in the rgb<->hsv conversion in the aeon dth that I didn't put into the rgbw light one (because I missed it somehow), so if you want it to work with the new app you'll want to go with aeon-led-bulb.

vseven:
@greens - So wouldn't it make sense to delete/obsolete the "RGBW Light" DTH, move the fingerprints in it over to "Aeon LED Bulb" and then rename the Aeon to "Z-Wave RGBW Bulb"?  There is already a "Zigbee-rgbw-bulb" so that would be more consistent.

I can put in the pulls if that makes sense.

greens:
We'd also have to run a script on our side to migrate existing installs using that DTH over to the different DTH. I'll keep it in mind, but unfortunately it takes more than just a PR.

wsus-script:
Sooooo....you want me to start the process with the Aeon as a source and you can mark the RGBW-Light as do not update or something and go from there?

vseven:
That's fine. Do you mind if I start the process and move stuff over to a new Z-Wave generic device to match the zigbee one? Not sure if you guys have a process on your end where you could just mark the current rgbw light dth as old but still in use until accounts can be migrated but it makes sense there should be a "zwave-rgbw-bulb".

greens:
I think we have to wait until the deviceType migration issues you experienced get fixed in the new app before we could do that. If we do move in the direction of a generic DTH, it would probably be rgbw-light, we'd just fix it up first.

But if you want to tag me to take a look at anything you do around overhauling current DTHs you can go ahead.

PR #3575, opened 2018-10-10
WWST-1618 added fingerprint for Satco 86-100 Plug-In Appliance Module

MAblewicz:

PR #3616, opened 2018-10-24
WWST-1618 added fingerprint for Satco 86-100 Plug-In Appliance Module…

greens:
… (#3575)


greens:
requesting fingerprint directly to staging. device certified under WWST-1583

PR #3619, opened 2018-10-25
ICP-6798 Add Tamper Alert to Zwave Motion Sensor

greens:
Most devices don't even send this so they'll always remain clear, but the BeSense does.


devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -88,6 +94,7 @@ def sensorValueEvent(value) {
 	if (value) {
 		createEvent(name: "motion", value: "active", descriptionText: "$device.displayName detected motion")
 	} else {
+		createEvent(name: "tamper", value: "clear")
```
tpmanley, line 97: Unfortunately this won't work when the DTH is locally executing. Take a look at https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy#L324 to see how we handled this for the Aeon Multisensor 6. Hopefully the same will work here.

greens: thanks. forgot about that.

Fixed*

tpmanley (approved):

PR #3617, opened 2018-10-25
WWST-1624 Add Sengled Element bulb with motion sensor

dkirker:

PR #3620, opened 2018-10-25
WWST-1624 Add power metering to Zigbee Dimmer with Motion Sensor

dkirker:


devicetypes/smartthings/zigbee-dimmer-with-motion-sensor.src/zigbee-dimmer-with-motion-sensor.groovy
```diff
@@ -21,6 +21,7 @@ metadata {
 		capability "Light"
 		capability "Switch"
 		capability "Switch Level"
+		capability "Power Meter"
```
tpmanley, line 24: For consistency with `zigbee-dimmer` and `zigbee-dimmer-power` we might want to change this to `zigbee-dimmer-power-with-motion`

tpmanley: It doesn't look like this has gone to prod yet so I think we should change it now.

varzac (approved):

k-morawski:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3617

PR #3624, opened 2018-10-26
ONEAPP-8001 Fan controller returns floats instead of ints

greens:


devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -106,7 +106,7 @@ def fanEvents(physicalgraph.zwave.Command cmd) {
 	def value = (cmd.value ? "on" : "off")
 	def result = [createEvent(name: "switch", value: value)]
 	result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)
-	result << createEvent(name: "fanSpeed", value: cmd.value/33)
+	result << createEvent(name: "fanSpeed", value: cmd.value/33 as int)
```
tpmanley: Would it be better to round than truncate?

greens: devices only ever send us multiples of 33, but I suppose it's more resilient

greens:
Fixed

tpmanley (approved):

Issue #3434, opened 2018-09-04
zigbee-switch-power divisor needs to be configurable


orangebucket:
The zigbee-switch-power DTH divides power values by 10 with the code commented with a TODO to make the divisor part of the configuration.

The Hive Active Plug and Salus SP600 are returning power values that are ten times too small. The TODO presumably explains why.

orangebucket:
Further testing suggested the devices were not adhering to the standards.

PR #3631, opened 2018-10-29
Update zigbee-switch-power.groovy

vseven:
I believe this would be the fix for https://github.com/SmartThingsCommunity/SmartThingsPublic/issues/3434 .

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -28,7 +28,12 @@ metadata {
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "0003, 000A, 0019", manufacturer: "Jasco Products", model: "45853", deviceJoinName: "GE ZigBee Plug-In Switch"
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "000A, 0019", manufacturer: "Jasco Products", model: "45856", deviceJoinName: "GE ZigBee In-Wall Switch"
     }
-
+    
+	preferences {
+		input title: "Power Correction", description: "Some sensors display the power value as a multiple of ten to get around a decimal place.  If you device does this you can add a divisor to offset the value.  A value of 1 makes no correction, a value of 10 will divide the sensors value by ten, etc.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+		input "powerDivisor", "number", title: "Divisor", description: "Divide power value by this number", range: "1..100", displayDuringSetup: false
+    }
+    
     tiles(scale: 2) {
         multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
             tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
@@ -56,7 +61,11 @@ def parse(String description) {
     if (event) {
         if (event.name == "power") {
             def powerValue
-            powerValue = (event.value as Integer)/10            //TODO: The divisor value needs to be set as part of configuration
+            if (powerDivisor) {
+			    powerValue = (event.value as Integer) / (int) powerDivisor
+            } else {
+                powerValue = (event.value as Integer)
+            }
             sendEvent(name: "power", value: powerValue)
         }
         else {


devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -56,7 +61,11 @@ def parse(String description) {
     if (event) {
         if (event.name == "power") {
             def powerValue
-            powerValue = (event.value as Integer)/10            //TODO: The divisor value needs to be set as part of configuration
+            if (powerDivisor) {
```
tpmanley, line 64: This is a good idea but unfortunately it won't work with local execution without a hub firmware change.

vseven:
Yes, but it as it sits the value is wrong for multiple devices. Hopefully it could be something slated for the next Hub update.

tpmanley:
@vseven , which devices? I don't think I've come across one that didn't use a divisor of 10 but knew it was possible

vseven:
See open issue linked above.  I was looking at different power monitoring switches and tried to do some homework before I purchased one, saw the outstanding open issue, and figured I'd submit on it since it looks like an easy fix.  The ones linked in the issue are mostly UK based and nothing I personally was going to buy but again looked like a easy fix.

I've dealt with a lot of power devices and in a industrial setting most do this so they don't have to worry about a decimal point and they expect you to do a divisor of 10 or even a 100 to get the decimal place back. Looks like a couple devices are just sending the value as is.

Side note: technically the power value should be a float since potentially dividing by 10 or 100 should leave a remainder.

greens:
If we were doing a full solution, wouldn't it be better to query the device for its divisor (or multiplier) instead of relying on the user to know it?

Or just to automatically set it based on mfr + model if we know it.

But as Tom said, none of this works without a hubcore revision to update local execution.

vseven:
Automatically would be preferred. In my experience with power meters they usually don't have a register where they keep that info, it's usually just in the manual.

greens:
You _should_ be able to do 

`zigbee.readAttribute(zigbee.SIMPLE_METERING_CLUSTER, 0x0302)`

and what you get back from that should be the divisor.

vseven:
I can change the pull and use that, if empty/null use 10 like the existing DTH?

vseven:
So you can easily remove the preference and do something like this:

```
def powerValue
def powerDivisor = zigbee.readAttribute(zigbee.SIMPLE_METERING_CLUSTER, 0x0302)
if (powerDivisor) {
    powerValue = (event.value as Integer) / (int) powerDivisor
} else {
    powerValue = (event.value as Integer) / 10
}
sendEvent(name: "power", value: powerValue)
```

but would you also check the register at 0x0301 for a multiplier and apply the same logic?  I've personally never seen a multiplier used but I suppose it's possible.

greens:
`zigbee.readAttribute()` doesn't return like that, you'll have to extract value out in the `parse()` function. (unless that was just done for illustrative purposes)

If you're going for it you might as well also add multiplier, though like you've said I've never seen it either.

vseven:
Yeah, it was psudo psudocode.  Ha.  I would think you'd pull it out and store it in a custom attribute and then use that as needed, wouldn't that make the most sense? The value is not going to change. I would even think you would read it on a install or updated for both divisor and multiplier and then use them as needed.

Of course all of this is moot if it's not going to be changed because of the whole local execution thing.

greens:
Yeah I would add a read in the `installed` method and store it in `state.divisor` and `state.multiplier`

vseven:
Think that's a little beyond me.  I would assume you'd have:
```
def installed() {
    zigbee.readAttribute(zigbee.SIMPLE_METERING_CLUSTER, 0x0301)  // poll multiplier
    zigbee.readAttribute(zigbee.SIMPLE_METERING_CLUSTER, 0x0302)  // poll divisor
}
```

then in your parse function, before the else where you have the `DID NOT PARSE` part you'd do a else (psudocode):

```
def mymap = zigbee.parseDescriptionAsMap(description)
if(mymap) {
     if(mymap.cluster == "0301") {  // assign multiplier}
     if(mymap.cluster == "0302") {  // assign divisor}
} else {
 // The did not parse part
}
```

Then in the actual powerValue assignment times or divide if those values are present.  But not sure if that's right so I'll let you guys or someone more experienced play with this if they want.  Sounds like its almost there.  Also I'll still say powerValue should be a float.

greens:
Yeah. Essentially.

orangebucket:
> @vseven , which devices? I don't think I've come across one that didn't use a divisor of 10 but knew it was possible

If I may politely intrude here as the creator of #3434, and having only just seen this discussion. The devices where I see a problem are the Salus SP600 and Hive Active Plug, which both report Computime as their manufacturer with model numbers SP600 and SLP2b respectively. The Hive brand is quite prominent in the UK and widely mentioned in the community, whereas the Salus brand is the sort of thing you find retailed by specialist heating engineers (but importantly the plug is relatively compact and can often be found for about £15 to £20 which is quite reasonable by UK standards, so those of us who saw it was ZHA 1.2 compliant and took a chance on it are punching the air in delight).

Having been enlightened by this thread, I managed to check attributes 0x0300, 0x0301 and 0x0302 of cluster 0x0702 on a couple of the Salus plugs and they were 0, 1 and 0x002710. I believe that means you still haven't come across any devices that don't use a divisor of ten.

However I also remain convinced that my 40W electric fan doesn't consume e.g. 3.4W and a USB charger doesn't consume 0.1W or 0.2W, and that multiplying those figures by ten produces something far more realistic.

So I rather fear that doing things 'properly' by looking at the units, multiplier and divisor wouldn't help with my issue, which is starting to smell a bit of device firmware not quite following the specs or just being wrong, though my diagnoses and knowledge are far too limited to say anything stronger than that. Obviously a user preference would delight me. I know I can use a custom device handler but I just prefer not to if I can get away with it.

Now I fear I may have just scored an 'own goal' here, but I wanted to clarify things.

greens:
@orangebucket what may be happening for you is that the device is reporting using the Electrical Measurement Cluster (0x0B04). I think it actually reports using a different magnitude so treating everything that gets parsed to "power" equally might be the root of the trouble. You should be able to verify if that's the cluster being used.

I think the real issue is that zigbee has two clusters that _can_ be used to report power and they're not incredibly consistent about formatting and units, but we may be treating them as though they are.

vseven:
> I think the real issue is that zigbee has two clusters that can be used to report power and they're not incredibly consistent about formatting and units, but we may be treating them as though they are.

Which brings this full circle.....why not just have a user preference and default it to 10 if not entered?  I know the whole local execution thing but temp offsets are a user preference and those work locally, right?

greens:
@vseven They do. The downside is that the new app doesn't support user preferences yet. But I suppose having it wrong in 1 client is better than having it wrong in both.

vseven:
In that case I'll stand by my initial commit except have a default preference of 10.  Maybe even make the range from 0.1 to 100 in case there is a sensor that needs a multiplier (would effectively make the range x10 through /100).

orangebucket:
> @orangebucket what may be happening for you is that the device is reporting using the Electrical Measurement Cluster (0x0B04). I think it actually reports using a different magnitude so treating everything that gets parsed to "power" equally might be the root of the trouble. You should be able to verify if that's the cluster being used.

That would involve me knowing what I am doing, and with Zigbee I am at the almost completely ignorant stage.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -56,7 +61,11 @@ def parse(String description) {
...
+			    powerValue = (event.value as Integer) / (int) powerDivisor
```
greens, line 65: you mixed tabs w/ spaces here

If you wanted to convert the whole thing to tabs I'd appreciate it.

greens:
@orangebucket you'd do 

`zigbee.parseDescriptionAsMap(description)` 

inside the `event.name=="power"` check in `parse()`. That returns a map and you could check (or just log) `map.clusterId == 0x0B04`.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
vseven: Apparently I deleted my branch and can't edit (?) so I'll throw the same changes into a new pull with tabs in place of spaces.  Leaving it otherwise exactly like this (although I think the range should be 0.1 to 100 in case of a multiplier even though rare).

(My bad, assumed it wouldn't be used)

PR #3632, opened 2018-10-29
ICP-6760 Add offlinePingable to parent and child dual switch (#3621)

greens:


greens:
Cherry-pick to staging that has been verified by SR-Pol

tpmanley (approved):

PR #3633, opened 2018-10-29
ICP-6885 Fixes error setting color when hue is 0

greens:
Issue arises because null and 0 are both treated as false

### DIFF
--- a/devicetypes/smartthings/aeon-led-bulb-6.src/aeon-led-bulb-6.groovy
+++ b/devicetypes/smartthings/aeon-led-bulb-6.src/aeon-led-bulb-6.groovy
@@ -249,11 +249,7 @@ def setColor(value) {
 		def c = value.hex.findAll(/[0-9a-fA-F]{2}/).collect { Integer.parseInt(it, 16) }
 		result << zwave.switchColorV3.switchColorSet(red: c[0], green: c[1], blue: c[2], warmWhite: 0, coldWhite: 0)
 	} else {
-		def hue = (value.hue == null) ? device.currentValue("hue") : value.hue
-		def saturation = (value.saturation == null) ? device.currentValue("saturation") : value.saturation
-		if(hue == null) hue = 13
-		if(saturation == null) saturation = 13
-		def rgb = huesatToRGB(hue, saturation)
+		def rgb = huesatToRGB(value.hue, value.saturation)
 		result << zwave.switchColorV3.switchColorSet(red: rgb[0], green: rgb[1], blue: rgb[2], warmWhite:0, coldWhite:0)
 	}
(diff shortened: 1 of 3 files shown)


dkirker (dismissed):

greens:
@vseven can you give this a shot for me?

vseven:
Tested the modified Aeon LED Bulb code, works as expected although based on your change I would think the following lines where you check for null are now redundant.

Worked for both red:
 1bd38040-78db-4d3e-a20b-c5ef35574607 8:01:47 PM: debug setColor: [hue:0, saturation:100] 

and gray:
 1bd38040-78db-4d3e-a20b-c5ef35574607 8:01:54 PM: debug setColor: [hue:0, saturation:0] 

Still submitting a new zwave-rgbw-bulb and controller but I'll use the same line

dkirker:
Hmmm.... Well, there is one case where those secondary checks *would* be needed, and that is if device.currentValue were null. Maybe those could just be replaced with a helper function "deviceCurrentValueWithDefault(state,default)", or make the ternary a little more nasty?

vseven:
If the last known value is null and somehow the passed in value is also null then the helper function will throw up an error.  I mean, it's your guys code, but personally the error being logged I would think is enough....I'd want to track down why null values are getting past in.  Otherwise those checks will reset the hue and saturation to an arbitrary number and might mask an error initially.

devicetypes/smartthings/aeon-led-bulb-6.src/aeon-led-bulb-6.groovy
```diff
@@ -249,8 +249,8 @@ def setColor(value) {
 		def c = value.hex.findAll(/[0-9a-fA-F]{2}/).collect { Integer.parseInt(it, 16) }
 		result << zwave.switchColorV3.switchColorSet(red: c[0], green: c[1], blue: c[2], warmWhite: 0, coldWhite: 0)
 	} else {
-		def hue = value.hue ?: device.currentValue("hue")
-		def saturation = value.saturation ?: device.currentValue("saturation")
+		def hue = (value.hue == null) ? device.currentValue("hue") : value.hue
```
tpmanley: The Color Control capability is defined as requiring hue and saturation so I don't think we need to guard against null inputs. Also other RGB DTHs like https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy#L214 don't do these checks. I'd say just remove these null checks and let the errors bubble up like @vseven said.

greens:
Fair enough. Updated. Needs a new approval.

tpmanley (approved):

PR #3634, opened 2018-10-29
WWST-1731 Zigbee Valve power source strings

greens:
I don't think OneApp handles capital letters


tpmanley (dismissed):
The capability definition uses lowercase so this makes sense

greens:
Had to fix this because bug is also present in appEngine. Working on a PR for that as well.

tpmanley (approved):

PR #3635, opened 2018-10-30
Wwst 1828

DCzarneckaS:
@MMateusiakS  @greens  @MAblewiczS


greens:
Based on the `profileId`, this is the wrong DTH to add this device to. I've incorporated the fingerprints here: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3637 so this PR can be closed.

PR #3638, opened 2018-10-30
Create zwave-rgbw-bulb.groovy

vseven:
Creation of a new "zwave-rgbw-bulb" to be the eventual replacement of the "rgbw-light" DTH which is not being maintained with recent changes and is for the most part a duplicate of "aeon-led-bulb".  Naming is also more consistent with other DTH's (zll-rgbw-bulb, zigbee-rgbw-bulb, etc).   Re: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3599#issuecomment-432804367 @greens  

Source code is based on the aeon as it was most up to date.  Modifications made:
- Updated copyright
- Merged all fingerprints from both "aeon-led-bulb" and "rgbw-light" into this DTH
- Added metadata for ocfDeviceType, mnmn, and vid with proper identifiers
- Switched standard tile over to a multiAttributeTile to more closely match "zll-rgbw-bulb"
- Removed tiles no longer needed after conversion to mutliAttributeTile
- Added capabilities "Light" & "Health Check"
- Added a "addDeviceWatch" routine to add our checkInterval for device health
- Fixed a issue that would prevent "Red" and "Grey" from being set through hue/sat per fix here https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3633#issuecomment-434410557

Needs full testing but since code is 98% from existing working DTH it should be minimal.  Also would be nice once approved to have this added to locally run devices like other "generic" Z-Wave and ZigBee DTH's.  Eventually I would think the aeon-rgbw-bulb and rgbw-light can both be replaced by this (and possibly other z-wave rgbw devices) and hopefully can be the base for a zwave-rgb-bulb by just removing the color temperature capability and slider.

(This replaced a closed pull request that was outdated with recent above changes)

### DIFF
--- a/devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
@@ -25,13 +25,13 @@ metadata {
 		capability "Refresh"
 		capability "Actuator"
 		capability "Sensor"
-    capability "Health Check"
-    capability "Light"
+    		capability "Health Check"
+    		capability "Light"
 
 		command "reset"
 
 		fingerprint inClusters: "0x33"
-    fingerprint inClusters: "0x26,0x33"
+    		fingerprint inClusters: "0x26,0x33"
 		fingerprint inClusters: "0x26,0x33,0x98"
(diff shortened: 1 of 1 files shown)


vseven:
Killing pull request.  Will update RGBW-Light in its place.

PR #3639, opened 2018-10-30
Create zwave-rgbw-controller.groovy

vseven:
Creation of a new "zwave-rgbw-bulb" to be the eventual replacement of the "rgbw-light" DTH which is not being maintained with recent changes and is for the most part a duplicate of "aeon-led-bulb".  Naming is also more consistent with other DTH's (zll-rgbw-bulb, zigbee-rgbw-bulb, etc).   Re: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3599#issuecomment-432804367 @greens  

Source code is based on the aeon as it was most up to date.  Modifications made:
- Updated copyright
- Merged all fingerprints from both "aeon-led-bulb" and "rgbw-light" into this DTH
- Added metadata for ocfDeviceType, mnmn, and vid with proper identifiers
- Switched standard tile over to a multiAttributeTile to more closely match "zll-rgbw-bulb"
- Removed tiles no longer needed after conversion to mutliAttributeTile
- Added capabilities "Light" & "Health Check"
- Added a "addDeviceWatch" routine to add our checkInterval for device health
- Fixed a issue that would prevent "Red" and "Grey" from being set through hue/sat per fix here https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3633#issuecomment-434410557

Needs full testing but since code is 98% from existing working DTH it should be minimal.  Also would be nice once approved to have this added to locally run devices like other "generic" Z-Wave and ZigBee DTH's.  Eventually I would think the aeon-rgbw-bulb and rgbw-light can both be replaced by this (and possibly other z-wave rgbw devices) and hopefully can be the base for a zwave-rgb-bulb by just removing the color temperature capability and slider.

(This replaced a closed pull request that was outdated with recent above changes)

### DIFF
--- a/devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
+++ b/devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
@@ -1,0 +1,241 @@
+/**
+ *  Copyright 2018 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ *  Z-wave RGBW Controller (White chanel is dimmed and doesn't support color temperature)
+ *
+ *  Author: SmartThings / vseven (Allan)
+ *  Date: 2018-10-26
+ */
+
+metadata {
+	definition (name: "ZWave RGBW Controller", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb") {
+		capability "Switch Level"
+		capability "Color Control"
+		capability "Switch"
+		capability "Refresh"
+		capability "Actuator"
+		capability "Sensor"
+    capability "Health Check"
+    capability "Light"
+
+		command "setWhiteLevel"
+		command "reset"
+    
+    attribute "whiteLevel", "number"
+
+		fingerprint mfr: "0330 ", prod: "0201", model: "D002", deviceJoinName: "RGBgenie RGBW Controller ZW-1002"
+	}
+
+	simulator {
+	}
+
+	tiles(scale: 2) {
+		multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true) {
+			tileAttribute("device.switch", key: "PRIMARY_CONTROL") {
+				attributeState("on", label:'${name}', action:"switch.off", icon:"st.lights.philips.hue-single", backgroundColor:"#00a0dc", nextState:"turningOff")
+				attributeState("off", label:'${name}', action:"switch.on", icon:"st.lights.philips.hue-single", backgroundColor:"#ffffff", nextState:"turningOn")
+				attributeState("turningOn", label:'${name}', action:"switch.off", icon:"st.lights.philips.hue-single", backgroundColor:"#00a0dc", nextState:"turningOff")
+				attributeState("turningOff", label:'${name}', action:"switch.on", icon:"st.lights.philips.hue-single", backgroundColor:"#ffffff", nextState:"turningOn")
+			}
+			tileAttribute ("device.level", key: "SLIDER_CONTROL") {
+				attributeState "level", action:"switch level.setLevel"
+			}
+			tileAttribute ("device.color", key: "COLOR_CONTROL") {
+				attributeState "color", action:"setColor"
+			}
+		}
+  }
+  controlTile("whiteSliderControl", "device.whiteLevel", "slider", width: 2, height: 2, inactiveLabel: false) {
+    state "whiteLevel", action:"setWhiteLevel", label:'White Level'
+ 	 }
+	standardTile("reset", "device.reset", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+		state "default", label:"Reset Color", action:"reset", icon:"st.lights.philips.hue-single"
+  	}
+	standardTile("refresh", "device.refresh", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+		state "default", label:"Refresh", action:"refresh.refresh", icon:"st.secondary.refresh"
+	}
+
+	main(["switch"])
+	details(["switch", "whiteSliderControl", "reset", "refresh"])
+}
+
+def updated() {
+     // Device-Watch simply pings if no device events received for 122 min (checkInterval)
+     log.debug("Adding checkInterval for health checks")
+     sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+     response(refresh())
+}
+
+def installed() {
+     // Device-Watch simply pings if no device events received for 122 min (checkInterval)
+     log.debug("Adding checkInterval for health checks")
+     sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+}
+
+def parse(description) {
+	def result = null
+	if (description.startsWith("Err 106")) {
+		state.sec = 0
+	} else if (description != "updated") {
+		def cmd = zwave.parse(description, [0x20: 1, 0x26: 3, 0x70: 1, 0x33:3])
+		if (cmd) {
+			result = zwaveEvent(cmd)
+			log.debug("'$description' parsed to $result")
+		} else {
+			log.debug("Couldn't zwave.parse '$description'")
+		}
+	}
+	result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd) {
+	dimmerEvents(cmd)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
+	dimmerEvents(cmd)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.switchmultilevelv3.SwitchMultilevelReport cmd) {
+	dimmerEvents(cmd)
+}
+
+private dimmerEvents(physicalgraph.zwave.Command cmd) {
+	def value = (cmd.value ? "on" : "off")
+	def result = [createEvent(name: "switch", value: value, descriptionText: "$device.displayName was turned $value")]
+	if (cmd.value) {
+		result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value , unit: "%")
+	}
+	return result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.hailv1.Hail cmd) {
+	response(command(zwave.switchMultilevelV1.switchMultilevelGet()))
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
+	def encapsulatedCommand = cmd.encapsulatedCommand([0x20: 1, 0x84: 1])
+	if (encapsulatedCommand) {
+		state.sec = 1
+		def result = zwaveEvent(encapsulatedCommand)
+		result = result.collect {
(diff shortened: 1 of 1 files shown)


vseven:
Pretty please?  @greens   😄 

Side note:  Any idea when "custom" tiles will come to the new app?  Mainly referring to "whiteSliderControl" that I used here which was how white channel was done on the Fibaro and I imaging any other RGBW "controller" as opposed to "bulb".

Side note 2: This should work for a Fibaro RGBW controller also but didn't have one to test with.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ *  Z-wave RGBW Controller (White chanel is dimmed and doesn't support color temperature)
+ *
+ *  Author: SmartThings / vseven (Allan)
+ *  Date: 2018-10-26
+ */
+
+metadata {
+	definition (name: "ZWave RGBW Controller", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb") {
```
greens, line 20: if it doesn't support color temperature it should just be "RGB"

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+			}
+			tileAttribute ("device.level", key: "SLIDER_CONTROL") {
+				attributeState "level", action:"switch level.setLevel"
+			}
+			tileAttribute ("device.color", key: "COLOR_CONTROL") {
+				attributeState "color", action:"setColor"
+			}
+		}
+  }
+  controlTile("whiteSliderControl", "device.whiteLevel", "slider", width: 2, height: 2, inactiveLabel: false) {
```
greens, line 57: spacing on this got pretty out-of whack because you pasted in something using spaces and the rest of the file uses tabs

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
greens: But it does support color temperature? Maybe it means something different to you, but we use "color temperature" and "white color" interchangeably.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
+	def encapsulatedCommand = cmd.encapsulatedCommand([0x20: 1, 0x84: 1])
+	if (encapsulatedCommand) {
+		state.sec = 1
+		def result = zwaveEvent(encapsulatedCommand)
+		result = result.collect {
+			if (it instanceof physicalgraph.device.HubAction && !it.toString().startsWith("9881")) {
+				response(cmd.CMD + "00" + it.toString())
```
greens, line 132: You don't have to do this anymore.

See how zwave-lock handles secure encapsulation.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.hailv1.Hail cmd) {
+	response(command(zwave.switchMultilevelV1.switchMultilevelGet()))
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
+	def encapsulatedCommand = cmd.encapsulatedCommand([0x20: 1, 0x84: 1])
+	if (encapsulatedCommand) {
+		state.sec = 1
```
greens, line 128: you don't need this anymore, as secure/insecure is now indicated by an "s" in the zwaveinfo.zw object.

see:

`zwaveInfo.zw.endsWith("s")`

in a couple of our DTHs.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: Color temperature to me means actual color temperature, 2700 K through 6500 K usually. However most rgbw controllers simply have a white channel that is 0 to 100% and acts like the R G & B channels.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+def setHue(value) {
+	log.debug "setHue($value)"
+	setColor(hue: value)
+}
+
+def setColor(value) {
+	def result = []
+	log.debug "setColor: ${value}"
+	if (value.hex) {
+		def c = value.hex.findAll(/[0-9a-fA-F]{2}/).collect { Integer.parseInt(it, 16) }
```
greens:
```suggestion
		def c = colorUtil.hexToRgb(value.hex)
```

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: Yeah, it seems the editor Within the device Handler on the account page treats things differently than GitHub. I can clean it up.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+	if(value.hex) sendEvent(name: "color", value: value.hex)
+	if(value.switch) sendEvent(name: "switch", value: value.switch)
+	if(value.saturation) sendEvent(name: "saturation", value: value.saturation)
+			
+	sendEvent(name: "whiteLevel", value: 0)
+	
+	commands(result)
+}
+
+def setWhiteLevel(percent) {
```
greens, line 201: Maybe the issue is that "color temperature" uses Kelvin instead of percent?

We've kind of hacked this in to the latest version of rgbw-light for reference (basically you just pretend its K in the range of 2700-6500 and find that percent:

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
```diff
@@ -0,0 +1,241 @@
...
+}
+
+def reset() {
+	log.debug "reset()"
+	sendEvent(name: "color", value: "#ffffff")
+	setWhiteLevel(99)
+}
+
+private command(physicalgraph.zwave.Command cmd) {
+	if (state.sec != 0) {
```
greens:
```suggestion
	if (zwaveInfo.zw.contains("s")) {
```

greens (commented):
Any reason you made a new DTH instead of just modernizing rgbw-light? Honestly I'd prefer that.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: For white I simply followed what you guys had in the Fibaro RGBW controller.  The problem with using color temperature is that it doesn't really support that capability so as you said its kinda a hack.  That and it will list in the connect app as color temp when its really dimmer control of the white channel.

Now the bad part of this is doing it "correctly" doesn't show a white slider in the connect app but does work correctly in classic.  Doing the hack route shows it in both but incorrectly.  I would think it should be done correctly and the connect app, I assume, will get updated to support custom tiles.

vseven:
Main reason is it really should be "z-wave rgbw bulb" to match the existing ZLL and ZigBee variants.  And since the new app doesn't deal with DTH changes the only thing to do would be create a new one and then move over once the new app works correctly.  I can make the changes there also once this is cleaned up, it should just be copy and paste at that point.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
greens: it's a hack, but it's a hack that works in the new app (which is a primary focus from our perspective).

greens:
I get what you're saying, but we'd rather have fewer DTHs than more consistently named ones.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: So remove the state.sec and the entire result part, leaving just the "zwaveEvent(encapsulatedCommand)" from what I can tell looking at the lock?

vseven (commented):
> but we'd rather have fewer DTHs than more consistently named ones.

Does the system handle a DTH rename?  I.e. could I kill my zwave-rgbw-bulb pull, put all the changes into the rgbw-light and after approved could you guys rename it and everyone never see the change?  I guess what I'm saying is the name a vanity name and the actual assigned DTH is some ID number on the back side or does it use the actual name?

greens:
It doesn't handle file renames well. Anyone with a currently-installed device using that DTH would have problems.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
greens: yep

vseven: Ok.  How do you handle line 86 then which is assigning state.sec = 0?  Remote that since you'd never check it again?  I.e. remove 86 & 87 then change 88 to just a if statement?

vseven:
Ok.  So since this is just for the "controller" which does act differently then a bulb (no color temp mainly) I'll still ask this is added once we resolve the issues.  Then I'll kill my pull for the "bulb" and update rgbw-light with the same changes so that's up to date.

vseven:
The more I look at this the more I'll coincide that until the Connect app supports custom tiles it doesn't make sense to have a separate DTH.  Once they do it definitely does since as mentioned RGBW controllers don't do "color temperature" but just do white dimming.  

With that said I'll close this pull and make the  changes in rgbw-light.  Once the app does support it I will create a "zwave-rgbw-controller" based on this but properly doing the white channel.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: nm, since you never check state.sec against anything there isn't a real good reason for setting it to 0.  Removed

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: Fixed.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: Fixed.

devicetypes/smartthings/zwave-rgbw-controller.src/zwave-rgbw-controller.groovy
vseven: Ok - Got a better answer for you after some testing.  On a RGBW "controller", the one I'm working on and a Fibaro and I imagine any other controller (non-bulb).  For bulbs color temperature is sent as a cool white and warm white value which are the inverse of each other.  But on the controller it just has a white channel.  So at 2700k with this DTH its sending the cool white channel 100% and the warm 0%.  The problem is this puts a RGBW controller to 100% white.  As you slide the slider "warmer" to 6500K the white channel drops to 0%.  So what you see on the screen is the opposite of what is actually happening.

For that reason you do need a RGBW Controller DTH in addition to a RGBW bulb DTH.  But like I said I'll let you review the changes to RGBW-Light and once its good copy/paste into a controller DTH and adjust that color temp to operate the white channel correctly.  I'll also remove he color temp capability since it doesn't have it.

With that said there are RGBWW controllers out there and analog RGBWW led strips that do have "cool" and "warm" white leds which would work correctly on the RGBW-Light DTH.  But I've never seen anyone using them and if someone was their fingerprint could just be added to RGBW-Light.

PR #3648, opened 2018-10-31
DVCSMP-4499 Z-Wave Lock: set manufacturer device data for LEE

mckeed:
Local execution for Z-Wave lock alarm report code 18 is currently relying on the manufacturer name being in device data, but it hasn't been getting set any more if everything is handled locally.


greens:
LE uses `state.manufacturer` over the zwInfo object?

PR #3651, opened 2018-11-01
WWST-1624 Add enrollResponse() to configure()

dkirker:
While it isn't necessary to handle the "enroll request" message in parse(), we do need to send an enrollResponse() to the bulb in order to receive the motion events.

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer-with-motion-sensor.src/zigbee-dimmer-with-motion-sensor.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer-with-motion-sensor.src/zigbee-dimmer-with-motion-sensor.groovy
@@ -157,5 +157,5 @@ def configure() {
 	setupHealthCheck()
 
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + refresh()
+	zigbee.onOffConfig(0, 300) + zigbee.levelConfig() + zigbee.enrollResponse() + refresh()
 }


dkirker:
@tpmanley @varzac So, in my playing with dealing with enrolling I removed the enrollResponse() in refresh() when I erroneously believed that it was not necessary. Unfortunately, it is mandatory in order for us to receive motion events. The bulb seems to default to this behavior of keeping the motion logical internal to the bulb and behaving as a motion-activated flood light would. In this instance, no motion events are sent.

Unless there is a reason that we shouldn't treat the bulb as a dimmer and a motion sensor that are "independent", I have added back the enrollResponse() so that we get the motion events. (Another option would be to "fake" motion events when we receive on and off in an unsolicited manner.)

I suspect without this PR that certification will fail. I figured that it is small and Raja has been testing it today that maybe we can deploy this tomorrow?

devicetypes/smartthings/zigbee-dimmer-with-motion-sensor.src/zigbee-dimmer-with-motion-sensor.groovy
```diff
@@ -139,7 +139,7 @@ def ping() {
 }
 
 def refresh() {
-	zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS)
+	zigbee.onOffRefresh() + zigbee.levelRefresh() + zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS) + zigbee.enrollResponse()
```
tpmanley: The `enrollResponse` should be sent by `configure` instead of `refresh` since it's used to configure the device.

tpmanley (dismissed):
Let me know once this has been tested and I can merge

tpmanley:
Also can you update the commit message since enrollResponse isn't in refresh anymore?

dkirker:
@tpmanley Done!

tpmanley (approved):

PR #3666, opened 2018-11-08
Update zigbee-switch-power.groovy

vseven:
Re: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3631

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -28,7 +28,12 @@ metadata {
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "0003, 000A, 0019", manufacturer: "Jasco Products", model: "45853", deviceJoinName: "GE ZigBee Plug-In Switch"
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "000A, 0019", manufacturer: "Jasco Products", model: "45856", deviceJoinName: "GE ZigBee In-Wall Switch"
     }
-
+    
+	preferences {
+		input title: "Power Correction", description: "Some sensors display the power value as a multiple of ten to get around a decimal place.  If you device does this you can add a divisor to offset the value.  A value of 1 makes no correction, a value of 10 will divide the sensors value by ten, etc.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+		input "powerDivisor", "number", title: "Divisor", description: "Divide power value by this number", range: "1..100", displayDuringSetup: false
+    }
+    
     tiles(scale: 2) {
         multiAttributeTile(name:"switch", type: "lighting", width: 6, height: 4, canChangeIcon: true){
             tileAttribute ("device.switch", key: "PRIMARY_CONTROL") {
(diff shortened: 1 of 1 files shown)


orangebucket:
Doesn't the committed code change the default divisor to 1, which would break most things?

Also typo 'you device' in preferences. I query whether the description is unnecessarily verbose and might say something more like 'Divisor for power value. Default 10.' I think that tells the punters what they need to know.

I also observe that Zigbee seems to use power values in kilowatts, so the devices would have to multiply their values by 10,000 and not 10 to store power values less than one watt in unsigned integers. ~~I find it somewhat counter-intuitive that the device handler receives power values in watts (x 10) and then applies a further divisor. Surely a division by 1,000 must already have taken place to get that far? Isn't that just an assumption too? What if the device manufacturers decided the nearest kilowatt was close enough for their needs and hadn't multiplied up?~~ _Sorry, that was absolute drivel._

Aside:  Following up  #3631 thread, the power events from the offending devices do seem to use cluster 0x0702, which also matches the cluster in the raw description in the devices list. Odd. Probably enough said on that though.

G,

vseven:
I could put a default value on the preference of 10 or I could say if value is null then value is 10.  Suppose having the preference default to 10 makes the most sense. And you're correct it is a little wordy but so are most other device handlers.

orangebucket:
> I could put a default value on the preference of 10 or I could say if value is null then value is 10. Suppose having the preference default to 10 makes the most sense. And you're correct it is a little wordy but so are most other device handlers.

It was mentioned previously that user preferences don't work with the new app. Might that also mean a default preference wouldn't work either?

vseven:
That is true....there isn't even a screen for user preferences yet.  However you can set it in "Classic" and it carries over to the new Connect app.

With that said....if null divide by 10 otherwise use the preference?  This way its not messing anyone else up and if the preference is never touched it will work as it currently does?

vseven:
Removed a big of verbiage, left the rest because it does explain whats going on.  Changed the code to say if the preference is entered use that (so a user like you can enter 1) but if no preference was set divide by 10.  This way the change shouldn't affect anyone unless they purposely tried to change it.

vseven:
@greens fixed the spacing and updated per above so the change shouldn't affect existing devices.

devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -28,7 +28,12 @@ metadata {
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "0003, 000A, 0019", manufacturer: "Jasco Products", model: "45853", deviceJoinName: "GE ZigBee Plug-In Switch"
         fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0702, 0B05", outClusters: "000A, 0019", manufacturer: "Jasco Products", model: "45856", deviceJoinName: "GE ZigBee In-Wall Switch"
     }
-
+    
+	preferences {
+		input title: "Power Correction", description: "Some sensors report the power value as a multiple of ten.  If your device does this you can add a divisor to offset the value.  A value of 1 makes no correction, a value of 10 will divide the sensors value by ten, etc.  If not entered the device will default to 10.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+		input "powerDivisor", "number", title: "Divisor", description: "Divide power value by this number", range: "1..100", displayDuringSetup: false
```
greens, line 34: the three lines above this are tabbed

maybe your editor has a "convert tabs to spaces" function?

PR #3671, opened 2018-11-08
Update rgbw-light.groovy

vseven:
@greens Per our conversation.  Basic changes:

- Added capability "Light"
- Made some basic changes per https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3639


vseven:
Also after testing it turns out I will have to put in a pull for a "controller".  Reason being is with the color temperature hack in this DTH the white channel on a "controller" is the inverse of the value.  So at 2700K its at 100% output, at 6500K its at 0% output.  Which of course doesn't work real well.

PR #3677, opened 2018-11-14
[WWST-1904; WWST-1903] Fingerprints for Aeotec Door/Window Sensor and LED Bulbs

PKacprowiczS:
@greens @MAblewiczS


MAblewicz (approved):
Only Fingerprint change. Look ok.

PR #3679, opened 2018-11-14
ICP-6568 Fix exception in Aeon Home Energy Meter gen5

greens:


PKacprowiczS (approved):
Jeez, I should've caught that, but live logging didn't show that exception...

tpmanley (approved):

PR #3684, opened 2018-11-16
[ICP-7144] Zigbee Valve checkinterval fix

PKacprowiczS:
@greens @MAblewiczS

### DIFF
--- a/devicetypes/smartthings/zigbee-valve.src/zigbee-valve.groovy
+++ b/devicetypes/smartthings/zigbee-valve.src/zigbee-valve.groovy
@@ -19,6 +19,7 @@ metadata {
         capability "Battery"
         capability "Configuration"
         capability "Power Source"
+        capability "Health Check"
         capability "Refresh"
         capability "Valve"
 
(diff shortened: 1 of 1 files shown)


tpmanley (changes requested):
`ping` will also need to be implemented

devicetypes/smartthings/zigbee-valve.src/zigbee-valve.groovy
```diff
@@ -140,3 +141,11 @@ def configure() {
     log.debug "Configuring Reporting and Bindings."
     refresh()
 }
+
+def installed() {
+    sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+}
+
+def ping() {
+    refresh()
```
tpmanley: `refresh` sends a lot of commands. We generally want `ping` to only send one or two commands since injecting a bunch of traffic into the network unnecessarily can degrade performance. I suggest changing this to just `zigbee.onOffRefresh()`

PKacprowiczS:
@tpmanley it should be right this time

tpmanley (approved):

PR #3719, opened 2018-11-28
ICP-7312 "fixes" fibaro mis-identifications

greens:


devicetypes/fibargroup/fibaro-co-sensor-zw5.src/fibaro-co-sensor-zw5.groovy
```diff
@@ -15,7 +15,8 @@ metadata {
 		attribute "coLevel", "number"
 
 		fingerprint mfr: "010F", prod: "1201", model: "1000"
-		fingerprint deviceId: "0x0701", inClusters:"0x5E,0x59,0x73,0x80,0x22,0x56,0x31,0x7A,0x5A,0x85,0x84,0x71,0x70,0x8E,0x9C,0x86,0x72"
+		fingerprint mfr: "010F", prod: "1201", model: "1001"
+		fingerprint mfr: "010F", prod: "1201"
```
greens, line 19: I'm not sure if just including these two fields is kosher, but I think it should be?

greens:
Have confirmed that this works.

devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -23,6 +23,9 @@ metadata {
 		capability "Temperature Alarm"
 
 		fingerprint mfr:"010F", prod:"0C02", model:"1002"
+		fingerprint mfr:"010F", prod:"0C02", model:"4002"
```
tpmanley, line 26: Is there any benefit of including some specific models vs the single fingerprint with must mfg/prod?

greens: lets us know which models we've specifically certified, I suppose.

dkirker: The other benefit I see is making sure we don't possibly hit edge cases with existing devices. I feel like if we remove the existing fingerprints then we need to re-test all devices that we removed. Though, maybe we should anyway.

tpmanley (approved):

dkirker (approved):
Looks like a good change. Probably will help lower support requests, too.

PR #3724, opened 2018-11-29
ONEAPP-7319 Send an initial button event to appease OCF

dkirker:
Button capability is momentary, so initial event is "pushed"

### DIFF
--- a/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
+++ b/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
@@ -245,6 +245,11 @@ private Map getButtonResult(buttonState, buttonNumber = 1) {
 
 def installed() {
     initialize()
+
+    // Initialize default states
+    device.currentValue("numberOfButtons")?.times {
+        sendEvent(name: "button", value: "pushed", data: [buttonNumber: it+1], displayed: false)
+    }
 }
 
 def updated() {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -245,6 +245,11 @@ private Map getButtonResult(buttonState, buttonNumber = 1) {
 
 def installed() {
     initialize()
+
+    // Initialize default states
+    device.currentValue("numberOfButtons")?.times {
+        sendEvent(name: "button", value: "pushed", data: [buttonNumber: num+1])
```
tpmanley: I think `num` needs to be `it` or you need to rename it with `num ->`

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -245,6 +245,11 @@ private Map getButtonResult(buttonState, buttonNumber = 1) {
 
 def installed() {
     initialize()
+
+    // Initialize default states
+    device.currentValue("numberOfButtons")?.times {
+        sendEvent(name: "button", value: "pushed", data: [buttonNumber: it+1])
```
tpmanley: Probably want to add `displayed: false`

dkirker: Yeah, not a bad idea.

dkirker: Should I add that to the "numberOfButton" events, too? Probably a good idea while I am here.

tpmanley: yeah i'd say so

dkirker: Done!

tpmanley (approved):

rappleg (approved):

PR #3725, opened 2018-11-30
ICP-7202

KKlimczukS:
@greens @MAblewiczS @MGoralczykS @DCzarneckaS @MMateusiakS @PKacprowiczS 
Please, could You take a look ?
 
DTH relied too much on NotificationReport (the state on/off was set to according to it). Meanwhile workaround for cleaning tamper was added and it caused switching siren OFF after few seconds. In case of Yale Siren tamper only activates the siren and doesn't change it's internal state (SwitchBinaryGet reported 0 when the siren was ON), that's why we shouldn't rely on it.

### DIFF
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -26,7 +26,6 @@ metadata {
 		capability "Sensor"
 		capability "Switch"
 		capability "Health Check"
-		capability "Tamper Alert"
 
 		fingerprint inClusters: "0x20,0x25,0x86,0x80,0x85,0x72,0x71"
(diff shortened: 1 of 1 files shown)


tpmanley:
Has this been tested while running locally?

greens:
Remove the "Tamper" capability (and all related messages) since it is now unused.

Fun that we got to have that big fight about it and generate new metadata that got rolled back for something that is getting completely removed anyway.

KKlimczukS:
Yes, it was tested with runLocally set to true. Wouldn't call it "a big fight" ;) - we expected that tamper would do more stuff than just physically turning the siren on. There are two tamper springs in the siren, but only one of them does anything (verified on two Yale devices that we have here).

greens (approved):

workingmonk:
updated PR title with correct ticket format

PR #3727, opened 2018-12-03
[ICP-7220] Fingerprint fix.

MGoralczykS:
@MAblewiczS @MMateusiakS @PKacprowiczS @DCzarneckaS @KKlimczukS 
There were two fingerprints. Diffrences between them were "deviceJoinName" and "outClasters"  I've deleted one fingerprint and added missing claster to "outClusters". Both bulbs was checked and them are pairing with correct name. 
 
Could you review this?


MAblewicz (approved):
Looks ok. Two FP (for same device) were merged into one.

PR #3738, opened 2018-12-04
ICP-7184 Query locks for battery level when power is restored

greens:
After consultation with Yale, they informed us that this notification is sent when power is restored. We'll use it to trigger a battery poll in the case that batteries were changed.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -626,6 +626,9 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch(cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -626,6 +626,9 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch(cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
```
tpmanley, line 629: Does this work with local execution?

greens: Yeah. LE only handles types 1, 7, 2, and 5. Everything else falls through. @kianooshST can you  confirm?

greens:
Verified by SR-Pol.

tpmanley (approved):

PR #3740, opened 2018-12-05
ICP-7369 ONEAPP-8773 add new version orvibo smoke sensor fingerprint.

BYDengS:
Hello,@greens ,@MAblewiczS ,and @ALamchaS .The ORVIBO smoke sensor have a new version device, and I have added the new version fingerprint in DTH.Please help review and merge it to China IOT Server.


tpmanley:
ICP-7369 looks like it might be a duplicate of ONEAPP-8773. At the very least both tickets should be referenced but you could also use the ONEAPP ticket instead since it has a lot of visibility and close the ICP ticket as a duplicate.

greens (approved):

greens:
published on staging and prod in CN

PR #3745, opened 2018-12-05
ICP-7323 ICP-7309 ICP-7308 ICP-7289 ICP-7282 ICP-7288 Stelpro Ki zigb…

greens:
…ee issues

* Missing a unit when reporting a setpoint change
* Possibly preventing operating state changes from being propagated
* Device seems to use a different C->F conversion than us

### DIFF
--- a/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
@@ -52,7 +52,6 @@ metadata {
 
 	preferences {
 		input("lock", "enum", title: "Do you want to lock your thermostat's physical keypad?", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: false)
-		input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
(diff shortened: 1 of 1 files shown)


devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -315,10 +315,6 @@ def parse(String description) {
 				} else {
 					map.value = "heating"
 				}
-
-				if (settings.heatdetails == "No") {
-					map.displayed = false
```
dkirker, line 320: This should only affect the event history, and in Classic. If this is actually preventing events from reaching the event pipeline, then there is a bug in data management.

See https://community.smartthings.com/t/stelpro-releases-the-first-z-wave-plus-line-voltage-thermostat-the-stzw402-stelpro-ki-thermostat/39278/179 for a reference to the purpose of this.

greens: Based on the live logs I was seeing, it wasn't getting switched anyway (but the events were). I was just throwing stuff at the wall.

greens: My guess is it's actually a ui metadata pointing at the wrong href.

dkirker: Turns out that the device was being configured to not report changes more frequently than 5 minutes, as well as the cadence of the duty cycles and the code that handles passing off the Zigbee messages. The ticket(s) include details.

PR #3748, opened 2018-12-06
ICP 7395

BYDengS:
These commit is to fix the bug that when set up ORVIBO gas sensor successfully, no status in ST app gas sensor plugin. And @greens @ALamchaS and @ALamchaS , please help review this commit and help to merge to China STG server.


greens:
Are you sure this is the right issue? ICP-7395 seems to be about a missing fingerprint. (Also ALamchaS is no longer with the company.)

BYDengS:
I have tested it by using the latest DTH.And found this issue. When I easy setup the gas sensor successfully, and don't trigger any event. The IDE and ST app plugin is no status for this gas sensor.And I found it is caused by DTH parse the zigbee command failed. So I changed the DTH. is it need to create a new JIRA bug to update the code?

BYDengS:
Ok,I will create a new JIRA Bug, and submit this change by using the bug's id. So I closed this request.

PR #3749, opened 2018-12-06
[ICP-7359] Add refresh() to check actual status when device is installed.

MGoralczykS:
@greens @MAblewiczS @MMateusiakS @KKlimczukS 
Could you review this?


greens:
Won't work. The device doesn't even know its own state until its been moved.

PR #3750, opened 2018-12-06
ICP-7174 ICP-7282 ICP-7283 ICP-7308 ICP-7309 ICP-7288 ICP-7289 Misc Stelpro fixes

dkirker:
ICP-7288 Change rounding method for temperature conversion
ICP-7308 ICP-7309 Tweaks to operating state logic (Not a full resolution)
ICP-7282 ICP-7283 Add heatingSetpoint temp unit
ICP-7174 Update Stelpro Ki logic errors

And 48eaf2b might fix part of ICP-7289


devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -141,6 +141,7 @@ def setupHealthCheck() {
 
 def configureSupportedRanges() {
 	sendEvent(name: "supportedThermostatModes", value: supportedThermostatModes, displayed: false)
+	// These are part of the deprecated Thermostat capability. Remove these when that capability is removed.
```
dkirker, line 144: I am checking to see if OneApp uses this in anyway. It didn't seem like its presence affected the range displayed in OneApp, but I want to make sure it isn't needed because of the reference to "Thermostat" in the capabilities.

This applies to all three device handlers.

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -310,13 +313,16 @@ def parse(String description) {
...
 					map.value = "idle"
 				} else {
 					map.value = "heating"
 				}
 
-				if (settings.heatdetails == "No") {
+				// If the user does not want to see the Idle and Heating events in the event history,
+				// don't show them. Otherwise, don't show them more frequently than 30 seconds.
+				if (settings.heatdetails == "No" ||
+					!secondsPast(device.currentState("thermostatOperatingState").getLastUpdated(), 30)) {
```
dkirker: I did this to try to compensate for the change to the operating state reporting intervals in configure(). I don't have a problem removing the extra 30 second logic, though.

cc @stelpro

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -276,23 +278,27 @@ def parse(String description) {
...
 					map.value = getTemperature(descMap.value)
+					map.unit = getTemperatureScale()
 					map.data = [heatingSetpointRange: heatingSetpointRange]
 				}
 			} else if (descMap.attrInt == ATTRIBUTE_PI_HEATING_STATE) {
 				def intVal = Integer.parseInt(descMap.value, 16)
 				log.debug "HEAT DEMAND"
 				map.name = "thermostatOperatingState"
-				if (intVal < 10) {
+				if (intVal < 0x10) {
```
dkirker: @stelpro Just to check, we are comparing the value with 16, correct? I changed the code you originally submitted which compared "10" with the hex string. I just want to make sure that I am reading that correctly. (I accidentally just removed the quotes and didn't add the "0x" part.)

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
stelpro: This is good idea.
The 30 seconds could be set to another value, but I suggest to keep it that way for now.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
stelpro: No, we were comparing with the 10 (decimal) value.

We don't want to recreate the same issue like in the Z-Wave DTH.
So we don't want to advertise EXACTLY when the heat is On or Off on a 15 seconds cycle.

The logic behind this is that we are considering the thermostat to be in heating mode when the heat percentage is above or equal to 10%.

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -291,22 +296,27 @@ def zwaveEvent(sensormultilevelv3.SensorMultilevelReport cmd) {
 
 		temp = map.value
 		// The specific values checked below represent ambient temperature alarm indicators
-		if (temp == "32765") {			// 0x7FFD
+		if (temp == 0x7ffd) {
 			map.name = "temperatureAlarm"
 			map.value = "freeze"
-			map.unit = ""
-		} else if (temp == "32767") {		// 0x7FFF
+			map.unit = null
```
tpmanley: Does setting `unit` to `null` work? Another option might be to do `map.remove('unit')` if the goal is to get rid of the unit. I'm  not sure if one way is preferable over another if they both work.

tpmanley: Actually I think it'd probably be more clear to just fully initialize the event for each specific case rather than setting value,  unit and name up front and overriding certain values. It might be a little extra code but it won't be much and I think it will be easier to understand.

tpmanley (dismissed):

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
dkirker: Awesome. I updated this back to 10 in a new commit. When I look into some of the deeper logic handling on setpoint changes I will check what events I am receiving. I *might* remove this check and improve the 30 seconds logic below to mitigate the "flapping" effect. I'm thinking something that pays attention to the reported temperature and the setpoint as well.

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
dkirker: Setting unit to null does work!

So far, since the logic was getting big, I moved the code out of parse() and put it in handleTemperature(). I am actually going to change the way parse() builds events because there will be a few times I want to pass more than one thing back from parse().

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -326,6 +302,45 @@ def parse(String description) {
...
 
+def handleTemperature(descMap) {
+	def map = [:]
+	def intVal = Integer.parseInt(descMap.value, 16)
+	map.name = "temperature"
+	map.unit = getTemperatureScale()
+	map.value = getTemperature(descMap.value)
+
+	// Handle special temperature flags where we need to change the event type
+	if (intVal == 0x7ffd) {
```
stelpro: Here, in order to take care of the bug in the device, I would simply rearrange the if-else statement to check for a negative value at the end.

If a negative temperature value is found (not 0x8000, which is an error code), then the Freeze alarm should be set.

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -337,6 +313,45 @@ def parse(String description) {
...
+		map.name = "temperatureAlarm"
+		map.value = "heat"
+		map.unit = null
+	} else if (intVal == 0x8000) {
+		map.name = null
+		map.value = null
+		map.unit = null
+		map.descriptionText = "Received a temperature error"
+	} else if (intVal > 0x8000) {
+		map.value = -(Math.round(2*(655.36 - map.value))/2)
```
greens: You should probably leave a comment explaining this special case. Also you probably want to be using `intVal` here instead of the converted `map.value`.

stelpro:
Donald,

Don't forget to add a check on the temperature to report a freezing alarm when the temperature is below 0°C

dkirker:
@stelpro I did some tweaking to re-organize the handleTemperature logic.

stelpro:
@dkirker that sounds good

greens (dismissed):
Looks good, but wait until all the bugs get confirmed first.

dkirker:
@greens Made one addition to the zigbee DTH to work around that bug. If Poland checks out, then we can merge.

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -362,11 +353,32 @@ def handleTemperature(descMap) {
...
 	}
 
 	map
 }
 
+// Due to a bug in this model's firmware, sometimes we don't get
+// an updated operating state; so we will force it.
+// TODO: Add firmware version check when change versions are known
+def handleOperatingStateBugfix(setpoint, temp) {
+	def currSetpoint = (setpoint != null) ? setpoint : device.currentValue("heatingSetpoint")
```
greens: this can just be a `?:`

greens: I guess you are avoiding the 0 degree corner case by making it an explicit null check

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -362,11 +353,32 @@ def handleTemperature(descMap) {
 			log.debug "CLEAR ALARM @ $map.value $map.unit (raw $intVal)"
 			sendEvent(name: "temperatureAlarm", value: "cleared")
 		}
+
+		handleOperatingStateBugfix(null, map.value)
```
greens: you could tidy up the function below if you just passed in the current values here instead of getting them down there

Then you could avoid the extra variables you're creating.

dkirker: It sounds like given some testing results, and the behavior of this bug, I might go about this a different way. More after my testing.

dkirker:
@greens  Alright, so the logic has changed a bit. Depending on what cert/QA says, this is probably the best we can do. I considered the route of managing the operating state ourselves, but I think that would add a bit more logic that I don't want to consider. This should at least make sure we are getting sane values, and that we have received the latest values.

dkirker:
@rappleg or @tpmanley or @varzac Any chance I can get a +1 from any of you?
@workingmonk If we want to target this for a 1/8 deploy to production, would I need to rebase this to staging, or will there be a push from master to staging before?

tpmanley (dismissed):

greens:
looks like you have merge conflicts now @dkirker

dkirker:
Yeah. Looks like a weather API change. Minor. Still trying to figure out the release schedule for next week. It looks like there was already a rolling down and up of branches. But waiting for Vinay to confirm on chat.

dkirker:
Alright, rebased to staging, and fixed indentation changes made from another commit.

greens (approved):

PR #3753, opened 2018-12-07
[WWST-2020] New DTH - Z-Wave Button

PKacprowiczS:
@greens @MAblewiczS @MMateusiakS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
+++ b/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
@@ -17,6 +17,7 @@
 metadata {
 	definition (name: "Z-Wave Button", namespace: "smartthings", author: "SmartThings") {
 		capability "Button"
+		capability "Battery"
 		capability "Sensor"
 		capability "Health Check"
 
--- a/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
+++ b/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
@@ -1,0 +1,128 @@
+/**
+ *  Z-Wave Button
+ *
+ *  Copyright 2018 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+metadata {
+	definition (name: "Z-Wave Button", namespace: "smartthings", author: "SmartThings") {
+		capability "Button"
+		capability "Sensor"
+		capability "Health Check"
+
+		fingerprint mfr: "010F", prod: "0F01", model: "1000", deviceJoinName: "Fibaro Button"
+	}
+
+	tiles(scale: 2) {
+		multiAttributeTile(name: "button", type: "generic", width: 6, height: 4, canChangeIcon: true) {
+			tileAttribute("device.button", key: "PRIMARY_CONTROL") {
+				attributeState "default", label: ' ', icon: "st.unknown.zwave.remote-controller", backgroundColor: "#ffffff"
+			}
+		}
+		valueTile("battery", "device.battery", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+			state "battery", label:'${currentValue}% battery', unit:""
+		}
+
+		main "button"
+		details(["button", "battery"])
+	}
+}
+
+def installed() {
+	sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	sendEvent(name: "numberOfButtons", value: 1)
+	response([
+			secure(zwave.batteryV1.batteryGet()),
+			"delay 2000",
+			secure(zwave.wakeUpV2.wakeUpNoMoreInformation())
+	])
+}
+
+def parse(String description) {
+	def result = []
+	if (description.startsWith("Err")) {
+		result = createEvent(descriptionText:description, isStateChange:true)
+	} else {
+		def cmd = zwave.parse(description)
+		if (cmd) {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+			secure(zwave.batteryV1.batteryGet()),
+			"delay 2000",
+			secure(zwave.wakeUpV1.wakeUpNoMoreInformation())
+	])
+}
+
+def configure() {
+	if(zwaveInfo.mfr?.contains("0086"))
+		[
+			secure(zwave.configurationV1.configurationSet(parameterNumber: 250, scaledConfigurationValue: 1)),
```
greens: can you leave a comment about what this parameter does?

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		log.warn "Unable to extract encapsulated cmd from $cmd"
+		createEvent(descriptionText: cmd.toString())
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.centralscenev1.CentralSceneNotification cmd) {
+	def value = eventsMap[(int) cmd.keyAttributes]
+	def event = createEvent(name: "button", value: value, descriptionText: "Button was ${value}")
+	[event, response(secure(zwave.wakeUpV1.wakeUpNoMoreInformation()))]
```
greens: You shouldn't need to send a `NoMoreInformation` here. That only needs to be sent in response to a `WakeUpNotification`, and that case should be handled by _that_ `zwaveParse` method.

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+		zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
+	} else {
+		cmd.format()
+	}
+}
+
+private getEventsMap() {[
+		0: "pushed",
+		1: "up_hold",
+		2: "down_hold",
```
greens, line 137: You should just map one of these to 'held' since it's more widely supported. I'd do whichever one is sent when the button is released.

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.centralscenev1.CentralSceneNotification cmd) {
+	def value = eventsMap[(int) cmd.keyAttributes]
+	def event = createEvent(name: "button", value: value, descriptionText: "Button was ${value}")
+	[event, response(secure(zwave.wakeUpV1.wakeUpNoMoreInformation()))]
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.sceneactivationv1.SceneActivationSet cmd) {
```
greens, line 90: what's the context for when this gets sent?

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+metadata {
+	definition (name: "Z-Wave Button", namespace: "smartthings", author: "SmartThings") {
+		capability "Button"
+		capability "Battery"
+		capability "Sensor"
+		capability "Health Check"
+		capability "Configuration"
+
+		fingerprint mfr: "010F", prod: "0F01", model: "1000", deviceJoinName: "Fibaro Button"
+		fingerprint mfr: "0086", prod: "0001", model: "0026", deviceJoinName: "Aeotec Panic Button"
```
greens: We already have this fingerprint in https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		main "button"
+		details(["button", "battery"])
+	}
+}
+
+def installed() {
+	sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	sendEvent(name: "numberOfButtons", value: 1)
```
greens, line 45: Because of how OneApp works, you'll need to also send a fake initial "pushed" event here.

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
```diff
@@ -0,0 +1,148 @@
...
+		zwaveEvent(encapsulatedCommand)
+	} else {
+		log.warn "Unable to extract encapsulated cmd from $cmd"
+		createEvent(descriptionText: cmd.toString())
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.centralscenev1.CentralSceneNotification cmd) {
+	def value = eventsMap[(int) cmd.keyAttributes]
+	def event = createEvent(name: "button", value: value, descriptionText: "Button was ${value}")
```
greens: Please include `data: [buttonNumber: 1]` in all the button events you send. I'm not quite sure if it's necessary, but we're including it for now.

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
greens: so this PR should also include removing that fingerprint from that DTH if that's the intention.

PKacprowiczS: Yup, that was intended. That is, if I understood Eric correctly in WWST-2025.

devicetypes/smartthings/zwave-button.src/zwave-button.groovy
PKacprowiczS: This is how Aeotec Panic Button reports its button events.

greens:
So I just realized that because of the way ui metadata works, if we want to have the extra events for the fibaro button, we shouldn't move the panic button to this DTH. UI-metadata will have to display all the possible events all the time, even though they can't be triggered by the panic button. Once HCS-299 is done we can move it, but until then, panic button should stay in its original DTH.

greens (approved):

greens:
Just checked. All the extra button presses aren't mapped/localized yet, so I'm going to try to get the ball moving on that.

PR #3757, opened 2018-12-10
ICP-7184 Query locks for battery level when power is restored (#3738)

greens:
After consultation with Yale, they informed us that this notification is sent when power is restored. We'll use it to trigger a battery poll in the case that batteries were changed.


greens:
cherry-pick to staging

tpmanley (approved):

PR #3770, opened 2018-12-12
[WWST-1943] Add fingerprint for innr candle white RB 245

MMateusiakS:
@MAblewiczS  fingerprint added
Raw Description	01 0104 0101 01 07 0000 0003 0004 0005 0006 0008 1000 01 0019


MAblewicz (approved):
Look Ok.

workingmonk:
@MAblewiczS the deviceJoinName capitalization is incorrect

MAblewicz:
@workingmonk @tpmanley @greens  is there a standard in naming devices. I checked different "Innr" bulbs (in this DTH and in the other) and there is no standard for this. Sometimes "innr" word is starting will capital letter, sometimes not.  Sometimes every word in name is starting with capital letters sometimes not. If we agree on one device naming standard i will create new issue and fix all Innr device names according to that standard.

greens:
@MAblewiczS looks like previously we did not capitalize the brand name "innr" but capitalized all the other words in the join name.

PR #3775, opened 2018-12-12
ONEAPP-8822 Prevent master code set reports from generating incorrect…

greens:
… events


greens:
I think this should be all we need. @vaish-rahul

PR #3788, opened 2018-12-14
ICP-7489 change vid for ozom smart siren as it doesn't support strobo…

greens:
…scope

Change-Id: Iee34c9893dd1787b91096487aab3c74a00a75a33


greens:
cherry-pick to staging

tpmanley (approved):

PR #3790, opened 2018-12-15
Wwst 2048: Add DTH for LUMI contact sensor

FenMei007S:
@tpmanley  @varzac  LUMI contact sensor also use 0500 to report battery normal and low


devicetypes/smartthings/zigbee-contact-sensor.src/zigbee-contact-sensor.groovy
```diff
@@ -0,0 +1,207 @@
...
+	def cluster = zigbee.parse(description)
+
+	if (shouldProcessMessage(cluster)) {
+		def msgStatus = cluster.data[2]
+
+		log.debug "parseCatchAllMessage: msgStatus: ${msgStatus}"
+		if (msgStatus == 0) {
+			switch(cluster.clusterId) {
+				case 0x0500:
+                	Map descMap = zigbee.parseDescriptionAsMap(description)
```
varzac, line 114: You have some indentation issues, you should make sure that you are consistently using tabs.

devicetypes/smartthings/zigbee-contact-sensor.src/zigbee-contact-sensor.groovy
```diff
@@ -0,0 +1,207 @@
...
+	    		 resultMap_battery.name = "battery"
+	    		 resultMap_battery.value = 50
+	    	 }
+	    	 else {
+	    		 if (zs.isBatterySet()) {
+	    			 // to generate low battery notification by the platform
+	    			 resultMap_battery.name = "battery"
+	    			 resultMap_battery.value = 5
+	    		 }
+	    		 else {
```
varzac, line 172: else should be on the same line as the `}` above

varzac:
Historically we have received a lot of complaints from customers when a device that they just joined reports < 100% battery (often times complaints about 67%).  This will report 50% immediately.  However, I'm not sure what else we could do for a device like this that only reports low.

FenMei007S:
@varazc,  LUMI sensor will add POWER CONFIGURE CLUSTER to support reporting battery percentage, so the battery 50% logic will be moved, we are waitting LUMI sending new devices to us.

varzac:
@FenMei007S Great, that will be much better.  Just some formatting issues then, otherwise it looks good.

PR #3799, opened 2018-12-17
Ps 001 - Salus SP600 Smart Plug/Socket

paulsheldon:
Adds fingerprint for Salus SP600 Smart Plug/Socket


tpmanley:
@paulsheldon , thanks for the PR! There are some comments on the community that the power reports using this DTH is off by a factor of 10. What has been your experience?

paulsheldon:
I'm testing that with a few appliances and will come back to you, early indications are yes maybe

PR #3807, opened 2018-12-19
ICP-6917 Applied new vid of UI metadata

vangoran:


vangoran:
@greens @tpmanley @kyunyongpark
According to the comments of ICP-6917, I applied new VID of UI metadata for Samsung Smart Doorlock, please review it.

dkirker:
@vangoran You should be fine removing the
`, mnmn: "SmartThings", vid: "SmartThings-smartthings-Samsung_Smart_Doorlock"`
part so all you are left with is
`definition (name: "Samsung Smart Doorlock", namespace: "Samsung SDS", author: "kyun.park") {`

The system will come up with the vid using the DTH name appended to "SmartThings-smartthings-" with spaces replaced as underscores. So:

Samsung Smart Doorlock

becomes

SmartThings-smartthings-Samsung_Smart_Doorlock

dkirker (changes requested):
Is this a device that is currently in use by the public? Or is it currently only with developers and certification? If it is only with developers and cert you should be fine removing the "mnmn" and "vid" parts.

vangoran:
> @vangoran You should be fine removing the
> `, mnmn: "SmartThings", vid: "SmartThings-smartthings-Samsung_Smart_Doorlock"`
> part so all you are left with is
> `definition (name: "Samsung Smart Doorlock", namespace: "Samsung SDS", author: "kyun.park") {`
> 
> The system will come up with the vid using the DTH name appended to "SmartThings-smartthings-" with spaces replaced as underscores. So:
> 
> Samsung Smart Doorlock
> 
> becomes
> 
> SmartThings-smartthings-Samsung_Smart_Doorlock

Okay, I will remove the useless definition parameters

dkirker (approved):
Since mnmn and vid were explicitly defined before, we should leave them explicitly defined, even if they match the default values.

PR #3809, opened 2018-12-19
[WWST-2182] Add fingerprint for Sengled Element Color Plus B22 - E11-U3E

DCzarneckaS:
@MMateusiakS @greens @MAblewiczS 
Raw : 
 01 0104 0102 00 0A 0000 0003 0004 0005 0006 0008 0300 0702 0B05 FC03 01 0019


MAblewicz (approved):
Looks good.

PR #3832, opened 2018-12-29
Added Shelly 2 device handler as roller shutter

dgasparri:
Added a device handler for shelly 2 as roller shutter using the REST API

http://shelly-api-docs.shelly.cloud/#shelly-family-overview


dgasparri:
Improved in a new commit

PR #3841, opened 2019-01-02
Device Handlers for the Shelly 1/2/4 as Switch and as Roller Shutter

dgasparri:
Device Handler for Shelly 2 as Roller Shutter
Device Handler for Shelly 1/2/4 as Switch

Using the REST API provided by the Shelly. Those devices do not currently support uPnP so a fixed IP address is required

http://shelly-api-docs.shelly.cloud/#common-http-api

They both work, just need minor fixes for the icons


akitamostodos:
Hello I use https://github.com/dgasparri/SmartThingsPublic/blob/master/devicetypes/shelly/shelly2.src/shelly2-as-roller-shutter.groovy

on 2.5 as roller and the preset button dosnt work for me, any idea? i have defined on level 50


lot of thanks

dgasparri:
Hi,

unfortunately there are serious limitations on the SmartThings hub. Samsung
has further restricted the development of interfaces only to "official"
partners, so to make the system even more close. I'm not endorsing it
anymore.

I hope that sooner or later, there will be a "Shelly official" Smartthing
App that will work.

Duccio

PR #3842, opened 2019-01-02
[ICP-7546] Fingerprint fix for Leedarson bulb

MMateusiakS:
Raw Description	01 0104 0101 01 09 0000 0003 0004 0005 0006 0008 0B05 1000 FEDC 02 000A 0019

@MAblewiczS, fingerprint fix for Leedarson bulb after software update


MAblewicz (approved):
Looks ok. Fingerprint fix only.

PR #3861, opened 2019-01-08
[WWST-2200] Modify fingerprint for Aurora Wireless Wall Remote

MMateusiakS:
raw desc:
01 0104 0830 01 03 0000 1000 0003 07 0003 0004 0005 0006 0008 1000 0019

@MAblewiczS fingerprint fix


MAblewicz (approved):
New fingerprint based on email from Thomas Lane (Aurora)

PR #3862, opened 2019-01-08
WWST-2091 Update to SmartWeather Station UI metadata

bflorian:
Change to UI metadata and OCF device type so that device displays current temperature properly on the One App dashboard and device list view.


bflorian:
No longer needed because problem is handed by generic metadata changes

PR #3894, opened 2019-01-20
CHAD-1670 CHAD-2223 Mark 2 Fibaro DTHs to run locally in 0.25

tpmanley:


cbaumler (dismissed):

varzac (dismissed):

varzac:
Is there a reason you left execute commands locally as false?  I thought in general we were marking everything as true unless there was an explicit reason not to?

tpmanley:
> Is there a reason you left execute commands locally as false? I thought in general we were marking everything as true unless there was an explicit reason not to?

Updated

tpmanley:
Please review again. The reviews were dismissed when I force pushed

varzac (dismissed):

cbaumler (dismissed):

tpmanley:
I removed the double switch since we were seeing some issues during test. @varzac please review again

tpmanley:
@varzac another review please

varzac (approved):

PR #3883, opened 2019-01-16
ICP-7871 Add Health Check to Nyce Motion Sensor

greens:

PR #3896, opened 2019-01-22
ICP-7871 Add Health Check to Nyce Motion Sensor (#3883)

greens:


tpmanley (approved):

tpmanley:
This change is already in staging so I think this PR can be closed

PR #3908, opened 2019-01-25
[WWST-1625] Add fingerprint for Sengled Element Classic.

MGoralczykS:
@MAblewiczS 
Could you review this? 
RAW: 01 0104 0101 00 08 0000 0003 0004 0005 0006 0008 0702 0B05 01 0019


MAblewicz (commented):
Looks Ok. Fingerprint change only.

MAblewicz (approved):
Looks Ok. Fingerprint change only.

PR #3910, opened 2019-01-26
WWST-2259 Add CT101 fingerprint with deviceJoinName

dkirker:


tpmanley (dismissed):

tpmanley:
I approved but put some questions on the ticket that I think we should resolve before merging

dkirker:
I'll check about the ticket this afternoon.

csuk07:
Let's change the join name to "Radio Thermostat CT101" and push to Prod. We can always change the name later. Addressed other issues in the JIRA ticket

dkirker:
Alright, changed to "Radio Thermostat"

tpmanley (approved):

PR #3915, opened 2019-01-28
ICP-7871 Add Health Check to Nyce motion sensor

greens:


devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
```diff
@@ -52,7 +52,8 @@ metadata {
 }
 
 def installed() {
-	sendEvent(name: "checkInterval", value: 60 * 12, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	// device report interval is 0x3600 seconds (230.4 minutes/3.84 hours) so checkinterval is ~that * 2 + 2 minutes
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60 * 4 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
tpmanley, line 56: The +2 minutes doesn't seem like enough to account for a little clock drift. I'd suggest something like 5% or 12 minutes

greens: the reporting interval is actually < 4 hours anyway, so clock drift should be covered.

tpmanley (approved):

PR #3916, opened 2019-01-28
ONEAPP-9503 Prevent automatic switch on/off events to keep states in …

greens:
…sync


tpmanley:
Could you add the hail support while you're in here?

greens:
@tpmanley right.

Fixed

tpmanley (approved):

PR #3920, opened 2019-01-29
[WWST-2275] Add FP for Sengled

DCzarneckaS:
@MAblewiczS

Raw Description | 01 0104 0101 00 08 0000 0003 0004 0005 0006 0008 0702 0B05 01 0019


MAblewicz (approved):
Fingeprint change only. OK

MAblewicz:
WWST-2275 Fingeprint fo Sengled Element Classic A19 100W (5000k) - E11-N14

PR #3924, opened 2019-01-30
[ICP-7995] Added batteryConfig() to configure() for keep device online when it'…

MGoralczykS:
…s powered.

@greens @tpmanley @MAblewiczS @PKacprowiczS @MMateusiakS 
It is fix for ICP-7995 when device goes disconnected after 1st ping. 
Could you review this?

### DIFF
--- a/devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
+++ b/devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
@@ -192,7 +192,7 @@ def configure() {
 		"raw 0x500 {01 23 00 00 00}", "delay 200",
 		"send 0x${device.deviceNetworkId} 1 1", "delay 1500",
 	]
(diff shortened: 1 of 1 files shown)


greens:
iirc from my discussion with Tom this device doesn't really listen to us very well. I have since put in a much longer checkInterval that should reflect how frequently the device checks in.

MGoralczykS:
I set checkInterval to 1h and there was strange behaviour. Device goes offline after 1st ping but on the 2nd ping device back online. This behaviour repeats every ping.

greens:
Is it the ping that's bringing it back online or just a periodic report?

MGoralczykS:
The second ping is bringing back device status

greens:
So I remembered that DH treats battery events differently. I don't think they can actually mark a device back online. Maybe change the ping to instead poll the IAS Zone?

MGoralczykS:
@greens 
Could you review this?

greens:
Does the device respond to this?

MGoralczykS:
Yes, it does.

greens (approved):

PR #3925, opened 2019-01-30
ICP-8046 Z-Wave Siren mistakenly always send 10 config commands

greens:
This rework attempts to move away from a counter to a flag-based system based on receiving a config report with a correct value in it.

### DIFF
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -123,7 +123,7 @@ def initialize() {
 
 	// if there's anything we need to send, send it now, and check again in 12s
 	if (cmds.size > 0) {
-		sendHubCommands(cmds)
+		sendHubCommand(cmds)
 		runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])
 	}
 }
@@ -317,6 +317,7 @@ def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport 
 	} else {
 		state.configured = true
 	}
+    return [:]
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd) {
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -317,7 +317,7 @@ def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport 
 	} else {
 		state.configured = true
 	}
-    return [:]
+	return [:]
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd) {
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -92,6 +92,7 @@ def installed() {
 
 def updated() {
 	log.debug "updated()"
+	state.configured = false
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 	runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])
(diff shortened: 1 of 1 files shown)


greens:
Still need to go grab a siren and test, but I'd like to have you guys check my math on this.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -300,6 +303,22 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
 	}
 }
 
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	def checkVal
+	// the last message sent by configure is a configuration get, so if we get a report, we succeeded in transmission
+	// and if the parameter 1 values match what we expect, then the configuration probably succeeded
+	if (isZipato()) {
+		checkVal = 3
```
kianooshkarami: 3?

kianooshkarami: I guess that is what we are setting this to in configuration.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -300,6 +303,22 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
...
+	} else if (isYale()) {
+		checkVal = state.alarmLength
+	}
+	if (checkVal) {
```
greens: this should probably be changed to an explicit null check in case `state.alarmLength` can be 0

kianooshkarami (dismissed):

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -300,6 +303,22 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
...
+		state.configured = (checkVal == cmd.scaledConfigurationValue)
+	} else {
+		state.configured = true
```
kianooshkarami, line 332: `state.configured` is still set to true if the checkVal is not set?

greens: yeah because if there's no checkVal there's nothing to configure.

greens:
Alright. I've tested this myself now and fixed a couple little bugs.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -300,6 +303,23 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
...
+		checkVal = 3
+	} else if (isYale()) {
+		checkVal = state.alarmLength
+	}
+	if (checkVal != null) {
+		state.configured = (checkVal == cmd.scaledConfigurationValue)
+	} else {
+		state.configured = true
+	}
+    return [:]
```
dkirker: I'm guessing the IDE gave you 4 spaces vs a tab? 😛

greens: fixed. ty

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -300,6 +303,23 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
...
 }
 
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	def checkVal
+	// the last message sent by configure is a configuration get, so if we get a report, we succeeded in transmission
+	// and if the parameter 1 values match what we expect, then the configuration probably succeeded
+	if (isZipato()) {
+		checkVal = 3
+	} else if (isYale()) {
+		checkVal = state.alarmLength
```
dkirker, line 327: I don't believe that settings are accessible through state like that. I think you can either use `alarmLength` or `settings.alarmLength`.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
dkirker: I feel like maybe we should define these somewhere so that they aren't magic numbers. I can see someone changing the value in one place but not elsewhere. Maybe as `def getZipatoVolume() { 3 }`, `def getZipatoDuration() { 2 }`, `def getZipatoSound() { 10 }`?

dkirker (changes requested):
Other than the two comments I just left, looks good.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
greens: it's set further up in `configure`

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
greens: magic numbers removed

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
dkirker: Ah! I should have known that.... Seeing as I have touched this code... 🤦‍♂️

dkirker (approved):
Looks good! Though it *might* make more sense to move the comments in the code block at line 153 into the getZipatoDefaults section. But otherwise, this should work well.

PR #3932, opened 2019-02-01
[WWST-2314] Uncomment fingerprint for Care Pendant and fix error while pairing.

MGoralczykS:
@greens @tpmanley @MAblewiczS 
Could you review this? 
There was an error while pairing -> 
"groovy.lang.GroovyRuntimeException: Ambiguous method overloading for method java.lang.Long#minus.
Cannot resolve which method to invoke for [null] due to overlapping prototypes between"

RAW: 01 0104 0402 00 05 0000 0001 0003 0020 0500 02 0003 0019

### DIFF
--- a/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
+++ b/devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
@@ -32,7 +32,7 @@ metadata {
 
         fingerprint inClusters: "0000, 0001, 0003, 0020, 0402, 0B05", outClusters: "0003, 0006, 0008, 0019", manufacturer: "OSRAM", model: "LIGHTIFY Dimming Switch", deviceJoinName: "OSRAM LIGHTIFY Dimming Switch"
         fingerprint inClusters: "0000, 0001, 0003, 0020, 0402, 0B05", outClusters: "0003, 0006, 0008, 0019", manufacturer: "CentraLite", model: "3130", deviceJoinName: "Centralite Zigbee Smart Switch"
-        //fingerprint inClusters: "0000, 0001, 0003, 0020, 0500", outClusters: "0003,0019", manufacturer: "CentraLite", model: "3455-L", deviceJoinName: "Iris Care Pendant"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -215,6 +215,7 @@ def configure() {
 private Map getButtonResult(buttonState, buttonNumber = 1) {
     if (buttonState == 'release') {
         log.debug "Button was value : $buttonState"
+        state.pressTime = now()
```
tpmanley: Making this change means the button will always be `pushed` and never `held`

tpmanley:
Make sure to verify press and hold are working with both cloud and local execution

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
greens: yeah I think you just want to add a null check for `state.pressTime`

MGoralczykS: @tpmanley button has 3 statuses. I made this change because when device is pairing there was an error about null value. 
should I create null check(in installed) or leave it in this method?

MGoralczykS:
> Make sure to verify press and hold are working with both cloud and local execution

This is working as a triger in local execution.

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
greens: Think about what the original code is intended to do. 

`state.pressTime` is set to `now()` when the button is pressed. When the button is released a separate event is sent. We subtract those two timestamps and then determine whether the button was pressed or held based on the delta. If you set `state.pressTime` to `now()` right before that subtraction every time, the delta will always be <1ms, so "held" will never be sent.

The error you're seeing is because a button release event was sent before `state.pressTime` was set. If `state.pressTime` is not set it makes the most sense to do nothing. So what you should be doing is:

* add a short circuit `[:]` return if `state.pressTime`is `null`
* verify that you are receiving the events you expect from the device.

MGoralczykS:
@greens @tpmanley I created new commit, could you review this?

greens (approved):
This looks good. And "pressed" and "held" are both sent as expected?

MGoralczykS:
Yes :)

PR #3934, opened 2019-02-01
ICP-5350 Rename child switch when parent is renamed

greens:


greens:
Ideally we should probably go back through all multi-endpoint devices and add similar logic, but that can wait for another day since at this point I think it's a not insignificant amount of changes.

tpmanley (approved):

dkirker (approved):

PR #3936, opened 2019-02-03
ICP-8138 Z-Wave Lock: Fix for pollCode stuck in infinite loop

rboy1:
CodeID is a String where as state.pollCode is an integer

@workingmonk @greens 

Fix for SmartLocks stuck in a loop reiterating the same user 1 updated


workingmonk:
cc @tpmanley

tpmanley (approved):

PR #3943, opened 2019-02-05
ICP-8141, ONEAPP-6820 Update vid for rest of IKEA bulbs

dkirker:


dkirker:
@tpmanley The vid was short enough that I was able to add the 2200K to it.

tpmanley (approved):

PR #3944, opened 2019-02-05
[WWST-2358] Add fingerprint to Aeotec NanoMote One EU/US ver

MGoralczykS:
@greens @tpmanley @MAblewiczS 
Could you review this? 

RAW EU: zw:Ss type:1801 mfr:0371 prod:0002 model:0004 ver:1.00 zwv:4.61 lib:03 cc:5E,55,98,9F sec:86,85,59,72,5A,73,80,84,5B,70,6C,7A secOut:26 role:04 ff:9600 ui:9600

RAW US: zw:Ss type:1801 mfr:0371 prod:0102 model:0004 ver:1.00 zwv:4.61 lib:03 cc:5E,55,98,9F sec:86,85,59,72,5A,73,80,84,5B,70,6C,7A secOut:26 role:04 ff:9600 ui:9600

### DIFF
--- a/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
+++ b/devicetypes/smartthings/zwave-button.src/zwave-button.groovy
@@ -15,7 +15,7 @@
  */
 
 metadata {
-	definition (name: "Z-Wave Button", namespace: "smartthings", author: "SmartThings") {
+	definition (name: "Z-Wave Button", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "x.com.st.d.remotecontroller" ,vid: "generic-button-4") {
 		capability "Button"
 		capability "Battery"
 		capability "Sensor"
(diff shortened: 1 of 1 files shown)


greens (commented):
This metadata won't expose the multiple press actions. Do these devices have "press 5x" actions or are they more simple pressed/held buttons?

@dkirker here's another device to add to the pile of "awaiting full translation of button capability"

MGoralczykS:
They have only pressed/held.

greens:
Then this probably fits better with the other Aeon remotes on `aeon-key-fob.groovy`.

MGoralczykS:
When I tried to put this on `aeon-key-fob.groovy` this device doesn't respond at any event. 
Also device paired with 4 child devices(it only have one button with press/held states). 
I made if condition to limit child devices to 1 but it still doesn't respond at any event.

greens:
It probably sends messages securely and they need to be unencapsulated. This DTH has a handler for secure messages, but the aeon-key-fob doesn't. It should be quick to add.

MGoralczykS:
This button communicates via Central State Operation - in aeon-key-fob I dont have this cc. 
I would be better if I leave this device in zwave-button DTH.  
Here -> https://products.z-wavealliance.org/products/2816 is a documentation. There is an information how this device communicates: 
```
"NanoMote One will send “Central Scene Notification” to the
associated nodes when the Button is short pressed, held and
released."
```

greens (approved):

PR #3945, opened 2019-02-05
[WWST-2327] Add FP for Zooz Power Switch

DCzarneckaS:
Just FP add.
Raw : zw:L type:1001 mfr:027A prod:0101 model:000D ver:1.02 zwv:4.38 lib:03 cc:5E,25,32,27,2C,2B,70,85,59,72,86,7A,73,5A role:05 ff:8700 ui:8700

@MAblewiczS  @MMateusiakS  @greens


MMateusiakS:
@greens  is capability Light still necessary? Maybe we should remove it from DTH?

greens:
I know for a fact this is a multi-endpoint device so this isn't going to be enough. 

You should either create a new DTH based on the old model in `zooz-power-strip.groovy` or you might be able to fit it in to `zwave-multi-metering-switch.groovy`. The second DTH is probably the best since I think the big distinction between the old Zooz power strip and this one is the addition of metering. See here if you need help with configuration: https://products.z-wavealliance.org/products/3158 (under "Configuration Parameters")

DCzarneckaS:
We dont talk about strip. This is PR for:
https://products.z-wavealliance.org/products/2517

greens:
Oh okay. My mistake.

@MMateusiakS "light" is still used for some automations and integrations so I don't feel comfortable removing it.

greens (approved):

MMateusiakS:
@greens this is what I need to know. Thanks!

PR #3955, opened 2019-02-07
[ICP-8156] Fix pairing without status.

MGoralczykS:
@greens @MAblewiczS @tpmanley @MMateusiakS @DCzarneckaS @PKacprowiczS 
Could you review this?

### DIFF
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -364,6 +364,7 @@ def ping() {
 
 def configure() {
 	log.debug "OnTime=${settings.OnTime} OnLevel=${settings.OnLevel} TempAdj=${settings.TempAdj}"
+	sendEvent(name: "motion", value: "inactive")
 	def cmds = commands([
 		zwave.configurationV1.configurationSet(parameterNumber: 1, size: 1, scaledConfigurationValue: validateSetting(settings.OnTime, 0, 127, 2)),
 		zwave.configurationV1.configurationSet(parameterNumber: 2, size: 1, scaledConfigurationValue: validateSetting(settings.OnLevel, -1, 99, -1)),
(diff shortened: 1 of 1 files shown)


devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -364,6 +364,7 @@ def ping() {
 
 def configure() {
 	log.debug "OnTime=${settings.OnTime} OnLevel=${settings.OnLevel} TempAdj=${settings.TempAdj}"
+	sendEvent(name: "motion", value: "inactive")
```
tpmanley: I recommend putting this in `installed` and setting `displayed: false` so it doesn't show up in the event feed

MGoralczykS: @tpmanley done :)

tpmanley (approved):

greens:
Why not just do a basic get and get the actual status of motion?

MGoralczykS:
@greens basicGet doesn't return actual status. 
There it is in refresh() but its only return color. Device only return move on Notification.

greens:
Can you do a `NotificationGet` for type 0x07? I'd assume that would return 0.

dkirker:
~Just a heads up on this device: For about 5 minutes after pairing it won't report motion. It is in some "test mode" so that you can see that it is working (it echos the motion using the LED, but does not send Z-Wave events). It *might* be best to just set motion to active.~

Actually, that wasn't it and that isn't 100% true. The device almost immediately reports motion when paired. I believe that it did not respond as expected to a NotificationGet? Trying in refresh():

`cmd << zwave.notificationV3.notificationGet(notificationType: notificationv3.NotificationReport.NOTIFICATION_TYPE_BURGLAR)`

returns

> [0C] NotificationReport(v1type:0, v1level:0, status:FF, type:7, event:254)

The event parameter is 254 regardless of motion or no motion.

greens:
254 just means unknown state. In that case I'm fine just synthesizing the event on add, but it's something I think we should avoid if at all possible.

dkirker:
@greens Agreed.

PR #3960, opened 2019-02-08
ICP-8122 Fibaro Single Switch 2 add initial poll

greens:


devicetypes/fibargroup/fibaro-single-switch-2-zw5.src/fibaro-single-switch-2-zw5.groovy
```diff
@@ -125,6 +126,7 @@ def ping() {
 def installed(){
     log.debug "installed()"
     sendEvent(name: "checkInterval", value: 1920, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+    response(refresh())
```
tpmanley, line 129: Seems odd to do this in installed. Shouldn't this be in configure?

greens: When I think initial state I think `installed`. Putting a refresh call in installed has been SOP for getting initial state for a while.

dkirker (approved):

PR #3961, opened 2019-02-09
WWST-1592 Remove old Aeon secure smart energy switch uk DTH

greens:
This has been migrated to the zwave-metered-switch DTH


greens:
Just cleaning up after migration. Don't know how sophisticated the process is or if we'll have to go through and delete this from superuser on every shard.

workingmonk:
Deleting this is step 1. I would target this against the production branch. After the merge, we will stop deploying again, but you will have to go in and manually delete it from superuser in all the prod and staging shards at least

greens:
@workingmonk my git foo isn't good enough to just get this single commit.

PR #3981, opened 2019-02-14
[WWST-2408] Add FP for Zooz Motion Sensor ZSE18

DCzarneckaS:
@MAblewiczS @greens @MMateusiakS 
Raw: 
zw:Ss type:0701 mfr:027A prod:0301 model:0012 ver:1.03 zwv:4.61 lib:03 cc:5E,55,98,9F sec:86,85,59,72,5A,73,80,71,30,84,70,6C,7A role:06 ff:8C07 ui:8C07


MAblewicz (approved):
Looks good. FP change only

PR #3987, opened 2019-02-18
[WWST-2310] New DTH Zigbee Multi Button.

PKacprowiczS:
@greens @MMateusiakS @KKlimczukS @MGoralczykS @DCzarneckaS

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -130,6 +130,10 @@ def refresh() {
 			zigbee.enrollResponse()
 }
 
+def ping() {
+	refresh()
+}
+
 def configure() {
 	def bindings = getModelBindings(device.getDataValue("model"))
 	return zigbee.onOffConfig() +
@@ -157,7 +161,7 @@ def updated() {
 def initialize() {
 	def numberOfButtons = modelNumberOfButtons[device.getDataValue("model")]
 	sendEvent(name: "numberOfButtons", value: numberOfButtons, displayed: false)
-	sendEvent(name: "DeviceWatch-Enroll", value: JsonOutput.toJson([protocol: "zigbee", scheme:"untracked"]), displayed: false)
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 	if(!childDevices) {
 		addChildButtons(numberOfButtons)
 	}
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -53,6 +53,9 @@ metadata {
 def parse(String description) {
 	def map = zigbee.getEvent(description)
 	def result = map ? map : parseAttrMessage(description)
+	if(result.name == "switch") {
+		result = createEvent(descriptionText: "Wake up event came in", isStateChange: true)
+	}
 	log.debug "Description ${description} parsed to ${result}"
 	return result
 }
@@ -128,6 +131,7 @@ def getBatteryPercentageResult(rawValue) {
 
 def refresh() {
 	return zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) +
+			zigbee.readAttribute(zigbee.ONOFF_CLUSTER, switchType)
 			zigbee.enrollResponse()
 }
 
@@ -144,19 +148,12 @@ def configure() {
 }
 
 def installed() {
-	initialize()
+	runIn(2, "initialize", [overwrite: true])
 	sendEvent(name: "button", value: "pushed", isStateChange: true)
-	if(childDevices) {
-		def event
-		for(def endpoint : 2..device.currentValue("numberOfButtons")) {
-			event = createEvent(name: "button", value: "pushed", isStateChange: true)
-			sendEventToChild(endpoint, event)
-		}
-	}
 }
 
 def updated() {
-	initialize()
+	runIn(2, "initialize", [overwrite: true])
 }
 
 def initialize() {
@@ -165,6 +162,13 @@ def initialize() {
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 	if(!childDevices) {
 		addChildButtons(numberOfButtons)
+	}
+	if(childDevices) {
+		def event
+		for(def endpoint : 2..device.currentValue("numberOfButtons")) {
+			event = createEvent(name: "button", value: "pushed", isStateChange: true)
+			sendEventToChild(endpoint, event)
+		}
 	}
 }
 
@@ -187,6 +191,7 @@ private addChildButtons(numberOfButtons) {
 }
 
 private getBatteryVoltage() { 0x0020 }
+private getSwitchType() { 0x0000 }
 private getHoldTime() { 1000 }
 private getButtonMap() {[
 		"3450-L" : [
@@ -208,19 +213,8 @@ private getModelNumberOfButtons() {[
 ]}
 private getModelBindings(model) {
 	def bindings = []
-	switch(model) {
-		case "3450-L":
-			for(def endpoint : 1..modelNumberOfButtons[model]) {
-				bindings += zigbee.addBinding(zigbee.ONOFF_CLUSTER, ["destEndpoint" : endpoint])
-			}
-			break
-		case "3450-L2":
-			for(def endpoint : 1..modelNumberOfButtons[model]) {
-				bindings += zigbee.addBinding(zigbee.ONOFF_CLUSTER, ["destEndpoint" : endpoint])
-			}
-			break
-		default:
-			break
+	for(def endpoint : 1..modelNumberOfButtons[model]) {
+		bindings += zigbee.addBinding(zigbee.ONOFF_CLUSTER, ["destEndpoint" : endpoint])
 	}
 	bindings
 }
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -132,7 +132,7 @@ def refresh() {
 
 def configure() {
 	def bindings = getModelBindings(device.getDataValue("model"))
-	return zigbee.onOffConfig() + zigbee.levelConfig() +
+	return zigbee.onOffConfig() +
 			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage, DataType.UINT8, 30, 21600, 0x01) +
 			zigbee.enrollResponse() +
 			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) + bindings
@@ -140,6 +140,7 @@ def configure() {
 
 def installed() {
 	initialize()
+	sendEvent(name: "button", value: "pushed", isStateChange: true)
 	if(childDevices) {
 		def event
 		for(def endpoint : 2..device.currentValue("numberOfButtons")) {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+	def map = zigbee.getEvent(description)
+	def result = map ? map : parseAttrMessage(description)
+	log.debug "Description ${description} parsed to ${result}"
+	return result
+}
+
+def parseAttrMessage(description) {
+	def descMap = zigbee.parseDescriptionAsMap(description)
+	def map = [:]
+	if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.commandInt != 0x07 && descMap?.value) {
```
greens: you don't seem to ever configure the reporting of the battery or to poll it

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+}
+
+def refresh() {
+	return zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) +
+			zigbee.enrollResponse()
+}
+
+def configure() {
+	def bindings = getModelBindings(device.getDataValue("model"))
+	return zigbee.onOffConfig() + zigbee.levelConfig() +
```
greens: `onOff` and `level` don't seem to be needed

greens: The `levelConfig` doesn't seem to be necessary.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+			zigbee.enrollResponse() +
+			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) + bindings
+}
+
+def installed() {
+	initialize()
+	if(childDevices) {
+		def event
+		for(def endpoint : 2..device.currentValue("numberOfButtons")) {
+			event = createEvent(name: "button", value: "pushed", isStateChange: true)
```
greens: missing a corresponding event for button 1, I think

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+
+private addChildButtons(numberOfButtons) {
+	for(def endpoint : 2..numberOfButtons) {
+		try {
+			String childDni = "${device.deviceNetworkId}:$endpoint"
+			def componentLabel = device.displayName[0..-2] + "${endpoint}"
+			addChildDevice("Child Button", childDni, device.getHub().getId(), [
+					completedSetup: true,
+					label         : componentLabel,
+					isComponent   : false,
```
greens: I think now we want `isComponent` to be true. @dkirker will have to help with the metadata when we get there

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+		def maxVolts = 3.0
+		def pct = (volts - minVolts) / (maxVolts - minVolts)
+		result.value = Math.min(100, (int)(pct * 100))
+		def linkText = getLinkText(device)
+		result.descriptionText = "${linkText} battery was ${result.value}%"
+		createEvent(result)
+	}
+}
+
+def refresh() {
```
greens, line 132: does this return anything? Does the device respond? If so it can have tracked device health and a ping method.

If not, we should probably remove the "Refresh" capability.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+
+def getButtonResult(buttonState, buttonNumber = 1) {
+	def event = [:]
+	if (buttonState == 'release') {
+		def timeDiff = now() - state.pressTime
+		if (timeDiff > 10000) {
+			return event
+		} else {
+			buttonState = timeDiff < holdTime ? "pushed" : "held"
+			def descriptionText = "${device.displayName[0..-2]} button $buttonNumber was $buttonState"
```
varzac: Is the display name not set by the user?  I don't think we should arbitrarily be stripping off characters.  I also think it is weird to model this as a "parent" device that is one of the 4 buttons and then 3 child devices that are the other 3.  I would either expect one parent with 4 children, or just one device that handles 4 different button presses.  But I have been a little bit removed from the discussion of how we plan to handle multi component devices, so I may just be mis-informed.

greens: This is the same pattern we used for aeon minimote. It saves us the headache of having an "empty" device in our apps.

Zach's right about the displayname, though. You can't guarantee it will always have a number at the end. This is just for a log message, though, so I don't think it's too dire.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
dkirker: Yes, isComponent would need to be true. Right now this will work in staging.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+		capability "Actuator"
+		capability "Battery"
+		capability "Button"
+		capability "Holdable Button"
+		capability "Configuration"
+		capability "Refresh"
+		capability "Sensor"
+		capability "Health Check"
+
+		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L", deviceJoinName: "Iris KeyFob 1"
```
dkirker, line 31: Is it intended that the deviceJoinName is "Iris KeyFob 1"?

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,208 @@
...
+	if(!childDevices) {
+		addChildButtons(numberOfButtons)
+	}
+}
+
+private addChildButtons(numberOfButtons) {
+	for(def endpoint : 2..numberOfButtons) {
+		try {
+			String childDni = "${device.deviceNetworkId}:$endpoint"
+			def componentLabel = device.displayName[0..-2] + "${endpoint}"
```
dkirker: Same note as above for the displayName.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
PKacprowiczS: Yes, device responds to that.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
PKacprowiczS: This is similar pattern I've seen and used myself in other multi-component devices, eg. Z-Wave Multi Metering Switch or Z-Wave Dual Switch.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
PKacprowiczS: But, is tracking device health a good idea for a device which is expected to be taken away from hub?

greens: Tracking device health if possible is always a good idea. If the user takes it out of range it would be good to reflect that as quickly as possible (within battery constraint reason). For example if you lose the device inside your home vs. away from home.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,226 @@
...
+				"04" : 2
+		]
+]}
+private getModelNumberOfButtons() {[
+		"3450-L" : 4,
+		"3450-L2" : 4
+]}
+private getModelBindings(model) {
+	def bindings = []
+	switch(model) {
```
greens: I would switch this instead on `modelNumberOfButtons()`.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,226 @@
...
+	return zigbee.onOffConfig() +
+			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage, DataType.UINT8, 30, 21600, 0x01) +
+			zigbee.enrollResponse() +
+			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) + bindings
+}
+
+def installed() {
+	initialize()
+	sendEvent(name: "button", value: "pushed", isStateChange: true)
+	if(childDevices) {
```
greens: this should probably be moved to initialize, and you should probably do a `runIn(2, "initialize", [overwrite: true])` here and in updated so you don't call it twice

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,226 @@
...
+		def pct = (volts - minVolts) / (maxVolts - minVolts)
+		result.value = Math.min(100, (int)(pct * 100))
+		def linkText = getLinkText(device)
+		result.descriptionText = "${linkText} battery was ${result.value}%"
+		createEvent(result)
+	}
+}
+
+def refresh() {
+	return zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, batteryVoltage) +
```
greens, line 133: I remembered that battery events don't keep devices online in device health. Is there another attribute you could read?

tpmanley: ^ this is currently true but it's supposedly changing very soon FYI

greens: "soon" can mean a lot of things.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
PKacprowiczS: So, instead of switching on `modelNumberOfButtons()` I've decided to just iterate with number of buttons. If any future device would work in different way, we can adjust it in no time.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
dkirker: Like with the other device handlers (or at least Z-Wave Dual Switch which I just checked), I would add a check for `device.displayName.endsWith('1')`.

I'd use:

`def componentLabel = (device.displayName.endsWith('1') ? device.displayName[0..-2] : device.displayName) + "${endpoint}"`. It might be better to do `.endsWith(' 1')` with the space since you are trimming two characters.

I'd probably do this above in the log message, as well, just to stay consistent.

PKacprowiczS: Done.

greens:
I just finished writing new metadata that uses our new multi-component stuff.

Can you add fingerprint-specific metadata for the two models that points to `generic-4-button`?

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -0,0 +1,220 @@
...
+		capability "Actuator"
+		capability "Battery"
+		capability "Button"
+		capability "Holdable Button"
+		capability "Configuration"
+		capability "Refresh"
+		capability "Sensor"
+		capability "Health Check"
+
+		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L", deviceJoinName: "Iris KeyFob 1"
```
greens: I just uploaded a new generic metadata using the new multi-component work to staging that should work for this device if you are on a staging environment as well.

Try adding a per-fingerprint vid pointing to `"generic-4-button"` and see if you can try it out in OneApp on a staging env.

greens (approved):

PR #3989, opened 2019-02-19
[WWST-2386] Add fingerprint to Zooz ZEN26 Switch

MGoralczykS:
@greens @MAblewiczS @MMateusiakS 
Could you review this? 

RAW: zw:L type:1001 mfr:027A prod:A000 model:A001 ver:1.00 zwv:5.03 lib:03 cc:5E,25,85,8E,59,55,86,72,5A,73,70,6C,9F,7A role:05 ff:8700 ui:8704


greens:
`zwave-switch` is mostly for GE devices. This should go in `zwave-switch-generic`. Also if this is a secure device it should probably go to `zwave-switch-secure` since those other two DTHs don't have secure-handling code in them. They only work because they're local.

MGoralczykS:
I added this to `zwave-switch-generic` because in RAW I didn't see secure CC. 
Could you review this?

greens:
Yeah it uses S2, which is CC 0x9F instead of S0 which is 0x98.

https://products.z-wavealliance.org/products/3147

MGoralczykS:
@greens 
It's working with this DTH :)

greens (approved):

PR #3991, opened 2019-02-19
DVCSMP-4643 RGBW Bulbs adding as Color Temperature Bulbs

greens:
Adds the "color temperature light" deviceId to the color temp bulb generic fingerprints to prevent mistaken fingerprint matches.

### DIFF
--- a/devicetypes/smartthings/zigbee-white-color-temperature-bulb.src/zigbee-white-color-temperature-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-white-color-temperature-bulb.src/zigbee-white-color-temperature-bulb.groovy
@@ -32,8 +32,8 @@ metadata {
 
 
 		// Generic
-		fingerprint profileId: "0104", deviceId: "010C", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300", outClusters: "0019"
-		fingerprint profileId: "0104", deviceId: "010C", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300, 0B04", outClusters: "0019"
+		fingerprint profileId: "0104", deviceId: "010C", inClusters: "0006, 0008, 0300", outClusters: "0019" // deviceId 010C = "White color temperature light"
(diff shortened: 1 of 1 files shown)


greens:
zll-rgbw-bulb, but there are a couple of device IDs for those devices. Should I also do the permutations of the generic fingerprint for all of those?

Was it also your intention to add generic fingerprints to zigbee-rgbw-bulb?

devicetypes/smartthings/zigbee-white-color-temperature-bulb.src/zigbee-white-color-temperature-bulb.groovy
```diff
@@ -32,8 +32,8 @@ metadata {
 
 
 		// Generic
-		fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300", outClusters: "0019"
-		fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300, 0B04", outClusters: "0019"
+		fingerprint profileId: "0104", deviceId: "010C", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300", outClusters: "0019"
```
tpmanley: I was thinking that for these generic deviceId specific fingerprints we should add a comment at the end of the fingerprint to identify the deviceId since most people won't know what that means. In this case it would be `Color temperature light`. Do you guys like that? An alternative is no comment or a new join name that says `Generic Color Temperature Light` or something like that.

tpmanley: I also think we should remove any clusters that are not used by the DTH. So in this case we only need 0006, 0008 and 0300.

varzac: I think a comment would be nice, but for the join name I think defaulting to the DTH name is fine since it's basically a generic name.

greens:
Updated.

tpmanley (approved):

varzac (approved):

PR #3992, opened 2019-02-19
CHAD-2936 Add local support for fibaro double switch 2

greens:


greens:
I'm very skeptical about this working for a multi-endpoint device. @kianooshST have you tried this out?

kianooshkarami:
> I'm very skeptical about this working for a multi-endpoint device. @kianooshST have you tried this out?

I think you may have answered your own question, but yeah we currently do not support local execution for multi-endpoint device

PR #3994, opened 2019-02-20
[WWST-2356] Modify DTH for Aeotec Panic Button DSA38EU

MMateusiakS:
@greens here is little update of DTH for Panic Button


greens:
Does the device not send the wakeup notification on add that triggers a battery get?

greens (approved):
It looks good. Have you actually confirmed that the device sends the `WakeUp` and that this will keep the device online using this DTH?

MMateusiakS:
I can confirm. Device sends WakeUpNotification.

PR #3998, opened 2019-02-21
WWST-2228: Add DTH for ezex smart sub meter(korea market)

FenMei007S:
@greens  @dkirker  @varzac  , pls review again, thank you : )


varzac (dismissed):

greens (dismissed):

greens:
@FenMei007S it looks like you have merge conflicts somehow

dkirker:
I'm guessing that it is because this is a continuation of the same branch from the previous PR, which is now in master, and that commit into master is different than its twin in the WWST-2228 branch.

dkirker:
For the sake of time, I resolved the merge commit, and also added a commit that updates the spacing to tabs.

dkirker:
@FenMei007S  Please be sure that the final device handler is tested

FenMei007S:
@greens @dkirker @varzac , I have tested this DTH,found that the default event value is not right.
`	def event = zigbee.getEvent(description)
	if (event) {
		log.info "event enter:$event"
        if(event.name=="power"){ 
            event.value =event.value/1000
        }else if(event.name=="energy"){
            event.value =event.value/1000000
        }
        log.info "event outer:$event"
		sendEvent(event)
	}
`
so,I changed the event value.Please have a review.

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -12,98 +12,107 @@
...
-        log.debug "Desc Map: $descMap"
-                
-        List attrData = [[clusterInt: descMap.clusterInt ,attrInt: descMap.attrInt, value: descMap.value]]
-        descMap.additionalAttrs.each {
-            attrData << [clusterInt: descMap.clusterInt, attrInt: it.attrInt, value: it.value]
+	log.debug "description is $description"
+	def event = zigbee.getEvent(description)
+	if (event) {
+		log.info "event enter:$event"
+        if(event.name=="power"){
```
dkirker: Please use tabs instead of spaces and add spaces in between the different tokens. For example:

```
		log.debug "event enter:$event"
		if (event.name == "power") {
			event.value = event.value/1000
			event.unit = "W"
		} else if (event.name == "energy") {
			event.value = event.value/1000000
			event.unit = "kWh"
		}
		log.debug "event outer:$event"
```
That way we can maintain consistency.

Also, please use log.debug instead of log.info, as log.info (especially for frequent events) generates a lot of extra data.

dkirker (changes requested):

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -12,98 +12,107 @@
...
+		descMap.additionalAttrs.each {
+			attrData << [clusterInt: descMap.clusterInt, attrInt: it.attrInt, value: it.value]
+		}
+		attrData.each {
+				def map = [:]
+
+				if (it.clusterInt == zigbee.SIMPLE_METERING_CLUSTER && it.attrInt == ATTRIBUTE_HISTORICAL_CONSUMPTION) {
+					log.debug "meter"
+					map.name = "power"
+					map.value = zigbee.convertHexToInt(it.value)/1000
```
dkirker: It just occurred to me, but it would be good to include the units, as well.
map.unit = "W"

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -12,98 +12,107 @@
...
+				} else if (it.clusterInt == zigbee.SIMPLE_METERING_CLUSTER && it.attrInt == ATTRIBUTE_READING_INFO_SET) {
+					log.debug "energy"
+					map.name = "energy"
+					map.value = zigbee.convertHexToInt(it.value)/1000000
```
dkirker: and
map.unit = "kWh"

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -12,98 +12,107 @@
...
-                
-        List attrData = [[clusterInt: descMap.clusterInt ,attrInt: descMap.attrInt, value: descMap.value]]
-        descMap.additionalAttrs.each {
-            attrData << [clusterInt: descMap.clusterInt, attrInt: it.attrInt, value: it.value]
+	log.debug "description is $description"
+	def event = zigbee.getEvent(description)
+	if (event) {
+		log.info "event enter:$event"
+        if(event.name=="power"){
+            event.value =event.value/1000
```
dkirker: I guess the Zigbee library is not adding the units, so it would probably be good to have them here, as well. (We'll need to fix that.)

For power:
event.unit = "W"

For energy:
event.unit = "kWh"

dkirker:
I occurred to me that there were no units. Even though there are default units (W for power and kWh for energy) defined in the capability, it is probably good practice to include them.  I added some comments in-line for that.

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -45,40 +45,53 @@ metadata {
...
+def getATTRIBUTE_READING_INFO_SET() { 0x0000 }
+def getATTRIBUTE_HISTORICAL_CONSUMPTION() { 0x0400 }
+
 def parse(String description) {
     log.debug "description is $description"
     def event = zigbee.getEvent(description)
     if (event) {
-        log.info event
+        log.info "event enter:$event"
+        if (event.name== "power") {
```
tpmanley: nit: space before `==` here and in the `else if` below

tpmanley (dismissed):
Just some minor comments. If you're going to want this deployed to STG right away please point the PR at the `staging` branch instead of the `master` branch.

dkirker:
It looks like the branch wasn't properly rebased to staging, so it has picked up a couple of extra commits. Also, please don't forget to use tabs for indenting.

greens (approved):

FenMei007S:
@dkirker Does the PR still have conflict when merage to master?

Issue #3986, opened 2019-02-17
Z-Wave Metering Dimmer does not properly implement Switch Level


andrewsayre:
The [Z-Wave Metering Dimmer](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy), and possibly others, do not implement `setLevel` as defined in the [SmartThings Capability Reference](https://smartthings.developer.samsung.com/docs/api-ref/capabilities.html#Switch-Level).  Specifically, they do not have an overload that accepts two numbers: `setLevel(level, rate)`.  While `rate` is not required, an overload should be provided that accepts the value and ignores it if not supported by the  device.

Using the new cloud API, there is no way of knowing whether the device supports `setLevel(level)` or `setLevel(level, rate)` and calling the latter results in an error because groovy can't find an overload to accept the parameters.

The fix is a simple adjustment from `setLevel(level)` to `setLevel(level, rate=0)` so Groovy generates the proper overloads.

Full list of DTHs that don't handle `setLevel` with two parameters (that I've found so far):
- GE Link Bulb
- Sylvania Ultra iQ
- ZigBee Accessory Dimmer
- ZigBee Dimmer Power
- ZigBee Dimmer with Motion Sensor
- ZigBee Dimmer
- ZLL Dimmer Bulb
- Z-Wave Metering Dimmer

tpmanley:
Thanks for the bug report. We'll get this fixed.

andrewsayre:
Thanks!  What kind of timeline are we looking at before it's updated and live?

tpmanley:
I'm hoping to see it go out early next week. One thing to note is many of the DTHs with the issue have `executeCommandsLocally: true` which means that if the device is running locally on the hub the commands like `setLevel` are sent directly to the hub instead of the DTH. The hub does support the optional second parameter so the commands execute correctly in that case. Most of the errors I'm seeing are with DTHs that don't have `executeCommandsLocally: true`, namely `GE Link Bulb` and `LIFX Color Bulb`. We still plan to fix all of them but it does mean the impact is a little less than how it originally appears.

andrewsayre:
I am finding users that have the Z-Wave Metering Dimmer, which is set to run locally, are affected and getting the error logged in the IDE about a missing method.

andrewsayre:
[...]

PR #3999, opened 2019-02-21
ICP-8341 Update setLevel to follow capability spec

dkirker:
Right now, I have all of the matches included. We can more than likely remove a few, but it is probably safe to just include them all.

### DIFF
--- a/devicetypes/capabilities/switch-level-capability.src/switch-level-capability.groovy
+++ b/devicetypes/capabilities/switch-level-capability.src/switch-level-capability.groovy
@@ -71,7 +71,7 @@ def off() {
 	'off'
 }
 
-def setLevel(value) {
+def setLevel(value, rate = null) {
 	"setLevel: $value"
 }
 
--- a/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
+++ b/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
@@ -370,7 +370,7 @@ def clearObstruction() {
     ] + configure()
 }
 
-def setLevel(value) {
+def setLevel(value, rate = null) {
     log.debug "setting level: ${value}"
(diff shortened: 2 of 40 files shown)


tpmanley:
I believe Keen needs to be updated too

devicetypes/capabilities/switch-level-capability.src/switch-level-capability.groovy
```diff
@@ -71,7 +71,7 @@ def off() {
 	'off'
 }
 
-def setLevel(value) {
+def setLevel(value, rate = null) {
```
tpmanley, line 74: I'm not familiar with how these files are used so I'm not sure if this is OK or not

tpmanley (approved):

devicetypes/capabilities/switch-level-capability.src/switch-level-capability.groovy
dkirker: I *think* they are capability examples. @bflorian could possibly confirm?

andrewsayre:
This fixes #3986

devicetypes/capabilities/switch-level-capability.src/switch-level-capability.groovy
andrewsayre: I understand them to be examples as well.

andrewsayre:
Any update on when this will go in?

dkirker:
@tpmanley Let's go over a QA plan tomorrow? I am sure this is a safe merge, but I want to make sure that it does get some level of verification.

rappleg (approved):

dkirker:
Once @workingmonk does the deploy tomorrow, we can go ahead and merge this. (Rebasing to master and merging now would be ugly.)

workingmonk:
@dkirker let me know after the deploy tomm

PR #4001, opened 2019-02-21
ICP-8390 Update Sylvania join name to new branding

greens:


greens:
I've been told it should only be for this device, not all the old lights we have.

tpmanley (approved):

PR #4002, opened 2019-02-22
[CHAD-1813] Arrival Sensor HA: Make the DTH run locally

tpmanley:
The DTH is now marked to run locally in 0.25. When it moves from cloud execution to local execution `updated` will be called which will stop the scheduled presence check running if it is currently running.

Also fixed an issue where `state.lastCheckin` is not set (presumably because the device never sent us a message). Currently it raises an exception but with the change it should just create a 'not present' event and then stop the scheduled check.

https://smartthings.atlassian.net/browse/CHAD-1813

### DIFF
--- a/devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
+++ b/devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
@@ -14,7 +14,8 @@ import groovy.json.JsonOutput
  *
  */
 metadata {
-    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings") {
+    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings",
+            runLocally: true, minHubCoreVersion: '000.025.00025', executeCommandsLocally: true) {
         capability "Tone"
         capability "Actuator"
         capability "Presence Sensor"
@@ -58,6 +59,7 @@ metadata {
 }
 
 def updated() {
+    stopTimer()
     startTimer()
 }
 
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
```diff
@@ -14,7 +14,8 @@ import groovy.json.JsonOutput
  *
  */
 metadata {
-    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings") {
+    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings",
+            runLocally: true, minHubCoreVersion: '000.025.0000', executeCommandsLocally: true) {
```
tpmanley: @varzac , what version first supported this device locally?

tpmanley:
Applied the `DO_NOT_MERGE` label until I finish testing and we figure out how we're going to roll this out since it's a special case.

devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
varzac: https://smartthings.atlassian.net/browse/CHAD-882 that and the changelog say 18.XX

greens (dismissed):

dkirker (dismissed):

tpmanley:
This has been pretty thoroughly tested and needs final review/approval

vladbel (dismissed):
This was tested and no issues discovered.
See https://smartthings.atlassian.net/browse/HCS-198 for more details

devicetypes/smartthings/arrival-sensor-ha.src/arrival-sensor-ha.groovy
```diff
@@ -14,7 +14,8 @@ import groovy.json.JsonOutput
  *
  */
 metadata {
-    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings") {
+    definition (name: "Arrival Sensor HA", namespace: "smartthings", author: "SmartThings",
+            runLocally: true, minHubCoreVersion: '000.018.00200', executeCommandsLocally: true) {
```
tpmanley: Since the latest round of testing was all done with 0.25.x I think we should update `minHubCoreVersion` to be on the safe side

greens: seems reasonable

tpmanley: updated

varzac (approved):

greens (approved):

PR #4003, opened 2019-02-22
WWST-2371 Add genericHandler designations

greens:


greens:
There's a lot of these and we can always add more.

tpmanley (approved):

bflorian (approved):

PR #4006, opened 2019-02-25
[WWST-1994] New DTH Z-Wave Basic Window Shade (Aeotec Nano Shutter)

PKacprowiczS:
@greens @dkirker @MMateusiakS @KKlimczukS @DCzarneckaS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
+++ b/devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
@@ -12,6 +12,8 @@
  *
  */
 
+import groovy.json.JsonOutput
+
 metadata {
 	definition (name: "Z-Wave Basic Window Shade", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.blind", mnmn: "SmartThings", vid: "generic-stateless-curtain") {
 		capability "Stateless Curtain Power Button"
@@ -42,14 +44,19 @@ metadata {
 	}
 
 	preferences {
-		input (
-				title: "Working direction",
+		section {
+			input(title: "Aeotec Nano Shutter settings",
 				description: "In case wiring is wrong, this setting can be changed to fix setup without any manual maintenance.",
-				name: "direction",
-				type: "enum",
-				defaultValue: "Normal",
-				options: ["Normal", "Reversed"]
-		)
+				displayDuringSetup: false,
+				type: "paragraph",
+				element: "paragraph")
+
+			input("reverseDirection", "bool",
+				title: "Reverse working direction",
+				defaultValue: false,
+				displayDuringSetup: false
+			)
+		}
 	}
 }
 
@@ -130,14 +137,14 @@ def ping() {
 def installed() {
 	log.debug "Installed ${device.displayName}"
 	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
-	sendEvent(name: "availableCurtainPowerButtons", value: ["open", "close", "pause"])
+	sendEvent(name: "availableCurtainPowerButtons", value: JsonOutput.toJson(["open", "close", "pause"]))
 	state.shadeState = "paused"
-	state.direction = direction ? direction : "Normal"
+	state.reverseDirection = reverseDirection ? reverseDirection : false
 }
 
 def updated() {
 	sendHubCommand(pause())
-	state.direction = direction ? direction : "Normal"
+	state.reverseDirection = reverseDirection ? reverseDirection : false
 }
 
 def configure() {
@@ -146,7 +153,7 @@ def configure() {
 }
 
 private secure(cmd) {
-	if(zwaveInfo.zw.endsWith("s")) {
+	if(zwaveInfo.zw.contains("s")) {
 		zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
 	} else {
 		cmd.format()
@@ -154,9 +161,9 @@ private secure(cmd) {
 }
 
 private getOpenValue() {
-	state.direction == "Normal" ? 0x00 : 0xFF
+	!state.reverseDirection ? 0x00 : 0xFF
 }
 
 private getCloseValue() {
-	state.direction == "Normal" ? 0xFF : 0x00
+	!state.reverseDirection ? 0xFF : 0x00
 }
--- a/devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
+++ b/devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
@@ -13,7 +13,7 @@
  */
 
 metadata {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+metadata {
+	definition (name: "Z-Wave Basic Window Shade", namespace: "smartthings", author: "SmartThings") {
+		capability "Stateless Curtain Power Button"
+		capability "Configuration"
+		capability "Actuator"
+		capability "Health Check"
+
+		command "open"
```
greens, line 24: @dkirker The `Stateless Curtain Power Button` capability is only proposed, but these commands are part of it, so they shouldn't need to be defined separately, right?

greens: Based on the commented-out code below @PKacprowiczS I'm guessing it didn't work without them, though?

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
```diff
@@ -0,0 +1,126 @@
...
+	result
+}
+
+def ping() {
+	secure(zwave.switchMultilevelV3.switchMultilevelGet())
+}
+
+def installed() {
+	log.debug "Installed ${device.displayName}"
+	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
greens: I'd remove `offlinePingable` as this isn't really a crucial device.

greens (commented):
I almost wonder if it's not just better to use the general WindowShade capability and just never send the "open" or "closed" states? I know it's at least mapped better in ocf.

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
dkirker: So, it appears that `Stateless Curtain Power Button` has the one command `setButton(button)` which takes one of `["open","close","pause"]`.

dkirker:
I believe part of the reason for using the current capability was for the "pause" command. This should work if the `setButton(button)` command is implemented that accepts an enum of `["open","close","pause"]`. I'd leave the `open()`, `close()`, and `pause()` commands in too for Classic and just have `setButton(button)` call them as necessary.

Also, you will want to do a `sendEvent(name: "availableCurtainPowerButtons", value: ["open", "close", "pause"])`

dkirker (changes requested):
Once my comments above have been added, the device should work from OneApp. However, it seems that Stateless Curtain Power Button has not been mapped yet.

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
greens: ah, makes sense

greens (dismissed):

dkirker (dismissed):
Looks good. I pinged the proper team on the ticket for the last step for the capability integration.

greens:
Do we know if stateless curtain power button works yet?

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
```diff
@@ -0,0 +1,132 @@
...
+}
+
+def ping() {
+	secure(zwave.switchMultilevelV3.switchMultilevelGet())
+}
+
+def installed() {
+	log.debug "Installed ${device.displayName}"
+	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	sendEvent(name: "availableCurtainPowerButtons", value: ["open", "close", "pause"])
```
greens: I _think_ you just need to make this JSON

PKacprowiczS:
@greens After every _pause()_ call device is sending by itself not-so-accurate report about its current state (e.g. when transiting from close -> pause, I've received _closing_ event) so I had to create this flag to ignore next response after calling _pause()_

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
greens: 👀 @PKacprowiczS

greens: so based on your comment on the ticket, this _doesn't_ need to be JSON?

PKacprowiczS: No, it doesn't.

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
```diff
@@ -0,0 +1,162 @@
...
+	state.direction = direction ? direction : "Normal"
+}
+
+def configure() {
+	log.debug "Configure..."
+	response(secure(zwave.configurationV1.configurationSet(parameterNumber: 80, size: 1, scaledConfigurationValue: 1)))
+}
+
+private secure(cmd) {
+	if(zwaveInfo.zw.endsWith("s")) {
```
greens: change this to `contains`

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
```diff
@@ -0,0 +1,162 @@
...
+			state "default", label:'Pause', action:"pause"
+		}
+
+		details(["open","close","pause"])
+	}
+
+	preferences {
+		input (
+				title: "Working direction",
+				description: "In case wiring is wrong, this setting can be changed to fix setup without any manual maintenance.",
```
greens, line 49: Would it make our lives easier if this was a boolean that asked whether or not you want to reverse working direction?

devicetypes/smartthings/zwave-basic-window-shade.src/zwave-basic-window-shade.groovy
dkirker: The capability is defined as JSON:

```
  availableCurtainPowerButtons:
    schema:
      type: object
      additionalProperties: false
      properties:
        value:
          type: array
          items:
            type: string
            enum:
              - open
              - close
              - pause
    type: JSON_OBJECT
```
Please try it as JSON and assuming that works then change it to that. If it doesn't,  there is a bug somewhere. But it should be a JSON array containing items of type string.

PKacprowiczS: It works for both JSON and array.

greens (approved):

PR #3717, opened 2018-11-28
WWST-1615 Adding fingerprint for Ozom Smart Carbon Monoxide Sensor

fyamanaka:
Dears,
@mckeed @tpmanley @MAblewiczS @greens 

Could you please approve this change ?
Note: This device does not has battery status monitoring once cluster 0x0001 is not present in device spec.

Best regards,

PR #4009, opened 2019-02-25
WWST-1615 Add fake battery event and different metadata for Ozom CO

dkirker:


dkirker:
So, I had thought the Ozom fingerprint was already added. Apparently, that PR was closed. Is there a reason the fingerprint in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3717 wasn't just put in a DTH for CO? I am considering moving it to something like that.

tpmanley (dismissed):

greens:
Did they end up in the Orvibo Gas Detector DTH which is the same as this one sans battery?

dkirker:
The last comment I see on #3717 is that #3717 was closed waiting for HCS-299. I'll go check. I commented on the ticket, as well, so if there are no objections I am going to move this to a proper Zigbee CO DTH.

dkirker:
@greens No. As far as I can tell the fingerprint is not in a merged PR. I just created a new DTH so that it uses the correct capabilities.

devicetypes/smartthings/zigbee-co-sensor.src/zigbee-co-sensor.groovy
```diff
@@ -0,0 +1,157 @@
...
+}
+
+def installed(){
+	log.debug "installed"
+
+	if (isOzomCO()) {
+		sendEvent(name: "battery", value: 100, unit: "%", displayed: false)
+	}
+
+	refresh()
```
greens: needs to be `response(refresh())` I think?

dkirker: Yeah, I think installed, updated, and uninstalled need it... This was copied from the Zigbee Smoke, which leads me to believe that it was never tested...

devicetypes/smartthings/zigbee-co-sensor.src/zigbee-co-sensor.groovy
```diff
@@ -0,0 +1,157 @@
+ /*
+  *  Copyright 2018 SmartThings
```
greens: you can change this too if you want

greens (dismissed):
Copyright date is off

devicetypes/smartthings/zigbee-co-sensor.src/zigbee-co-sensor.groovy
dkirker: Eh. The file I copied 98% of this from is 2018. 🤷‍♂️

dkirker:
Bumped the copyright date. @greens since this is pointed to staging we should probably wait until tomorrow to merge. I could try to rebase, but I am sure it would just get ugly.

greens (approved):
Merge whenever.

workingmonk:
bumping up copyright is actually a bad practice. older copyright shows that we created the file in 2018 and have the copyright since then. we could leave it as is or do `2016-2019`

greens:
@workingmonk It's a new DTH.

PR #4013, opened 2019-02-26
[ICP-8158] Set the highest temperature value.

MGoralczykS:
@greens @tpmanley @dkirker 
Could you review this?


dkirker (changes requested):
I left a comment on the ticket on why I think this is a bad idea.

PR #4014, opened 2019-02-26
[WWST-2406] Updated Aeon Multisensor DTH to support Zooz 4-in-1 sensor

PKacprowiczS:
@greens @MMateusiakS @KKlimczukS @DCzarneckaS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
@@ -21,9 +21,11 @@ metadata {
 		capability "Sensor"
 		capability "Battery"
 		capability "Health Check"
+		//capability "Tamper Alert" waiting for LE fix for Zooz 4-in-1
 
 		fingerprint deviceId: "0x0701", inClusters: "0x5E,0x86,0x72,0x59,0x85,0x73,0x71,0x84,0x80,0x30,0x31,0x70,0x98,0x7A", outClusters:"0x5A"
 		fingerprint mfr:"0086", prod:"0102", model:"004A", deviceJoinName: "Aeotec MultiSensor (Gen 5)"
+		fingerprint mfr:"027A", prod:"2021", model:"2101", deviceJoinName: "Zooz 4-in-1 sensor"
 	}
 
 	simulator {
@@ -92,9 +94,15 @@ metadata {
 		standardTile("configure", "device.configure", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
 			state "configure", label:'', action:"configure", icon:"st.secondary.configure"
 		}
+		/* waiting for LE fix for Zooz 4-in-1
+		valueTile("tamper", "device.tamper", height: 2, width: 2, decoration: "flat") {
+			state "clear", label: 'tamper clear', backgroundColor: "#ffffff"
+			state "detected", label: 'tampered', backgroundColor: "#ff0000"
+		}
+		*/
 
 		main(["motion", "temperature", "humidity", "illuminance"])
-		details(["motion", "temperature", "humidity", "illuminance", "battery", "configure"])
+		details(["motion", "temperature", "humidity", "illuminance", "battery", "configure"/*, "tamper"*/])
 	}
 }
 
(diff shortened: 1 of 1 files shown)


greens:
Looks like this device supports tamper as well, so multisensor 6 is probably a better fit.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -236,4 +263,20 @@ private setConfigured(configure) {
 
 private isConfigured() {
 	getDataValue("configured") == "true"
+}
+
+private sendCheckInterval() {
+	def checkInterval
+	if(zwaveInfo.mfr == "027A") checkInterval = 8 * 60 * 60 + 2 * 60
```
kianooshkarami: The formula going forward for Z-Wave sleepy devices should be `(2 * Wake up Interval ) + ( 10 minutes) `

kianooshkarami: I would suggest to set the `def wakeupInterval` and then before sending the check interval do ` checkInterval = 2 * checkInterval + (10 * 60)`

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
```diff
@@ -110,17 +122,32 @@ def parse(String description)
...
+		createEvent(descriptionText: cmd.toString())
+	}
+}
+
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv1.WakeUpNotification cmd)
 {
-	[descriptionText: "${device.displayName} woke up", isStateChange: false]
+	def cmds = []
+	def result = createEvent(descriptionText: "$device.displayName woke up", isStateChange: false)
+	if (!state.lastbatt || (now() - state.lastbatt) >= 10 * 60 * 60 * 1000) {
```
kianooshkarami: Going forward and Per Z-Wave Plus V2 rules, we need to ask all the sleepy nodes for their battery level when they wake up. So I would say, let's just keep it simple, and just ask for Battery level every time the device wakes up

PKacprowiczS:
> Looks like this device supports tamper as well, so multisensor 6 is probably a better fit.

@greens but multisensor 6 has also UV index and power source, which this device lacks.

devicetypes/smartthings/aeon-multisensor.src/aeon-multisensor.groovy
PKacprowiczS: Wasn't 4h a default wakeUpInterval time value for sleepy devices, or I have missed something in the meantime?

kianooshkarami: It does not appear that we are configuring the Wake up interval to be something other than 4 hours, I thought we were, sorry about the confusion. Having said that, we should have the same checkInterval for all devices using this DTH. 

We should use this formula going forward for all Z-Wave sleepy devices `checkInterval = 2 * checkInterval + (10 * 60)`.

greens:
I think I meant gen5. Most of the modernization you did in this PR (battery/wakeup/security) is already in gen 5.

PKacprowiczS:
In that case you're right, but gen 5 doesn't support tamper either, so now I'm not sure if I should add this device to existing DTH or create brand new with support for all features.

greens:
You could add tamper and the fingerprint to the gen5 DTH and then generate and assign a device-specific metadata that includes tamper.

That way the Aeon device would not display the tamper tile in OneApp while the Zooz device would.

greens (commented):
probably just revert all the changes on `aeon-multisensor`

PKacprowiczS:
@greens reverted and squashed

devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
```diff
@@ -21,9 +21,11 @@ metadata {
 		capability "Sensor"
 		capability "Battery"
 		capability "Health Check"
+		//capability "Tamper Alert" waiting for LE fix for Zooz 4-in-1
```
greens: Sorry I didn't realize. This isn't the right approach. 

We should just give the Zooz fingerprint-specific metadata pointing to "generic-motion-6", which includes everything except Tamper

greens (approved):

dkirker:
The function "clearTamper" is not defined...

greens:
@dkirker that was fixed in the later PR: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11083

dkirker:
@greens I see. *that* one is still missing `clearTamper`... That also make more sense to the exception that Steven was seeing.

PR #4015, opened 2019-02-26
ICP-8161 Make initial humidity request more robust

dkirker:
Sometimes the first humidity request gets returned as "0%", so request twice to make sure we get a refreshed value on install/configure.


tpmanley (dismissed):

varzac (dismissed):

greens:
While we're here can we just group all the repeated code in `poll` and `configure`?

greens (dismissed):
This does add the updateWeather to configure when it wasn't there before if you're okay with it.

dkirker:
Yeah. It'll likely do nothing on pairing since settings.zipcode is more than likely null.

dkirker:
Just a little tweak to the old PR, and an addition of humidity-less vids for the two devices that don't support a humidity reading. (A note, a comment from Stelpro seemed to indicate that the devices report back '0' when requested, but don't actually have the sensors, so I am not going to put checks around each call for humidity. This will let us at least fill in the attribute with something.)

varzac (approved):

greens (approved):

dkirker:
I will be adding some logic to not display the increasing or decreasing temperature if we have received an alarm event and only show the alarm temperature. This seems to be tripping up certification.

PR #4021, opened 2019-02-26
HCS-391 name translation causing metadata switch

greens:


tpmanley (commented):
Does this work?

greens:
idk. Just trying this first to avoid having to re-deploy strings.

greens:
I guess the strings don't get pulled down when you self-publish, so I'll try it this way and have Nate deploy the strings.

I think this join name should be inoffensive enough.

PR #4023, opened 2019-02-27
HDP-146 Reduce cloud executions

greens:


greens:
You guys have been around longer so maybe you can tell me why the smartpower outlet isn't already marked for runLocally and executeLocally

varzac (approved):

tpmanley (approved):
I believe the reason that SmartPower Outlet wasn't marked to execute commands locally is it was one of the first devices we supported in LEE and `executeCommandsLocally` hadn't been defined at the time.

PR #4025, opened 2019-02-28
null: Door Device Handlers

drobinson142:
Submitting 'Door Device Handlers' for publication


tpmanley:
@drobinson142 , we already have support for the `MCT-340 E` in the `SmartSense Open/Closed Sensor` DTH. If you could see if the `MCT-340 SMA` also works with that DTH I'd be willing to add a the fingerprint to that DTH but would rather not add a new DTH unless there is some new functionality.

PR #4038, opened 2019-03-04
ICP-8469 Child devices do not display during setup screen

greens:
Changing "completedSetup" to "false" should give us the correct behavior


tpmanley (approved):

greens:
didn't work

PR #4045, opened 2019-03-06
[ICP-8606] Update Devices Names for all Innr devices 

DCzarneckaS:
@MMateusiakS  @greens @MAblewiczS 
Update Devices Names for all Innr devices


greens (approved):
Thanks for taking care of this.

PR #4046, opened 2019-03-06
[ICP-8448] Z-Wave Button event missed isStateChange flag

PKacprowiczS:
@greens @MMateusiakS @MGoralczykS 
This probably caused missed events in IDE and issues with automation.


greens:
While you're here you should also add the missing `mnmn` field to the definition.

greens (approved):

PR #4051, opened 2019-03-07
[DVCSMP-4674] Add check for poll command before calling it

vl4ds:


vl4ds:
@gelicia @workingmonk  please review, thanks!

tpmanley (approved):

PR #4054, opened 2019-03-08
[ICP-8666] Add fingerprint for submodel of ozom CO-8ZBS

MMateusiakS:
Just adding fingerprint for OZOM device. @dkirker  made DTH for this, but we got different submodel in SRPOL. @dkirker commited this DTH directly to STG, so just to make this quite conflictless I submit his code with fingerprint od submodel 15T.

@greens, I guess DTH was reviewed previously and from my side I just added second FP.

### DIFF
--- a/devicetypes/smartthings/zigbee-co-sensor.src/zigbee-co-sensor.groovy
+++ b/devicetypes/smartthings/zigbee-co-sensor.src/zigbee-co-sensor.groovy
@@ -1,0 +1,158 @@
+ /*
+  *  Copyright 2019 SmartThings
+  *
+  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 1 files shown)


greens:
Donald's staging changes should have been rolled up to master already. I'm investigating why that didn't happen and holding off on this until then.

dkirker:
@greens My changes were opened against staging and merged in to there after Tuesday's deploy. I assume there wasn't a roll up/roll down that happened after. (I know @workingmonk does them on Tuesday, and I know sometimes they happen on Thursday, but I don't think there was one this Thursday.)

dkirker:
@MMateusiakS If you open a PR against staging you can slip the fingerprint in and it will go out with the deploy tomorrow.

PR #4066, opened 2019-03-11
[ICP-8666] Add fingerprint with submodel of Ozom CO-8ZBS

MMateusiakS:
@dkirker as we discussed in ICP-8666 I am doing PR against STG.


dkirker (dismissed):

dkirker (approved):
I tweaked the isOzomCO() method. I think I had originally intended to do an && and not a ||, and I also added the model # for the other fingerprint.

PR #4067, opened 2019-03-11
[CHAD-2943] Z-Wave Lock: Fix unreleased regression with call to requestCode

tpmanley:
The regression was introduced here:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4010

The problem was called out here:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4010#issuecomment-471463993


tpmanley:
@vaish-rahul thanks for catching this issue

kianooshkarami (approved):
Thanks

vaish-rahul:
@tpmanley With this change in place, two events will be generated:

1. When AlarmReport is received.
2. When UserCodeReport is received.

This will cause problem in Smart Lock Guest Access service which I have mentioned in detail in the JIRA ticket DVCSMP-4694.

PR #4068, opened 2019-03-12
ICP-8166 Ignore 0% battery events from Utilitech siren

dkirker:


devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -363,7 +363,13 @@ def handleSwitchValue(value) {
 
 def zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
 	def map = [name: "battery", unit: "%"]
-	if (cmd.batteryLevel == 0xFF) {
+
+	// The Utilitech siren always sends low battery events (0xFF) below 20%,
+	// so we will ignore 0% events that sometimes seem to come before valid events.
+	if (cmd.batteryLevel == 0 && isUtilitech()) {
```
tpmanley, line 369: I didn't check but does this work even if the device is running locally?

greens: We don't parse zwave battery events locally because of all the non-standard stuff like this.

greens (approved):

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
dkirker: Yeah, what @greens said.

PR #4070, opened 2019-03-12
[CHAD-2943] Revert changes since still not working correctly

tpmanley:
There are still problems with these changes and I don't want to risk delaying the release so just reverting the changes to give us more time to get the correct fix for the original problem.

### DIFF
--- a/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
+++ b/devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
@@ -248,7 +248,7 @@ def parse(String description) {
 			)
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


tpmanley:
Github is being weird but I'm trying to reopen this PR. There are issues with the changes that we need more time to work out even though I thought for a minute there that we might have a quick fix.

tpmanley:
@kianooshST suggested we make this change instead of reverting: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4071

tpmanley:
We're going with https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4071 instead

PR #4076, opened 2019-03-13
[WWST-2361] Added support for Aeotec NanoMote Quad

PKacprowiczS:
@greens @MMateusiakS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
+++ b/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
@@ -58,10 +58,10 @@ def updated() {
 def initialize() {
 	def numberOfButtons = prodNumberOfButtons[zwaveInfo.prod]
 	sendEvent(name: "numberOfButtons", value: numberOfButtons, displayed: false)
-	if(zwaveInfo.mfr?.contains("0371")) {
+	if(isUntrackedAeotec()) {
 		sendEvent(name: "DeviceWatch-Enroll", value: JsonOutput.toJson([protocol: "zwave", scheme:"untracked"]), displayed: false)
 	} else {
-		sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+		sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 10 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	}
 	if(!childDevices) {
 		addChildButtons(numberOfButtons)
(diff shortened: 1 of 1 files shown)


greens:
generic-4-button doesn't support double-pressed. Do the nanomotes?

PKacprowiczS:
Nope, they only support pressed and held. I've left double because it is supported by Fibaro devices and it's already working part of Button Capability.

devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -54,7 +58,11 @@ def updated() {
 def initialize() {
 	def numberOfButtons = prodNumberOfButtons[zwaveInfo.prod]
 	sendEvent(name: "numberOfButtons", value: numberOfButtons, displayed: false)
-	sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	if(zwaveInfo.mfr?.contains("0371")) {
```
kianooshkarami: Can we have a helper function that defines which manufacturer this is?

devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -54,7 +58,11 @@ def updated() {
...
+		sendEvent(name: "DeviceWatch-Enroll", value: JsonOutput.toJson([protocol: "zwave", scheme:"untracked"]), displayed: false)
+	} else {
+		sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
```
kianooshkarami: Is the wake up interval 4 hours? 

Sorry for repeating myself across multiple PRs, but the check interval needs to be 
```

func calculate_check_interval(w: wakeup_interval)
{
    check_interval = 2 * w + 10 minutes;
}
```

The 10 minutes should be the offset used going forward, the reason being that if we miss the first wake up interval (at the radio level), the device could stay up for 5 minutes before going back to sleep and wake up 4 hours or so later. Also most of these low power devices are running on Low power internal clocks which can be inaccurate, so the 4 hours wake up interval could sleep by seconds or more, so the 10 minutes should be a decent cushion that when elapsed we can feel safe to mark the device offline

devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -74,8 +82,11 @@ def initialize() {
 
 def configure() {
 	def cmds = []
-	for(def parameter : 21..26) {
-		cmds += secure(zwave.configurationV1.configurationSet(parameterNumber: parameter, scaledConfigurationValue: 15)) //Makes Fibaro KeyFob buttons send all kind of supported events
+	if(zwaveInfo.mfr?.contains("010F")) {
```
kianooshkarami: Please use a helper function to identify the manufacturer

devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
PKacprowiczS: Sorry, totally forgot about it. Now it should be ok.

greens (approved):

PR #4079, opened 2019-03-15
WWST-3653: Add korean for door lock

FenMei007S:
@tpmanley @greens  it is for korea market


greens (approved):

greens:
I think we should add this for now, but there was still no resolution for HCS-391.

PR #4096, opened 2019-03-19
ICP-8494 Set isStateChange:true for hue and saturation events

dkirker:


dkirker:
Issue fixed using another method.

PR #4099, opened 2019-03-20
Curb definition block for oneapp compatibiity

munds:
Adds code to the definition block so that it shows up in the right place for one app


ghost (approved):

ghost (approved):
Approved.

PR #4100, opened 2019-03-20
DVCSMP-4714 weather icons

bflorian:
Fixed problem with leading zeros in icon strings. Also added refresh when app is updated (as in changing zip code)

### DIFF
--- a/devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
+++ b/devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
@@ -70,16 +70,16 @@ metadata {
         }
 
         standardTile("weatherIcon", "device.weatherIcon", decoration: "flat", height: 2, width: 2) {
-            state "00", icon:"https://smartthings-twc-icons.s3.amazonaws.com/00.png", label: ""
-            state "01", icon:"https://smartthings-twc-icons.s3.amazonaws.com/01.png", label: ""
-            state "02", icon:"https://smartthings-twc-icons.s3.amazonaws.com/02.png", label: ""
-            state "03", icon:"https://smartthings-twc-icons.s3.amazonaws.com/03.png", label: ""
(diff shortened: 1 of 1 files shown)


tpmanley:
I'm still getting a lux of 10000 for an `obs.iconCode` of 26

tpmanley:
It's happening because `obs.iconCode` is an Integer. I added this code to the DTH:

`log.info "obs.iconCode is ${obs.iconCode.getClass()}"`

And got this: 

`java.lang.SecurityException: Invoking class java.lang.Integer.getClass() is not allowed @line 375 (estimateLux)`

bflorian:
How about now @tpmanley ? Sandboxing prevents getting the Class object, which lets you do nasty things

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -347,11 +351,11 @@ private estimateLux(obs, sunriseDate, sunsetDate) {
...
     else {
         //day
-        switch(obs.iconCode) {
-            case '04':
+        switch(obs.iconCode as String) {
+            case '4':
                 lux = 200
                 break
-            case ['05', '06', '07', '08', '09', '10',
+            case ['5', '6', '7', '8', '9', '10',
```
tpmanley: I still need to test this, but should 16 be in the list too? Also, if we changed the cases to integers we could do `case 5..26:` which seems like a nice improvement

bflorian: 16 looks like snow so it seems like it should be included. I switched to using integer ranges.

tpmanley (approved):

tpmanley:
@bflorian I think this can be merged

tpmanley:
@bflorian @workingmonk any reason not to merge this, either now or after the next week's deploy (since this is pointed at `staging`)?

rboy1:
This patch would work too, it also needs a fix to the 2 other attributes to convert to String while reporting them.

PR #4101, opened 2019-03-21
ICP-8430 Update Stelpro outdoor temperature weather API logic

dkirker:
The new API does not allow for non-U.S. zip codes; however, it allows for any geolocation. To make sure that this works globally, and to require no extra steps for existing installations (except for the possible ones that do not have a geolocation defined for their SmartThings location), we'll keep track of a failed API call on the defined zipcode with a flag so that we don't continue to make requests in the future that we know will fail.

### DIFF
--- a/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
@@ -49,8 +49,14 @@ metadata {
 	simulator { }
 
 	preferences {
-		input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
-		input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+		section {
+			input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
+		}
+		section {
+			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter a zipcode below.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+			input title: "", description: "Only U.S. based zipcodes are supported by the weather provider. However, if you are outside of the U.S. then enter your zipcode below and make sure that your SmartThings location has a Geolocation configured.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+		}
 	}
 
 	tiles(scale : 2) {
@@ -161,6 +167,7 @@ def updated() {
 
 	unschedule(scheduledUpdateWeather)
 	if (settings.zipcode) {
+		state.invalidZip = false // Reset and validate the zip-code later
 		runEvery1Hour(scheduledUpdateWeather)
 		scheduledUpdateWeather()
 	}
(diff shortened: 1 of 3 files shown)


dkirker:
cc @Stelpro Let me know how the text for the device preference description looks.

stelpro:
Hi Donald,

May I suggest that the procedure is the same for all user without any regards to their country?
For instance, instead of specifying that the weather provider does not support non-US zip code, you could just write: "Enter your zip code or postal code and make sure your SmartThings hub has its geolocation enabled".

Or something similar.

I feel it sounds a little less as a patch that way...
What do you think?

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -161,6 +167,7 @@ def updated() {
 
 	unschedule(scheduledUpdateWeather)
 	if (settings.zipcode) {
+		state.invalidZip = false // Reset and validate the zip-code later
```
tpmanley, line 169: Unless I'm not understanding this change correctly, I think setting `state.invalidZip = false` just means it will always use lat/lon and not zip due to the check on line 220

stelpro: The check on line 220 verifies if the zip code is not invalid (double negative), so it's a check for validity.
If the zipcode is valid, getTwcConditions() is invoked

tpmanley: 🤦‍♂️ it's too early, thanks :)

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -349,21 +356,50 @@ def updateWeather() {
...
+					// Restrict to two decimal places for the API
+					java.text.DecimalFormat formatter = new java.text.DecimalFormat()
+					formatter.setMaximumFractionDigits(2)
+					formatter.setMinimumFractionDigits(0)
+
+					def lat = formatter.format(location.latitude)
+					def lon = formatter.format(location.longitude)
+
+					def latlon = "${lat},${lon}"
+					weather = getTwcConditions(latlon)
```
greens: `weather = getTwcConditions(sprintf("%.2f, %.2f", lat, lon)`

greens:
Any particular reason this needs to go directly to staging?

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -49,8 +49,14 @@ metadata {
...
 
 	preferences {
-		input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
-		input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+		section {
+			input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
+		}
+		section {
+			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter a zipcode below.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
+			input title: "", description: "Only U.S. based zipcodes are supported by the weather provider. However, if you are outside of the U.S. then enter your zipcode below and make sure that your SmartThings location has a Geolocation configured.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
```
greens: I would say "Only US ZIP codes are supported by the weather provider. However, if you are outside the US, you can enter your _postal code_ to make sure that your SmartThings location supports other geolocation services."

stelpro:
@greens,

The reason we need to fix this is because the outdoor temperature is no longer displayed on the Stelpro thermostat since the weather API change.

dkirker:
@greens As @stelpro noted, the weather API change unknowingly created a regression, so Canadian users are no longer seeing the outdoor temperature displayed on their thermostats. I wanted to get this released on Tuesday, but I also wanted to continue testing and tweak the logic to not hit the API (or our cache of the data) with known bad location data.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
dkirker: I'll give this a test. Thanks

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
dkirker: I think @stelpro 's suggestion above might be a bit better and possibly result in less confusion. I'm looking at how to make sure that the "SmartThings hub has its geolocation enabled" part makes sense. (I need to play with the configuration setting in the new ST app a little more to see what wording it uses to best be generic for both.)

dkirker:
I have updated the logic and the display text for the preferences.

greens (approved):

greens:
Just ping me when you want to merge.

dkirker:
@greens totally missed your last comment. Let's go ahead and merge this, and I will work with Bashi today.

stelpro:
For info, I tried the weather in the SmartThings Classic app, and it works.
However, I do not receive it in the new app...
Is it normal?

PR #4107, opened 2019-03-22
[WWST-2344] Added support for Aeotec KeyFob

PKacprowiczS:
@greens @MMateusiakS @MGoralczykS @DCzarneckaS @ZWozniakS


greens:
So this means that for the nanomote wakeup might not have been parsed. Is it possible that this means  it might actually be okay to give it proper health check?

PKacprowiczS:
I've also thought about it, but nanomote still has wakeup interval value at 0, so it won't report anything, anyway.

greens (approved):

PR #4110, opened 2019-03-22
ICP-8495 Battery percentage incorrect

greens:


dkirker (approved):
Looks good if it works out as per the comments on the ticket.

I think the only two things I would modify are consistent spaces around the `if`s, and switching the clusters, commands, and attributes to constants. (@tpmanley Maybe this is stuff we want to add into the Zigbee class?)

PR #4114, opened 2019-03-25
ICP-8874 Fix common exception in Springs Remote DTH

greens:


devicetypes/smartthings/springs-window-fashions-remote.src/springs-window-fashions-remote.groovy
```diff
@@ -82,7 +82,7 @@ def zwaveEvent(physicalgraph.zwave.commands.crc16encapv1.Crc16Encap cmd) {
 }
 
 def zwaveEvent(physicalgraph.zwave.Command cmd) {
-    createEvent(isStateChange: true, dscriptionText: "$device.displayName: $cmd")
+    createEvent(isStateChange: true, descriptionText: "$device.displayName: $cmd")
```
tpmanley: Why is this DTH creating events for anything besides battery? Any reason we shouldn't just remove the `createEvent` from everything except the `BatteryReport` handler?

greens:
That should clean it up a bit generally and get rid of messages we didn't need to be sending.

dkirker (approved):

PR #4121, opened 2019-03-27
[WWST-2398] Add FP and change addChild method for Zooz Zen Strip v2

DCzarneckaS:
@greens @MMateusiakS 
RAW: 
 zw:L  type:1004 mfr:027A prod:A000 model:A004 ver:2.00 zwv:5.03 lib:03  cc:5E,25,85,8E,59,55,86,72,5A,73,70,71,32,9F,60,6C,7A role:05 ff:8800  ui:8800 epc:7 ep:['1001 5E,85,59,8E,25']

### DIFF
--- a/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
+++ b/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
@@ -27,6 +27,7 @@ metadata {
 		fingerprint mfr:"0086", prod:"0003", model:"0084", deviceJoinName: "Aeotec Nano Switch 1"
 		fingerprint mfr:"0086", prod:"0103", model:"0084", deviceJoinName: "Aeotec Nano Switch 1"
 		fingerprint mfr: "0000", cc: "0x5E,0x25,0x27,0x32,0x81,0x71,0x60,0x8E,0x2C,0x2B,0x70,0x86,0x72,0x73,0x85,0x59,0x98,0x7A,0x5A", ccOut:"0x82", ui:"0x8700", deviceJoinName: "Aeotec Nano Switch 1"
+		fingerprint manufacturer: "027A", prod: "A000", model: "A004", deviceJoinName: "Zooz ZEN Power Strip"
 	}
 
 	tiles(scale: 2){
@@ -142,7 +143,11 @@ private createMeterEventMap(cmd) {
 
 def zwaveEvent(physicalgraph.zwave.commands.multichannelv3.MultiChannelEndPointReport cmd, ep = null) {
 	if(!childDevices) {
-		addChildSwitches(cmd.endPoints)
+    	if (zwaveInfo.mfr.equals("027A") && zwaveInfo.model.equals("A004")) {
+        	addChildSwitches(5)
+        } else {
+			addChildSwitches(cmd.endPoints)
+    	}
 	}
 	response([
 			resetAll(),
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
```diff
@@ -27,6 +27,7 @@ metadata {
 		fingerprint mfr:"0086", prod:"0003", model:"0084", deviceJoinName: "Aeotec Nano Switch 1"
 		fingerprint mfr:"0086", prod:"0103", model:"0084", deviceJoinName: "Aeotec Nano Switch 1"
 		fingerprint mfr: "0000", cc: "0x5E,0x25,0x27,0x32,0x81,0x71,0x60,0x8E,0x2C,0x2B,0x70,0x86,0x72,0x73,0x85,0x59,0x98,0x7A,0x5A", ccOut:"0x82", ui:"0x8700", deviceJoinName: "Aeotec Nano Switch 1"
+		fingerprint manufacturer: "027A", prod: "A000", model: "A004", deviceJoinName: "Zooz ZEN Power Strip"
```
tpmanley: Use `mfr` like the other fingerprints

devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
```diff
@@ -142,7 +143,11 @@ private createMeterEventMap(cmd) {
 
 def zwaveEvent(physicalgraph.zwave.commands.multichannelv3.MultiChannelEndPointReport cmd, ep = null) {
 	if(!childDevices) {
-		addChildSwitches(cmd.endPoints)
+		if (zwaveInfo.mfr.equals("027A") && zwaveInfo.model.equals("A004")) {
```
kianooshkarami: Please use helper function to identify the devices. It helps in readability. 

Also, This change should have a comment that documents the reason behind this change.

tpmanley: I think @kianooshST was asking for a helper function like `isZoozZenStripV2()` which would check the `mfr` and `model`. For the comment something like `cmd.endPoints includes the USB ports but we don't want to expose them as child devices since they cannot be controlled so hardcode to just include the outlets` would be more descriptive (please verify that's accurate too).

devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
```diff
@@ -140,15 +142,37 @@ private createMeterEventMap(cmd) {
 	eventMap
 }
 
+// cmd.endPoints includes the USB ports but we don't want to expose them as child devices since they cannot be controlled so hardcode to just include the outlets
 def zwaveEvent(physicalgraph.zwave.commands.multichannelv3.MultiChannelEndPointReport cmd, ep = null) {
 	if(!childDevices) {
-		addChildSwitches(cmd.endPoints)
+		if (isZoozZenStripV2()) {
+			addChildSwitches(5)
+		} else if (isZoozDoublePlug()) {
```
PKacprowiczS, line 150: Does Double Plug report wrong number of endpoints that this exception here is necessary?

DCzarneckaS: " // cmd.endPoints includes the USB ports but we don't want to expose them as child devices since they cannot be controlled so hardcode to just include the outlets "

greens (approved):

PR #4123, opened 2019-03-28
ONEAPP-9767, The lock capibility have no autolockerror status

FenMei007S:
@tpmanley  As Insung cho metioned , autolockerror is just for classic app, and it will be handle as null,since lock capibility have no status of autolockerror,  this why oneapp show error popup.i changed it unlocked with timeout


tpmanley (approved):

dkirker:
I think this was the case I was trying to work with in ICP-1772. Since the OCF side doesn't have an "unknown" state or a "jammed" state we can't relay any of that back. There's further comments on the ticket (I was pushing to get that defined on the OCF side).

PR #4127, opened 2019-03-29
ICP-8276 Only clear temperature alarm when the temperature is leaving the alarm zone

dkirker:


devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -330,20 +330,34 @@ def handleTemperature(descMap) {
...
+				if ((lastAlarm == "heat" &&
+						map.value < (map.unit == "C" ? 50 : 122) &&
+						lastTemp > map.value) ||
+					(lastAlarm == "freeze" &&
+						map.value > (map.unit == "C" ? 0 : 32) &&
+						lastTemp < map.value)) {
+							sendEvent(name: "temperatureAlarm", value: "cleared")
+				}
+			}
+		} /*else {
```
dkirker: I'm planning to remove the commented code after feedback on the ticket.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -330,20 +330,34 @@ def handleTemperature(descMap) {
...
+		def lastTemp = device.currentValue("temperature")
+		def lastAlarm = device.currentValue("temperatureAlarm")
+		if (lastAlarm != "cleared") {
+			if (lastTemp && device.currentState("temperature").unit == map.unit) {
+				// Check to see if we are coming out of our alarm state and only clear then
+				// NOTE: A thermostat might send us an alarm *before* it has completed sending us previous measurements,
+				// so it might appear that the alarm is no longer valid. We need to check the trajectory of the temperature
+				// to verify this.
+				if ((lastAlarm == "heat" &&
+						map.value < (map.unit == "C" ? 50 : 122) &&
```
tpmanley: Keep the old comment so we know where the 50 comes from

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -330,20 +330,34 @@ def handleTemperature(descMap) {
...
+						lastTemp > map.value) ||
```
tpmanley: Personally I think it would be easier to understand if you switch the order here (and also below): ex: `map.value < lastTemp`

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -538,7 +552,7 @@ def configure() {
 	log.debug "binding to Thermostat cluster"
 	requests += zigbee.addBinding(THERMOSTAT_CLUSTER)
 	// Configure Thermostat Cluster
-	requests += zigbee.configureReporting(THERMOSTAT_CLUSTER, ATTRIBUTE_LOCAL_TEMP, DataType.INT16, 10, 60, 50)
+	requests += zigbee.configureReporting(THERMOSTAT_CLUSTER, ATTRIBUTE_LOCAL_TEMP, DataType.INT16, 1, 60, 50)
```
greens: 10s was too frequent? This seems like a bad idea.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
dkirker: Oh, so that it does the check for the freeze alarm before the heat alarm?

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
dkirker: This was a check to see if 10 was too infrequent. The issue, at least with this device, was that the Zigbee system was queuing up temperature reports from the thermostat system and then sending them to us about every 10 seconds. The firmware didn't seem to be clearing the queue if a new temperature report came in -- but the temperature alarm report would go to the front of the queue. So the result was we would get the alarm report, but then keep getting temperature reports up to 50C until the queue cleared, and the reports appeared to come at 10 second intervals. This doesn't seem to really help that.

I am going to revert this and just tweak the logic above.

greens (dismissed):
reporting interval looks less scary now.

Did you have any trouble like what you described when you tested it?

dkirker:
Changing to 1 second didn't seem to flush the queue quicker (though, I didn't try repairing the device). I just created a commit that will hide temperature readings when we receive an indication of an alarm. I need to test it, and make sure that if we receive a temp that is > HEAT_ALARM_TEMP or < FREEZE_ALARM_TEMP we start sending those values.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -333,18 +336,26 @@ def handleTemperature(descMap) {
...
 				// Check to see if we are coming out of our alarm state and only clear then
 				// NOTE: A thermostat might send us an alarm *before* it has completed sending us previous measurements,
 				// so it might appear that the alarm is no longer valid. We need to check the trajectory of the temperature
 				// to verify this.
 				if ((lastAlarm == "freeze" &&
-						map.value > (map.unit == "C" ? 0 : 32) &&
+						map.value > FREEZE_ALARM_TEMP &&
 						lastTemp < map.value) ||
 					(lastAlarm == "heat" &&
-						map.value < (map.unit == "C" ? 50 : 122) &&
```
greens: isn't it possible that the device-reported unit is different than the user's units (`getTemperatureScale()`)?

dkirker: They should match. map.unit is set to getTemperatureScale, and Zigbee devices report in C and we convert map.value when map.unit is set.

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
dkirker: Updated the "magic numbers" to "constants".

dkirker:
I am confident that this code takes care of the original bug and edge cases associated with it. This also contains the original check code to handle a bug I found in the device where if it is coming out of a temperature alarm (at least an overheating alarm) and drops below the alarm temperature but then goes back above the device doesn't actually send us a new alarm event.

greens (approved):

PR #4130, opened 2019-03-31
[CHAD-3119] Unify Command Class Version Mapping when Parsing Z-Wave Messages

kianooshkarami:
When parsing a Z-Wave message received from devices, a given DTH would need to map that message to a specific version of a command class as different versions of the command class may have different fields and require special parsing. The initial design of the Z-Wave messages could possibly include the encapsulation the message was received with. However, starting in 0.25.X Hub Firmware Release, the hub started removing the encapsulation the message was received with in preparation to handle S2 devices. The encapsulation removal allows most DTHs to support S2 seamlessly without having to worry about encapsulation.

However, with removal of encapsulation of the received messages, we also need to ensure that all the mapping used for encapsulated messages are also used in `zwave.parse()` method as well, and as such, the changes in this PR is to ensure SmartThings DTHs have the same source of command class version mapping regardless of encapsulation.

For more information, please visit: https://community.smartthings.com/t/z-wave-dths-changes-to-encapsulation-handling-starting-in-0-25-x-hub-firmware-release/159433

### DIFF
--- a/devicetypes/smartthings/fibaro-dimmer.src/fibaro-dimmer.groovy
+++ b/devicetypes/smartthings/fibaro-dimmer.src/fibaro-dimmer.groovy
@@ -73,6 +73,17 @@ metadata {
 	}
 }
 
+/**
+ * Mapping of command classes and associated versions used for this DTH
+ */
+private getCommandClassVersions() {
+	[
+		0x26: 1,  // Switch Multilevel
+		0x70: 2,  // Configuration
+		0x72: 2   // Manufacturer Specific
+	]
+}
+
 def parse(String description) {
 	def item1 = [
 		canBeCurrentState: false,
@@ -83,7 +94,7 @@ def parse(String description) {
 		value:  description
 	]
 	def result
-	def cmd = zwave.parse(description, [0x26: 1, 0x70: 2, 072: 2])
+	def cmd = zwave.parse(description, commandClassVersions)
     //log.debug "cmd: ${cmd}"
     
     if (cmd) {
@@ -141,6 +152,10 @@ def createEvent(physicalgraph.zwave.commands.switchmultilevelv1.SwitchMultilevel
 		result[i].type = "digital"
 	}
 	result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerSpecificReport cmd) {
+	log.trace "[DTH] Executing 'zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerSpecificReport)' with cmd = $cmd"
 }
 
 def doCreateEvent(physicalgraph.zwave.Command cmd, Map item1) {
--- a/devicetypes/smartthings/fibaro-door-window-sensor.src/fibaro-door-window-sensor.groovy
+++ b/devicetypes/smartthings/fibaro-door-window-sensor.src/fibaro-door-window-sensor.groovy
@@ -105,11 +105,28 @@
 	}
 }
 
+/**
+ * Mapping of command classes and associated versions used for this DTH
+ */
+private getCommandClassVersions() {
+	[
+		0x30: 1,  // Sensor Binary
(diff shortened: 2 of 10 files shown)


devicetypes/smartthings/fibaro-dimmer.src/fibaro-dimmer.groovy
```diff
@@ -83,7 +94,7 @@ def parse(String description) {
 		value:  description
 	]
 	def result
-	def cmd = zwave.parse(description, [0x26: 1, 0x70: 2, 072: 2])
```
tpmanley, line 86: Interesting, this had a bug `072`

tpmanley: Although this DTH doesn't have a handler for `manufacturerspecificv2`. Perhaps that CC should be removed from the list that are parsed?

devicetypes/smartthings/fibaro-door-window-sensor.src/fibaro-door-window-sensor.groovy
```diff
@@ -105,11 +105,28 @@
 	}
 }
 
+/**
+ * Mapping of command classes and associated versions used for this DTH
+ */
+private getCommandClassVersions() {
+	[
+		0x30: 1,  // Sensor Binary
+		0x31: 2,  // Sensor MultiLevel
```
tpmanley, line 114: 0x31 wasn't in the CC list before but it did have a handler. It seems to make sense to add it but I'm not clear why it wasn't included before.

tpmanley:
If I counted correctly only 9 of the 24 changed DTHs appear to be broken right now. The rest already had all the correct CC in the list passed to `zwave.parse`. I'd suggest breaking this into two PRs so we can get the 9 DTH fixes released ASAP. That first PR can be against the `production` branch so we can hotfix release as soon as they are tested.

devicetypes/smartthings/fibaro-dimmer.src/fibaro-dimmer.groovy
kianooshkarami: Aah I thought it did. Will remove

cbaumler (dismissed):

kianooshkarami:
@tpmanley @greens @dkirker @cbaumler @eddysun1990 

I went through the files again and found 10 of them that we need to fix, and as such the PR has been updated with the 10 files.

devicetypes/smartthings/fibaro-dimmer.src/fibaro-dimmer.groovy
kianooshkarami: Added a `manufacturerspecificv2` handler as well

greens (approved):
This looks good. Some of these are moot (like specifying `0x98:1`), and I wonder if anyone else has any thoughts on doing this rather than just changing most handlers to use the default version from appengine?

tpmanley:
Can you add the Jira to the PR title?

devicetypes/smartthings/fibaro-door-window-sensor.src/fibaro-door-window-sensor.groovy
tpmanley: @kianooshST any idea why this wasn't included before?

kianooshkarami: No, sorry, I do not know. I can remove it, I feel like we should include it now.

tpmanley: I think it's fine to leave in as long as the tests cover its usage

jmagnuson:
August Pro lock looks good with `Z-Wave Lock Without Codes`.

greens (approved):

tpmanley (approved):

PR #4136, opened 2019-04-03
ONEAPP-10946 Prevent edge case in Mobile Presence

SungyeonWoo:
1. When 'not present' event received (left case)
     -> If occupancy value is not 'unoccupied', occupancy value should be 'unoccupied' before posting 'not present'
2. When 'occupied' event received (inside case)
     -> If presence value is not 'present', presence value should be 'present' before posting 'occupied'


destiny4man (approved):
Looks good

PR #4137, opened 2019-04-03
WWST-2086 - Null pointer exception

rboy1:
If the user doesn’t enter a zipcode it causes a null pointer exception

@juano2310


juano2310:
Thanks a lot!

juano2310 (approved):

PR #4138, opened 2019-04-03
null: Smart Basement

paulsmithkc:
Submitting 'Smart Basement' for publication


paulsmithkc:
New smartthings app supports this functionality without needing a custom smartapp.

PR #4140, opened 2019-04-04
[WWST-2486] Add FP for IKEA Surte light panel

DCzarneckaS:
@greens @tpmanley @MMateusiakS @PKacprowiczS 
RAW: 01 C05E 0220 02 09 0000 0003 0004 0005 0006 0008 0300 0B05 1000 04 0005 0019 0020 1000


greens (approved):
So @DCzarneckaS does the color temperature range just depend on the driver you have attached?

PR #4141, opened 2019-04-04
[DVCSMP-4721]Replace iasZoneConfig with just zone status config

varzac:
This is because the iasZoneConfig zigbee library call also configures
reporting for the IAS CIE address which we don't need for any of these
devices.

### DIFF
--- a/devicetypes/smartthings/Orvibo-Contact-Sensor.src/Orvibo-Contact-Sensor.groovy
+++ b/devicetypes/smartthings/Orvibo-Contact-Sensor.src/Orvibo-Contact-Sensor.groovy
@@ -131,7 +131,7 @@ def configure() {
 	log.debug "Configuring Reporting, IAS CIE, and Bindings."
 	//The electricity attribute is reported without bind and reporting CFG. The TI plan reports the power once in about 10 minutes; the NXP plan reports the electricity once in 20 minutes
 	if (getDataValue("manufacturer") == "Aurora") {
-		cmds = zigbee.iasZoneConfig(30, 60 * 5) + zigbee.batteryConfig()
+		cmds = zigbee.enrollResponse() + zigbee.configureReporting(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS, DataType.BITMAP16, 30, 60 * 5, null) + zigbee.batteryConfig()
 	}
 	cmds += refresh()
 	cmds
--- a/devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
+++ b/devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
@@ -166,10 +166,7 @@ private Map parseIasMessage(String description) {
 def refresh()
 {
 	log.debug "refresh called"
-	[
-		"st rattr 0x${device.deviceNetworkId} 1 1 0x20"
-
-	]
+	return zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020) //battery read
 }
 
 def ping() {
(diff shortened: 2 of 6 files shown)


varzac:
We can decide if we care enough to make this change.

devicetypes/smartthings/Orvibo-Contact-Sensor.src/Orvibo-Contact-Sensor.groovy
```diff
@@ -131,7 +131,7 @@ def configure() {
 	log.debug "Configuring Reporting, IAS CIE, and Bindings."
 	//The electricity attribute is reported without bind and reporting CFG. The TI plan reports the power once in about 10 minutes; the NXP plan reports the electricity once in 20 minutes
 	if (getDataValue("manufacturer") == "Aurora") {
-		cmds = zigbee.iasZoneConfig(30, 60 * 5) + zigbee.batteryConfig()
+		cmds = zigbee.configureReporting(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS, DataType.BITMAP16, 30, 60 * 5, null) + zigbee.batteryConfig()
```
tpmanley: `iasZoneConfig` also calls `enrollResponse`. I think at the very least you should pull that into the DTH so this PR doesn't change the order of configuration related messages and only removes the reporting of the CIE Address. Looking at all the DTHs changed in this PR I think the `configure` and `refresh` functions can be optimized further to reduce redundant commands and decrease the time it takes to get current status on initial join but I think those could be handled in a separate PR to decrease risk.

varzac: I left it out mostly because it was duplicated elsewhere.  In this case `enrollResponse` is still called on configure, but you are right that it changes the order.  I could re-arrange things to try to maintain order, but I didn't think it would be a big issue for the enroll response command to come at a different point.  Do you think we should be duplicating and sending multiple versions?

devicetypes/smartthings/zigbee-valve.src/zigbee-valve.groovy
```diff
@@ -158,7 +158,7 @@ def refresh() {
     cmds += zigbee.configureReporting(CLUSTER_BASIC, BASIC_ATTR_POWER_SOURCE, DataType.ENUM8, 5, 21600, 1)
     if (device.getDataValue("model") == "E253-KR0B0ZX-HA") {
         cmds += zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS)
-        cmds += zigbee.iasZoneConfig()
+        cmds += zigbee.configureReporting(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS, DataType.BITMAP16, 0, 60 * 60, null)
```
varzac, line 162: This one actually doesn't have the enrollResponse so I will add that.

devicetypes/smartthings/Orvibo-Contact-Sensor.src/Orvibo-Contact-Sensor.groovy
tpmanley: If the device isn't enrolled in the IAS Zone before the ZoneStatus is read it returns the incorrect status. I am not sure the same issue exists with reporting of ZoneStatus but I wanted to err on the side of caution. Like I said I think these functions can be reworked more massively to improve them even further which we could do now or later, but they would need to get much more thorough testing if we change the order of the functions.

devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
```diff
@@ -166,66 +166,21 @@ private Map parseIasMessage(String description) {
...
 def ping() {
 	refresh()
 }
 
 def configure() {
-
-	String zigbeeEui = swapEndianHex(device.hub.zigbeeEui)
 	log.debug "Configuring Reporting, IAS CIE, and Bindings."
-	def configCmds = [
-		"zcl global write 0x500 0x10 0xf0 {${zigbeeEui}}", "delay 200",
```
greens, line 184: Isn't this configuring the CIE address? Is it just unused by the device?

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
```diff
@@ -182,7 +182,11 @@ def configure() {
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 
 	log.debug "Configuring Reporting, IAS CIE, and Bindings."
-	def cmds = refresh() + zigbee.iasZoneConfig(30, 60 * 5) + zigbee.batteryConfig() + zigbee.temperatureConfig(30, 60 * 30) + zigbee.enrollResponse()
+	def cmds = refresh() +
+			zigbee.configureReporting(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS, DataType.BITMAP16, 30, 60 * 5, null) +
+			zigbee.batteryConfig() +
+			zigbee.temperatureConfig(30, 60 * 30) +
+			zigbee.enrollResponse()
```
greens, line 189: should `enrollResponse()` be moved to the front here like Tom mentioned?

devicetypes/smartthings/nyce-motion-sensor.src/nyce-motion-sensor.groovy
varzac: That is done as a part of the `enrollResponse` call as well.

devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy
varzac: `enrollResponse` is also called in `refresh` which is the first thing that happens here.  So previously we were sending 3 sets of `enrollResponse` commands, and I figured I'd drop it to 2

greens: 👍

greens (approved):

tpmanley (approved):
Need a ticket in the title

PR #4143, opened 2019-04-06
Spruce Scheduler Weather Api and Sensor model number update

natec007:
[213632f]Critical Update Spruce Scheduler to use new ST TWC weather api. Weather api function was partially implemented by SmartThings team, however several functions were not completed. Primary changes in weather function. Also fixed 1 bug and added version number.

[c3df129] Add updated device model number and fingerprint for Gen 3 Spruce Sensor. Update device health to 3 hours to reduce "false alarms" of inactivity.

### DIFF
--- a/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
+++ b/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
@@ -1,5 +1,5 @@
 /**
- *  Spruce Sensor -updated with SLP model number 5/2017
+ *  Spruce Sensor -updated with SLP3 model number 3/2019
  *
  *  Copyright 2014 Plaid Systems
  *
@@ -19,6 +19,10 @@
  -------5/2017 Updates--------
  -Add fingerprints for SLP
  -add device health, check every 60mins + 2mins
+ 
+ -------3/2019 Updates--------
+ -Add fingerprints for SLP3
+ -change device health from 62mins to 3 hours
  */
  
 metadata {
@@ -41,6 +45,7 @@ metadata {
         
         fingerprint profileId: "0104", inClusters: "0000,0001,0003,0402,0405", outClusters: "0003, 0019", manufacturer: "PLAID SYSTEMS", model: "PS-SPRZMS-01", deviceJoinName: "Spruce Sensor"
         fingerprint profileId: "0104", inClusters: "0000,0001,0003,0402,0405", outClusters: "0003, 0019", manufacturer: "PLAID SYSTEMS", model: "PS-SPRZMS-SLP1", deviceJoinName: "Spruce Sensor"
+        fingerprint profileId: "0104", inClusters: "0000,0001,0003,0402,0405", outClusters: "0003, 0019", manufacturer: "PLAID SYSTEMS", model: "PS-SPRZMS-SLP3", deviceJoinName: "Spruce Sensor"
 	}
 
 	preferences {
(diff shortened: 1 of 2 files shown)


devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
```diff
@@ -316,8 +321,8 @@ def updated(){
             sendEvent(name: 'configuration',value: 0, descriptionText: "Settings changed and will update at next report. Measure interval set to ${interval} mins")
     	}
     }
-    //check every 1 hour + 2mins
-    sendEvent(name: "checkInterval", value: 1 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+    //check every 3 hours
+    sendEvent(name: "checkInterval", value: 3 * 60 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
tpmanley: It looks like attribute 0x0000 on cluster 0x0405 is configured for reporting every 610 seconds ~ 10 minutes. Why the increase of checkInterval to 3 hours?

natec007: Yes, the default interval is 610 s, however, in the preferences the user can change the interval to 2 hours, so I had made it 3 to cover the longest interval.  If its preferable to keep it shorter, 2 hours would work, but would prefer to make it a little longer than the previous 1 hour, if possible.
line 342:
> //update intervals
def intervalUpdate(){
	log.debug "intervalUpdate"
    def minReport = 10
    def maxReport = 610
    if (interval != null) {
    	minReport = interval
        maxReport = interval * 61
    }
    [    
    	"zcl global send-me-a-report 0x405 0x0000 0x21 $minReport $maxReport {6400}", "delay 500",
    	"send 0x${device.deviceNetworkId} 1 1", "delay 500",
        "zcl global send-me-a-report 1 0x0000 0x21 0x0C 0 {0500}", "delay 500",
        "send 0x${device.deviceNetworkId} 1 1", "delay 500",
    ]    
}

greens: You could programmatically change the checkInterval based on that preference. I know we do that in `aeon-multisensor-6` (see the bottom of the `configure()` method). For longer `checkIntervals` we prefer `reportingInterval * 2 + 10 * 60`.

natec007: Sounds good, changed install value back to previous 62 min value, and added variable interval for when user changes poll interval.  However, did make minimum value 62 mins.  B/c some of the previous model sensors, would change the "measure" interval but would still only report after an hour.

greens (approved):
looks good to me

dkirker (approved):

dkirker:
Merging.

@natec007 Do you have any tips for testing this? It should be in staging tomorrow, and I want to work with our QA so that they can verify this.

Also, Happy Earth Day!

natec007:
@dkirker 
The changes are to the weather api.  Setting a schedule to run a few minutes into the future and then checking the logs will show the weather forecast and calculated weather variables when the schedule executes.  The weather will also display in the Recently.  In the Spruce Scheduler, the following fields would be set:

- Schedule Settings: Select device, set time +5mins, set today as day
-Weather Settings: Enable Seasonal Weather
-Zone Summary: Set 1 or more zones, then
- Each Zone x:  Set Sprinkler type and Landscape to anything other than off
All other values can be default
Save and that is all that is needed for test
Thanks!

PR #4147, opened 2019-04-07
[DVCSMP-4721] SmartSense Open/Close Sensor/NYCE Motion Sensor: Add missing DataType imports

tpmanley:
This recent PR introduced the use of DataType in these DTHs but forgot to
import it:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4141

This bug has not yet gone to staging.

https://smartthings.atlassian.net/browse/DVCSMP-4721


varzac (approved):
🔔 on me

PR #4158, opened 2019-04-09
DVCSMP-4722 Move SmartSense Button to public repo

greens:


tpmanley (approved):

greens:
I looked into it but couldn't find a particularly not-gross way of preserving git history. I can look more into doing the gross stuff though if you'd prefer.

tpmanley:
Nope, we have the history on the other repo so this is fine

PR #4181, opened 2019-04-19
[ICP-9255] Add attribute "scene" so that scene ID will be reported to SmartApps

eddysun1990:
* Added "Scene" to attribute

https://smartthings.atlassian.net/browse/ICP-9255


devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
```diff
@@ -14,7 +14,7 @@ metadata {
 		command "reset"
 		command "clearError"
 
-		attribute "errorMode", "string"
+		attribute "errorMode", "string", "scene"
```
tpmanley: You need to make a separate attribute for `scene`

greens (approved):

PR #4183, opened 2019-04-19
ICP-9214 Fix fan event handling for GE controller

dkirker:
This is a first pass. Not fully ready for review yet.

### DIFF
--- a/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
+++ b/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
@@ -108,21 +108,33 @@ def zwaveEvent(physicalgraph.zwave.commands.hailv1.Hail cmd) {
 
 def fanEvents(physicalgraph.zwave.Command cmd) {
 	def value = (cmd.value ? "on" : "off")
-	def result = [createEvent(name: "switch", value: value)]
-	result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)
+	def result = [createEvent(name: "switch", value: value),
+				  createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)]
+
 	def fan_level = Math.ceil(cmd.value/33) as int
-//	if (cmd.value < 33 && cmd.value >= 1) fan_level = 1 //sometimes we get "1" when the device is on low
+
+	// The GE, Honeywell, and Leviton treat 33 as medium, so account for that
+	if (cmd.value == 33) {
+		fan_level = 2
+	}
 	result << createEvent(name: "fanSpeed", value: fan_level)
+
 	return result
 }
 
 def on() {
 	state.lastOnCommand = now()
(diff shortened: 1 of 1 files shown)


dkirker:
Waiting for feedback from QA about the other GE device and the Honeywell device.

dkirker:
Updated given the testing/verification that I did with the Leviton and Honeywell models

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -108,21 +108,33 @@ def zwaveEvent(physicalgraph.zwave.commands.hailv1.Hail cmd) {
...
-	def result = [createEvent(name: "switch", value: value)]
-	result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)
+	def result = [createEvent(name: "switch", value: value),
+				  createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value)]
+
 	def fan_level = Math.ceil(cmd.value/33) as int
-//	if (cmd.value < 33 && cmd.value >= 1) fan_level = 1 //sometimes we get "1" when the device is on low
+
+	// The GE, Honeywell, and Leviton treat 33 as medium, so account for that
+	if (cmd.value == 33) {
```
tpmanley: What other values do they report? I just wonder if it would be better to change the code above to do `Math.ceil((cmd.value + 1)/33) as int`

greens: Have never been able to get a consistent answer about that, hence all the changes to this DTH

greens (dismissed):

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
dkirker: 1-32 = 1,
33-66 = 2,
67 - 99 or 100 = 3

Bashi realized that calling setLevel could be used as an indicator as to what that value means to the bulb.

dkirker: I suspect the firmware does a check on the values and not some math.

if 1 <= i <= 32

else if 33 <= i <= 66

else

🤷‍♂

tpmanley: Personally I think the range checks would be preferable over the use of ceil and the special check for 33

dkirker: @tpmanley I am going to give this a whirl today. Made some updates.

tpmanley (approved):

greens (approved):

kianooshkarami (approved):

dkirker:
@tpmanley I had pointed this at staging in anticipation to have it in prod by now. We should probably merge this tomorrow after the deploy. I am doing a test of the code today to verify my recent change.

dkirker:
@tpmanley  Verified yesterday. Since the deploy happened today we can merge this so it is in staging. Could you please merge?

PR #4186, opened 2019-04-22
ICP-9241: ezex smart submeter not setting checkInterval

FenMei007S:
@tpmanley , plz review it.

### DIFF
--- a/devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
+++ b/devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
@@ -18,6 +18,7 @@ metadata {
         capability "Refresh"
         capability "Health Check"
         capability "Sensor"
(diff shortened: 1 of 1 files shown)


tpmanley (dismissed):

tpmanley:
Has this already been QA tested? If we merge to `staging` today it will go to PRD tomorrow

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -50,6 +51,14 @@ def parse(String description) {
     def event = zigbee.getEvent(description)
     if (event) {
         log.info event
+        if (event.name== "power") {
```
tpmanley: nit: should have a space before the `==` here and below

greens (approved):

PR #4195, opened 2019-04-24
[WWST-2442] Added fingerprint for Qubino Smart Plug

PKacprowiczS:
@greens @dkirker @MMateusiakS @KKlimczukS @MGoralczykS @ZWozniakS @DCzarneckaS 
Since Diana is on a sick leave, I've jumped into her place and added the fingerprint

### DIFF
--- a/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
+++ b/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
@@ -43,7 +43,7 @@ metadata {
 		//zw:L type:1001 mfr:0258 prod:0003 model:1087 ver:3.94 zwv:4.05 lib:03 cc:5E,72,86,85,59,5A,73,70,25,27,71,32,20 role:05 ff:8700 ui:8700
 		fingerprint mfr: "0258", prod: "0003", model: "1087", deviceJoinName: "NEO Coolcam Power Plug"  //EU
 		fingerprint mfr: "027A", prod: "0101", model: "000D", deviceJoinName: "Zooz Power Switch"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
```diff
@@ -43,6 +43,7 @@ metadata {
 		//zw:L type:1001 mfr:0258 prod:0003 model:1087 ver:3.94 zwv:4.05 lib:03 cc:5E,72,86,85,59,5A,73,70,25,27,71,32,20 role:05 ff:8700 ui:8700
 		fingerprint mfr: "0258", prod: "0003", model: "1087", deviceJoinName: "NEO Coolcam Power Plug"  //EU
 		fingerprint mfr: "027A", prod: "0101", model: "000D", deviceJoinName: "Zooz Power Switch"
+		fingerprint mfr: "0159", prod: "0002", model: "0054", deviceJoinName: "Qubino Smart Plug", mnmm: "SmartThings", vid: "generic-switch-power"
```
greens: The custom metadata is because the device only supports power and not energy?

PKacprowiczS: Well, I was told to just paste fingerprint from Diana's PR to another DTH, so I didn't check it.
However, as you had asked, I've checked it, and apparently device does report energy consumption. Quote from documentation: "The energy consumption (kWh) is reported regardless of the values, set in the parameters 40 and 42. The energy consumption will be reported, when it increases for at least 0,1 kWh."

greens: lol

thanks for checking

greens (approved):

PR #4200, opened 2019-04-25
WWST-2306 IKEA TRADFRI wireless dimmer

dkirker:


devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -23,6 +23,8 @@ metadata {
 		capability "Configuration"
 
 		fingerprint profileId: "0104", inClusters: "0000,1000,0003", outClusters: "0003,0004,0005,0006,0008,1000,0019", manufacturer: "Aurora", model: "Remote50AU", deviceJoinName: "Aurora Wireless Wall Remote"
+		fingerprint endpointId: "01", profileId: "0104", deviceId: "0810", inClusters: "0000, 0001, 0003, 0009, 0B05, 1000", outClusters: "0003, 0004, 0006, 0008, 0019, 1000", manufacturer: "IKEA of Sweden", model: "TRADFRI wireless dimmer", deviceJoinName: "IKEA TRÅDFRI Wireless dimmer"
```
tpmanley: Might be a good time to simplify to just a single fingerprint with `fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI wireless dimmer", deviceJoinName: "IKEA TRÅDFRI Wireless dimmer"`

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -56,39 +72,57 @@ def parse(String description) {
...
+			} else if (descMap.commandInt == LEVEL_MOVE_COMMAND) {
 				sendEvent(name: "level", value: descMap.data[0] == "00" ? 100 : STEP)
 				sendEvent(name: "switch", value: "on" )
+
 				log.debug "step to ${descMap.data}"
-			} else if (descMap.commandInt == 0x03) {
+			} else if (descMap.commandInt == LEVEL_STOP_COMMAND) {
 				log.debug "stop move"
-			}
+			} else if (descMap.commandInt == LEVEL_MOVE_LEVEL_ONOFF_COMMAND) {
```
tpmanley: When does it send this command?

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: Done!

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: This seems to be what it sends for the on and off events.

greens (dismissed):

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -44,62 +45,121 @@ metadata {
...
 		else {
-			sendEvent(event)
+			results << createEvent(event)
 		}
 	} else {
 		def descMap = zigbee.parseDescriptionAsMap(description)
-		if (descMap && descMap.clusterInt == 0x0006) {
-			if (descMap.commandInt == 0x01 || descMap.commandInt == 0x00) {
+
+		if (descMap && descMap.clusterInt == zigbee.ONOFF_CLUSTER) {
```
tpmanley: Could you add some comments describing the messages each of the different devices use and how they map to capabilities? Some of these changes don't seem to make sense but I think it would be a lot easier to understand with some additional context.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -44,62 +45,121 @@ metadata {
...
-				sendEvent(name: "button", value: "pushed", data: [buttonNumber: 1], isStateChange: true)
-			} else if (descMap.commandInt == 0x04) {
-				sendEvent(name: "button", value: "held", data: [buttonNumber: 1], isStateChange: true)
+			} else if (descMap.commandInt == LEVEL_MOVE_LEVEL_ONOFF_COMMAND) {
+				if (descMap.data[0] == "00") {
+					results << createEvent(name: "switch", value: "off", isStateChange: true)
+				} else if (descMap.data[0] == "FF") {
+					results << createEvent(name: "switch", value: "on", isStateChange: true)
+				} /*else {
+					results << createEvent(name: "switch", value: "on", isStateChange: true)
```
tpmanley: This should be updated or removed

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: @tpmanley Done! Let me know if you think a few more are needed? I didn't add a comment for the ONOFF logic since I am not sure how the Aurora remote handles it (I assume it only sends one type of packet and thus only supports a "toggle" action?), but set the default action to be to create the event based off of the command the device sends us.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: Fixed the code up!

dkirker:
Added some code in anticipation of https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4275

I'll probably need to rebase this after it is merged.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -42,64 +43,140 @@ metadata {
...
+def getLEVEL_STEP_COMMAND() { 0x02 }
+def getLEVEL_STOP_COMMAND() { 0x03 }
+def getLEVEL_MOVE_LEVEL_ONOFF_COMMAND() { 0x04 }
+def getLEVEL_MOVE_ONOFF_COMMAND() { 0x05 }
+def getLEVEL_STEP_ONOFF_COMMAND() { 0x06 }
+def getLEVEL_STOP_ONOFF_COMMAND() { 0x07 }
+def getLEVEL_DIRECTION_UP() { "00" }
+def getLEVEL_DIRECTION_DOWN() { "01" }
+
+def getUINT8_STR() { "20" }
```
tpmanley, line 66: Rather than defining this I think it'd be better to just convert `DataType.UINT8` to a hex string at the place you need to use it

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -42,64 +43,140 @@ metadata {
...
 		else {
-			sendEvent(event)
+			results << createEvent(event)
 		}
 	} else {
 		def descMap = zigbee.parseDescriptionAsMap(description)
-		if (descMap && descMap.clusterInt == 0x0006) {
-			if (descMap.commandInt == 0x01 || descMap.commandInt == 0x00) {
+
+		if (descMap && descMap.clusterInt == zigbee.ONOFF_CLUSTER) {
```
tpmanley: You can remove the check of `descMap` since it will always be a map and simplify the if conditionals to just `descMap.clusterInt == zigbee.ONOFF_CLUSTER`

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
tpmanley: I think it's still a bit hard to understand. What do you think about breaking this into separate functions for each model? There might be a little bit of duplicated code but I think it will be a lot more readable and easier to support/extend.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -42,64 +43,140 @@ metadata {
...
+			} else if (isIkeaDimmer() && (descMap.commandInt == LEVEL_MOVE_COMMAND || descMap.commandInt == LEVEL_MOVE_ONOFF_COMMAND)) {
+				// For the IKEA Dimmer we are going to treat Level Move and Level Move with On/Off as Level Step
+				results += handleStepEvent(descMap)
+			} else if (descMap.commandInt == LEVEL_MOVE_COMMAND) {
+				results << createEvent(name: "level", value: descMap.data[0] == LEVEL_DIRECTION_UP ? 100 : STEP)
+				results << createEvent(name: "switch", value: "on")
+
+				log.debug "step to ${descMap.data}"
+			} else if (descMap.commandInt == LEVEL_STOP_COMMAND || descMap.commandInt == LEVEL_STOP_ONOFF_COMMAND) {
+				log.debug "stop move"
```
tpmanley: If we don't need to create an event in this case I think it would be good to explicitly state that in a comment

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -42,64 +43,140 @@ metadata {
...
 def parse(String description) {
 	log.debug "description is $description"
+	def results = []
 
 	def event = zigbee.getEvent(description)
 	if (event) {
 		if (event.name=="level" && event.value==0) {}
 		else {
-			sendEvent(event)
+			results << createEvent(event)
```
tpmanley, line 88: Does `zigbee.getEvent` create events for any of the devices this DTH supports? It seems like it might not

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -23,6 +23,7 @@ metadata {
 		capability "Configuration"
 
 		fingerprint profileId: "0104", inClusters: "0000,1000,0003", outClusters: "0003,0004,0005,0006,0008,1000,0019", manufacturer: "Aurora", model: "Remote50AU", deviceJoinName: "Aurora Wireless Wall Remote"
+		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI wireless dimmer", deviceJoinName: "IKEA TRÅDFRI Wireless dimmer"
```
tpmanley: I'm realizing that although the fingerprints are a lot shorter, it was useful to see the cluster list. Thoughts on including it as a comment at the end of or above the fingerprint?

greens: if we're going to have the whole thing anyway we might as well just include it in the fingerprint.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: Hmmmmm I think I'd do that if I were getting dynamic values, but since it is only this one value (which the spec isn't going to change) I feel like it would be computationally simpler to leave it as a static declaration. It just seems like adding `String.format("%02X", value)` throws more complexity in to it. :/

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: I re-organized some code that hopefully makes it a little easier. I feel like the duplicate code would be too great if I handled it per device. There are really only a couple of areas where the implementations differ.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: I don't believe that it does. I'd need to review zigbee.getEvent to see.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -23,6 +23,7 @@ metadata {
 		capability "Configuration"
 
 		fingerprint profileId: "0104", inClusters: "0000,1000,0003", outClusters: "0003,0004,0005,0006,0008,1000,0019", manufacturer: "Aurora", model: "Remote50AU", deviceJoinName: "Aurora Wireless Wall Remote"
+		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI wireless dimmer", deviceJoinName: "IKEA TRÅDFRI Wireless dimmer" // 01 [0104 or C05E] 0810 02 06 0000 0001 0003 0009 0B05 1000 06 0003 0004 0006 0008 0019 1000
```
tpmanley, line 26: First one I've seen in the wild using this format. I think I like this format. What do you guys think? Still OK with this?
@dkirker @greens @varzac .

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
tpmanley: 🤷‍♂ I don't care strongly. It's used in one place so a single `String.format` doesn't seem like a big deal, but I'm fine either way

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -43,64 +44,159 @@ metadata {
...
+		results += handleStepEvent(descMap)
+	} else if (descMap.commandInt == LEVEL_MOVE_COMMAND || descMap.commandInt == LEVEL_MOVE_ONOFF_COMMAND) {
+		// For the IKEA Dimmer we are going to treat Level Move and Level Move with On/Off as Level Step
+		if (isIkeaDimmer()) {
+			results += handleStepEvent(descMap)
+		} else {
+			results << createEvent(name: "level", value: descMap.data[0] == LEVEL_DIRECTION_UP ? 100 : STEP)
+			results << createEvent(name: "switch", value: "on")
+
+			log.debug "step to ${descMap.data}"
```
tpmanley: Not clear that this log message is useful

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -43,64 +44,159 @@ metadata {
...
+
+def handleStepEvent(descMap) {
+	def results = []
+	def currentLevel = device.currentValue("level") as Integer ?: 0
+	def value = null
+
+	log.debug "move to ${descMap.data}"
+
+	if (descMap.data[0] == LEVEL_DIRECTION_UP) {
+		log.debug "move up"
```
tpmanley: These log messages would be more useful if they including some data, like `STEP` and the new `value`

tpmanley (dismissed):
Couple of log message comments but otherwise looks good

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: I kinda want to clean the comment up a little bit. Something more like:

`// 01 [0104 or C05E] 0810 02 in: 0000 0001 0003 0009 0B05 1000 out: 0003 0004 0006 0008 0019 1000`

Since I'm not a computer I don't really care too much about the length of each array of clusters; replacing the lengths with keys makes it a little more readable (otherwise, the hex values seem to blur together).

But having the data can be useful.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
dkirker: Cool. I'll leave it static, for now. I'd like to think that I'm doing my part shaving off some CPU processes which would cut down CPU load a little. 😮

dkirker:
@tpmanley Alright, how does that look?

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -43,64 +44,160 @@ metadata {
...
+	log.debug "handleStepEvent - ${descMap.data}"
+
+	if (descMap.data[0] == LEVEL_DIRECTION_UP) {
+		value = Math.min(currentLevel + STEP, 100)
+	} else if (descMap.data[0] == LEVEL_DIRECTION_DOWN) {
+		value = Math.max(currentLevel - STEP, 0)
+	}
+
+	if (value != null) {
+		log.debug "Step ${descMap.data[0] == LEVEL_DIRECTION_UP ? "up" : "down"} by $STEP to $value"
```
tpmanley, line 187: Could move this up to the if/else if and then you'd know it was up or down already

tpmanley (approved):

PR #4205, opened 2019-04-26
[ICP-9238] Fix for configure reporting method

MMateusiakS:
@greens missing zigbee class


tpmanley:
I think this PR should be against staging so it can go to prod next week since it's definitely a bug and a simple fix.

tpmanley (dismissed):

greens:
@tpmanley re-targeted to staging after a fun learning experience with git

greens (approved):

tpmanley (approved):
At least you had fun 😁

PR #4207, opened 2019-04-26
[ICP-8219] Zigbee Plugin Motion Sensor: Removing failing call to refresh from installed

tpmanley:
At the time "installed" is called the device isn't sufficiently initialized
so it's not possible to send the device Zigbee commands. If Zigbee commands
are attempted like they were before this change they will cause an exception
which will prevent the device from being initialized correctly.

https://smartthings.atlassian.net/browse/ICP-8219


tpmanley:
cc @FenMei007S

dkirker (approved):

PR #4218, opened 2019-05-01
DVCSMP-4734 Locally running Fibaro DTHs fail due to failure to overri…

eddysun1990:
…de runIn() for locally devices

* Scheduled event will not occur in runIn() in locally running devices, unless it is overridden by "forceForLocallyExecuting" so the scheduled event occurs.
* Added 'overwrite' inside runIn()

https://smartthings.atlassian.net/browse/DVCSMP-4734


kianooshkarami (approved):

greens:
Needs a rebase since @kianooshST changed the base branch

PR #4220, opened 2019-05-01
DVCSMP-4734 Locally Running Fibaro DTHs Failure

eddysun1990:
* Scheduled event will not occur in runIn() in locally running devices, unless it is overridden by "forceForLocallyExecuting" so the scheduled event occurs.
* Added 'overwrite' inside runIn()


kianooshkarami (approved):

greens (approved):

tpmanley (approved):

tpmanley:
I'll deploy this once it's passed QA

dkirker (commented):
I'm late to the game but 👍 looks good. `overwrite` is  `true` by default (at least according to the docs, and possibly the code the last time that I saw it), but it doesn't hurt to include it.

PR #4226, opened 2019-05-03
[CHAD-1813] Arrival Sensor HA: Make the DTH run locally

tpmanley:
The DTH is now marked to run locally on 0.18.20+. When it moves from cloud
execution to local execution `updated` will be called which will stop the
scheduled presence check running if it is currently running.

Also fixed an issue where `state.lastCheckin` is not set (presumably
because the device never sent us a message). Currently it raises an exception
but with the change it should just create a 'not present' event and then
stop the scheduled check.

https://smartthings.atlassian.net/browse/CHAD-1813


tpmanley:
The reason we had to revert this change last week is addressed in DM in staging so I think this can go in for next week.

varzac (approved):

greens (approved):

PR #4231, opened 2019-05-03
WWST-4222 Early fw versions of fibaro dimmer 2 do not support param 38

greens:
Reports from users indicate v3.03 does not have support for this param, causing sync to never complete.

### DIFF
--- a/devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
+++ b/devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
@@ -173,6 +173,7 @@ def syncStart() {
 			// this parameter (38) is not supported on some earlier firmware versions, so we'll mark it as already synced
 			if ("$it.key" == "levelCorrection" && (zwaveInfo.ver as float) <= REDUCED_CONFIGURATION_VERSION) {
 				state."$it.key".state = "synced"
+                syncNeeded = settings.size() > 1
 			}
 		}
 	}
--- a/devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
+++ b/devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
@@ -166,8 +166,7 @@ def syncStart() {
 		if(settings."$it.key" != null) {
 			if (state."$it.key" == null) { state."$it.key" = [value: null, state: "synced"] }
 			// this parameter (38) is not supported on some earlier firmware versions, so we'll mark it as already synced
-			if (!zwaveInfo.ver ||
-				("$it.key" == "levelCorrection" && (zwaveInfo.ver as float) <= REDUCED_CONFIGURATION_VERSION)) {
+			if ("$it.key" == "levelCorrection" && (!zwaveInfo.ver || (zwaveInfo.ver as float) <= REDUCED_CONFIGURATION_VERSION)) {
 				state."$it.key".state = "synced"
 			} else if (state."$it.key".value != settings."$it.key" as Integer || state."$it.key".state in ["notSynced","inProgress"]) {
 				state."$it.key".value = settings."$it.key" as Integer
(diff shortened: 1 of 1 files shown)


greens:
This will probably need to wait until the prod hotfix is rolled to master to avoid conflicts.

tpmanley:
I just rolled `staging` down to `master` (I had previously rolled `production` down to `staging`) so you should be able to rebase

devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
```diff
@@ -170,6 +170,10 @@ def syncStart() {
 				state."$it.key".state = "notSynced"
 				syncNeeded = true
 			}
+			// this parameter (38) is not supported on some earlier firmware versions, so we'll mark it as already synced
+			if ("$it.key" == "levelCorrection" && (zwaveInfo.ver as float) <= REDUCED_CONFIGURATION_VERSION) {
```
tpmanley: This could result in `syncNeeded` being incorrectly set to true if `levelCorrection` was the only remaining parameter that required sync. I suggest doing this check up above so `syncNeeded` is only set to `true` if the parameter needs to be synced.

greens:
Should be good to go now.

devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
tpmanley: This still seems odd. Why not just move this up into the loop above and only set `syncNeeded` if sync is needed?

greens: I guess I'm misunderstanding what you're asking? There's no loop above this. Did you mean the if statement above? If so, this seems like the more prudent place to put it since it only adds one extra comparison.

tpmanley: Sorry, I was reading the code wrong and making incorrect comments. What I suggest is moving this new `if` statement to above the `if (state."$it.key".value != settings."$it.key" as Integer || state."$it.key".state in ["notSynced","inProgress"])` and then turn that into an `else if`.

greens: Gotcha. Fixed now. And Fibaro got back to us with the correct fw version number.

tpmanley: Changes look good. Is there any risk that `zwaveInfo.ver` won't be set? Just wondering if we should handle that being null and assume the FW version does not support this parameter.

greens: Don't expect it will ever be the case, but it can't hurt. Null check added.

devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy
```diff
@@ -165,7 +165,11 @@ def syncStart() {
 	parameterMap().each {
 		if(settings."$it.key" != null) {
 			if (state."$it.key" == null) { state."$it.key" = [value: null, state: "synced"] }
-			if (state."$it.key".value != settings."$it.key" as Integer || state."$it.key".state in ["notSynced","inProgress"]) {
+			// this parameter (38) is not supported on some earlier firmware versions, so we'll mark it as already synced
+			if (!zwaveInfo.ver ||
```
tpmanley: I think we want `if ("$it.key" == "levelCorrection" && (!zwaveInfo.ver || (zwaveInfo.ver as float) <= REDUCED_CONFIGURATION_VERSION))` since AFAIK the other settings can be synced with any FW version

greens: very true. good catch

tpmanley (approved):

PR #4246, opened 2019-05-07
ICP-9233 Proper null check for Stelpro Maestro temp alarm handling

dkirker:


dkirker:
Currently verifying fix -- but I suspect this is it.

dkirker:
Poland has indicated that the issue is gone with this fix.

dkirker:
@tpmanley If we're fine using Poland to do the verification in staging, I can rebase this to staging so that it goes out on Tuesday.

tpmanley (dismissed):
I'm fine with targeting staging

dkirker:
@tpmanley Rebased!

tpmanley (approved):

PR #4247, opened 2019-05-08
[DARC-996] Changing isComponent flag to true for multi switch

PKacprowiczS:
@greens @MMateusiakS


greens:
You also need to add `mcdSync: true` to the definition of the DTH and the metadata must be updated. @dkirker is the best resource for help with that.

Note that you'll need device-specific metadata for a 2-switch device vs a 4-switch device when using multi-components

tpmanley:
The plan is to not make this DTH an MCD device so I'm going to close this PR

PR #4249, opened 2019-05-08
WWST-4050 Eaton 5-Scene Keypad switch to multi-component

greens:
set multi-component flag and device should now point to custom metadata


greens:
this is verified by eng

dkirker (approved):

PR #4278, opened 2019-05-16
ICP-8376 Add supportedButtonValues to Z-Wave Button

dkirker:


dkirker:
A mobile client bug needs to be addressed for this to work. However, it should be better than just forcing an incomplete solution with a specific fingerprint vid.

tpmanley (approved):

greens (approved):

PR #4279, opened 2019-05-16
[ICP-9547-2] Replacing not used method.

ZWozniakS:
@greens @dkirker can you please review? sorry I didn't notice one extra use of method


dkirker (approved):

dkirker:
D'oh! Thanks!

PR #4283, opened 2019-05-17
[ICP-9570] Handle set level zero value

MMateusiakS:
@greens to keep consistency in set level capability


devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -213,7 +213,11 @@ def setGenericName(value){
 }
 
 def setLevel(value, rate = null) {
-	zigbee.setLevel(value)
+	if (value == 0) {
```
tpmanley, line 216: We actually take care of this inside `zigbee.setLevel` so you don't need to do it here

MMateusiakS: I hurried up with PR:)

PR #4284, opened 2019-05-17
WWST-4245 Add default zwave command handler to DTHs missing it

greens:
This should severely reduce the number of missingMethodExceptions we see in the logs.


greens:
I got this list via 

`grep -L "zwaveEvent(physicalgraph.zwave.Command" $(grep -lr "physicalgraph.zwave" .)`

let me know if you think I missed any

kianooshkarami (commented):
I would add a log statement about the command received, thoughts?

greens:
Trading in exceptions for log statements. I don't feel particularly strongly about it either way.

kianooshkarami:
I was suggesting the logging on the principle that I would like to have the ability to view commands that are not processed by the DTH, I understand the newly added function solves the missing method exception, I also would like to know if we need to add missing handlers for some devices :)

greens:
updated

dkirker (approved):

PR #4291, opened 2019-05-20
[ICP-9587] Changed deviceJoinName and deleted doubled fingerprint

ZWozniakS:
@greens @MMateusiakS There is a problem with two Muller's Licht bulbs: Tint GU10 Spot White, Tint Candle White which shares the same DTH and the same fingerprint as well. The same fingerprints are making them indistinguishable and that is why both bulbs are pairing with the first fingerprint we've made (with the deviceJoinName for GU10 Spot White). Muller Licht did not give different model names in fingerprints for these bulbs so our solution, for now, is to leave only one fingerprint for both bulbs and give them one universal name "Muller Licht Tint White Bulb". What do you think about this? Maybe we should contact with Muller Licht and ask if they don't want to update model names in those fingerprints?


greens (approved):
This is the best solution. I doubt the manufacturer would issue a change.

PR #4295, opened 2019-05-21
[WWST-3670] Added fingerprint for Aurora Smart Double Socket and edited configure()

ZWozniakS:
Could you please review? @greens @MMateusiakS

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
+++ b/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
@@ -28,6 +28,7 @@ metadata {
 		command "childOff", ["string"]
 
 		fingerprint profileId: "0104", inClusters: "0000, 0005, 0004, 0006", outClusters: "0000", manufacturer: "ORVIBO", model: "074b3ffba5a045b7afd94c47079dd553", deviceJoinName: "Switch 1"
+		fingerprint profileId: "0104", inClusters: "0006, 0005, 0004, 0000, 0003, 0B04, 0008", outClusters: "0019", manufacturer: "Aurora", model: "DoubleSocket50AU", deviceJoinName: "Aurora Smart Double Socket 1"
 	}
 	// simulator metadata
 	simulator {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -163,6 +164,21 @@ def configureHealthCheck() {
...
+		def cmd = []
+		cmd << childOff("${device.deviceNetworkId}:02}")
+		cmd << off()
+
+		cmd.each { sendHubCommand(it) }
+	}
 }
+
+private Boolean isOrvibo() {
+	device.getDataValue("model") == "074b3ffba5a045b7afd94c47079dd553"
```
greens, line 182: might be more legible to change this to `isAurora()`

greens:
I think that while you're working on it, it might be best to do a general cleanup of this DTH. There are a lot of magic numbers (actually magic strings) that should be replaced with `zigbee.CLUSTER_ON_OFF`

Also why do you turn the aurora off on install?

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
KKlimczukS: I would leave isOrvibo() - since it's a bit unusual device behaviour -> 
`//the orvibo switch will send out device anounce message at ervery 2 mins as heart beat,setting 0x0099 to 1 will disable it.
return zigbee.writeAttribute(0x0000, 0x0099, 0x20, 0x01, [mfgCode: 0x0000])`

in `else` statement would be: `return response(refresh())` for this Aurora (and probably other devices that would be supported by this DTH in the future) .
@greens What do You think about it Steven ?

KKlimczukS:
> I think that while you're working on it, it might be best to do a general cleanup of this DTH. There are a lot of magic numbers (actually magic strings) that should be replaced with `zigbee.CLUSTER_ON_OFF`
> 
> Also why do you turn the aurora off on install?

Yup, we will do some refactoring here. All changes will be in another pull request soon -> 
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/4323

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
greens: sure

PR #4300, opened 2019-05-22
[ICP-9427] Handle IAS message reporting

MMateusiakS:
@greens handle IAS message

### DIFF
--- a/devicetypes/smartthings/zigbee-lock-without-codes.src/zigbee-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zigbee-lock-without-codes.src/zigbee-lock-without-codes.groovy
@@ -160,8 +160,10 @@ def unlock() {
 def parse(String description) {
 	def result = null
 	if (description) {
-		if (description.startsWith('read attr -')) {
+		if (description?.startsWith('read attr -')) {
 			result = parseAttributeResponse(description)
+		} else if (description?.startsWith('zone report')) {
+			result = parseIasMessage(description)
 		} else {
 			result = parseCommandResponse(description)
 		}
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-lock-without-codes.src/zigbee-lock-without-codes.groovy
```diff
@@ -212,6 +214,12 @@ private def parseAttributeResponse(String description) {
 	return result
 }
 
+private def parseIasMessage(String description) {
+	ZoneStatus zs = zigbee.parseZoneStatus(description)
+	def responseMap = [ name: "battery", value: zs.isBatterySet() ? 5 : 50, isStateChange: true]
```
greens: why are we forcing `isStateChange`?

devicetypes/smartthings/zigbee-lock-without-codes.src/zigbee-lock-without-codes.groovy
```diff
@@ -160,8 +160,10 @@ def unlock() {
 def parse(String description) {
 	def result = null
 	if (description) {
-		if (description.startsWith('read attr -')) {
+		if (description?.startsWith('read attr -')) {
 			result = parseAttributeResponse(description)
+		} else if (description?.startsWith('zone report')) {
```
lukhong, line 165: Just one question. Isn't there any chance to come 'zone status'?
Then I think it will miss handling it.

MMateusiakS: I haven't notice 'zone status' what I was in logs was only 'zone report'

devicetypes/smartthings/zigbee-lock-without-codes.src/zigbee-lock-without-codes.groovy
MMateusiakS: to see that I got event about battery state

greens: But we wouldn't want it typically. Filling the event logs with battery 50% doesn't fix anything, does it?

MMateusiakS: I accept your argument. Let's go ahead and merge PR.

greens (approved):

PR #4324, opened 2019-05-30
[WWST-4403] Add HomeSeer Water Leak Sensor fingerprint

z-michel:
Added fingerprint for HomeSeer LS-100+ Water Leak Sensor


kianooshkarami (dismissed):

z-michel:
Updated PR and commit title to WWST JIRA

greens (dismissed):

kianooshkarami:
HomeSeer's temperature reporting is disabled by default and to enable temperature reporting we would have to perhaps have device specific DTH code. Personally, I would like to see this device to have its own DTH, but I think this change is a great start.

greens:
We have a z-wave water/temp/light sensor DTH, but nothing for just water and temp.

kianooshkarami (approved):

PR #4412, opened 2019-06-12
ICP-9647 OneApp expecting level events after setLevels, even if value…

greens:
… does not change

OneApp will timeout waiting for a level event that won't occur. This is fine for physical devices since we poll them after sets anyway, but for these emulated devices we'll need to fake it to match.

### DIFF
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -81,7 +81,7 @@ def parse(String description) {
 					if (value == 0) {
 						sendEvent(name: "switch", value: "off")
 					} else {
-						sendEven t(name: "level", value: value)
+						sendEvent(name: "level", value: value)
 					}
 				}
 			} else if (descMap.commandInt == 0x01) {
--- a/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
@@ -120,7 +120,11 @@ def setLevel(value, rate = null) {
 	} else {
 		sendEvent(name: "switch", value: "on")
 	}
-	sendEvent(name: "level", value: value)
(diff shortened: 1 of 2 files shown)


tpmanley (dismissed):

dkirker:
Looks good, but this will conflict with the code that is in staging.

dkirker:
Also, does this actually work if the level event is not a `isStateChange: true` type thing? I had been told that an event that wasn't a state change would get dropped through the pipeline to OCF. (Maybe that has changed. We had issues with lock:unlocked -> lock() -> error locking -> lock:unlocked resulting in a "network error" in the client.)

greens:
Entirely possible. I've asked SR-Pol to test.

dkirker:
That cause for that delay seems odd... 

Also, I realized this shouldn't conflict with the PR that was merged in staging. But, I did realize, in the Battery Accessory Dimmer  the Sengled event handling has the possibility of turning the switch on with a brightness of 0. Case 1 and case 6 just send on events (in handleSengledSwitchEvents in staging) while the handleSwitchEvent code (in staging) does a check to see if level is 0 and sets it to a default when an on event comes in. I hadn't actually caught this when doing the IKEA dimmer work as we don't have all of the switches in Mountain View. :-S

greens:
This should be current now.

devicetypes/smartthings/zigbee-accessory-dimmer.src/zigbee-accessory-dimmer.groovy
```diff
@@ -115,10 +115,16 @@ def on() {
 def setLevel(value, rate = null) {
 	if (value == 0) {
 		sendEvent(name: "switch", value: "off")
+		// OneApp expects a level event when the dimmer value is changed
+		value = device.currentValue("level")
 	} else {
 		sendEvent(name: "switch", value: "on")
-		sendEvent(name: "level", value: value)
 	}
+	runIn(1, delayedSend, [data: createEvent(name: "level", value: value), overwrite: true])
```
tpmanley, line 123: Remind me why can't the event be sent right away?

greens: The "setLevel" command makes it up to the cloud _after_ the level event otherwise, so OneApp doesn't associate them and satisfy its own response requirements.

tpmanley (approved):

PR #4198, opened 2019-04-25
[WWST-4011] Fingerprint for the Sengled Smart LED Motion Sensor PAR38 Bulb

ganeshasprabhu:
DTH for WWST-4011 for Sengled-PAR38-Light
git username  : ganeshasprabhu

PR #5043, opened 2019-06-15
Revert #4198 "[WWST-4011] Fingerprint for the Sengled Smart LED Motion Sensor PAR38 Bulb" (f49ba79)

dkirker:


dkirker:
See discussion on ICP-9360

tpmanley (approved):

PR #5053, opened 2019-06-15
[DVCSMP-4768] First pass as normalizing and simplifying device preferences

tpmanley:
I left this as separate commits to make it easier to review. The intent is to address some inconsistencies between similar device preferences and remove stuff that I don't think is necessary. Fewer inconsistencies and unnecessary preferences means less that needs to get translated.

### DIFF
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -108,7 +108,7 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
+			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature by this many degrees", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
 		}
 		section {
 			input title: "Associated Devices", description: "The below preferences control associated node 2 devices.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -108,7 +108,6 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input title: "Temperature Offset", description: "This feature allows you to correct any temperature variations by selecting an offset. Ex: If your sensor consistently reports a temp that's 5 degrees too warm, you'd enter '-5'. If 3 degrees too cold, enter '+3'.", displayDuringSetup: true, type: "paragraph", element: "paragraph"
 			input "TempAdj", "number", title: "Degrees", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
 		}
 		section {
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -108,7 +108,7 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input "TempAdj", "number", title: "Degrees", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
+			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
 		}
 		section {
 			input title: "Associated Devices", description: "The below preferences control associated node 2 devices.", displayDuringSetup: false, type: "paragraph", element: "paragraph"
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -108,7 +108,7 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature by this many degrees", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
+			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
 		}
 		section {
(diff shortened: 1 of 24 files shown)


devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -108,8 +108,7 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input title: "Temperature Offset", description: "This feature allows you to correct any temperature variations by selecting an offset. Ex: If your sensor consistently reports a temp that's 5 degrees too warm, you'd enter '-5'. If 3 degrees too cold, enter '+3'.", displayDuringSetup: true, type: "paragraph", element: "paragraph"
-			input "TempAdj", "number", title: "Degrees", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
+			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature by this many degrees", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
```
varzac: all the other ones use `tempOffset` instead of `TempAdj` do we want to make this match?

tpmanley: I thought about it but that change would mean I'd have to also run a script to call `updated` on all existing devices which didn't seem worth the effort.

varzac (dismissed):

devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
dkirker: Actually, this one might be worth holding off on. At least, I'd probably remove the description as that seems like it is not accurate for what the device preference actually is (I might have added that and not fully updated it). It is possible that the offset here would need to be converted if we were switching this to degrees, since it sounds like the preference is for tenths of a degree... (Unless you already investigated this and it checks out.)

dkirker:
I left a comment on the other PR about removing the description text, since that seems useful. But, I am not sure if there will be more to improve that later?

dkirker (commented):
I left a comment on the other PR about removing the description text, since that seems useful. But, I am not sure if there will be more to improve that later?

devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
tpmanley: Good catch. Will fix.

tpmanley: Pushed a fix in ac4442611d2f599819f8e082f7edaa4973788a12

devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -108,8 +108,7 @@ metadata {
 			input "TempMin", "number", title: "Temperature Report Frequency", description: "Temperature report sent every N minutes [0-127]", range: "0..127", defaultValue: 6, displayDuringSetup: true, required: false
 		}
 		section {
-			input title: "Temperature Offset", description: "This feature allows you to correct any temperature variations by selecting an offset. Ex: If your sensor consistently reports a temp that's 5 degrees too warm, you'd enter '-5'. If 3 degrees too cold, enter '+3'.", displayDuringSetup: true, type: "paragraph", element: "paragraph"
-			input "TempAdj", "number", title: "Degrees", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
+			input "TempAdj", "number", title: "Temperature Offset", description: "Adjust temperature up/down N tenths of a degree F [(-127)-(+128)]", range: "-127..128", defaultValue: 0, displayDuringSetup: true, required: false
```
greens, line 111: so this one actually does something distinct?

tpmanley: Yeah, the offset is in 10ths of degree so updating it to be consistent with other DTHs would be more complex of a change so was just going to leave it as is.

greens (dismissed):

greens:
I think this is as consolidated as I was able to achieve. Ready for review.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
```diff
@@ -46,7 +46,7 @@ metadata {
 	}
 
 	preferences {
-		input "ledIndicator", "enum", title: "LED Indicator", description: "Turn LED indicator... ", required: false, options:["on": "When On", "off": "When Off", "never": "Never"], defaultValue: "off"
+		input "ledIndicator", "enum", title: "LED Indicator", description: "Turn LED indicator on... ", required: false, options:["on": "When On", "off": "When Off", "never": "Never"], defaultValue: "off"
```
tpmanley, line 49: I don't think we want the extra ` on` do we?

devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy
```diff
@@ -77,7 +77,7 @@ metadata {
         }
 
         preferences {
-            input "preset", "number", title: "Default half-open position (1-100)", defaultValue: 50, required: false, displayDuringSetup: false
+            input "preset", "number", title: "Preset position", defaultValue: 50, range: "1..100", required: false, displayDuringSetup: false
```
tpmanley, line 80: I think `position` should be capitalized since it's a title

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -68,11 +68,11 @@ metadata {
 
 		// Yale siren only
 		preferences {
-			input name: "alarmLength", type: "number", title: "Alarm length (1-10 min)", range: "1..10"
+			input name: "alarmLength", type: "number", title: "Alarm length", range: "1..10"
```
tpmanley, line 71: `Length` should be capitalized. Also maybe change it to `Duration`? Finally, might want to add a description and include that the units is minutes.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -68,11 +68,11 @@ metadata {
...
 			// defaultValue: 10
 			input name: "alarmLEDflash", type: "bool", title: "Alarm LED flash"
 			// defaultValue: false
-			input name: "comfortLED", type: "number", title: "Comfort LED (0-25 x 10 sec.)", range: "0..25"
+			input name: "comfortLED", type: "number", title: "Comfort LED (x10 sec.)", range: "0..25"
```
tpmanley, line 75: What does this do? We're removing units from the title in most other cases so might make sense to do it here too and add a description that both clarifies what this is and indicates the units.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
greens: I may have misunderstood, but I think it previously would have read as "Turn LED indicator...[when on]"

Whereas with this fix it should be "Turn LED indicator on...[when on]" which I think more accurately explains what it does

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
greens: I'm honestly not sure what it does.

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
greens: I think these have already been sent of for translation.

devicetypes/smartthings/dimmer-switch.src/dimmer-switch.groovy
tpmanley: I think you're right. Not sure what I was thinking originally

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
tpmanley: The changes in this PR have been sent off or the original strings?

greens: the changes in this PR

tpmanley:
👍 (I can't approve my own PR)

greens (approved):

PR #4500, opened 2019-06-13
Fix lux reporting and other attributes

rboy1:
`iconcode` is provided as number, the switch case statement assumes it's a string and always defaults to 10000, fix switch case statements
`percentPrecip` and `forecastIcon` attributes are a string, convert numbers to string while reporting

@juano2310 @bflorian

PR #5235, opened 2019-06-17
SmartWeather Station Tile: Fix attributes reporting

rboy1:
The values returned by TWC are numbers, the attributes are expecting strings

Resubmitting with new baseline, #4500 

@tpmanley


tpmanley:
@rboy1 is this causing any visible problems? It seems to be working OK to me without this change

rboy1:
Not yet @tpmanley. The attribute is defined as type string. The data from TWC is returned as a number. It may be working due to a quirk or how the data is currently handled by the platform. To be safe it should converted to a string and reported. I didn’t see any downside to it.

tpmanley (approved):
Thanks for confirming. Just wanted to make sure I wasn't missing something and I agree it's a good thing to fix.

tpmanley:
I tested this myself and it's a minor change so I am going to skip creating a ticket for it.

PR #5382, opened 2019-06-18
WWST-2325 - DTH upgrade - Aurora Smart PIR Sensor

KKlimczukS:
@greens @tpmanley @MAblewiczS @ZWozniakS @MGoralczykS @PKacprowiczS 

Raw data of this one is quite short: 01 C0C9 0001 01 03 0003 0005 0006 00
No manufacturer and no model in device’s details in graphapi.
I’ve managed to pair this device using this fingerprint:
fingerprint profileId: "C0C9", deviceId: "0001", inClusters: "0003,0005,0006" , deviceJoinName: "Aurora Smart PIR Sensor"

Please take a look.

### DIFF
--- a/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
+++ b/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
@@ -17,8 +17,8 @@ metadata {
 		capability "Sensor"
 
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0009,0500"
-		//raw description 23 0104 0402 00 06 0000 0001 0003 000F 0020 0500 02 000A 0019
-		fingerprint endpoint: "23", profileId: "0104", inClusters: "0000,0001,0003,000F,0020,0500", outClusters: "000A,0019", deviceJoinName: "Aurora Smart PIR Sensor"
+		//raw description 22 0104 0107 00 03 0000 0003 0406 00
+		fingerprint mfr: "Aurora", model: "MotionSensor51AU", deviceJoinName: "Aurora Smart PIR Sensor"
 	}
 
 	// simulator metadata
--- a/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
+++ b/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
@@ -17,8 +17,8 @@ metadata {
 		capability "Sensor"
 
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0009,0500"
-		//raw data: 01 C0C9 0001 01 03 0003 0005 0006 00
-		fingerprint profileId: "C0C9", deviceId: "0001", inClusters: "0003,0005,0006",  deviceJoinName: "Aurora Smart PIR Sensor"
+		//raw description 23 0104 0402 00 06 0000 0001 0003 000F 0020 0500 02 000A 0019
+		fingerprint endpoint: "23", profileId: "0104", inClusters: "0000,0001,0003,000F,0020,0500", outClusters: "000A,0019", deviceJoinName: "Aurora Smart PIR Sensor"
 	}
 
 	// simulator metadata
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/motion-detector.src/motion-detector.groovy
```diff
@@ -17,6 +17,8 @@ metadata {
 		capability "Sensor"
 
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0009,0500"
+		//raw data: 01 C0C9 0001 01 03 0003 0005 0006 00
+		fingerprint profileId: "C0C9", deviceId: "0001", inClusters: "0003,0005,0006",  deviceJoinName: "Aurora Smart PIR Sensor"
```
greens: @tpmanley this skeeves me out. I think we should ask for new samples. It seems absurd that they wouldn't include manufacturer _or_ model.

tpmanley: I didn't even look past that weird profile but yeah I agree

KKlimczukS: ok, let's wait for upgraded samples / firmware update.

devicetypes/smartthings/motion-detector.src/motion-detector.groovy
```diff
@@ -17,6 +17,8 @@ metadata {
 		capability "Sensor"
 
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0009,0500"
+		//raw description 23 0104 0402 00 06 0000 0001 0003 000F 0020 0500 02 000A 0019
+		fingerprint endpoint: "23", profileId: "0104", inClusters: "0000,0001,0003,000F,0020,0500", outClusters: "000A,0019", deviceJoinName: "Aurora Smart PIR Sensor"
```
tpmanley: We don't have many fingerprints with `endpoint`. Did you check to make sure this fingerprint is only applying to devices with an endpoint of 0x23? For example, you should be able to change the endpoint here to something else (ex: 22) and then verify the device joins as a Thing and not as a `Motion Detector`.

tpmanley: Actually now that I read the ticket can't you add a normal fingerprint with manufacturer: Aurora model: MotionSensor51AU?

greens: Yeah. @KKlimczukS this is the fingerprint I use to get the device to add with the actual mfr/model in the IDE. Once you know that you can just make it an mfr/model fingerprint as usual.

KKlimczukS: Turned out that device is paired with an endpoint 22 when i used a new fingerprint: 
fingerprint mfr: "Aurora", model: "MotionSensor51AU", deviceJoinName: "Aurora Smart PIR Sensor" (so it doesn't matter if I put there any other endpoint number).
It's possible to pair with specific endpoint only when You use old way fingerprint (put specific clusters in fingerprint), but writing an endpoint number in fingerprint doesn't matter either. 
Anyway, looking for endpoint in hub logs helped to connect with the device in a way that we could get real manufacturer and model number to create a new fingerprint.

devicetypes/smartthings/motion-detector.src/motion-detector.groovy
```diff
@@ -17,6 +17,8 @@ metadata {
 		capability "Sensor"
 
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0009,0500"
+		//raw description 22 0104 0107 00 03 0000 0003 0406 00
+		fingerprint mfr: "Aurora", model: "MotionSensor51AU", deviceJoinName: "Aurora Smart PIR Sensor"
```
tpmanley: For Zigbee fingerprints we use `manufacturer` instead of `mfr`

KKlimczukS: ok, fixed.

greens (approved):

PR #5822, opened 2019-06-19
DVCSMP-3664 Fix most common DTH exceptions

greens:


greens:
The nightly e-mail finally pushed me over the edge.

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -72,7 +72,7 @@ def parse(String description) {
     else {
         if ((description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))) {
             def descMap = zigbee.parseDescriptionAsMap(description)
-            if (descMap.clusterInt == 0x0001 && descMap.attrInt == 0x0020) {
+            if (descMap.clusterInt == 0x0001 && descMap.attrInt == 0x0020 && descMap.value != null) {
```
tpmanley, line 75: Do you know when this is `null`?

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -292,7 +292,7 @@ def parseDescriptionAsMap(description) {
     def result = map.get('result')
     def list = [];
 
-  if (getDataLengthByType(map.get('encoding')) < map.get('value').length()) {
+  if (getDataLengthByType(map.get('encoding')) != null && getDataLengthByType(map.get('encoding')) < map.get('value').length()) {
```
dkirker, line 295: hmmmmm Is there the chance that `map.get('encoding')` can be null? In which case getDataLengthByType will have an issue when it goes for `map.get(t)` (map.get(null)). I almost wonder if it is possible to just dump this function altogether and use the one in the Zigbee library? I'm not quite sure how much custom stuff is in here. I see a few other uses without null protection, like `map.get('value').length()` that has me concerned...

Now that I am looking at the files as a whole, I almost want to re-write the entire thing.....

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
greens: Maybe for the reporting config response?

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
greens: It might be that `map.get('encoding')` is `null`. I'm mostly just doing naive-as-possible fixes since I don't want to touch that DTH any more than I have to. It's very nasty.

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -334,13 +334,15 @@ def flipHexStringEndianness(s)
...
-	 "2a":3,	"2b":4,	"2c":5,	"2d":6,	"2e":7,	"2f":8,	"30":1,	"31":2,	"38":2,	"39":4,	"40":8,	"e0":4,	"e1":4,	"e2":4,
-	 "e8":2,	"e9":2,	"ea":4,	"f0":8,	"f1":16]
+	def map = ["08":1,"09":2, "0a":3, "0b":4, "0c":5, "0d":6, "0e":7, "0f":8, "10":1, "18":1, "19":2, "1a":3, "1b":4,
+				"1c":5, "1d":6, "1e":7, "1f":8, "20":1, "21":2, "22":3, "23":4, "24":5, "25":6, "26":7, "27":8, "28":1,
+				"29":2, "2a":3, "2b":4, "2c":5, "2d":6, "2e":7, "2f":8, "30":1, "31":2, "38":2, "39":4, "40":8, "e0":4,
+				"e1":4, "e2":4, "e8":2, "e9":2, "ea":4, "f0":8, "f1":16]
 
 	// return number of hex chars
-	 return map.get(t) * 2
+    def num_of_hex_chars = map.get(t)
```
greens: looks like my tabs are inconsistent here. will fix

tpmanley (approved):

PR #5859, opened 2019-06-19
DVCSMP-3664 Fix most common DTH exceptions

greens:


devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -235,6 +235,6 @@ private crcEncap(physicalgraph.zwave.Command cmd) {
 	zwave.crc16EncapV1.crc16Encap().encapsulate(cmd).format()
 }
 
-def zwaveEvent(physicalgraph.zwave.Command cmd) {
-	log.warn "Unhandled command: ${cmd}"
+def zwaveEvent(physicalgraph.zwave.Command cmd, int endpointId) {
```
dkirker: Would it make more sense to do something like:

```
def zwaveEvent(physicalgraph.zwave. Command cmd, endpoint=null)
```

to match the others? The handling of secure messages just calls `zwaveEvent(cmd)`

greens: fixed to unify them

dkirker:
I'd think that this would be fine:

```
def zwaveEvent(physicalgraph.zwave.Command cmd, endpoint = null) {
	[descriptionText: "$device.displayName: $cmd endpoint: $endpoint", isStateChange: true]
}
```
You'd just see `"My Device: physicalgraph.zwave.SomeUnhandledCommand() endpoint: null"` If there is no endpoint.

But fine either way.

dkirker (dismissed):

devicetypes/smartthings/zwave-dual-switch.src/zwave-dual-switch.groovy
```diff
@@ -160,8 +160,9 @@ def zwaveEvent(physicalgraph.zwave.commands.crc16encapv1.Crc16Encap cmd) {
 	[:]
 }
 
-def zwaveEvent(physicalgraph.zwave.Command cmd) {
-	[descriptionText: "$device.displayName: $cmd", isStateChange: true]
+def zwaveEvent(physicalgraph.zwave.Command cmd, endpoint = null) {
+	if (endpoint == null) [descriptionText: "$device.displayName: $cmd", isStateChange: true]
```
tpmanley: I think we may want to log a message rather than create an event in this case

greens:
switched to a log message

tpmanley (approved):

PR #6144, opened 2019-06-21
WWST-2347 Re-enable mcd devices

greens:
move aeotec key fob fingerprint as well

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -18,7 +18,7 @@ import groovy.json.JsonOutput
 import physicalgraph.zigbee.zcl.DataType
 
 metadata {
-	definition (name: "Zigbee Multi Button", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "generic-4-button") {
+	definition (name: "Zigbee Multi Button", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
 		capability "Actuator"
 		capability "Battery"
 		capability "Button"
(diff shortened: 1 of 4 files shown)


greens:
these should be ready for primetime now.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -18,7 +18,7 @@ import groovy.json.JsonOutput
 import physicalgraph.zigbee.zcl.DataType
 
 metadata {
-	definition (name: "Zigbee Multi Button", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "generic-4-button") {
+	definition (name: "Zigbee Multi Button", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
```
tpmanley, line 21: I think we want to remove the `runIn` from `installed` because the child components need to be created on initial device creation. It's OK to do it from `updated` though

greens: thought that was fixed now?

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -148,8 +148,8 @@ def configure() {
 }
 
 def installed() {
-	runIn(2, "initialize", [overwrite: true])
 	sendEvent(name: "button", value: "pushed", isStateChange: true)
```
tpmanley: Minor but I think we want to set `displayed` to false on these initial events

tpmanley (approved):

PR #6665, opened 2019-06-25
WWST-2347 MCD multi-buttons

greens:
Enables multi-buttons. There was a conflict with the metadata where not having a component labeled "button1" was causing UI inconsistencies.

### DIFF
--- a/devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
+++ b/devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
@@ -23,6 +23,7 @@ metadata {
 		capability "Health Check"
 
 		fingerprint deviceId: "0x0101", inClusters: "0x86,0x72,0x70,0x80,0x84,0x85"
+		fingerprint mfr: "0086", prod: "0101", model: "0058", deviceJoinName: "Aeotec Key Fob"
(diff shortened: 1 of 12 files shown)


tpmanley:
Do you want to change the join name from `Iris KeyFob 1` -> `Iris KeyFob`?

tpmanley (dismissed):

greens:
Still need to change this. Bashi pointed out that the Fibaro devices support "double pressed", but the metadata we're using here does not.

...sigh

greens:
@dkirker this is ready to go. I've uploaded the generic-6-2 metadata and requested it to prod, along with an MR. I believe you were running into this problem with another device where the indexing was based on having or not having a main component. For simplicity's sake I've decided to mirror what we are using for the wallmotes.

dkirker:
I left a comment on the metadata MR around `supportedButtonValues`. But yeah, I think the issue was because the device type was treating the parent as button 1 and creating n-1 children.

tpmanley (approved):

dkirker (approved):
Looks good!

Just to confirm, the fingerprints were all commented in both device handlers (i.e. neither device handler had fingerprints that were being used), so we don't need to worry about customers having existing devices that we need to send `supportedButtonValues` for?

greens:
Correct. This never went live. I'm going to hold this off until after the deploy, though.

PR #6680, opened 2019-06-25
ICP-9520 Lux thermostat enforces its own setpoint separation

greens:
when you change a setpoint to be very close to another one, the thermostat automatically adjusts the other setpoint to be some buffer away. This reporting was not a whole number.

### DIFF
--- a/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
+++ b/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
@@ -441,7 +441,7 @@ def fanOn() {
 def setCoolingSetpoint(degrees) {
 	if (degrees != null) {
 		def celsius = (temperatureScale == "C") ? degrees : fahrenheitToCelsius(degrees)
-        celsius = (celsius as Double).round(2)
+		celsius = (celsius as Double).round(2)
 		return zigbee.writeAttribute(THERMOSTAT_CLUSTER, COOLING_SETPOINT, DataType.INT16, hex(celsius * 100)) +
 				zigbee.readAttribute(THERMOSTAT_CLUSTER, COOLING_SETPOINT)
 	}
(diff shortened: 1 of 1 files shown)


greens:
Have sought guidance on when we're expecting whole numbers. I thought we did .5 increments in Celsius, but maybe not?

dkirker:
I'm pretty sure that we had agreed on 0.5 some time back. However, the step values in the metadata still don't work properly.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -152,17 +152,17 @@ private parseAttrMessage(description) {
 			if (it.attribute == LOCAL_TEMPERATURE) {
 				log.debug "TEMP"
 				map.name = "temperature"
-				map.value = getTemperature(it.value)
+				map.value = Math,round(getTemperature(it.value))
```
dkirker: The comma there is going to cause some problems.

greens:
SR-Pol says this is what they want.

devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy
```diff
@@ -152,17 +152,17 @@ private parseAttrMessage(description) {
 			if (it.attribute == LOCAL_TEMPERATURE) {
 				log.debug "TEMP"
 				map.name = "temperature"
-				map.value = getTemperature(it.value)
+				map.value = Math.round(getTemperature(it.value))
```
dkirker: Was glancing at the code earlier this morning and realized the `Math.round` would probably be best on line 344 in `getTemperature()`. Then it would only be in one place and would be consistently used for all temp conversions.

greens: agreed. updated

dkirker (approved):

dkirker:
Makes sense re the rounding. I went back and checked the other ticket related to rounding and it was for Ecobee, which does display decimal values.

LGTM

PR #7139, opened 2019-06-27
ICP-9908 Improve battery handling for IKEA Wireless Dimmer

dkirker:


devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -291,7 +291,10 @@ def handleBatteryEvents(descMap) {
 			batteryValue = Math.round(rawValue / 2)
 		}
 
-		if (batteryValue != null) {
+		// Some special logic for the IKEA Dimmer:
+		// Only report the event if we have a legal battery reading, and the difference between this reading and the last reading isn't > 20%.
+		// Sometimes we receive spurious low battery events.
+		if (batteryValue != null && !(isIkeaDimmer() && device.currentValue("battery") && (device.currentValue("battery") as int) - batteryValue > 20)) {
```
tpmanley: I'm not familiar with the values typically reported from this device but for other devices we've implemented an algorithm where the device needs to report a new value 2 (or more) times before it would be considered valid. That way if the value actually changes it does get reported eventually.

greens: Based on the issue it looks like these reports come too frequently to rely on that.

greens: what if we kept a running average of the last, say, ten values?

dkirker: A running average could work. It seems that the battery values fluctuate based on the usage at the time. I seem to recall that the device only responds to battery events when the device is being used.

Some of the reported values were: 37, 11, 37, 11, 37, 11, 17, 8, 11, 0, 11, 17, 11, 17.

I've never actually experienced reporting as frequent as certification has.

greens (dismissed):
I guess an average of the most recent 2 values is probably the easiest way to do it. I suppose this will be more reactive if we start getting 0 readings.

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -340,7 +340,8 @@ def configure() {
 			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_VOLTAGE_ATTR, DataType.UINT8, 0, 21600, null)
```
dkirker: Was there a reason that the original DTH didn't use zigbee.batteryConfig? @MMateusiakS

dkirker: cc @tpmanley

greens (dismissed):
This looks like a smart change.

tpmanley (dismissed):

dkirker:
This is still pointed at staging (I believe that was to get it in to staging after a deploy had occurred). If we would rather have this go to staging on July 16 instead of being merged and then deployed then, I can either rebase to master, or just wait until after the next deploy to merge.

greens:
I'd feel better if it was pointed back at master.

dkirker:
Done

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -332,15 +333,21 @@ def installed() {
...
 	if (isCentraliteSwitch()) {
 		zigbee.addBinding(zigbee.ONOFF_CLUSTER) + zigbee.addBinding(zigbee.LEVEL_CONTROL_CLUSTER) +
 			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_VOLTAGE_ATTR) +
-			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_VOLTAGE_ATTR, DataType.UINT8, 0, 21600, null)
+			zigbee.batteryConfig(0, 21600, null)
 	} else {
 		zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_PERCENT_ATTR) +
-			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_PERCENT_ATTR, DataType.UINT8, 30, 10 * 60, null)
+			// Report no more frequently than 30 seconds, no less frequently than 6 hours, and when there is a change of 10% (expressed as half percents)
+			zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, BATTERY_PERCENT_ATTR, DataType.UINT8, 30, 21600, 20)
```
tpmanley: The report interval is longer than the checkInterval which will make the sengled switch get marked offline.

dkirker: I'm going to up the checkInterval, if that sounds good. I think 6 hours is plenty short enough, especially for battery devices.

tpmanley: Does the sengled respond to pings? If both the report interval and check interval are both 6 hours then one missed report could cause it to go offline if it doesn't respond to pings.

dkirker: I was going to set the checkInterval to 2 * reportInterval + ~2 mins to 10 mins. Unless you think reportInterval should be 3 hours and checkIntervale 6 hours + buffer? Or shorter?

tpmanley: I do think `checkInterval` should be 2 * reportInterval + 2-10 minutes. Typically we'd do 5 minutes for report interval but that impacts battery life so I think it's a good idea to increase it. I'd say go with a 3 hour reportInterval and 6:10 checkInterval for now and also run it by product for their thoughts.

dkirker: Made some updates. I realized the IKEA dimmer does report to us, we just can't ping it. So now the reportInterval and checkInterval are all in sync. I'm surprised that some of it wasn't caught before. Anyway, I have this running at home, so we'll see if it marks offline.

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -332,15 +332,20 @@ def installed() {
 }
 
 def configure() {
-	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 10 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	def offlinePingable = isIkeaDimmer() ? "0" : "1" // We can't ping the IKEA dimmer, so tell device helath this
```
tpmanley: sp: "health"

tpmanley (dismissed):

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
dkirker: Fixed

tpmanley (approved):

PR #8660, opened 2019-07-01
Configurable acceleration sensitivity

rboy1:
Add support for configuring acceleration sensitivity for 2nd gen and 3rd gen chipsets. Default sensitivity can be too high for most uses as it leads to false triggering from nearby vibrations and insufficient for specialized use cases where careful handling of sensor doesn't trigger it.

@tpmanley to follow up on our discussion. Tested with the v2 sensor, please test on the v3 sensor.

### DIFF
--- a/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
@@ -71,7 +71,7 @@ metadata {
 		}
 		section {
 			input title: "Motion Sensitivity", description: "This feature allows you to adjust the sensitivity of the sensor to touch, vibration or movement", displayDuringSetup: false, type: "paragraph", element: "paragraph"
-			input "motionSensitivity", "enum", title: "Sensitivity", description: "Adjust motion sensitivity", options: ["1":"Lowest", "30":"Low", "60":"Medium", "100":"High", "130":"Highest"], defaultValue: "100", displayDuringSetup: false
+			input "motionSensitivity", "enum", title: "Sensitivity", description: "Adjust motion sensitivity", options: ["Lowest", "Low", "Medium", "High", "Highest"], defaultValue: "High", displayDuringSetup: false
 		}
 		section {
 			input("garageSensor", "enum", title: "Do you want to use this sensor on a garage door?", description: "Tap to set", options: ["Yes", "No"], defaultValue: "No", required: false, displayDuringSetup: false)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -467,6 +475,11 @@ def updated() {
 			sendEvent(name: 'status', value: 'closed', descriptionText: descriptionText, translatable: true)
 		}
 	}
+    
+	// Update configuration
+	configCmds += configure()
```
tpmanley, line 477: Can you pull the setting of the acceleration threshold out into a separate function and just call that here rather than `configure`?

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -451,8 +458,9 @@ def configure() {
 
 def updated() {
 	log.debug "updated called"
-	log.info "garage value : $garageSensor"
-	if (garageSensor == "Yes") {
+	log.info "garage value : $garageSensor, motion sensitivity : $motionSensitivity"    
+    	def configCmds = []
```
tpmanley, line 459: nit: Indentation looks off here

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
```diff
@@ -415,10 +419,13 @@ def configure() {
...
 		configCmds += zigbee.writeAttribute(0xFC02, 0x0000, 0x20, 0x01, [mfgCode: manufacturerCode])
 		// passed as little-endian as a bug-workaround
-		configCmds += zigbee.writeAttribute(0xFC02, 0x0002, 0x21, "7602", [mfgCode: manufacturerCode])
-	} else if (device.getDataValue("manufacturer") == "Samjin") {
+		int sensitivity = (0x0276 * 100 / ((motionSensitivity as Integer) ?: 100)) as int // Motion sensitivity is inverse
+		configCmds += zigbee.writeAttribute(0xFC02, 0x0002, 0x21, DataType.pack(sensitivity, DataType.UINT16, true), [mfgCode: manufacturerCode])
+	} else if (device.getDataValue("manufacturer") == "Samjin") { // Current version
 		log.debug "Refreshing Values for manufacturer: Samjin "
-		configCmds += zigbee.writeAttribute(0xFC02, 0x0000, 0x20, 0x14, [mfgCode: manufacturerCode])
+		def y = (180 / -99 * (((motionSensitivity as Integer) ?: 100) - 100) as int) // Motion sensitivity is inverse
```
tpmanley: This math is a little hard to follow. Since there are only five possible inputs what do you think about just creating a map to the desired value? I think that will be easier to understand.

rboy1: It's linear inverse transformation, the equation was simplified. I can extract it into a separate function.
The reason I thought about using the transformation function instead of just values is so that the sensitivity can be mapped into a simple range from 1% to 150%. (100% being the current DTH value). This way it allows for an easy scaling to sensitivity and the function can extended to a free input range (instead of an enum) or a sliding scale UI if support in future. 

Either way it fine, let me know and I can change it accordingly.

greens: Mostly agree with Tom here. Since the input values are currently constrained it would make sense to just have a map.

But if you do want to keep this in for future changes, I at least think you should break it out into a separate function where you explain the function a bit more in the comments.

devicetypes/smartthings/smartsense-multi-sensor.src/smartsense-multi-sensor.groovy
rboy1: Done Tom!

rboy1:
@greens @tpmanley made the recommended changes and switched it to a map

PR #9129, opened 2019-07-03
Spruce Scheduler weatherapi correct location for Canada

natec007:
- Update location to accept and use location in Canada (lat & long)
- Update UI to remove separate zipcode page and put entry field on same weather page
- ln 1316 Remove availabledays = 7 days to correct default of 0 days selected


dkirker:
@natec007 I haven't had a chance to review, but will be having a look.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -147,39 +151,25 @@ def globalPage() {
...
-            input(name: 'zipcode', type: 'text', title: 'Zipcode. Default value is current Zip code',
-                defaultValue: getPWSID(), required: false, submitOnChange: true )
-        }
-    }
-}
-
 private String getPWSID() {
     String PWSID = location.zipCode
-    if (zipcode) PWSID = zipcode
+    if (!location.zipCode.isNumber()) PWSID = "${location.latitude.floatValue()},${location.longitude.floatValue()}"
```
dkirker: It might be best to do:

`!location.zipCode?.isNumber()` in case `location.zipCode` is null.

dkirker: I would even have checks on the latitude and longitude.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -812,8 +791,8 @@ private String getname(String i) {
 }
 
 private String zipString() {
-    if (!settings.zipcode) return "${location.zipCode}"
-    if (!settings.zipcode.isNumber()) return "pws:${settings.zipcode}"
+    if (!settings.zipcode && location.zipCode.isNumber()) return "${location.zipCode}"
+    if (!settings.zipcode) return "${location.latitude.floatValue()},${location.longitude.floatValue()}"//"pws:${settings.zipcode}"
```
dkirker: Same note as above re location fields.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -2056,6 +2035,8 @@ def getRainToday() {
 boolean isWeather(){
     if (!settings.isRain && !settings.isSeason) return false
     String wzipcode = zipString()
+    log.debug wzipcode
+    //log.debug "${getTwcLocation("${wzipcode}")}"
```
dkirker: And, of course, as above, you'll have to check the return of zipString incase someone is using ST without a location.

dkirker (changes requested):
@natec007 Sorry about the review delay!

I left a few inline comments about handling the properties of the location object, just in case someone is using Spruce on a ST location that does not have a zip or lat+long.

natec007:
@dkirker No problem, thanks for checking it out.

Added the recommended checks for location, and a return false for isWeather() in the event of location error.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
natec007: Check longitude and latitude are available before setting

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
natec007: Check getTwcLocation(wzipcode)?.location?.city and check if getTwcForecast and getTwcConditions return otherwise return false and ignore isWeather()

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -147,40 +152,27 @@ def globalPage() {
...
-}
-
-private String getPWSID() {
-    String PWSID = location.zipCode
-    if (zipcode) PWSID = zipcode
-    return PWSID
+private String getDefaultLocation() {
+    String DefaultLocation = "Not set"
+    if(location?.zipCode) DefaultLocation = location.zipCode
+    if (!location.zipCode?.isNumber() && location?.latitude && location?.longitude) DefaultLocation = "${location.latitude.floatValue()},${location.longitude.floatValue()}"
```
dkirker: D'oh! Forgot a `?` on `location.zipCode?.isNumber()` so that it is `location?.zipCode?.isNumber()`.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -811,10 +792,11 @@ private String getname(String i) {
     if (settings."name${i}") return settings."name${i}" else return "Zone ${i}"
 }
 
-private String zipString() {
-    if (!settings.zipcode) return "${location.zipCode}"
-    if (!settings.zipcode.isNumber()) return "pws:${settings.zipcode}"
-    else return settings.zipcode
+private String zipString() {    
+    if (settings?.zipcode) return settings.zipcode
+    if (!settings.zipcode && location.zipCode?.isNumber()) return "${location.zipCode}"
```
dkirker: D'oh! Forgot a `?` on `location.zipCode?.isNumber()` so that it is `location?.zipCode?.isNumber()`.

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
```diff
@@ -2056,10 +2038,17 @@ def getRainToday() {
...
-    String city = getTwcLocation(wzipcode).location.city ?: wzipcode
+    log.debug wzipcode
+    //log.debug "${getTwcLocation("${wzipcode}")}"
+    String city = getTwcLocation(wzipcode)?.location?.city ?: wzipcode
     def forecastData = getTwcForecast(wzipcode)
     def conditionsData = getTwcConditions(wzipcode)
-   
+	
+    if (!forecastData || !conditionsData) {
+        note('warning', "${app.label}: Please check Zipcode/PWS setting, error: null", 'a')
```
dkirker, line 2096: Just a heads up with this logic: getTwc* has the potential to throw exceptions on errors. I believe if the location is invalid you will get an exception. I have some cascading logic to work with the possibility that we have an invalid location here (while the code linked is a DTH and this is a SmartApp, I believe the same applies to both): https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy#L383

I wonder if moving the code to get the weather in to a helper function that handles that logic like in the Stelpro code and then returns an object with `{conditionsData: ..., forecastData: ...}` or `null` that can be used by `getRainToday()` and `isWeather()` would be the best way. Something like:

```
def getWeather(...) {
    ...
    return {conditionsData: ..., forecastData: ...} or null
}

def getRainToday() {
    def conditionsData = getWeather()?.conditionsData

    ...
}

def isWeather() {
    def weather = getWeather()
    def conditionsData = weather?.conditionsData
    def forecastData = weather?.forecastData

    ...
}
```

dkirker (commented):
@natec007  I left a few comments. It hadn't occurred to me to mention that the *Twc* functions throw exceptions... Sorry!

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
natec007: OK, added ?

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
natec007: OK, added ?

smartapps/plaidsystems/spruce-scheduler.src/spruce-scheduler.groovy
natec007: Broke out forecast and conditions and city to try-catch to make sure any errors are caught.  Functions will return null which causes weather function to return without attempting to make weather adjustments.  City returns "not set" as weather is sometimes set without a specific city name.
Thanks

natec007:
@dkirker I ran tests by changing the zipstring to return various values or nulls to simulate "bad" values.  Data, null, errors caught appropriately and weather functions returned values or skipped as needed.  The try-catch returns null if exceptiosn are thrown. Thanks for the comments

dkirker:
@natec007 Awesome! I'll have a look!

dkirker (approved):

dkirker:
@natec007 Apologies for not getting back sooner. Feel free to nag me here or by email in the future! It's a little too easy to get busy. 😔

natec007:
@dkirker No problem, understand getting busy, thanks!

PR #10709, opened 2019-07-09
DVCSMP-3664 fix more dth exceptions

greens:


devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
```diff
@@ -133,7 +133,7 @@ def parse(String description) {
  	
     //check in configuration change
     if (!device.latestValue('configuration')) result = poll()
-    if (device.latestValue('configuration').toInteger() != interval && interval != null) {  	
+    if (device.latestValue('configuration') as float != interval && interval != null) {
```
tpmanley, line 136: Comparisons of equality with floats don't seem like a great idea. Is this ever not an integer?

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -292,7 +292,7 @@ def parseDescriptionAsMap(description) {
     def result = map.get('result')
     def list = [];
 
-  if (getDataLengthByType(map.get('encoding')) != null && getDataLengthByType(map.get('encoding')) < map.get('value').length()) {
+  if (encoding != null && getDataLengthByType(encoding) != null && getDataLengthByType(encoding) < value.length()) {
```
tpmanley, line 295: Could fix the messed up indentation in this method while you're in here if you want

devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
greens: log messages show it as being `"10.0"`, which shockingly does not cast correctly.

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
greens: there's so much that is so bad I'm scared to touch it.

dkirker: The file definitely has a case of mixed indentation... The memories of using the IDE and submitting through it...

I'd probably just do the whole file at once, and probably not as part of this PR.

devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
dkirker: Line 320 has `device.latestValue('configuration').toInteger()` as well. Is the exception around `device.latestValue('configuration')` being Null and trying to call toInteger() on it? In which case, `device.latestValue('configuration')?.toInteger()` is probably the safest route.

Also, I suspect that "configuration" needs to be defined as an attribute in the metadata block. (It doesn't seem to expose that attribute to the controller SmartApp, so this may not be necessary, but probably safe to do anyway.)

`attribute "configuration", "number"`

greens: the log message is a NumberFormatException, not an NPE. thanks for pointing out that other occurence

greens: "configuration" is a preference

greens:
ping

dkirker (approved):

PR #10724, opened 2019-07-09
Publication #01

SmartThings:
Do I need to publish the Custom Device Handler?
For some reason it is not firing when the App calls the 'bridge'.


dcolley (approved):
Boilerplate from smartthings-mqtt-bridge

PR #11034, opened 2019-07-11
ONEAPP-13424 Eaton 5-Scene Keypad network errors

greens:
- explicitly list metadata
- synthesize initial state since setup takes so long (as part of setup all switches are set to off anyway)

### DIFF
--- a/devicetypes/smartthings/eaton-5-scene-keypad.src/eaton-5-scene-keypad.groovy
+++ b/devicetypes/smartthings/eaton-5-scene-keypad.src/eaton-5-scene-keypad.groovy
@@ -12,7 +12,7 @@
  *
  */
 metadata {
-	definition(name: "Eaton 5-Scene Keypad", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
+	definition(name: "Eaton 5-Scene Keypad", namespace: "smartthings", author: "SmartThings", mcdSync: true, mnmn: "SmartThings", vid: "SmartThings-smartthings-Eaton_5-Scene_Keypad") {
 		capability "Actuator"
 		capability "Health Check"
 		capability "Refresh"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/eaton-5-scene-keypad.src/eaton-5-scene-keypad.groovy
```diff
@@ -61,6 +61,7 @@ def installed() {
 
 	// Device-Watch simply pings if no device events received for checkInterval duration of 32min = 2 * 15min + 2min lag time
 	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "switch", value: "off")
```
tpmanley, line 64: Could add displayed false

tpmanley (approved):

devicetypes/smartthings/eaton-5-scene-keypad.src/eaton-5-scene-keypad.groovy
```diff
@@ -12,7 +12,7 @@
  *
  */
 metadata {
-	definition(name: "Eaton 5-Scene Keypad", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
+	definition(name: "Eaton 5-Scene Keypad", namespace: "smartthings", author: "SmartThings", mcdSync: true, mnmn: "SmartThings", vid: "SmartThings-smartthings-Eaton_5-Scene_Keypad") {
```
dkirker, line 15: Is the explicit vid necessary? I just noticed the metadata wasn't in production, so the client was definitely using the default generic. Was there a problem with the correct vid being generated when metadata is staging? (I just want to make sure that there isn't an underlying problem where this doesn't work without an explicit mnmn and vid that might be affecting other devices, too.)

greens: There's an older one that is `5_Scene` instead of `5-Scene`. I asked Bashi to test first without the explicit vid but it wasn't getting picked up.

PR #11038, opened 2019-07-12
Salus miniSmartplug SX885ZB Device Handler

computime-jasonchan:
We are selling Zigbee Smartplug in US.
And we would like our customer able to use Smartthing to control our products.

Jason


tpmanley:
Thanks for the PR @computime-jasonchan . Someone should be reaching out to you shortly

PR #11047, opened 2019-07-15
ONEAPP-14063 Send supportedButtonValues for Aeon Key Fob and Panic Button

dkirker:
With some added formatting clean-up

### DIFF
--- a/devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
+++ b/devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
@@ -56,13 +56,25 @@ metadata {
 
 def parse(String description) {
 	def results = []
+
+	if (!device.currentState("supportedButtonValues")) {
+		sendEvent(name: "supportedButtonValues", value: JsonOutput.toJson(["pushed", "held"]), displayed: false)
+
+		if (childDevices) {
+			childDevices.each {
+				it.sendEvent(name: "supportedButtonValues", value: JsonOutput.toJson(["pushed", "held"]), displayed: false)
+			}
+		}
+	}
+
 	if (description.startsWith("Err")) {
 		results = createEvent(descriptionText:description, displayed:true)
 	} else {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
```diff
@@ -56,13 +56,25 @@ metadata {
 
 def parse(String description) {
 	def results = []
+
+	if (!device.currentState("supportedButtonValues")) {
```
tpmanley, line 60: You could put this in `buttonEvent` instead of `parse` because we do a similar check for `numberOfButtons` in `buttonEvent`, and plus it would mean this check would only happen on button presses and not on every message.

tpmanley (approved):

devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
dkirker: I thought about that. My concern is that the change is related to something that, if mobile doesn't get my requested change for rule builder's behavior in for the upcoming release, one won't be able to use the device in automations until this event is sent.

Maybe what can be done is that I'll follow this PR up with a PR to move that code into the button handler when the next release is ready? Or, I suppose I could just put it in the WakeUpNotification handler -- I assume all devices send it?

dkirker:
Will have QA verify, will set target branch based on QA timeline and merge then to time with next mobile release.

devicetypes/smartthings/aeon-key-fob.src/aeon-key-fob.groovy
tpmanley: OK that makes sense. It would be nice to remove this eventually. Perhaps it could be moved to `updated` after a few weeks so it hasn't taken effect for some reason (dead battery) the user at least has a way to manually recover it.

tpmanley (approved):

PR #11031, opened 2019-07-11
WWST-4444 merge the wistar DTH to zigbee-window-shade

BYDengS:
@dkirker and @greens , please help review this commit.

PR #11054, opened 2019-07-17
Remove fingerprint from Zigbee Curtain

dkirker:
This commit was accidentally removed from PR #11031


dkirker:
@tpmanley It appears that a commit that I had done in #11031 was removed, so the now deprecated Zigbee Curtain DTH ends up conflicting with the proper DTH. I suspect that we can look at full on removing this DTH, but for now I have removed the fingerprint and left a note in the fingerprint area.

dkirker:
cc @BYDengS

tpmanley (approved):

PR #11056, opened 2019-07-17
ICP-9742 Bounds check on heating setpoint so that we can properly handle C and F setpoint ranges with rounding

dkirker:


tpmanley (approved):

dkirker:
I've tested. Waiting to hear back from Poland before merging.

PR #11083, opened 2019-07-23
[ICP-9962] Zooz 4-in-1 sensor reports illuminance in percentage instead lux

PKacprowiczS:
Conversion method taken from community DTH, I'm not sure about it tho.
However, it was "recommended" by manufacturer, so I'm fine leaving it that way.

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
@@ -21,11 +21,9 @@ metadata {
 		capability "Sensor"
 		capability "Battery"
 		capability "Health Check"
-		capability "Tamper Alert"
 
 		fingerprint deviceId: "0x0701", inClusters: "0x5E,0x86,0x72,0x59,0x85,0x73,0x71,0x84,0x80,0x30,0x31,0x70,0x98,0x7A", outClusters:"0x5A"
 		fingerprint mfr:"0086", prod:"0102", model:"004A", deviceJoinName: "Aeotec MultiSensor (Gen 5)"
-		fingerprint mfr:"027A", prod:"2021", model:"2101", deviceJoinName: "Zooz 4-in-1 sensor", mnmn: "SmartThings", vid: "generic-motion-6"
 	}
 
 	simulator {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/aeon-multisensor-gen5.src/aeon-multisensor-gen5.groovy
```diff
@@ -181,7 +181,7 @@ def zwaveEvent(physicalgraph.zwave.commands.sensormultilevelv5.SensorMultilevelR
 			break;
 		case 3:
 			map.name = "illuminance"
-			map.value = cmd.scaledSensorValue.toInteger()
+			map.value = isZooz() ? getLuxFromPercentage(cmd.scaledSensorValue.toInteger()) : cmd.scaledSensorValue.toInteger()
```
tpmanley: This capability is handled locally for this dth so this code path won't be hit

greens: @tpmanley It should get bounced back up to cloud correctly if they've set their scale bit correctly. The LE code checks for the scale bit.

@PKacprowiczS can you expand this case to add handling based on `cmd.scale` (instead of just device)? `0x0` is percentage and `0x1` is lux. I understand devices will probably have different scales in the future (and probably won't be put on this DTH) but it looks more standardized and we can confirm that they're sending correct z-wave packets.

tpmanley: Illuminance is a smart lighting trigger so it won't work if it executes in the DTH

greens: the trigger won't fire at all, even if the hub is online?

tpmanley: Right, if the behavior is believed to run locally it will not run in the cloud even if the DTH generates the trigger event

greens: 😞 so what should we do? This mapping of %=>lux doesn't seem robust enough to include in hubcore and breaking this whole device into a new DTH because of this seems like a waste, but we can't leave this as-is since illumination triggers just won't work.

So a new DTH seems like our only option...

tpmanley: Unless @varzac has any other ideas I think a new DTH is the best option

varzac: Yeah, unfortunately I'd say new DTH is the only option.  I can't think of a better solution.

greens (dismissed):

greens:
Looks good. Let me know when you're ready to merge.

greens (approved):

greens:
@PKacprowiczS is this good to go? If so I'll make sure it goes to staging this week.

PR #11068, opened 2019-07-18
ONEAPP-2033 Add fingerprint-level ocfDeviceTypes

greens:
ICP-3935
ICP-3010
ICP-4151
ICP-4325
ICP-4579
ICP-4612
ICP-4615
ICP-4685
ICP-4488
ICP-5459
ICP-7011
ICP-7099
ICP-7442
ICP-7507
ICP-8147
ICP-6456
ICP-10197
ICP-8602
ICP-5569
ICP-3821
ICP-3781
ICP-3967
ICP-9498
ICP-9576
ICP-10221

Adds ocfDeviceType to several fingerprints on DTHs that have different icons at a device-level

PR #11085, opened 2019-07-23
Wwst 4738

FenMei007S:
@tpmanley plz review it


tpmanley:
@FenMei007S this looks like you should be able to add a fingerprint to the `Zigbee Switch` DTH instead of making a new DTH. If you did it because you wanted to set `ocfDeviceType: "oic.d.smartplug"` you can now do that at the fingerprint level (see https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11068)

PR #11089, opened 2019-07-23
ICP-8170 Set divisor for Aurora energy readings

greens:
Should leverage code changes made in hubcore 26 to allow the multiplier/divisor fields to be set in the DTH but still used in local execution.


greens:
An ideal solution probably interrogates the device for the multiplier and divisor attributes.

tpmanley (dismissed):

tpmanley (approved):

PR #11101, opened 2019-07-26
DVCSMP-3664 Fix more DTH exceptions

greens:


greens:
With my previous fixes I didn't notice that `encoding` was getting re-assigned.

greens:
I'm also skeptical of the `continue` so open to suggestions.

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -308,6 +308,7 @@ def parseDescriptionAsMap(description) {
        encoding = raw[index..(index+1)]
        index+= 2;
        len =getDataLengthByType(encoding)
+       if (!len) continue
```
tpmanley: This should fix the exception, but it's possible there is another bug in this multi-attribute handling. Ideally this DTH would be updated to use the zigbee library more. I would at least add a comment or log a message that there is unhandled data being dropped.

tpmanley: This is an example of the message that causes the exception: `read attr - raw: B116010201141100293F081200292907 dni: B116 endpoint: 01 cluster: 0201 size: 20 attrId: 0011 result: success encoding: 29 value: 0729290012083f`

tpmanley: Actually now that I look more closely there must be a bug in the multi attribute handling. The raw message includes two attributes: 0x0011 and 0x0012. Both are int16's with the first having a value of 0x083F and the second having a value of 0x0729.

tpmanley: I think the best thing to do is switch to the zigbee library version which handles both read attributes and report attributes correctly

dkirker:
At first glance this looks good. Checking over it again.

Just to make sure I set my scope properly: we're only treating this PR as fixes to the exceptions and NOT a complete overhaul to the DTH? (The reason, if it has become the latter, there are some suggestions that I would make that I wouldn't for the former.)

tpmanley (dismissed):

devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -191,160 +183,96 @@ def parse(String description) {
 	def result = []
 
 	if (description?.startsWith("read attr -")) {
-
-		//TODO: Parse RAW strings for multiple attributes
-		def descMap = parseDescriptionAsMap(description)
+		def descMap = zigbee.parseDescriptionAsMap(description)
```
tpmanley, line 187: These changes are more major than the other DTH exception fixes so I'd suggest creating a separate ticket to help ensure it gets tested by QA.

greens:
Yeah. These changes have me worried too. I'm going to verify we have a Fidure on hand to test with.

dkirker:
We should have at least one. I think I saw one on one of the racks near the device lab door (it might even be near where "F" brand devices would be).

greens:
no regressions reported by Bashi other than that for whatever reason it was adding as a thing. Fingerprint of the device that I added matches the generic, but it wasn't adding.

dkirker (approved):

PR #11107, opened 2019-07-29
CHAD-3493 Z-Wave Cert: Add generic support for locks based on command…

greens:
… classes

### DIFF
--- a/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
@@ -23,6 +23,7 @@ metadata {
 		capability "Health Check"
 		capability "Configuration"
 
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -23,6 +23,7 @@ metadata {
 		capability "Health Check"
 		capability "Configuration"
 
+		fingerprint inClusters: "0x62, 0x71, 0x80, 0x85"
```
cbaumler: How does this work exactly? If a device supports one or more of these command classes, then it gets assigned this DTH?

cbaumler:
@eddysun1990 Since you have a development board set up, can you test this with the SDK 7.11.0 Door Lock sample app that PepperOne used?

eddysun1990:
Yes I will do that.

eddysun1990 (approved):

PR #11113, opened 2019-07-30
CHAD-3494 Meter supported get

greens:
Adds meter support to generic device and a meter supported get.


cbaumler (approved):

greens:
@kianooshST had some reservations about this I think

PR #11114, opened 2019-07-30
CHAD-3496 Send sensor multilevel supported get to fibaro motion sensor

greens:


cbaumler (dismissed):

greens:
Not sure if we also need to send a SensorMultilevelSupportedGetScale for each reported sensor type...

cbaumler:
Probably do need to do that.

image

cbaumler (approved):

tpmanley:
@greens what should we do with this old PR?

PR #11123, opened 2019-08-01
OnOff ZigBee Controller

SmartThings:
ZigBee controller


tpmanley:
This change needs to be rebased on the latest `master` so the existing fingerprints are not modified. Also the new fingerprint `fingerprint profileId: "0104", deviceId: "2016DP3419", manufacturer: "xiaomi", model: "ZNCZ02LM", deviceJoinName: "Mi Power Plug"` doesn't look valid. The deviceID should be 4 ascii-hex characters. I'm just going to close this PR and I think it would be easiest to open a new PR with the correct fingerprint added to the code in the latest `master` branch.

PR #11125, opened 2019-08-01
WWST-4867 - AXIS Gear Retrofit Blind Controller

AXISLabsShare:
contains device handler for WWST certification under SmartThingsPublic/devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy


devicetypes/axis/axis-gear-v2.src/axis-gear-v2.groovy
```diff
@@ -0,0 +1,196 @@
+metadata {
+    definition (name: "AXIS Gear-V2", namespace: "axis", author: "AXIS Labs") {  
+        capability "Actuator"
+        capability "Configuration"
+        capability "Switch"
+        capability "Switch Level"
+        capability "Refresh"        
+        capability "Battery"
+        capability "HealthCheck"
```
greens: two words

devicetypes/axis/axis-gear-v2.src/axis-gear-v2.groovy
```diff
@@ -0,0 +1,196 @@
...
+	//sendEvent(name:"level", value:0, displayed:true)
+    setLevel(100)  
+}
+
+//Send Command through setLevel()
+def off() {
+	setLevel(0)
+}
+//Command to set the blind position (%) and log the event
+def setLevel(value) {
```
greens: to match the capability prototype (and have this function correctly with smartapps), please add a `rate = null` default argument to the signature

devicetypes/axis/axis-gear-v2.src/axis-gear-v2.groovy
```diff
@@ -0,0 +1,196 @@
...
+	sendEvent(name:"level", value: value, displayed:true)
+    def L = Math.round(value);
+    def i = Integer.valueOf(L.intValue());
+    setWindowShade(i)
+	zigbee.setLevel(i)
```
greens: you should be able to call `zigbee.setLevel(value)` like we do in our other DTHs without the need for rounding or casting and calling `setWindowShade(value as int)` should do all of the casting you would need, I think.

greens: If you've configured reporting correctly, shouldn't the device report its level after it is set, making the call to `setWindowShade()` redundant?

greens (changes requested):
The gear, gear v2, and gear v3 all seem identical to me and should at least be implemented as one single DTH with multiple fingerprints.

Alternatively, I think they would fit easily in our `zigbee-window-shade` DTH with only a few slight manufacturer-specific tweaks.

As a general comment, if you could please make all of your whitespace (spaces and tabs specifically) and formatting (spacing for if statements, function capitalization) consistent I would appreciate it.

devicetypes/axis/axis-gear-v2.src/axis-gear-v2.groovy
AXISLabsShare: > The gear, gear v2, and gear v3 all seem identical to me and should at least be implemented as one single DTH with multiple fingerprints.
> 
> Alternatively, I think they would fit easily in our `zigbee-window-shade` DTH with only a few slight manufacturer-specific tweaks.
> 
> As a general comment, if you could please make all of your whitespace (spaces and tabs specifically) and formatting (spacing for if statements, function capitalization) consistent I would appreciate it.

the DTH that should be reviewed is v5.1 as Gear, GearV2 and GearV3 were older device handlers. V5.1 should work with older Gears that don't use window covering commands as well as newer versions of firmware that do.

As a result the older versions can be removed if required.

greens: Please remove them from the PR.

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
```diff
@@ -0,0 +1,420 @@
+metadata {
+    definition (name: "AXIS Gear-V5.1", namespace: "axis", author: "AXIS Labs") {  
+        capability "Window Shade"
+        capability "Window Shade Preset"
+        capability "Switch Level"
+        capability "Battery"
+		capability "Refresh"
```
greens: Repeating the general request to make whitespace consistent. Here's an example of a line using tabs immediately following a line using spaces.

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
```diff
@@ -0,0 +1,420 @@
...
+
+            return zigbee.command(CLUSTER_WINDOWCOVERING, WINDOWCOVERING_CMD_STOP)         
+        }
+        else {
+            sendEvent(name: "windowShade", value: "stoppingNS")
+            return zigbee.readAttribute(CLUSTER_LEVEL, LEVEL_ATTR_LEVEL, [delay:5000])        
+        }
+    }
+    else {
+    	if (state.currentVersion >= 1093){
```
greens: this repeated use of `1093` should be replaced with a function `getMIN_WINDOW_COVERING_VERSION` like you've done for the clusters.

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
```diff
@@ -0,0 +1,420 @@
...
+
+    def event = zigbee.getEvent(description)
+    if (event) {
+        sendEvent(event)
+    }
+
+    Map map = [:]
+    if (description?.startsWith('read attr -')) {
+        map = parseReportAttributeMessage(description)
+    } else if (description?.startsWith('attr report -')) {
```
greens: this condition could be combined with the above

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
```diff
@@ -0,0 +1,420 @@
...
+            tileAttribute("device.windowShade", key: "PRIMARY_CONTROL") {
+                attributeState("open",  label: 'Open',  action:"close", icon:"http://i.imgur.com/4TbsR54.png", backgroundColor:"#ffcc33", nextState: "closing")
+                attributeState("partially open", label: 'Partially Open', action:"close", icon:"http://i.imgur.com/vBA17WL.png", backgroundColor:"#ffcc33", nextState: "closing")
+                attributeState("closed", label: 'Closed', action:"open",  icon:"http://i.imgur.com/mtHdMse.png", backgroundColor:"#bbbbdd", nextState: "opening")
+                attributeState("opening", label: 'Opening', action: "stop", icon: "http://i.imgur.com/vBA17WL.png", backgroundColor: "#ffcc33", nextState: "stopping")
+                attributeState("closing", label: 'Closing', action: "stop", icon: "http://i.imgur.com/vBA17WL.png", backgroundColor: "#bbbbdd", nextState: "stopping") 
+                attributeState("stopping", label: 'Stopping',  icon: "http://i.imgur.com/vBA17WL.png", backgroundColor: "#ff7777") 
+                attributeState("stoppingNS", label: 'Stopping Not Supported',  icon: "http://i.imgur.com/vBA17WL.png", backgroundColor: "#ff7777")
```
greens: These states "stopping" and "stoppingNS" will not be visible in the new app and behavior may be strange because of it.

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
```diff
@@ -0,0 +1,420 @@
...
+    sendEvent(name: "version", value: "Checking Version ... ")     
+	return zigbee.readAttribute(CLUSTER_BASIC, BASIC_ATTR_SWBUILDID)
+}
+
+//configure reporting
+def configure() {   
+	state.currentVersion = 0
+	sendEvent(name: "windowShade", value: "unknown")
+    log.debug "Configuring Reporting and Bindings."
+    sendEvent(name: "checkInterval", value: 60 * 5, displayed: true, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
greens: Generally when we set the check interval we like it to allow for the hub to have missed one periodic report + some fuzz time before marking the device offline. Right now you're configuring all your attributes to report once every hour, so unless you continually mess with the device or it has some other out-of-band reporting, you're going to be pinging this device every five minutes. I don't expect that would do great things for your battery life, and it doesn't do great things for anyone to rely on frequent pings to keep the device online. Based on your reporting intervals I would instead recommend a checkInterval time of 2 hours + 10 minutes `(2 * 60 * 60 + 10 * 60)`.

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
AXISLabsShare: the SmartThings Classic app is still available and these states do work in that version. In the new app, I have confirmed that it does not go through these states. Instead of stopping when opening and closing, the app will start to close or open respectively

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
AXISLabsShare: noted

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
AXISLabsShare: Thanks for the recommendation. Will update

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
AXISLabsShare: noted

devicetypes/axis/axis-gear-v5.1.src/axis-gear-v5.1.groovy
AXISLabsShare: the discrepancies between the tabs and spacing are due to the IDE for the devices handlers. We have went through the DTH and replaced spaces with tabs for consistency. When creating a new line in the IDE, it creates an indentation with spaces rather than tabs.

AXISLabsShare (commented):
axis-gear-st.groovy will be used for ST moving forward

greens:
Did you intend to leave the v3 DTH? I notice it has a different fingerprint than the models included in the ST DTH. I'd again suggest deleting the DTH from this PR and moving the fingerprint to the `zigbee-window-shade` DTH.

devicetypes/axis/axis-gear-st.src/axis-gear-st.groovy
```diff
@@ -0,0 +1,408 @@
...
+	log.debug "parse() refresh"
+	def cmds_refresh = null
+    
+	if (state.currentVersion >= MIN_WINDOW_COVERING_VERSION){
+		cmds_refresh = zigbee.readAttribute(CLUSTER_WINDOWCOVERING, WINDOWCOVERING_ATTR_LIFTPERCENTAGE) +
+							zigbee.readAttribute(CLUSTER_POWER, POWER_ATTR_BATTERY) +
+							zigbee.readAttribute(CLUSTER_BASIC, BASIC_ATTR_SWBUILDID)
+    }
+    else {
+		cmds_refresh = zigbee.readAttribute(CLUSTER_LEVEL, LEVEL_ATTR_LEVEL) +
```
greens: Only the first command is different here. You could condense this.

devicetypes/axis/axis-gear-st.src/axis-gear-st.groovy
```diff
@@ -0,0 +1,408 @@
...
+	log.info "configure() --- cmds: $cmds"
+	return attrs_refresh + cmds
+}
+
+def parse(String description) {
+	log.trace "parse() --- description: $description"
+	
+	def event = zigbee.getEvent(description)
+	if (event) {
+		sendEvent(event)
```
greens, line 329: I believe this will result in you sending duplicate `level` events. What might be a better alternative is to do some inspections of level events returned by this function here and then generate the supplementary windowShade events and then only conditionally call `parseReportAttributeMessage` if no event was returned.

devicetypes/axis/axis-gear-st.src/axis-gear-st.groovy
AXISLabsShare: change made to condense

devicetypes/axis/axis-gear-st.src/axis-gear-st.groovy
AXISLabsShare: change made to only process the onoff events to correctly report when used via device operation (off indicates closed position). All other attribute reports are handled in parseReportAttributeMessage

greens (approved):

greens:
Merged. This should be on our staging environment tomorrow and then live on production next Tuesday.

PR #11127, opened 2019-08-01
 CHAD-3513 Send a setpoint get after a setpoint set

greens:
I would assume this was covered by association, but maybe not? Might also just be a mapping/plugin error...


cbaumler:
Tested on Honeywell TH8320ZW1007.

cbaumler (approved):

PR #11133, opened 2019-08-05
[WWST-4933][WWST-4884]: add fingerprint for GDKES switch and outlet

FenMei007S:


FenMei007S:
@tpmanley @greens  plz review it

greens (approved):

PR #11157, opened 2019-08-07
[ICP-10363] Fibaro Temperature Sensor (for FGT) doesn't go offline

PKacprowiczS:
@greens @dkirker @KKlimczukS @MGoralczykS @ZWozniakS

### DIFF
--- a/devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
+++ b/devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
@@ -331,7 +331,7 @@ def addChild() {
 	String componentLabel =	 "Fibaro Temperature Sensor"
 	String ch = "ch2"
 
-	addChildDevice("Child Temperature Sensor", childDni, device.hub.id,[completedSetup: true, label: componentLabel, isComponent: false, componentName: ch, componentLabel: componentLabel])
+	addChildDevice("Child Temperature Sensorr", childDni, device.hub.id,[completedSetup: true, label: componentLabel, isComponent: false, componentName: ch, componentLabel: componentLabel])
 }
 
 private getMaxHeatingSetpointTemperature() {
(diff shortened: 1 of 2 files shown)


greens:
I think it might be worth adding that configurationGet to the `refreshChild` method as well.

devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
```diff
@@ -328,4 +334,9 @@ private getMaxHeatingSetpointTemperature() {
 
 private getMinHeatingSetpointTemperature() {
 	temperatureScale == 'C' ? 10 : 50
+}
+
+private changeTemperatureSensorStatus(status) {
+	def map = [name: "healthStatus", value: status]
```
PKacprowiczS: @greens this event doesn't mark device offline. What should I send instead?

greens: `name: "DeviceWatch-DeviceStatus", value: "offline"`

I think that should work. I don't know how it meshes with devices that are actively pinged, but give it a shot.

devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
```diff
@@ -299,7 +306,8 @@ def switchMode() {
 def sendEventToChild(event) {
 	String childDni = "${device.deviceNetworkId}:2"
 	def child = childDevices.find { it.deviceNetworkId == childDni }
-	child?.sendEvent(event)
+	if (state.isChildOnline)
+		child?.sendEvent(event)
```
greens, line 310: were you having a problem with this being sent too often?

dkirker:
It is looking good! One question: are we relying on ping() to be called for us to check on the child sensor?

devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
PKacprowiczS: Kind of, but it's more like, after I've received Multilevel Report from endpoint 2, I was sending an event to child, and it was bringing it back online every time, meaning that sending "offline" event haven't done a trick. So this flag is a workaround to store real child status in parent device and prevent sending 'fake' events to additional temperature sensor.

PKacprowiczS:
@dkirker yep

greens:
@PKacprowiczS want to change it back so we can merge?

PKacprowiczS:
@greens I wanted to wait for all devices' issues to be verified, but it seems those are more related to its design, not DTH. Now names are back to normal.

greens (approved):

PR #11161, opened 2019-08-08
[ICP-10547]: Orvibo 3-gang health check doesn't work, and init status is incorrect

FenMei007S:
health check not work for switch 3, because of no configreport for switch 3. plz review it. @tpmanley  @greens


greens:
Please add the Jira ticket number

greens (approved):

PR #11164, opened 2019-08-09
IKEA-Tradfri

Pr0bably:
all  IKEA-Tradfri


tpmanley:
Thanks for the submission @Pr0bably ! We actually already have support for most if not all of the devices in this DTH and our platform doesn't allow us to have multiple DTHs with the same fingerprint so we won't be able to accept this. Also we want everything that we add going forward to work in both the classic and the new app and unfortunately this DTH is using some features that would only work in the classic app.

PR #11171, opened 2019-08-12
ICP-9040 Check to see if operating state from thermostat is correct when we set mode to off

dkirker:
This change checks to see if the current mode is off and forces the operatingState to be idle if it told us it was heating. Sometimes the thermostat indicates that it is heating using the somewhat arbitrary heat demand value even if the mode is off or the temperature is not currently valid for heating.

### DIFF
--- a/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
@@ -418,7 +418,7 @@ def validateOperatingStateBugfix(map) {
 	if (state.rawSetpoint != null && state.rawTemp != null) {
 		def oldVal = map.value
 
-		if (state.rawSetpoint <= state.rawTemp) {
+		if (state.rawSetpoint <= state.rawTemp || device.currentValue("thermostatMode") == "off") {
 			map.value = "idle"
 		} else {
 			map.value = "heating"
@@ -429,6 +429,7 @@ def validateOperatingStateBugfix(map) {
 			map.data = [correctedValue: true]
 		}
 	}
+
 	map
 }
 


devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -418,7 +418,8 @@ def validateOperatingStateBugfix(map) {
 	if (state.rawSetpoint != null && state.rawTemp != null) {
 		def oldVal = map.value
 
-		if (state.rawSetpoint <= state.rawTemp) {
+		if (state.rawSetpoint <= state.rawTemp ||
+			(device.currentValue("thermostatMode") && device.currentValue("thermostatMode") == "off")) {
```
tpmanley: The `(device.currentValue("thermostatMode")` check shouldn't be necessary since if it doesn't have a current value it won't be equal to "off".

tpmanley:
@dkirker can you explain how this fixes the problem?

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
dkirker: No, but wouldn't trying to do `null == "off"` result in a NPE?

dkirker:
@tpmanley Bleh, updated the description.

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
tpmanley: No, you can safely check for equality against null. You can verify this in the Groovy console: https://groovyconsole.appspot.com/

tpmanley (dismissed):

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
dkirker: Alright, updated that!

tpmanley (approved):

PR #11182, opened 2019-08-14
Fix ICP-10512 when device is offline, the state has been changed to opening/closing and can't change back.

BYDengS:
@greens @dkirker @tpmanley  please help review this commit, this commit is to fix the bug when device is offline, the state has been changed to opening/closing and can't change back. Thank you.

### DIFF
--- a/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
+++ b/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
@@ -135,26 +135,16 @@ def supportsLiftPercentage() {
 
 def close() {
     log.info "close()"
-    sendEvent(name: "windowShade", value: "closing")
     zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_CLOSE)
 }
 
 def open() {
     log.info "open()"
-    sendEvent(name: "windowShade", value: "opening")
     zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_OPEN)
 }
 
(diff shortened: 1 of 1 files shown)


greens:
Will the device report "opening" and "closing" still? Isn't that why you had to alter the DTH in the first place?

BYDengS:
@greens the issue is that when device is offline or power off, the device state also changed successfully. And I found that  `sendEvent(name: "windowShade", value: "XXX")` will be send in `close(),open() and setlevel()` even if device is offline. So I delete it.
After delete the `sendEvent(name: "windowShade", value: "XXX")` event,  the windowShade capability works well. But for `setLevel()` function, Wistar device is good, Dooya device will send the opening/closing state after device finish the operation. and wait a while, changed to `partially open`. it seems a little strange.

BYDengS:
@greens As I said before, the Dooya will report the state just after it finish the operation. is there any good way to optimize this situation?

greens:
So the Dooya only reports the level once the operation has finished. And the issue is that because there are no in-progress reports, you cannot create "opening"/"closing" events. _I_ believe this operation is fine, since this is the way the z-wave shades currently work. (No "opening" and "closing" events are generated.) 

If the team is refusing to certify these devices without opening/closing events I don't think there's anything you can do other than convince them that they're not necessary.

greens (approved):

dkirker:
The `levelEventHandler` should handle sending opening or closing anyway if the device is sending level events.

But yes, good idea to not artificially send open, closed, opening, and closing in the commands.

PR #11187, opened 2019-08-15
One user's mislabeled child devices are causing 90% of our exceptions…

greens:
…. This should fix it.

### DIFF
--- a/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
+++ b/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
@@ -224,23 +224,25 @@ def reset(endpoint = 1) {
 
 def childOnOff(deviceNetworkId, value) {
 	def switchId = getSwitchId(deviceNetworkId)
-	sendHubCommand onOffCmd(value, switchId)
+	if (switchId != null) sendHubCommand onOffCmd(value, switchId)
 }
 
 def childRefresh(deviceNetworkId) {
 	def switchId = getSwitchId(deviceNetworkId)
-	sendHubCommand refresh(switchId)
+	if (switchId != null) sendHubCommand refresh(switchId)
 }
 
 def childReset(deviceNetworkId) {
 	def switchId = getSwitchId(deviceNetworkId)
-	log.debug "Child reset switchId: ${switchId}"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
```diff
@@ -223,21 +223,26 @@ def reset(endpoint = 1) {
...
 def childReset(deviceNetworkId) {
-	def switchId = deviceNetworkId?.split(":")[1] as Integer
+	def switchId = getSwitchId(deviceNetworkId)
 	log.debug "Child reset switchId: ${switchId}"
 	sendHubCommand reset(switchId)
 }
 
+def getSwitchId(deviceNetworkId) {
+	def split = deviceNetworkId?.split(":")
+	return (split.length > 1) ? split[1] as Integer : 1
```
kianooshkarami: otherwise 1?

kianooshkarami (commented):
Can you show me how this happens? I am trying to understand this better

tpmanley (dismissed):

devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
greens: it's the default argument to all the functions

greens: and considering it's not working now it can't be much worse

kianooshkarami: I rather fail out and not send any message

greens:
@kianooshST I _believe_ last time I checked that the user seemed to have created their child devices with a different DTH that did not give them the DNIs we expect here, and then switched the DTH they were using.

devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy
greens: fair enough, I'll fix it

kianooshkarami (approved):

PR #11188, opened 2019-08-16
[ICP 10644]: Multi switch click event not take endpoint info

FenMei007S:
@tpmanley  @greens


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -63,11 +63,13 @@ metadata {
 def installed() {
 	createChildDevices()
         refresh()
+	updateDataValue("onOff", "catchall")
```
tpmanley: This needs to be before the refresh or else the refresh commands will not be sent to the hub. Also please fix the indentation in these two functions.

tpmanley (approved):

PR #11191, opened 2019-08-17
[ICP-9111, ICP-10109, ICP-10123, ICP-10121, ICP-10136, ICP-10646] Z-Wave Multi Metering Switch: Fix problem with automations and child devices

tpmanley:
Even though this isn't a MCD DTH it was setting the componentName and componentLabel on the child devices. The platform expects that a single component device always calls that component "main". Since these child devices had a component named something other than "main", for example "switch2", it was causing errors during operations like creating automations.

https://smartthings.atlassian.net/browse/ICP-10646


dkirker (dismissed):

dkirker (approved):

dkirker:
I verified this as a fix on Thursday evening/Friday with the dual outlet at my desk before packing to move desks.

PR #11195, opened 2019-08-19
[WWST-4312, WWST-4318] Added FP for Muller Licht bulbs (GU10 + Candle) with color temp range…

ZWozniakS:
… 1800-6500 K
@greens Can you please review? Thanks!

### DIFF
--- a/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
@@ -47,7 +47,7 @@ metadata {
 		fingerprint profileId: "0104", inClusters: "0000, 0004, 0003, 0005, 0006, 0008, 0300, 1000", outClusters: "0019", manufacturer: "innr", model: "BY 285 C", deviceJoinName: "Innr Smart Bulb Color"
 		fingerprint manufacturer: "innr", model: "RB 250 C", deviceJoinName: "Innr Smart Candle Colour", mnmn:"SmartThings", vid: "generic-rgbw-color-bulb-1800K-6500K"
 		// Müller Licht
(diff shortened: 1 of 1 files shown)


csuk07:
@ZWozniakS can you re-target this to staging?

dkirker (commented):
@ZWozniakS Just to confirm, the added clusters are correct and there are no bulbs that have the shorter cluster list?

ZWozniakS:
@dkirker Yes, this extra clusters and the whole FP are identical for GU10 and Candle bulbs, they have the same Raw Description. This is correct for these bulbs. There is E27 bulb which has shorter cluster list but we decide to make it as CST.

greens (approved):

dkirker:
@ZWozniakS I see. So the fingerprint that was originally there was for the device which will now be CST? Do we want to leave it in, even if it is CST, so that it pairs with the correct DTH? (I am guessing this conversation has already been had?) @tpmanley

dkirker (approved):

greens:
@dkirker there's two devices which share a fingerprint. One will have the correct join name and be WWST. One will have an incorrect join name and be CST.

dkirker:
Ok, so I should have caught this and stopped the merge, however, the cluster list is wrong:

```
fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300, OBO5, 1000, FEDC", outClusters: "000A, 0019", manufacturer: "MLI", model: "ZBT-ExtendedColor", deviceJoinName: "Müller Licht Bulb White+Color", mnmn:"SmartThings", vid: "generic-rgbw-color-bulb-1800K-6500K"
```

The `OBO5` won't work... It should be `0B05`.

greens:
Crap. Okay. I'll fix it.

greens:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11208

dkirker:
I mean... It *might* pair correctly still. Not sure if the fingerprinting stuff blows up on invalid clusters, or just drops them. 🤪

ZWozniakS:
Oh my... I am sorry about this cluster.
@dkirker @greens The E27 bulb is still pairing with this new fingerprint with extra clusters, the only problem was with color temperature range but we decide that is much better to have correct temp range for two bulbs than for only one. Because of this inconsistency, the decision was to make E27 CST. (GU10, Candle - 1800-6500K, E27 2200-6500K)

PR #11196, opened 2019-08-20
ICP-9962 Added missing clearTamper() method for Zooz 4-in-1

PKacprowiczS:
@greens @dkirker

### DIFF
--- a/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
+++ b/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
@@ -204,6 +204,10 @@ def configure() {
 	secureSequence(request) + ["delay 20000", zwave.wakeUpV2.wakeUpNoMoreInformation().format()]
 }
 
+def clearTamper() {
+	sendEvent(name: "tamper", value: "clear")
+}
+
 private getLuxFromPercentage(percentageValue) {
 	def multiplier = luxConversionData.find {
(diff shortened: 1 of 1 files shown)


dkirker (dismissed):

dkirker:
Actually, @PKacprowiczS could I get you to make 1 change? On https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/290a9d4377111257644cbbc0e5841c5516e0739c/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy#L174 there is an unschedule, you can remove that. The runIn will overwrite/auto-unschedule any existing calls by default. That said, you can add an `overwrite: true` flag to the call to runIn in the next line, like so:

```
runIn(10, clearTamper, [overwrite: true, forceForLocallyExecuting: true])
```

While this DTH isn't marked as locally executing, I think we can go ahead and just leave the `forceForLocallyExecuting` in.

dkirker (approved):

dkirker:
@PKacprowiczS And just to confirm, you have run this and verified that the code works?

PKacprowiczS:
@dkirker yup, tamper alert is being cleared. One thing that's left is metadata, but we're still waiting for  being able to upload it.

dkirker:
@PKacprowiczS If you create an MR for the metadata definitions to the correct repository I will upload those in the morning and get them to production.

PKacprowiczS:
@dkirker we don't have access to that new repo on codesamsung...

PR #11207, opened 2019-08-21
WWST-2347 enable mcd for aeon minimote

greens:


greens:
@dkirker it looks like the metadata for this on prod isn't the version that's checked in to the repo. Think you can take care of that?

dkirker:
@greens Yea

tpmanley:
Added DO_NOT_MERGE for now

dkirker (commented):
Will need `sendEvent(name: "supportedButtonValues", value: ["pushed","held"].encodeAsJson(), displayed: false)` for the parent and the child devices.

dkirker (dismissed):

greens:
Uploaded some changes, but they only work with metadata and plugin -> staging.

The bright side is that since there's barely any functionality now in new app, there will be no degradation of service.

tpmanley:
Have you checked that existing devices that are updated to this new version get correctly synced to OCF?

greens:
@tpmanley Yeah. Added device on classic with old DTH, set up smart lighting automations in classic for components. Then I switched to the new DTH (which obviously runs update automatically, so we'll need to do that manually when we deploy) and the old automations still worked. 

The main point to make is that the components are currently invisible in OneApp, but once `updated()` is called with this new DTH (and the staging plugin is pushed to prod, and the staging metadata is pushed to prod) the device displays and functions correctly, even if it was installed prior.

greens:
ready to go

tpmanley (approved):

PR #11220, opened 2019-08-27
Allow user to select icon for generic child switch

rboy1:
The default icon is a light bulb but the child DTH is a generic switch. It can (is) used by many DTH's using multi component devices for switch endpoints. This allows the user to select their own icon there by allow it to be more meaningful when used a child switch or outlet.

@greens this CR is based on user feedback


greens (approved):
Fine with me.

PR #11235, opened 2019-08-29
[DVCSMP-3664] Aeon Minimote: Fix NPE during button event handling if child isn't found

greens:


devicetypes/smartthings/aeon-minimote.src/aeon-minimote.groovy
```diff
@@ -76,10 +76,10 @@ def buttonEvent(button, held) {
 		log.error "Child device $childDni not found"
 	}
 	if (held) {
-		child?.sendEvent(name: "button", value: "held", data: [buttonNumber: 1], descriptionText: "$child.displayName was held", isStateChange: true)
+		if (child) child.sendEvent(name: "button", value: "held", data: [buttonNumber: 1], descriptionText: "$child.displayName was held", isStateChange: true)
```
tpmanley, line 79: Can you explain what this fixes? The code looks equivalent to me

tpmanley:
Please add a jira and the dth name to the PR title

devicetypes/smartthings/aeon-minimote.src/aeon-minimote.groovy
greens: Arguments are evaluated right to left. `child.displayName` gets evaluated, but `child` is null, hence NPE

tpmanley:
Please make the PR title a little more descriptive so we can see at a glance what this is changing. I'd suggest something like `[DVCSMP-3664] Aeon Minimote: Fix NPE during button event handling if child isn't found`.

tpmanley (approved):

PR #11240, opened 2019-08-30
[ICP-10773] Added more supported button states for Fibaro Button

PKacprowiczS:
@greens @dkirker @KKlimczukS @MGoralczykS @ZWozniakS


greens:
@dkirker are these translated yet? If so (and supportedButtonStates works as advertised) could we add them to the metadata?

greens (approved):

PR #11246, opened 2019-09-03
[ICP-10684] Disabled possibilty to turn off Sengled by decreasing dimmer value

ZWozniakS:
@greens Can you please review? Thanks!
@dkirker @PKacprowiczS @KKlimczukS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
@@ -144,17 +144,18 @@ def handleSengledSwitchEvents(descMap) {
 				//move down
 				if (descMap.data[2] == '02') {
 					//long press of 'DIM' button
-					value = Math.max(currentLevel - DOUBLE_STEP, 0)
+					value = Math.max(currentLevel - DOUBLE_STEP, 1)
 				} else if (descMap.data[2] == '01') {
 					//short press of 'DIM' button
-					value = Math.max(currentLevel - STEP, 0)
+					value = Math.max(currentLevel - STEP, 01)
 				} else {
 					log.info "Invalid value ${descMap.data[2]} received for descMap.data[2]"
 				}
 
-				if (value == 0) {
-					results << createEvent(name: "switch", value: "off")
+				if (device.currentValue("level") == 1) {
+					results << createEvent(name: "level", value: value)
 				} else {
+					results << createEvent(name: "switch", value: "on")
 					results << createEvent(name: "level", value: value)
 				}
 				break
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -144,17 +144,18 @@ def handleSengledSwitchEvents(descMap) {
 				//move down
 				if (descMap.data[2] == '02') {
 					//long press of 'DIM' button
-					value = Math.max(currentLevel - DOUBLE_STEP, 0)
+					value = Math.max(currentLevel - DOUBLE_STEP, 1)
 				} else if (descMap.data[2] == '01') {
 					//short press of 'DIM' button
-					value = Math.max(currentLevel - STEP, 0)
+					value = Math.max(currentLevel - STEP, 01)
```
PKacprowiczS: I'm guessing it shouldn't be '01' here.

ZWozniakS: Sure, thanks!

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
```diff
@@ -144,17 +144,18 @@ def handleSengledSwitchEvents(descMap) {
...
 					//short press of 'DIM' button
-					value = Math.max(currentLevel - STEP, 0)
+					value = Math.max(currentLevel - STEP, 1)
 				} else {
 					log.info "Invalid value ${descMap.data[2]} received for descMap.data[2]"
 				}
 
-				if (value == 0) {
-					results << createEvent(name: "switch", value: "off")
+				if (device.currentValue("level") == 1) {
```
dkirker, line 155: I'm a little confused by what you are trying to do here. I think you want to leave this logic in place as is, as this is the code to handle events from the device. This would prevent the device from ever turning off, at least by using the level control on the device. Unless that is what we are trying to achieve.

dkirker (commented):
I left a comment in line. I am confused by this change.

devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy
ZWozniakS: Yes, that is correct, I don’t want to let the device to turn off by decreasing the dimmer value as metadata set the dimmer range between 1-100%. Also I did not want to hardcode this “if” with value == 1, because then if user want to decrease dimmer value while device is off, the device won’t turn on while current value would be 2, 3, 4, 5, 6 as the dimmer level is changing by 5 degrees - one step (or current value 2 - 11 while decreasing by 10 (double step). I just don’t want to turn on the device while the user wants to decrease dimmer value while device is off and the current value is 1%.

ZWozniakS: I know that you are involved in the dimmer value discussion, I am not sure if there is any agreement on this topic, how the dimmer should work. Maybe it would be better if I wait for the results first and then make changes.

PR #11247, opened 2019-09-03
Iblind

SmartThings:
Iblind


dualtronic:
Can this be merged with the existing [Z-wave window shade](https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy) DTH? They are very similar.

PR #11253, opened 2019-09-04
DevWs

devropes:


MAblewicz (commented):
@devropes Could you check if your device will work with zigbee-range-extender DTH (https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zigbee-range-extender.src). Maybe just adding FP there is enough instead of creation of new DTH.

devropes:
@MAblewiczS Yes. I believe it should work. I've justed tested it out on my device and it seems alright. So could you just add my fingerprint profile to that DTH?

Also when this DTH is published, will it be made available under the Rooms Beautiful brand in the app?

devropes:
@MAblewiczS Do you know if my FP has been added the zigbee-range-extender DTH?

dkirker:
@devropes I created a PR here to add the Rooms Beautiful Range Extender to the Zigbee Range Extender DTH: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11579

It will pair with the deviceJoinName that you gave it, "Range Extender".

PR #11257, opened 2019-09-04
DevWs for RoomsBeautiful containing containing Curtain

devropes:


shrakuma:
@devropes  can you please let us know if you have tested this integration with the SmartThings new app?
The custom commands are not supported yet in the new app. 
cc @tpmanley @MAblewiczS @greens  can you please recommend if any, generic DTh that could handle this

devropes:
@shrakuma
Hi. The custom commands are meant to work for the "Smartthings Classic
App". They allow pause/continue of a curtain motor. I know these features
don't work yet in the new "Samsung Connect app", but they are crucial
functions of my device. I haven't yet found a way to to accomplish this in
the "Samsung Connect App". So I'll likely direct customers to use the
"Smartthings Classic App" for full functionality of the Curtain device.
However, the open/close, refresh, and battery functions will work just fine
in the new "Samsung Connect App". I've thoroughly tested this, and the
Curtain device works fine in both apps.

greens:
@devropes I think you'll find everything you need in the "Window Shade" capability as it is used here:

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy

My guess is that you could most likely even just add your device to the above DTH with minimal tinkering.

devropes:
@greens 
I've tried using the "Window Shade" capability. The only problem with it is that it always shows this error message on the Samsung Connect app as shown in the below image. That is why I couldn't use it.

Additionally, I checked the example DTH you provided, and the window shade functionality was just the same as the switch. I could see the contPause functionality was separate as it was in my DTH.

Though my device supports the window shade cluster, it also works with the switch cluster as well for ZigBee.

@shrakuma
Even though custom commands aren't yet supported in the new app, they still work in the Smartthings classic app. The DTH provided by @greens seems to show that this is ok.

image

greens:
@devropes in order to use the Window Shade capability you would have to define `open()` and `close()` methods and send `"windowShade"` events.

I'm a bit confused by your comments about `zigbee-window-shade`, what problem did you have with it?

devropes:
@greens As shown in the screenshot in the previous comment, I got a "network error" when using your sample DTH. I did define open() & close(). Because of this "network error", I didn't use the "Window Shade" capability.

greens:
@devropes did you get that network error after adding the fingerprint for your device to the zigbee-window-shade DTH or did you switch DTHs after the fact? If you switched after it might not have configured correctly.

devropes:
@greens I tried both options and got the error.

I also want to add that my DTH also has a preference setting (for the Smartthings classic app) so that users can invert the direction of the curtain. There are a few important differences between the existing DTH and mine including the way the battery percent is handled, open/close commands, and pause/continue commands.

devropes:
@greens Do you know if the device handler is ready for publication yet?

shrakuma:
@devropes  We are running through some issues with the window shade. can we chat over email? Please email me at shravan.kumar@smartthings.com 
Additionally, would it be possible for you to ship one of your devices so that we can take a closer look at the DTH development on our end? If so, here's the shipment details - https://smartthings.developer.samsung.com/docs/devices/publishing/ship-your-device.html

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,249 @@
...
+		capability "Battery"
+		capability "Configuration"
+		capability "Refresh"
+		capability "Window Shade"
+		capability "Health Check"
+        
+        attribute("replay", "enum")
+        attribute("battLife", "enum")
+
+		command "pause"
```
dkirker: This is now part of the Window Shade capability, so there shouldn't be a need to explicitly declare it. We are, however, trying to improve how the controls are displayed in the SmartThings app (not Classic). Right now the tile brings up a list of controls (open, close, pause).

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,249 @@
...
+def installed() {
+	sendEvent(name: "supportedWindowShadeCommands", value: JsonOutput.toJson(["open", "close", "pause"]), displayed: true)
+	sendEvent(name: "battery", value: 100)
+	sendEvent(name: "battLife", value: "full")
+	response(refresh())
+}
+
+def updated() {
+	// Needed because updated() is being called twice
+    def time
```
dkirker: What you can actually do here if this is still an issue is:

```
def updated() {
    runIn(2, "finishConfiguration", [overwrite: true])
}

def finishConfiguration() {
        if (invert.value == false)
            response(normal())
        else if(invert.value == true)
            response(reverse())
}
```

The `runIn(2, ` will mean that `finishConfiguration` will run after 2 seconds (pick whatever timeout makes sense), and if `updated` is called multiple times before `finishConfiguration` runs then each new call will just replace the previous scheduled run of `finishConfiguration`.

An alternate is to just let it run multiple times.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,249 @@
...
+    	sendEvent(name: "switch", value: "off")
+        log.warn ("reverse-off")
+        zigbee.writeAttribute(0xFC00, 0x0000, DataType.BOOLEAN, 0x01)
+    }
+    else{
+    	sendEvent(name: "switch", value: "on")
+        log.warn ("reverse-on")
+        zigbee.writeAttribute(0xFC00, 0x0000, DataType.BOOLEAN, 0x01)
+    }
+}
```
dkirker: Do you also need to send updated windowShade events here?

dkirker (commented):
I left a few operational comments in-line without getting in to general code review.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker If I remove the lines that say attribute "replay" and "battLife" and command "pause", will there be any effect on how my DTH will perform? Also is my command "cont" a part of the Window Shade Capability, or do I still have manually define it?

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker I still have to define the sendEvents because if the user reverses the direction of the curtain, the app should update to the correct direction. I've updated the code to sendEvent to windowShade instead of switch now.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker I've updated this in the code now.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
dkirker: @devropes Go ahead and leave the two `attribute` lines. Those are still needed for using the device in Classic.

dkirker: One note: The general workflow for the Window Shade capability would be (for example):

=> `open()`
<= `windowShade: opening`
=> `pause()`
<= `windowShade: open`
*user decides to continue opening*
=> `open()`
<= `windowShade: opening`
=> `pause()`
<= `windowShade: open`
*user decides to close it some*
=> `close()`
<= `windowShade: closing`
*"ahh, that is good right there!"*
=> `pause()`
<= `windowShade: open`

This would work with this integration, correct?

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
dkirker: @devropes I'd go ahead and have the `switch` and `windowShade` events. That should keep everything happy.

dkirker:
@devropes Apologies for the delay. I added a couple of comments. Aside from that, I think it looks good!

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker Yes, that would work. Does my DTH implement that windowShade capability correctly?

dkirker: @devropes So as far as I can tell, it looks good. :)

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
dkirker: Actually, since you aren't declaring the Switch capability, then the changed code is fine as is.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,260 @@
...
+ *	Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *	for the specific language governing permissions and limitations under the License.
+ */
+
+import groovy.json.JsonOutput
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition(name: "Curtain", namespace: "Rooms Beautiful", author: "Alex Feng", ocfDeviceType: "oic.d.blind", mnmn: "SmartThings", vid: "generic-shade") {
```
dkirker: One last change: I am uploading some metadata that better aligns with your capability set. It will use the vid `generic-shade-4`. If you could please set to that then I think this will be all ready!

devropes: @dkirker I've just committed the change.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,260 @@
...
+ *	Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+ *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *	for the specific language governing permissions and limitations under the License.
+ */
+
+import groovy.json.JsonOutput
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+	definition(name: "Curtain", namespace: "Rooms Beautiful", author: "Alex Feng", ocfDeviceType: "oic.d.blind", mnmn: "SmartThings", vid: "generic-shade-4") {
```
dkirker: Please use `Rooms Beautiful Curtain` as the DTH name. You'll also need to change the file name from `devicetypes/rooms-beautiful/curtain.src/curtain.groovy ` to `devicetypes/rooms-beautiful/rooms-beautiful-curtain.src/rooms-beautiful-curtain.groovy `

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
```diff
@@ -0,0 +1,260 @@
...
+		capability "Battery"
+		capability "Configuration"
+		capability "Refresh"
+		capability "Window Shade"
+		capability "Health Check"
+        
+        attribute("replay", "enum")
+        attribute("battLife", "enum")
+
+		command "pause"
```
dkirker: Please go ahead and remove this line `command "pause"` since this is now in the Window Shade capability (but please leave both of the attribute lines above).

dkirker (commented):
@devropes I received a couple of other requests from a colleague. I've put two comments inline, and then there was also a request to update the device handler to use consistent indenting (whether it is 4 spaces or a tab -- we have been migrating to using tabs).

I'd say after that we will be good to go! Thanks!

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker I've now changed this in the DTH and committed it.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker I've now changed this in the DTH and committed it.

devicetypes/rooms-beautiful/curtain.src/curtain.groovy
devropes: @dkirker In regards to the indenting, on the Smartthings IDE, the indenting looks good, but on GitHub, it looks messed up a bit. I'm not sure how to fix that.

dkirker: @devropes I should be able to assist with that. If this is the old SmartThings IDE, it unfortunately mixes spaces and tabs. :(

devropes: @dkirker I use graph.api.smartthings.com as the IDE. Is this there a newer IDE than this?
And I can fix the tabbing directly on the github website for now.
Also, should I remove curtain.src/curtain.groovy from the repo?

devropes: @dkirker Actually I will probably need assistance with the indenting. It doesn't seem like editing in GitHub helps.

dkirker: @devropes I'll take care of the indenting. I can also deal with the duplicate files at the same time.

dkirker (approved):

dkirker:
@devropes I updated the indentation and removed the now duplicate curtain.groovy. If this looks good to you I will merge!

devropes:
@dkirker It all looks good. Thanks.

devropes:
Hi Donald,

Thanks for all your help. I wanted to ask when the Smartthings app will be
updated with the Rooms Beautiful product? I haven't seen my curtain in the
app yet.

Thanks,
Alex

PR #11261, opened 2019-09-06
[ICP-10695] Aeotec Wallmote: Wrong buttonNumber in button events

tpmanley:
The button events on the components were using the number of the button (1-4) when it should have always been 1. This was causing automations like those setup with Smart Lighting to not trigger.

https://smartthings.atlassian.net/browse/ICP-10695


tpmanley:
This bug affects PROD so I targeted this to `staging` with the intent of releasing it next week.

greens (approved):

kianooshkarami (approved):

PR #11270, opened 2019-09-10
ICP-8575 Update thermostatMode with supportedThermostatModes to work around OCF mapping bug

dkirker:


dkirker:
I marked this WIP because I need to test it today, and verify making sure that supportedThermostatModes is JSON.

dkirker:
Last time I tested this, it did *not* have the intended effect... Should probably just close this.

PR #11276, opened 2019-09-11
DevWs for Most Target Limited containing containing SmartSense Open/Closed Sensor

adurosmart:


MAblewicz (approved):
Apart of identation  (tabs instead of spaces) look ok.

PR #11277, opened 2019-09-11
DevWs for Most Target Limited containing containing ZLL RGBW Bulb

adurosmart:


tpmanley:
Hi @adurosmart , I recently merged some formatting changes to this DTH which is making it look like there are a ton of changes in this PR. I made this change with the hopes of having fewer conflicts when adding fingerprints going forward. I'd suggest updating your branch to the latest `master` branch and then re-adding your new fingerprint to the section for your product line.

PR #11279, opened 2019-09-11
WWST-5302 add eWeLink Motion Sensor fingerprint

BYDengS:
@greens @dkirker @tpmanley  please help review this PR. Thank you:)

### DIFF
--- a/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
+++ b/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
@@ -68,16 +68,19 @@ def installed(){
 def parse(String description) {
 	log.debug "description(): $description"
 	def map = zigbee.getEvent(description)
+	ZoneStatus zs
 	if (!map) {
 		if (description?.startsWith('zone status')) {
-			map = parseIasMessage(description)
+			zs = zigbee.parseZoneStatus(description)
+			map = parseIasMessage(zs)
 		} else {
 			def descMap = zigbee.parseDescriptionAsMap(description)
 			if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER) {
 				map = batteyHandler(description)
 			} else if (descMap?.clusterInt == zigbee.IAS_ZONE_CLUSTER && descMap.commandInt != 0x07 && descMap.value) {
 				log.debug "parseDescriptionAsMap: $descMap.value"
-				map = getMotionResult(descMap.value)
+				zs = new ZoneStatus(zigbee.convertToInt(descMap.value, 16))
+				map = parseIasMessage(zs)
 			}
 		}
 	}
@@ -101,8 +104,7 @@ def batteyHandler(String description){
 	return map
 }
 
-def parseIasMessage(String description) {
-	ZoneStatus zs = zigbee.parseZoneStatus(description)
+def parseIasMessage(ZoneStatus zs) {
 	Boolean motionActive = zs.isAlarm1Set() || zs.isAlarm2Set()
 	if (!supportsRestoreNotify()) {
 		if (motionActive) {
--- a/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
+++ b/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
@@ -25,6 +25,7 @@ metadata {
 		capability "Refresh"
 		capability "Health Check"
 		capability "Sensor"
+		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0500", outClusters: "0003", manufacturer: "eWeLink", model: "MS01", deviceJoinName: "eWeLink Motion Sensor"
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001", manufacturer:"ORVIBO", model:"895a2d80097f4ae2b2d40500d5e03dcc", deviceJoinName: "Orvibo Motion Sensor"
         fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001,FFFF", manufacturer:"Megaman", model:"PS601/z1", deviceJoinName: "INGENIUM ZB PIR Sensor"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
```diff
@@ -67,9 +68,14 @@ def installed(){
 def parse(String description) {
 	log.debug "description(): $description"
 	def map = zigbee.getEvent(description)
+	def manufacturer = getDataValue("manufacturer")
 	if(!map){
 		if (description?.startsWith('zone status')) {
-			map = parseIasMessage(description)
+			if (manufacturer == "eWeLink"){
```
tpmanley: I think we may be able to avoid this manufacturer specific check which would be preferable. Does the eWeLink report the ZoneStatus message with the Restore Notify (bit 5) set? If so, and the other devices that use this DTH have Restore Notify cleared then we can just look at that bit to know whether we need to report motion stopped automatically or if the device will do it.

tpmanley: @Duncankn can you let us know whether the PS601/z1 reports ZoneStatus with Restore Notify cleared?

Duncankn: @tpmanley our device doesn't support Zone status with restore notify

BYDengS: @tpmanley  I don't know well about the " Restore Notify", for the eWeLink device, if there is no motion during 1 min, the device will report 0x0000, and if there is detected motion, device will report 0x0001, is there anything I should modify?

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
```diff
@@ -104,6 +110,18 @@ def parseIasMessage(String description) {
 	return getMotionResult(motionActive)
 }
 
+def parseWeLinKMotion(String description) {
```
tpmanley: Rather than creating this new function, I'd like to see an update to the existing `parseIasMessage`. Since the eWeLink link device does report when motion stops but doesn't set the Restore Notify bit in the ZoneStatus message you'll need to do something like this:

```
def parseIasMessage(String description) {
	ZoneStatus zs = zigbee.parseZoneStatus(description)
	Boolean motionActive = zs.isAlarm1Set() || zs.isAlarm2Set()
	if (!(zs.isRestoreReportsSet() || supportsRestoreNotify())) {
		if (motionActive) {
			def timeout = 20
			log.debug "Stopping motion in ${timeout} seconds"
			runIn(timeout, stopMotion)
		}
	}
	return getMotionResult(motionActive)
}

def supportsRestoreNotify() {
     getDataValue("manufacturer") == "eWeLink"
}
```

Implementing it in this way makes it easier for us to add support for additional devices to this DTH in the future.

BYDengS:
@tpmanley I have update the DTH, and test it with real Device, it looks no problem, please have a code review, thanks.

BYDengS:
@tpmanley I have update the DTH, and test it with real Device, it looks no problem, please have a code review, thanks.

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
```diff
@@ -67,11 +68,17 @@ def installed(){
...
+	if (!map) {
 		if (description?.startsWith('zone status')) {
 			map = parseIasMessage(description)
 		} else {
-			map = batteyHandler(description)
+			def descMap = zigbee.parseDescriptionAsMap(description)
+			if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER) {
+				map = batteyHandler(description)
+			} else if (descMap?.clusterInt == zigbee.IAS_ZONE_CLUSTER && descMap.commandInt != 0x07 && descMap.value) {
+				log.debug "parseDescriptionAsMap: $descMap.value"
```
tpmanley, line 81: It's better to refactor `parseIasMessage ` and this a bit so you parse the value into a `ZoneStatus`: `def zs = new ZoneStatus(zigbee.convertToInt(descMap.value, 16))` and then pass it into a common function for interpreting the ZoneStatus. That way, whether it receives a ZoneStatusChangeNotification or a ZoneStatus attribute report or read attribute response it will handle the message in the same way (including scheduling the motion stopped event for devices that need it).

BYDengS:
@tpmanley  I update it, plz review.

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
```diff
@@ -67,11 +68,20 @@ def installed(){
...
 		if (description?.startsWith('zone status')) {
-			map = parseIasMessage(description)
+			zs = zigbee.parseZoneStatus(description)
+			map = parseIasMessage(zs)
 		} else {
-			map = batteyHandler(description)
+			def descMap = zigbee.parseDescriptionAsMap(description)
+			if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER) {
+				map = batteyHandler(description)
+			} else if (descMap?.clusterInt == zigbee.IAS_ZONE_CLUSTER && descMap.commandInt != 0x07 && descMap.value) {
```
varzac, line 80: This check should probably be 
`descMap?.clusterInt == zigbee.IAS_ZONE_CLUSTER && (descMap.commandInt == 0x01 || descMap.commandInt == 0x0A)` as the following code is only valid if this is a read attr or report attr message.

varzac (approved):

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
tpmanley: I agree but I know there are a lot of DTHs that do the same thing. I think we can leave it for now and could fix them all up at the same time

tpmanley (approved):

PR #11280, opened 2019-09-11
DevWs for Most Target Limited containing containing ZLL White Color Temperature Bulb  and 2 more

adurosmart:


tpmanley:
@adurosmart can you update your local branch to the latest `master` and then re-add your new fingerprints to the appropriate section? I recently reorganized these DTHs to avoid future merge conflicts so you need to grab those latest changes.

PR #11285, opened 2019-09-12
[WWST-5165] Everspring Radiator Thermostat integration

PKacprowiczS:
metadata for this device is at WWST-5519
please share your thoughts

### DIFF
--- a/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
@@ -1,0 +1,262 @@
+/**
+ *	Copyright 2019 SmartThings
+ *
+ *	Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *	in compliance with the License. You may obtain a copy of the License at:
+ *
+ *		http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *	Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+
+ *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *	for the specific language governing permissions and limitations under the License.
+ *
+ */
+metadata {
+	definition (name: "Z-Wave Radiator Thermostat", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.thermostat") {
+		capability "Thermostat Mode"
+		capability "Refresh"
+		capability "Battery"
+		capability "Thermostat Heating Setpoint"
+		capability "Health Check"
+		capability "Thermostat"
+		capability "Temperature Measurement"
+
+		command "setThermostatSetpointUp"
+		command "setThermostatSetpointDown"
+		command "switchMode"
+
+		fingerprint mfr: "0060", prod: "0015", model: "0001", deviceJoinName: "Everspring Thermostatic Radiator Valve", mnmn: "SmartThings", vid: "generic-radiator-thermostat"
+	}
+
+	tiles(scale: 2) {
+		multiAttributeTile(name:"thermostat", type:"general", width:6, height:4, canChangeIcon: false)  {
+			tileAttribute("device.heatingSetpoint", key: "VALUE_CONTROL") {
+				attributeState("VALUE_UP", action: "setThermostatSetpointUp")
+				attributeState("VALUE_DOWN", action: "setThermostatSetpointDown")
+			}
+			tileAttribute("device.thermostatMode", key: "PRIMARY_CONTROL") {
+				attributeState("off", action:"switchMode", nextState:"...", icon: "st.thermostat.heating-cooling-off")
+				attributeState("heat", action:"switchMode", nextState:"...", icon: "st.thermostat.heat")
+				attributeState("emergency heat", action:"switchMode", nextState:"...", icon: "st.thermostat.emergency-heat")
+			}
+			tileAttribute("device.temperature", key: "SECONDARY_CONTROL") {
+				attributeState("temperature", label:'${currentValue}°', icon: "st.alarm.temperature.normal",
+						backgroundColors:[
+								// Celsius
+								[value: 0, color: "#153591"],
+								[value: 7, color: "#1e9cbb"],
+								[value: 15, color: "#90d2a7"],
+								[value: 23, color: "#44b621"],
+								[value: 28, color: "#f1d801"],
+								[value: 35, color: "#d04e00"],
+								[value: 37, color: "#bc2323"],
+								// Fahrenheit
+								[value: 40, color: "#153591"],
+								[value: 44, color: "#1e9cbb"],
+								[value: 59, color: "#90d2a7"],
+								[value: 74, color: "#44b621"],
+								[value: 84, color: "#f1d801"],
+								[value: 95, color: "#d04e00"],
+								[value: 96, color: "#bc2323"]
+						]
+				)
+			}
+		}
+
+		valueTile("battery", "device.battery", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+			state "battery", label: 'Battery:\n${currentValue}%', unit: "%"
+		}
+		standardTile("refresh", "command.refresh", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+			state "refresh", label: 'refresh', action: "refresh.refresh", icon: "st.secondary.refresh-icon"
+		}
+		main "thermostat"
+		details(["thermostat", "battery", "refresh"])
+	}
+}
+
+def initialize() {
+	sendEvent(name: "checkInterval", value: 4 * 60 * 60 + 24 * 60 , displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "supportedThermostatModes", value: thermostatSupportedModes, displayed: false)
+	response(refresh())
+}
+
+def installed() {
+	initialize()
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+
+def emergencyHeat() {
+	setThermostatMode("emergency heat")
+}
+
+def off() {
+	setThermostatMode("off")
+}
+
+def setHeatingSetpoint(setpoint) {
```
greens, line 200: just for safety's sake, you should enforce the setpoint limits here as well

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+			break
+	}
+	createEvent(map)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport cmd) {
+	createEvent(name: "heatingSetpoint", value: convertTemperatureIfNeeded(cmd.scaledValue, 'C', cmd.precision), unit: temperatureScale)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.sensormultilevelv5.SensorMultilevelReport cmd) {
```
greens, line 153: does the thermostat always report in Celsius, even if you change it to F mode (if possible)

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+	def map = [name: "temperature", value: convertTemperatureIfNeeded(cmd.scaledSensorValue, 'C', cmd.precision), unit: temperatureScale]
+	createEvent(map)
+}
+
+def zwaveEvent(physicalgraph.zwave.Command cmd) {
+	log.warn "Unhandled command: ${cmd}"
+	[:]
+}
+
+def setThermostatMode(String mode) {
```
dkirker, line 163: Since turning off the thermostat when an unsupported mode is sent in to this command would be bad behavior, you should enforce a check on mode. Like here: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/f6f0b3ff7b31c83db0c78883f8383ca48ab04cb1/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy#L366

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+
+private getMinHeatingSetpointTemperature() {
+	temperatureScale == 'C' ? 10 : 50
+}
+
+private getThermostatSupportedModes() {
+	if (isEverspringRadiatorThermostat()) {
+		["off", "heat", "emergency heat"]
+	} else {
+		[]
```
dkirker: We should probably have some sort of default here, even if there are no other fingerprints in the DTH, since this DTH is designed to be generic. I'd just have the else return `["off","heat"]`.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+		capability "Temperature Measurement"
+
+		command "setThermostatSetpointUp"
+		command "setThermostatSetpointDown"
+		command "switchMode"
+
+		fingerprint mfr: "0060", prod: "0015", model: "0001", deviceJoinName: "Everspring Thermostatic Radiator Valve", mnmn: "SmartThings", vid: "generic-radiator-thermostat"
+	}
+
+	tiles(scale: 2) {
```
dkirker, line 30: There's actually a better way to do the tiles now so that you don't need to do `switchMode` and `setThermostatSetpointUp` and `setThermostatSetpointDown`. Check out: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/f6f0b3ff7b31c83db0c78883f8383ca48ab04cb1/devicetypes/smartthings/zigbee-thermostat.src/zigbee-thermostat.groovy#L42

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,262 @@
...
+			state "refresh", label: 'refresh', action: "refresh.refresh", icon: "st.secondary.refresh-icon"
+		}
+		main "thermostat"
+		details(["thermostat", "battery", "refresh"])
+	}
+}
+
+def initialize() {
+	sendEvent(name: "checkInterval", value: 4 * 60 * 60 + 24 * 60 , displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "supportedThermostatModes", value: thermostatSupportedModes, displayed: false)
```
dkirker: Also call here
```
sendEvent(name: "heatingSetpointRange", value: [minHeatingSetpointTemperature, maxHeatingSetpointTemperature], displayed: false)
```

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
PKacprowiczS: right now we have only EU version of this device, afaik right now there's no US version available, so all instructions mention only Celsius and I don't see any option to change temperature unit as well

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
PKacprowiczS: It works well, but when I try to change thermostat mode, I see only one (heat) on the list, despite all of them being sent as an event (and all of them are listed in IDE in supportedThermostatModes list).

greens: any idea on this @dkirker ? does the metadata only specify one mode, maybe?

dkirker: Make sure the payload for `supportedThermostatModes` is encoded as JSON.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
dkirker: Also, encode supportedThermostatModes as JSON here and elsewhere. (Apologies for not catching that earlier.)

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
dkirker: Since this DTH is designed as a "generic" DTH, I would base it off of the scale that the device sends us. If the device for some reason is not including that data properly in the packet, then I would use `isEverspringRadiatorThermostat` to help set it, but not assume the scale to be "C". If someone dumps a new fingerprint in this DTH and isn't diligent they will hit a bug here.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -0,0 +1,254 @@
...
+
+def off() {
+	setThermostatMode("off")
+}
+
+def setHeatingSetpoint(setpoint) {
+	setpoint = temperatureScale == 'C' ? setpoint : fahrenheitToCelsius(setpoint)
+	setpoint = Math.max(Math.min(setpoint, maxHeatingSetpointTemperature), minHeatingSetpointTemperature)
+	[
+			secure(zwave.thermostatSetpointV2.thermostatSetpointSet([precision: 1, scale: 0, scaledValue: setpoint, setpointType: 1, size: 2])),
```
dkirker, line 204: @greens  Given my comment above about the scale, does it make sense to convert to a scale that we know the device is using, or just force 'C'? If another device is added to this DTH and it is in "F" mode, if it receives a thermostat setpoint set with a value where the scale is "C" will it "do the right thing"? (I assume yes without pulling up the specs, but you know what assuming leads to...)

dkirker (commented):
I left a couple of new comments.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
greens: I'd just leave a note near the fingerprints. Adding that conversion logic is a significant lift.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
PKacprowiczS: Yep, that've worked.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
PKacprowiczS: @greens I've added a note, but I'm not sure if it is what you've though about

greens (approved):

PR #11296, opened 2019-09-16
ICP-10783, Fix to handle notification report for Dawon switch on/off

lukhong:
WWST-5338

### DIFF
--- a/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
+++ b/devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
@@ -199,16 +199,6 @@ def zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerS
 	result << createEvent(descriptionText: "$device.displayName MSR: $msr", isStateChange: false)
 }
 
-def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cmd){
-	if ((cmd.notificationType == 0x08) && zwaveInfo?.mfr?.equals("018C")) {
-		if (cmd.event == 0x02) {
-			createEvent(name: "switch", value: "off")
-		} else if (cmd.event == 0x03) {
-			createEvent(name: "switch", value: "on")
-		}
-	}
-}
-
 def zwaveEvent(physicalgraph.zwave.Command cmd) {
 	log.debug "${device.displayName}: Unhandled: $cmd"
 	[:]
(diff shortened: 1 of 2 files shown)


greens:
Only issue I have with this is that it makes the `runLocally` untrue. @kianooshST : thoughts? This device for some reason chose not to use any of the switch command classes to report its state.

lukhong:
@kianooshST Do you have any opinion on this?

devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
```diff
@@ -199,6 +199,16 @@ def zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerS
 	result << createEvent(descriptionText: "$device.displayName MSR: $msr", isStateChange: false)
 }
 
+def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cmd){
+	if ((cmd.notificationType == 0x08) && zwaveInfo?.mfr?.equals("018C")) {
+		if (cmd.event == 0x02) {
```
kianooshkarami: Looking at the Notification Type of Power Management (0x08) and the following state/events:

0x02: AC Mains Disconnected
0x03: AC Mains re-connected

I think this makes sense. It would be great to add a comment in the code.

The only think I would ask is that, does it makes sense to have this  for all devices that have "018C" manufacturer or just do it for this particular device.

kianooshkarami (dismissed):

kianooshkarami:
> Only issue I have with this is that it makes the runLocally untrue

because of this reason, I would suggest to create a new DTH just for this device. @greens Thoughts?

greens:
I find it frustrating to make a whole new DTH for such a little thing, but I guess we should commit to actually be running locally when we say we are.

@lukhong can you instead make a new PR that is a copy of `zwave-metering-switch`, but with these changes and without the already-existing fingerprints?

kianooshkarami:
@greens I do not disagree with your assessment about a new DTH that is almost the same.

devicetypes/smartthings/zwave-metering-switch.src/zwave-metering-switch.groovy
lukhong: This Korean company is just started to work with SmartThings. Currently requested devices that have a switch seem to have the same approach. So I can say having 018C mfr in the code makes sense currently.

lukhong:
Of course, I can create a new DTH for this mfr's devices and will request a review.

lukhong:
@greens I've created a new DTH for Dawon device.
- Removed some codes existed for other mfr's.
- Changed name, added vid to use the existing ui-metadata of Z-Wave metering switch.
- Removed FPs in the generic DTH.

kianooshkarami (approved):

greens (approved):

PR #11301, opened 2019-09-16
ICP-10824 - Added fake 'no motion' event to pairing process for Aurora Smart PIR sensor

KKlimczukS:
Added fake 'no motion' event to pairing process for Aurora Smart PIR sensor. 
Removed 'Configuration' capability, because polling this device in pairing process didn't help.

@tpmanley @greens @MAblewiczS @ZWozniakS @MGoralczykS @PKacprowiczS 
Please take a look.

### DIFF
--- a/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
+++ b/devicetypes/smartthings/motion-detector.src/motion-detector.groovy
@@ -46,7 +46,7 @@ def installed() {
 	if(isAuroraMotionSensor51AU()) {
 		// Aurora Smart PIR Sensor doesn't report when there is no motion during pairing process
 		// reports are sent only if there is motion detected, so fake event is needed here
-		sendEvent(name: "motion", value: "inactive", descriptionText: "${device.displayName} - initially no motion" )
+		sendEvent(name: "motion", value: "inactive", displayed: false)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/motion-detector.src/motion-detector.groovy
```diff
@@ -44,6 +43,11 @@ metadata {
 
 def installed() {
 	initialize()
+	if(isAuroraMotionSensor51AU()) {
+		// Aurora Smart PIR Sensor doesn't report when there is no motion during pairing process
+		// reports are sent only if there is motion detected, so fake event is needed here
+		sendEvent(name: "motion", value: "inactive", descriptionText: "${device.displayName} - initially no motion" )
```
tpmanley: You don't need the `descriptionText` but you should set `displayed: false`

devicetypes/smartthings/motion-detector.src/motion-detector.groovy
```diff
@@ -85,11 +89,9 @@ def parse(String description) {
 }
 
 def isTracked() {
-	return device.getDataValue("model") == "MotionSensor51AU"
+	return isAuroraMotionSensor51AU()
 }
 
-def configure() {
```
tpmanley, line 91: This device uses IAS Zone to report motion, right? I expect you need `enrollResponse` to get it enrolled so wondering why that's not here.

KKlimczukS: In my self published DTH i used zigbee.iasZoneConfig(0,60) - enrollResponse() is called inside that method. Anyway, configure() method is removed here, because it doesn't make a difference (it wasn't in original DTH)

devicetypes/smartthings/motion-detector.src/motion-detector.groovy
KKlimczukS: ok, fixed

tpmanley (approved):

PR #11303, opened 2019-09-17
ICP-10896 Axis Gear pause support

greens:
The Axis Gear already had a command to stop the blinds, but this should hook it up to the Window Shade capability so that it can be used in OneApp.


tpmanley (approved):

dkirker:
Looks good! I should be deploying an updated generic-shade-2 to production today. (I think it might have had "pause" in the alternatives before a recent change, but there will be a new one.)

PR #11330, opened 2019-09-26
[ICP-11101] Child devices of hub connected devices not being added to room on join

tpmanley:
Some hub connected DTHs were not setting the hubId when creating child
devices. This was resulting in the child devices not being shown in the
join screen in the app which had the side effect that they were not being
added to the room the user selected when enabling join.

https://smartthings.atlassian.net/browse/ICP-11101

### DIFF
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
@@ -254,8 +254,14 @@ private channelNumber(String dni) {
 private void createChildDevices() {
 	state.oldLabel = device.label
 	for (i in 1..2) {
-		addChildDevice("Switch Child Device", "${device.deviceNetworkId}-ep${i}", null, [completedSetup: true, label: "${device.displayName} (CH${i})",
-																						 isComponent: true, componentName: "ep$i", componentLabel: "Channel $i"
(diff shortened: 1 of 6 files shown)


devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
```diff
@@ -254,8 +254,14 @@ private channelNumber(String dni) {
 private void createChildDevices() {
 	state.oldLabel = device.label
 	for (i in 1..2) {
-		addChildDevice("Switch Child Device", "${device.deviceNetworkId}-ep${i}", null, [completedSetup: true, label: "${device.displayName} (CH${i})",
-																						 isComponent: true, componentName: "ep$i", componentLabel: "Channel $i"
+		addChildDevice("Switch Child Device",
+				"${device.deviceNetworkId}-ep${i}",
+				device.hubId,
```
tpmanley, line 259: This is the main change. The rest is just formatting. Same in the other files

greens:
`zwave-device-multichannel` uses `device.getHub().getId()` and also has this bug so perhaps that's not correct and should also be changed to `device.hubId`?

tpmanley:
> zwave-device-multichannel uses device.getHub().getId() and also has this bug so perhaps that's not correct and should also be changed to device.hubId?

Those two should be equivalent. It might be a different issue.

greens (approved):

PR #11341, opened 2019-09-27
ICP-11103 - sending a supportedWindowShadeCommands event

KKlimczukS:
There were some changes on the plugin side, so sending a supportedWindowShadeCommands event is needed now for shades to work.

@tpmanley @MAblewiczS @ZWozniakS @MGoralczykS @PKacprowiczS 

Please, take a look.

### DIFF
--- a/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
+++ b/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
@@ -11,6 +11,7 @@
  *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
  *	for the specific language governing permissions and limitations under the License.
  */
+import groovy.json.JsonOutput
 import physicalgraph.zigbee.zcl.DataType
 
 metadata {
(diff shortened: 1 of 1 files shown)


dkirker (dismissed):

dkirker:
Do we need to get existing devices to send this event? We could either add a call in `parse()` to check if the attribute is set and if not send the event, or run a script to update it.

Also, does the mobile client not work if that attribute is not set?

cc @tpmanley

tpmanley:
If the plugin now requires this I would suggest creating the event in `installed` and also creating it in `parse` if it hasn't already been created. That way we won't have to run a script to update existing devices.

KKlimczukS:
@tpmanley @dkirker - according to Your suggestions - there are new changes in DTH.

devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
```diff
@@ -82,9 +83,20 @@ private List<Map> collectAttributes(Map descMap) {
...
+	log.debug "installed"
+	sendEvent(name: "supportedWindowShadeCommands", value: JsonOutput.toJson(["open", "close", "pause"]))
+}
+
 // Parse incoming device messages to generate events
 def parse(String description) {
 	log.debug "description:- ${description}"
+
+	// update existing devices
+	if (!state.supportedWindowShadeCommands) {
```
tpmanley, line 96: This will need to be ` if (!device.currentValue("supportedWindowShadeCommands"))`

KKlimczukS: instead of  
`if (!state.supportedWindowShadeCommands) ` ?

KKlimczukS:
@tpmanley 
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11364/
includes changes from this PR. I think we can close this one and focus on PR: 11364

tpmanley:
I agree that we can close this PR. I addressed the same issue in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11355 and also generated the necessary event on all existing devices.

PR #11345, opened 2019-09-30
DVCSMP-3664 Fix NPE in zigbee metering plug

greens:


greens:
I believe this is picking up the config reporting response.

tpmanley (approved):

PR #11353, opened 2019-10-01
DevWs for ICL containing containing ZLL White Color Temperature Bulb

SmartThings:


tpmanley:
Can you rebase this change on the latest `master` branch and add the new fingerprint to a manufacturer-specific section similar to the existing fingerprints?

PR #11355, opened 2019-10-03
[CHAD-3868] Fix window shade commands not being visible on device view

tpmanley:
The latest device plugin expects that devices implementing the Window
Shade capability have created the supportedWindowShadeCommands event to
indicate which commands they support. If that event hasnt' been generated
then the plugin doesn't render the device properly.


dkirker (dismissed):

dkirker:
I did just notice that the Axis Gear DTH doesn't have the `displayed: false` for its event. Is this something we'd want to add? https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/a6f456e0968b741cd1b6f4528de0d86166670803/devicetypes/axis/axis-gear-st.src/axis-gear-st.groovy#L311

tpmanley:
> I did just notice that the Axis Gear DTH doesn't have the `displayed: false` for its event. Is this something we'd want to add?

done

greens (approved):

PR #11366, opened 2019-10-06
Ajax Online RGBCCT

SmartThings:
This is based on "Generic Zigbee Bulb"

Please note this my first handler so please let me know if there is anything I need to change.

Also, I would like to have the temperature control from 2200k to 6500k. (i don't have any coding knowledge). Please help me if you can. 

Thank you 

Ajay

### DIFF
--- a/devicetypes/smartthings/aj-rgbcct.src/aj-rgbcct.groovy
+++ b/devicetypes/smartthings/aj-rgbcct.src/aj-rgbcct.groovy
@@ -1,0 +1,272 @@
+/**
+ *  Copyright 2017 SmartThings
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License. You may obtain a copy of the License at:
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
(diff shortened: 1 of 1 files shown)


tpmanley:
@Ajax-online It appears you're trying to add two new fingerprints to the existing "Zigbee RGBW Bulb" DTH. To add a fingerprint the most important information to include is the Manufacturer and Model which are the values reported by the device in the ManufacturerName and ModelName attributes on the Basic cluster. Then add lines like this to this part of the existing DTH: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy#L38.

`fingerprint manufacturer: "<ManufacturerName>", model: "<ModelName>", deviceJoinName: "Ajax Light", mnmn: "SmartThings", vid: "generic-color-temperature-bulb-2200K-6500K" `

The `vid` is what specifies the color temperature range showed in the app. We'll actually need to create this one on our end because it doesn't currently exist. I'll let you know when that has been done.

tpmanley:
@Ajax-online I noticed the newer PR https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/32385 so I'm going to close this one. The same comment still applies.

Ajax-online:
Thank you 
I am still having difficulty

Ajay

PR #11368, opened 2019-10-07
[ICP-11183] Z-Wave Dimmer Switch Generic: Fix icon for the ilumin Dimmable Bulb

MGoralczykS:


tpmanley (approved):

tpmanley:
@MGoralczykS in the future can you make the commit message/PR title be a little more descriptive, similar to what I just changed it to? When changing a single DTH like this I like to follow the pattern of `[Ticket] <DTH Name>: <change summary>`. The benefit is at a glance we can see what files are being changed and get a sense of the scope of the change.

PR #11369, opened 2019-10-07
[ICP-11031] Sending meterGet for Watss after changing dimmer value

ZWozniakS:
@greens @dkirker Could you please review? 

### DIFF
--- a/devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy
+++ b/devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy
@@ -221,7 +221,8 @@ def setLevel(level, rate = null) {
 	if(level > 99) level = 99
 	encapSequence([
 		zwave.basicV1.basicSet(value: level),
-		zwave.switchMultilevelV1.switchMultilevelGet()
+		zwave.switchMultilevelV1.switchMultilevelGet(),
+		meterGet(scale: 2)
 	], 5000)
 }
 
--- a/devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy
+++ b/devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy
@@ -222,7 +222,6 @@ def setLevel(level, rate = null) {
 	encapSequence([
 		zwave.basicV1.basicSet(value: level),
 		zwave.switchMultilevelV1.switchMultilevelGet(),
-		meterGet(scale: 2)
 	], 5000)
 }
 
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-metering-dimmer.src/zwave-metering-dimmer.groovy
```diff
@@ -221,7 +221,8 @@ def setLevel(level, rate = null) {
 	if(level > 99) level = 99
 	encapSequence([
 		zwave.basicV1.basicSet(value: level),
-		zwave.switchMultilevelV1.switchMultilevelGet()
+		zwave.switchMultilevelV1.switchMultilevelGet(),
+		meterGet(scale: 2)
```
tpmanley: The `setLevel` command will be executed locally and not in the DTH on all hubs except for Hub v1

cc @kianooshST

greens: Yeah, @ZWozniakS this isn't going to fix your problem because this command is run locally.

Important question: when you tested with this, did the device send the updated values when you polled them like this? Something we've run into before is that devices just don't measure these values that frequently, or they do so on a timer. If the device hasn't updated its own value internally there's even less we could do.

ZWozniakS: I was checking this with @KKlimczukS and it works (with runLocally and executeCommandsLocally flags set on false) but still, there will always be a delay after the dimmer value is set, here it is ~10 seconds. When I set the delay interval to 2 seconds it was working faster as expected but not all the time because of lags in the cloud... I don't think it is possible to update this value with an immediate effect. Maybe this change, if possible, should be done in the code that works in local execution mode?

greens: Based on other code in this DTH, we are setting a reporting interval of 5 minutes. It looks like that code was written for other Aeotec devices, but would still execute for this one. You'll need to edit it to be more model-specific, and add additional configuration for these devices based on this: https://products.z-wavealliance.org/products/2095/configs

Specifically, I think you'll want to set parameter 102 to 4, 103 to 8, 112 to 300, and 113 to 300. You may also need to tinker with the values of parameters 91 and 92.

ZWozniakS: What is the idea of setting parameters for groups 2 and 3? Would this change anything when the report interval is still set to 5 minutes? I think that it is still much too long if we want to have an impact on the frequency of reporting watts changes. But it looks quite good with the parameters 91 and 92. I have set 91 to 5 and 92 to 1% and the result is that when I am changing the dimmer value I am getting the new watts report faster but in two stages. For example, the dimmer is set to 21% and the Power Meter is 10 W, then I am changing dimmer to 96% and while it is being set I am getting two Power Meter values - first 54 W and then 59 W (and the second one is final). I am not sure if this is satisfying us.

greens: If it works, it's fine with me.

The idea with setting parameters for groups 2 and 3 is that the docs made it seem like the device doesn't have a group 1, or is now using group 1 for something else.

ZWozniakS: Ok, but in fact, we are getting watts reports from group 1 so I am not sure if we want to double that with report from group 2. Maybe reports from group 1 and 3 (with kWh) will be enough?

greens (approved):

PR #11380, opened 2019-10-09
[WWST-4627] New DTH for Qubino On/Off Flush Thermostat 2

PKacprowiczS:

### DIFF
--- a/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
+++ b/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
@@ -13,7 +13,7 @@
  *
  */
 metadata {
-	definition (name: "Qubino Flush Thermostat", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "qubino-thermostat", ocfDeviceType: "oic.d.thermostat") {
+	definition (name: "Qubino Flush Thermostat", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.thermostat") {
 		capability "Thermostat"
 		capability "Thermostat Mode"
 		capability "Thermostat Heating Setpoint"
@@ -317,7 +317,7 @@ def changeMode() {
 }
 
 private secure(cmd) {
-	if (zwaveInfo.zw.endsWith("s")) {
+	if (zwaveInfo.zw.contains("s")) {
 		zwave.securityV1.securityMessageEncapsulation().encapsulate(cmd).format()
 	} else {
 		cmd.format()
--- a/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
+++ b/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
@@ -1,0 +1,325 @@
+/**
+ *	Copyright 2019 SmartThings
+ *
+ *	Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ *	in compliance with the License. You may obtain a copy of the License at:
+ *
+ *		http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *	Unless required by applicable law or agreed to in writing, software distributed under the License is distributed
+
+ *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *	for the specific language governing permissions and limitations under the License.
+ *
+ */
+metadata {
+	definition (name: "Qubino Flush Thermostat", namespace: "smartthings", author: "SmartThings", mnmn: "SmartThings", vid: "qubino-thermostat", ocfDeviceType: "oic.d.thermostat") {
+		capability "Thermostat"
+		capability "Thermostat Mode"
+		capability "Thermostat Heating Setpoint"
+		capability "Thermostat Cooling Setpoint"
+		capability "Thermostat Operating State"
+		capability "Temperature Measurement"
+		capability "Power Meter"
+		capability "Energy Meter"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -0,0 +1,325 @@
...
+
+def zwaveEvent(physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport cmd) {
+	def map = [:]
+	switch (cmd.setpointType) {
+		case physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport.SETPOINT_TYPE_HEATING_1:
+			map.name = "heatingSetpoint"
+			break
+		case physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport.SETPOINT_TYPE_COOLING_1:
+			map.name = "coolingSetpoint"
+			break
```
kianooshkarami, line 173: How about if the `setPointType` is something else?

PKacprowiczS: This device supports only those two, so if somehow setPointType is different than expected, this method would return nothing useful.
By asking, you mean to add some warning about not supported setPoint, or...?

devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -0,0 +1,325 @@
...
+	def deviceTemperatureScale = cmd.scale ? 'F' : 'C'
+	createEvent(name: "temperature", value: convertTemperatureIfNeeded(cmd.scaledSensorValue, deviceTemperatureScale, cmd.precision), unit: temperatureScale)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	state.supportedModes = ["off"]
+	//this device doesn't act like normal thermostat, it can support either 'cool' or 'heat' after configuration
+	if (cmd.parameterNumber == 59 && !state.isThermostatModeSet) {
+		state.supportedModes.add(cmd.scaledConfigurationValue ? "cool" : "heat")
+		state.noSupportedMode = cmd.scaledConfigurationValue ? "heat" : "cool"
```
greens: this doesn't seem to be used elsewhere

devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -0,0 +1,325 @@
...
+		valueTile("power", "device.power", width: 2, height: 2) {
+			state "default", label:'${currentValue} W'
+		}
+		valueTile("energy", "device.energy", width: 2, height: 2) {
+			state "default", label:'${currentValue} kWh'
+		}
+		standardTile("changeMode", "device.changeMode", width: 2 , height: 2, inactiveLabel: false, decoration: "flat") {
+			state("default", action: "changeMode", label: 'Push to switch mode', nextState: "unpair")
+			state("unpair", label: 'Unpair and pair device again')
+		}
```
greens, line 98: this should be a preference instead

Why do they need to re-pair? Does updating `supportedThermostatModes` not work? And does the config value persist across remove/re-pair?

greens: or at least a preference as well as a tile

PKacprowiczS: Quote from device manual: 
“This parameter determines how the device will operate if it will operate in the heating mode or in the cooling mode. (...) After parameter change, first exclude device (without setting parameters to defaultvalue) and then re include the device!”
Device can't support three modes at single setup, we need to re-configure it if we want it to work in 'cool' mode ('heat' is default)

greens: I'm wondering if they say that in their documentation because they don't expect controllers to be able to handle it. It's entirely possible that it would work fine for us.

PKacprowiczS: I've tested it, and it works only using steps defined by Qubino in their manual. After changing parameter value and polling for Supported Modes there's no change. 
Only after re-pairing we're able to use another mode.

devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -0,0 +1,339 @@
...
+def changeMode() {
+	if (state.supportedModes.contains("heat")) {
+		sendHubCommand(zwave.configurationV1.configurationSet(parameterNumber: 59, scaledConfigurationValue: 1))
+	} else {
+		sendHubCommand(zwave.configurationV1.configurationSet(parameterNumber: 59, scaledConfigurationValue: 0))
+	}
+}
+
+private secure(cmd) {
+	if (zwaveInfo.zw.endsWith("s")) {
```
greens: This needs to be `contains` now because of changes made to the zw object.

At some point I need to go back and update all the old DTHs as well.

greens:
Did you change the vid because you just uploaded device-specific metadata or because you're relying on the fallback metadata?

PKacprowiczS:
@greens I've followed @dkirker suggestion and I've changed device specific metadata vid to _SmartThings-smartthings-Qubino_Flush_Thermostat_

greens (approved):
I'm approving this, but I'm expecting us to have a lot of trouble passing cert.

PR #11387, opened 2019-10-10
DevWs for GLEDOPTO CO.,LTD containing containing aaron  and 1 more

mosanwen:


MAblewicz:
Device Type Handler for that device is not Ready. All method that are repsonsible for Handling Zigbee  empty (there is just Log.debug used). Please send proper DTH with next request.  Cosider using generic DTH by adding just fingerprint. We suggest  using https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zigbee-rgb-bulb.src  or https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zll-rgb-bulb.src DTH.

MAblewicz:
Also create new Hub Connected SmartThing Compatible review using developer workspace with new version of DTH.

PR #11390, opened 2019-10-11
[ICP-11070] Modified zwave-siren DTH, add capability Tamper for Everspring Siren.

MGoralczykS:
@greens @KKlimczukS @ZWozniakS @PKacprowiczS 
Could you review this?


greens:
A better solution would be to add the tamper capability to the Z-Wave Siren DTH and then just set different metadata for devices that support it vs those that do not.

kianooshkarami (commented):
Is this going to show tamper for all Sirens?

MGoralczykS:
I think it isn't. It's going to show tamper for sirens which have it in metadata (only for OneApp), and for all sirens on Classic.

MGoralczykS:
@kianooshST, @greens Could you check it?

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -394,6 +403,13 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 		}
 		result << createEvent([name: "switch", value: isActive ? "on" : "off", displayed: true])
 		result << createEvent([name: "alarm", value: isActive ? "both" : "off", displayed: true])
+	} else if (cmd.notificationType == 0x07) { //Tamper Alert
+		switch (cmd.event) {
+			case 0x03:
+				result << createEvent([name: "tamper", value: "detected"])
+				result << createEvent([name: "alarm", value: "both"])
```
greens: The way I interpreted the table you posted is that the device will send an event parameter when tamper is activated, but will omit it when the tamper switch has been released (cleared).

Is that what the device sends?

MGoralczykS: Yes, it is

greens: Then the event with the payload should cause tamper detected and the event with no payload should cause tamper clear.

MGoralczykS: So there is only an option to clear tamper by off event. What should we do in this situation?

greens: You can leave the current tamper clear in the "off" event, but I think you should be handling both of the tamper events the device claims to send in the documentation you showed me.

MGoralczykS:
Also, I have a question, at 407 line should I add an exception for the Everspring device?

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -394,6 +403,14 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 		}
 		result << createEvent([name: "switch", value: isActive ? "on" : "off", displayed: true])
 		result << createEvent([name: "alarm", value: isActive ? "both" : "off", displayed: true])
+	} else if (cmd.notificationType == 0x07) { //Tamper Alert
+		switch (cmd.event) {
+			case 0x00: //Tamper switch is pressed more than 3 sec
+			case 0x03: //Tamper switch is pressed more than 3 sec and released
+				result << createEvent([name: "tamper", value: "detected"])
+				result << createEvent([name: "alarm", value: "both"])
```
greens: I don't think you understand my comment. Or maybe I didn't understand the documentation. It is my impression that the 0x03 event is meant to be a tamper clear and the 0x00 event is meant to indicate tamper. Do I have that correct, or is that not your impression?

MGoralczykS: I did it wrong in previous PR. 

I understand this in this way: 
When I hold tamper (0x00) it should be clear/clear tampered status, when I recive 0x03 I should turn on tamper.
If I hang the siren, tamper will be pressed more than 3sec(status clear), tamper should be turned on when somebody(release tamper button) will try to take it.

greens: Okay. I'll let you make the changes then.

MGoralczykS: Ok I need you help. 
I made some changes in notificationReport : 
`else if (cmd.notificationType == 0x07) { //Tamper Alert
    	if(cmd.event == 0x00) {
    		log.debug "0x00 ------------------"
                        result << secure(zwave.basicV1.basicSet(value: 0x00))
			result << "delay 3000"
			result << secure(zwave.basicV1.basicGet())
                        result += "delay 3000"
                        result += sendEvent(name: "tamper", value: "clear")
        	        result += sendEvent(name: "alarm", value: "off")
         } else if(cmd.event == 0x03) {
    		log.debug "0x03 ------------------"
			result += createEvent([name: "tamper", value: "detected"])
			result += createEvent([name: "alarm", value: "both"])
    	}
	}`

And this part of code 
`               result << secure(zwave.basicV1.basicSet(value: 0x00))
		result << "delay 3000"
		result << secure(zwave.basicV1.basicGet())
`
is taken from off() and here it doesn't work. When I press button and call off() (from app) device turn off. 

Statuses in app are changing but on siren it doesn't.

greens: Well for one you've put `sendEvent` where I think you meant to put `createEvent`. You've also mixed the operators `<<` and `+=`. I don't think that matters, but for style reasons you should be consistent at least in the same function.

I think the thing I'm having trouble with is when the device sends the different alarm notifications. When you "tamper" the device the alarm also turns on? But when you "untamper" the device, the alarm does not turn off, so you're trying to do that via z-wave?

greens: Perhaps the device only allows you to clear tamper manually (by pressing a button on the device)?

If so, I might have been mistaken in telling you to handle both notifications.

MGoralczykS: The alarm turns off but light signals don't. 
About operators I know it was just for my tests. 

When I cleared tamper manually (pressing off()) it was working correctly. I will try with those  notifications.

MGoralczykS: I checked this with operators, with createEvent and there was the same result all the time. I can only turn off light effect when manually call off().

greens: Okay. Based on what you've described now, your original way of handling it was correct.

greens: Sorry.

MGoralczykS: so I should make a commit where handle only 0x03 event and reset status after manual calling off()?

greens: No I think what you had before where you only cleared tamper after you received an "off" from the device.

MGoralczykS:

greens (approved):

greens (changes requested):
I guess I misremembered how this was previously written. I thought you sent clear when the device sends "off", not when the user presses "off".

Would that work?

MGoralczykS:
@greens That won't work. It only works when I clear tampe after user press "off".

greens:
I thought you said in your previous version that pressing off only turned off the siren but not the lights?

MGoralczykS:
In that way, yes :)

greens:
@PKacprowiczS can you help me with this? I don't think I understand how this device's clearing of tamper works.

PKacprowiczS:
@greens device's tamper itself works like this:
- when device's cover is taken away, tamper alert is being sent, lights and sound goes on
- when we put cover back and press the button to 'arm' a device, tamper is cleared, sound is turned off, but lights are still going on

We can turn off the lights by sending off() command, and Michal's initial approach was to send it automatically when we're receiving tamper clear message (another words, device have been armed).
Although, after a talk we've decided that forcing user to manually turn off lights by user is better option, as he will be reviewing a potential threat by himself.

MGoralczykS:
@greens I was talking with Przemek and Konrad and the result from our conversation is -> tamper should be cleared only when the user hangs the siren on the wall and tamper button is pushed (on device). In other cases, tamper status should be always "on".

greens (approved):

PR #11391, opened 2019-10-11
[ICP-11308] Send motion event in installed() to fix pairing without status

MGoralczykS:
@greens @KKlimczukS @ZWozniakS 
Could you review this?

### DIFF
--- a/devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
+++ b/devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
@@ -67,7 +67,6 @@ def installed() {
 // Device wakes up every 4 hours, this interval allows us to miss one wakeup notification before marking offline
 	sendEvent(name: "checkInterval", value: 8 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	sendEvent(name: "tamper", value: "clear", displayed: false)
-	sendEvent(name: "motion", value: "inactive", displayed: false)
 	response(initialPoll())
 }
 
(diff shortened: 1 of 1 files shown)


greens:
There's already an initial poll in this DTH. Is that command not being sent or not being responded to? Perhaps this device uses a different command class to report motion? I'd prefer we investigate that before we synthesize the event.

MGoralczykS:

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -273,6 +273,7 @@ def initialPoll() {
 	}
 	request << zwave.batteryV1.batteryGet()
 	request << zwave.sensorBinaryV2.sensorBinaryGet(sensorType: 0x0C) //motion
+	request << zwave.notificationV3.notificationGet(notificationType: 0x07, event: 0x08) //motion for Everspiring
```
kianooshkarami, line 276: Technically, we are using Notification CC V3 here and the event "0x07: Motion Detection" is for V2 of the command class and later and event "0x08: Motion Detection" was added in V4.

My preference would be to use to 0x07, but odes the everSpring motion sensor respond to that instead of 0x08?

MGoralczykS: When I set 0x07 device still pairing without motion status. With 0x08 this issue doesn't occur.

kianooshkarami: Strange implementation.

kianooshkarami (approved):

greens (approved):

PR #11392, opened 2019-10-11
Ikea bulb

Lars1983:
Ikea bulb


tpmanley:
@Lars1983 There are a lot of whitespaces changes but the only code change is the removal of `fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0300, FC0F", outClusters: "0019", manufacturer: "OSRAM", model: "LIGHTIFY Edge-lit flushmount", deviceJoinName: "SYLVANIA SMART+ Flush Mount"`. Was that what you intended or where you trying to add a new fingerprint?

PR #11412, opened 2019-10-15
DevWs for GLEDOPTO CO.,LTD containing containing aaron

mosanwen:


MAblewicz:
Device Type Handler for that device is not Ready. All method that are responsible for Handling Zigbee messages empty (there is just Log.debug inside). Please send proper DTH with next request. Cosider using generic DTH by adding just fingerprint. We suggest using https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zigbee-rgb-bulb.src or https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zll-rgb-bulb.src DTH.

MAblewicz (changes requested):
Please check my comments here or in Developers Workspace

PR #11418, opened 2019-10-15
[WWST-5851] DTH for Aeotec Range Extender AU region

BSSandeepSharma:
Only region details are added in comment


kianooshkarami (approved):
I went through the https://products.z-wavealliance.org/products and validated each entry. 

Go ahead and merge

greens:
@kianooshST don't forget to squash commits in the future

kianooshkarami:
I prefer to maintain separate commits when each commit is self contained for the functionality added.

greens:
Then it should have had each Jira ticket listed.

kianooshkarami:
It does..., Try clicking each commit(PR) in this case.

PR #11425, opened 2019-10-15
DevWs for Ajaxonline Ltd containing containing Ajax Filament

SmartThings:


MAblewicz (changes requested):
@Ajax-online 
Please add you fingerprint to generic DTH  (add only fingerprint lines there) without copying entire files and providing it as new one (devicetypes/ajax-online/aj-rgbcct.src/aj-rgbcct.groovy). Currently you are not only adding you device to ecosystem but also duplicating fingerprints ofother partners.
Probably you were changing: https://github.com/SmartThingsCommunity/SmartThingsPublic/tree/master/devicetypes/smartthings/zigbee-white-color-temperature-bulb.src
Just add your fingerprints to correct file and publish it using Developers Workspace.

PR #11431, opened 2019-10-16
ICP-10136 - calling sendHubCommand() once with all commands (for all endpoints)

KKlimczukS:
@greens @kianooshST @MAblewiczS @ZWozniakS @MGoralczykS @PKacprowiczS 

Code refactoring - before: sendHubCommand () was invoked many times with a set of commands for each endpoint separately (the delay only worked on one separate sendHubCommand() call)
Now it is called once with a set of commands for all endpoints (the delay separates all commands at once).

Please, could You take a look ?


greens:
Looks good, but will wait for verification from QA.

greens (approved):

PR #11440, opened 2019-10-18
[ICP-10897] Added updateColor() method in configure()

ZWozniakS:
@greens @tpmanley could you please review this? 

### DIFF
--- a/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
@@ -204,6 +204,7 @@ def configure() {
 
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
 	refresh()
+	updateColor()
 }
 
 def setColorTemperature(value) {
--- a/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
@@ -75,7 +75,7 @@ metadata {
 
 		// Sengled
 		fingerprint profileId: "0104", inClusters: "0000,0003,0004,0005,0006,0008,0300,0702,0B05,FC03,FC04", outClusters: "0019", manufacturer: "sengled", model: "E11-N1EA", deviceJoinName: "Sengled Multicolor"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -204,6 +204,7 @@ def configure() {
 
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
 	refresh()
+	updateColor()
```
tpmanley: You need to return a list of all the commands you want to send to the device. As written this will cause the `refresh` commands to not be sent so the device will not be configured.

Calling `updateColor` here will not fix the issue. All that does is generate hue and saturation events based on the value in `state` but there won't be any values there when the device is first configured.

What I suggest you try instead is read the Color Mode attribute (attr id 0x0008) in `refresh` right after reading the saturation. Then add handling for the ReadAttributeResponse to `parse`. If the value is 0x00 (CurrentHue and CurrentSaturation) then call `updateColor` (will likely need to use `runIn`). If the value is 0x02 then generate a colorTemperature event with the current color temperature. This is going to require careful testing and likely some tweaking to get it working correctly.

ZWozniakS: @tpmanley thank you kindly for the review! I've done what you were asking but the issue still occurred. I wrote more in a comment in Jira ticket.

ZWozniakS: @tpmanley I've added the last changes by adding initial settings in installed for this bulb. Could you please review? 
And thanks for the support! :)

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -150,6 +151,15 @@ def parse(String description) {
 				state.saturationValue = Math.round(zigbee.convertHexToInt(zigbeeMap.value) / 0xfe * 100)
 				runIn(5, updateColor, [overwrite: true])
 			}
+			else if (zigbeeMap.attrInt == ATTRIBUTE_COLOR_MODE){ //Color Mode Attribute
```
tpmanley: Since this didn't work I think it's best to remove it

ZWozniakS: Sure, done!

tpmanley (approved):

PR #11441, opened 2019-10-18
[ONEAPP-18302] Z-Wave Siren: Fix infinite loop in initialization

kianooshkarami:
If a Z-Wave siren using this DTH fails to initialize correctly or drops offline, the DTH attempts will generate 3 or more commands every 12 seconds attempting to initialize the device correctly. This leads to the hub's buffer overflowing since it will exhaust its resources attempting to communicate the messages that are bound to fail leading to the eventual queue becoming full and dropping messages.

As a result, I have chosen to cap the attempt to initialize the device after 3 attempts.

https://smartthings.atlassian.net/browse/ONEAPP-18302

### DIFF
--- a/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
+++ b/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
@@ -84,23 +84,46 @@ metadata {
 	}
 }
 
+// Perform a periodic check to ensure that initialization of the device was successful
+def getINIT_VERIFY_CHECK_PERIODIC_SECS() {30}
+def getINIT_VERIFY_CHECK_MAX_ATTEMPTS() {3}
+
 def installed() {
 	log.debug "installed()"
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	state.initializeAttempts = 0
 	initialize()
 }
 
 def updated() {
 	log.debug "updated()"
 	state.configured = false
+	state.initializeAttempts = 0
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
-	runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])
+	log.debug "updated(): Schedule in ${INIT_VERIFY_CHECK_PERIODIC_SECS} secs to verify initilization"
+	runIn(INIT_VERIFY_CHECK_PERIODIC_SECS, "initializeCallback", [overwrite: true, forceForLocallyExecuting: true])
+}
+
+def initializeCallback() {
+	log.debug "initializeCallback()"
+	state.initializeVerifyTimerPending = false
+	initialize()
 }
 
 def initialize() {
-	log.debug "initialize()"
+	if (state.initializeVerifyTimerPending) {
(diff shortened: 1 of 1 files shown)


dkirker (dismissed):

sangsoolee (dismissed):
I confirmed that it works as intended.

jaemin11lee (dismissed):

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -84,23 +84,41 @@ metadata {
 	}
 }
 
+// Perform a periodic check to ensure that initialization of the device was successful
+def getInitVerifyCheckPeriodicSecs() {
```
tpmanley: We typically name these functions in all caps to make them look like constants, for example `getINIT_VERIFY_CHECK_PERIOD_SECS`

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -84,23 +84,41 @@ metadata {
...
+
+def getInitVerifyCheckMaxAttempts() {
+	3
+}
+
 def installed() {
 	log.debug "installed()"
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	state.initPeriodCheckAttempts = 0;
```
tpmanley: nit: `;` can be removed

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
```diff
@@ -84,23 +84,41 @@ metadata {
...
 	initialize()
 }
 
 def updated() {
 	log.debug "updated()"
 	state.configured = false
+	state.initPeriodCheckAttempts = 0;
 	// Device-Watch simply pings if no device events received for 122min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, isStateChanged: true, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
-	runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])
```
tpmanley, line 99: Why did you remove the `runIn`? Part of the reason it's there is because `updated` can be called multiple times and this ensures the commands in `initialize` are only sent once (due to `overwrite: true`)

kianooshkarami: The previous implementation would trigger an initialization step in 12 seconds. I removed the 12 second and call `initialize()` right away. This in my opinion is cleaner since the delay in the first place was needed to perform initialization and also there will not be a timer created if there is no need for initialization. 

In `initialize()`, the `runIn()` will still perform `overwrite: True` when creating the timer. `runIn(initVerifyCheckPeriodicSecs, "initialize", [overwrite: true, forceForLocallyExecuting: true])`

Perhaps, your question is that if `updated` is called while `runIn()` timer is pending, in that case, we sort of do double initialization. I guess I can revert it to `runIn(12, "initialize", [overwrite: true, forceForLocallyExecuting: true])` at `updated()` for that case. 

If there is

devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy
kianooshkarami: Aaah, good point. Missed it.

cbaumler (dismissed):

cbaumler (dismissed):

cbaumler (dismissed):
LGTM

greens (dismissed):
Looks good to me. Deployment is tomorrow anyway so there's no reason to hotfix to prod.

I've linked an older issue I had that describes this one to your ticket.

tpmanley (dismissed):

tpmanley (approved):

PR #11451, opened 2019-10-22
[ICP-10361] Fibaro Heat Controller - additional temperature sensor connectivity check

PKacprowiczS:

### DIFF
--- a/devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
+++ b/devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
@@ -309,10 +309,10 @@ def switchMode() {
 	}
 }
 
-def sendEventToChild(event) {
+def sendEventToChild(event, forced = false) {
 	String childDni = "${device.deviceNetworkId}:2"
 	def child = childDevices.find { it.deviceNetworkId == childDni }
-	if (state.isChildOnline)
+	if (state.isChildOnline || forced)
 		child?.sendEvent(event)
 }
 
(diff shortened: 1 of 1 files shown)


greens (dismissed):
Looks good pending verification.

dkirker (dismissed):

kianooshkarami (dismissed):

devicetypes/smartthings/fibaro-heat-controller.src/fibaro-heat-controller.groovy
```diff
@@ -303,10 +309,10 @@ def switchMode() {
 	}
 }
 
-def sendEventToChild(event) {
+def sendEventToChild(event, forced = false) {
```
greens, line 312: Can you explain why this was needed? Were there battery reports causing the device to be marked online or something?

PKacprowiczS: More like the opposite. If device is marked as offline, _sendEventToChild()_ won't send 'online' event due _isChildOnline_ flag being set to false. At least I'm thinking that it would be this way, cause, as I wrote on JIRA, currently I'm unable to set device offline.

greens: I'm just wondering the circumstances you're experiencing where you need to _not_ send a message. The one place where you're forcing the send is immediately after the device _should_ have been marked online anyway, so the "force" flag seems unnecessary.

PKacprowiczS: I don't want temperature messages being sent to child when it's already marked as offline.
I'm forcing only events related to device health, and for simplicity sake, I'm doing it for all of them.

greens (approved):

PR #11478, opened 2019-10-28
[ICP-11209] Added on command to the setColorTemp()

ZWozniakS:

### DIFF
--- a/devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
+++ b/devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
@@ -192,9 +192,14 @@ def setColorTemperature(temp) {
 	def warmValue = temp < 5000 ? 255 : 0
 	def coldValue = temp >= 5000 ? 255 : 0
 	def parameterNumber = temp < 5000 ? WARM_WHITE_CONFIG : COLD_WHITE_CONFIG
-	def cmds = [zwave.configurationV1.configurationSet([parameterNumber: parameterNumber, size: 2, scaledConfigurationValue: temp]),
-				zwave.switchColorV3.switchColorSet(warmWhite: warmValue, coldWhite: coldValue)]
-	commands(cmds) + "delay 7000" + commands(queryAllColors(), 500)
+	def results = []
+	results << zwave.configurationV1.configurationSet([parameterNumber: parameterNumber, size: 2, scaledConfigurationValue: temp])
+	results << zwave.switchColorV3.switchColorSet(warmWhite: warmValue, coldWhite: coldValue)
+	if (device.currentValue("switch") != "on") {
+		results << zwave.basicV1.basicSet(value: 0xFF)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-multiwhite-bulb.src/aeon-multiwhite-bulb.groovy
```diff
@@ -192,9 +192,14 @@ def setColorTemperature(temp) {
...
 	def coldValue = temp >= 5000 ? 255 : 0
 	def parameterNumber = temp < 5000 ? WARM_WHITE_CONFIG : COLD_WHITE_CONFIG
-	def cmds = [zwave.configurationV1.configurationSet([parameterNumber: parameterNumber, size: 2, scaledConfigurationValue: temp]),
-				zwave.switchColorV3.switchColorSet(warmWhite: warmValue, coldWhite: coldValue)]
-	commands(cmds) + "delay 7000" + commands(queryAllColors(), 500)
+	def results = []
+	results << zwave.configurationV1.configurationSet([parameterNumber: parameterNumber, size: 2, scaledConfigurationValue: temp])
+	results << zwave.switchColorV3.switchColorSet(warmWhite: warmValue, coldWhite: coldValue)
+	if (device.currentValue("switch") != "on") {
+		results << zwave.basicV1.basicSet(value: 0xFF)
```
greens, line 199: My instinct would be that the `on` should come before the color set.

dkirker: How does the device behave in either scenario? If I do the color set followed by on does it come on with the color that was set just prior? I imagine if I turn on and then do color set the bulb comes on at the last color and then changes?

ZWozniakS: Yes, that is correct @dkirker. That is why I put "on" command after setting the color temperature, because then the bulb immediately turns on the light at the selected color temperature, without waiting for a change from the previous value. 
The same logic guided me here: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/11479. If I put "on" command before setting the color, first the bulb would turn on with the last color and then changes to which I choose. I can imagine that this situation would be reported as a bug. But when I put "on" command after, the bulb switch to the desired color value right away.

ZWozniakS: @greens, you mentioned in Jira ticket that there were problems with zigbee devices in similar cases so if you think that putting "on" command first would cause less problems I can change it. We can simply inform testers that this is how it should work to avoid more serious issues.

greens: Well I guess if we don't have any device where it's misbehaving yet this is fine. We can always update it later if we run into one of those devices where it doesn't work.

greens (approved):

PR #11483, opened 2019-10-28
ICP-10136 - Zooz Power Strip: calling sendHubCommand() once with all …

greens:
…commands (for all endpoints) (#11431)

* ICP-10136 - calling sendHubCommand() once with all commands (for all endpoints)

* ICP-10136 - decreased delay between commands send from refresh() to 200 milliseconds


greens:
Going to be cherry-picking this all the way up to prod because this DTH is currently broken on prod and is throwing a ton of exceptions.

tpmanley (approved):

PR #11484, opened 2019-10-28
ICP-10136 - Zooz Power Strip: calling sendHubCommand() once with all …

greens:
…commands (for all endpoints) (#11431) (#11483)

* ICP-10136 - calling sendHubCommand() once with all commands (for all endpoints)

* ICP-10136 - decreased delay between commands send from refresh() to 200 milliseconds


tpmanley (approved):
Changes look reasonable. I'm assuming this has been well tested though.

PR #11497, opened 2019-11-05
WWST-4999 - Support for "Ikea Open/Close remote E1766"

KKlimczukS:
@tpmanley @greens @MAblewiczS @ZWozniakS @MGoralczykS @PKacprowiczS 

Support for "Ikea Open/Close remote E1766". This IKEA button operates on WINDOW COVERING CLUSTER (UP, DOWN, CLOSE), but only UP and DOWN reports are handled here (events are mapped as "pushed" for "button" capability - more details in WWST-4999). 

### DIFF
--- a/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
+++ b/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
@@ -29,8 +29,8 @@ metadata {
 
 		fingerprint inClusters: "0000, 0001, 0003, 0009, 0B05, 1000", outClusters: "0003, 0004, 0005, 0006, 0008, 0019, 1000", manufacturer: "IKEA of Sweden", model: "TRADFRI remote control", deviceJoinName: "IKEA TRÅDFRI Remote", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_Remote_Control"
 		fingerprint inClusters: "0000, 0001, 0003, 0009, 0102, 1000, FC7C", outClusters: "0003, 0004, 0006, 0008, 0019, 0102, 1000", manufacturer:"IKEA of Sweden", model: "TRADFRI on/off switch", deviceJoinName: "IKEA TRÅDFRI On/Off switch", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_On/Off_Switch"
-		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI open/close remote", deviceJoinName: "Ikea Open/Close remote", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000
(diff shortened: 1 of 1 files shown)


greens:
How is this device intended to be used? Is it meant to be paired with Ikea blinds itself and then used to control them directly? I'm concerned that just mapping the up/down buttons to 1/2 isn't the best use of this device. 

We've done "virtual dimmers" before. Would it be more useful to make this a "virtual shade" so the metaphor of the remote is preserved? Would we even expect people to add this to our platform and not just directly with the shades its meant to control?

greens (commented):
Alright. I get it now. How's this working with the custom metadata? Does it display "up" and "down" in all the views/menus?

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -82,13 +89,22 @@ private getIkeaOnOffSwitchNames() {
 	]
 }
 
+private getIkeaOpenCloseRemoteNames() {
+	[
+			"Up", // Up button
+			"Down" // Down button
```
greens: nit: this is a bit too tabbed for me, personally

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -29,6 +29,7 @@ metadata {
 
 		fingerprint inClusters: "0000, 0001, 0003, 0009, 0B05, 1000", outClusters: "0003, 0004, 0005, 0006, 0008, 0019, 1000", manufacturer: "IKEA of Sweden", model: "TRADFRI remote control", deviceJoinName: "IKEA TRÅDFRI Remote", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_Remote_Control"
 		fingerprint inClusters: "0000, 0001, 0003, 0009, 0102, 1000, FC7C", outClusters: "0003, 0004, 0006, 0008, 0019, 0102, 1000", manufacturer:"IKEA of Sweden", model: "TRADFRI on/off switch", deviceJoinName: "IKEA TRÅDFRI On/Off switch", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_On/Off_Switch"
+		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI open/close remote", deviceJoinName: "Ikea Open/Close remote E1766", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000
```
tpmanley: We're seeing a lot of these devices join with a manufacturer name of just `KE`. We're still investigating why that's happening but I'd suggest including two fingerprints for now with both variants of the manufacturer.

KKlimczukS: ok, done

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
KKlimczukS: Fixed

tpmanley (dismissed):
Changes look good but I have one more question about the join name which I put on the ticket

tpmanley (approved):

maxi1134:
Hey guys! Thanks for this device handler, Although the current state of the button doesn't seem to be reported to SmartThings

image

PR #11653, opened 2019-11-10
DevWs for Most Target Limited containing containing Zigbee Multi Button

adurosmart:


adurosmart:
@greens  
Help me to review this commit.
Thank you very much.

greens (commented):
We've moved away from the `buttonNumber` style of multi-button device. Instead we add child buttons as components. Please see `zigbee-multi-button` and instead consider integrating the support for these devices into that DTH.

adurosmart:
@greens  i commit new code in adurosmart:devws-for-most-target-limited-co_136，my device can't support Battery capability,Sensor capability,Holdable Button capability,Sensor capability,Health Check capability.

PR #11740, opened 2019-11-13
DevWs for Most Target Limited containing containing ZigBee RGBW Bulb

adurosmart:


shrakuma:
@adurosmart there seems to be a lot of removals and re-adds. can you please let us know what FPs were being added here or does it need to be rebased?
cc @greens @MAblewiczS @tpmanley

adurosmart:
@greens @MAblewiczS @tpmanley  
No changes to the code, we just want to change the display name of the product, so can I close the request?

shrakuma:
@adurosmart if this change is needed. could you please rebase and submit just the change with the display name change in the PR

PR #11745, opened 2019-11-13
ICP-11490 ilumin bulb reporting intermediate level change values

greens:
I think we're just querying for the level value too soon after setting it. This just adds in a small static delay to account for transmission time.


greens:
Verified by Poland. This brings the delay into line with the other z-wave dimmer DTHs.

tpmanley (approved):

PR #11746, opened 2019-11-13
[ICP-10853] Axis Gear ST: Add support for Window Shade Preset in the new app

AXISLabsShare:


devicetypes/axis/axis-gear.src/axis-gear.groovy
```diff
@@ -0,0 +1,200 @@
+metadata {
+    definition (name: "AXIS Gear", namespace: "axis", author: "AXIS Labs") {
```
tpmanley: Did you intend to checkin this new DTH or only the changes to the current DTH?

AXISLabsShare: removed the old device handler. only the current DTH should have changes applied to it

tpmanley (approved):

PR #11963, opened 2019-11-14
DevWs for NIE-TECH CO., LTD. containing containing Z-Wave Dual Switch

sky-nie:


MAblewicz (changes requested):
Change identation from spaces to tabs on that line. Everything else looks fine.

MAblewicz (approved):
Now its good. Approved

PR #12031, opened 2019-11-14
[ICP-11467] Reduced delay time in on/off methods

ZWozniakS:
I couldn't find information about why this delay time was 3500 so I was trying carefully to find the highest lower value while this issue didn't occur.


greens:
I think you need to look at this one a little more. From the video it looks like the device is turning off, but then immediately turning back on. The delay shouldn't have anything to do with that.

ZWozniakS:
@greens I couldn't reproduce this issue again to make other changes, the same as cert tester couldn't, so we are closing this issue. It probably was issue with cloud.

PR #12253, opened 2019-11-16
Inovelli Switch

SmartThings:
Light Switch


tpmanley:
I'm going to close this PR because we're working directly with Inovelli on the integration.

PR #12333, opened 2019-11-20
[ICP-11638] Zigbee Metering Plug: fix honyar outlet's USB port can't be controlled issue

FenMei007S:
@tpmanley , honyar's outlet use endpont 2 to handle usb port, plz review it

### DIFF
--- a/devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
+++ b/devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
@@ -110,11 +110,20 @@ def parse(String description) {
 }
 
 def off() {
-    zigbee.off()
+    def cmds = zigbee.off()
+    if (device.getDataValue("model") == "HY0105") {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
```diff
@@ -111,10 +111,16 @@ def parse(String description) {
 
 def off() {
     zigbee.off()
+    if (device.getDataValue("model") == "HY0105") {
+	zigbee.command(zigbee.ONOFF_CLUSTER, 0x00, "", [destEndpoint: 0x02])
```
tpmanley: You need to return all the commands in a list. As written this command will be the only one executed and `zigbee.off()` will not be executed. Rather than sending two commands I think you could instead send a single command to endpoint 0xFF. That should turn off all the endpoints that support the OnOff cluster.

FenMei007S:
@tpmanley ,return cmds list, and i tryed using 0XFF, when turn on device, the device will turn on,then turn off, turn on again. returning cmds list have no this issue

tpmanley (approved):

PR #12334, opened 2019-11-20
[ICP-11598] POPP Radiator Thermostat 'always online' workaround

PKacprowiczS:
In this PR I'm preventing "cached setpoint" event from being sent whenever user wants to change it through app, now it'll be sent only if device is online.

This also for ICP-11632.
When changing setpoint manually on a device, it sends _ThermostatSetpointReport_ with chosen value, but then it also sends _WakeUpNotification_, where in result, user's manually set value is being overwritten. Small code fix in _ThermostatSetpointReport_ handler prevents it.

### DIFF
--- a/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
@@ -136,8 +136,10 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
 
 def zwaveEvent(physicalgraph.zwave.commands.multicmdv1.MultiCmdEncap cmd) {
 	cmd.encapsulatedCommands().collect { encapsulatedCommand ->
-		isPoppRadiatorThermostat() ? zwaveEvent(encapsulatedCommand, true) : zwaveEvent(encapsulatedCommand) 	//in case any future device would support MultiCmdEncap
-	}.flatten()																									//and won't need any special handler, like POPP does
+		isPoppRadiatorThermostat() ? zwaveEvent(encapsulatedCommand, true) : zwaveEvent(encapsulatedCommand) 	
+		//in case any future device would support MultiCmdEncap
+		//and won't need any special handler, like POPP does
+	}.flatten()
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -176,7 +176,9 @@ def zwaveEvent(physicalgraph.zwave.commands.thermostatmodev2.ThermostatModeRepor
 
 def zwaveEvent(physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport cmd) {
 	def deviceTemperatureScale = cmd.scale ? 'F' : 'C'
-	createEvent(name: "heatingSetpoint", value: convertTemperatureIfNeeded(cmd.scaledValue, deviceTemperatureScale, cmd.precision), unit: temperatureScale)
+	def setpoint = Float.parseFloat(convertTemperatureIfNeeded(cmd.scaledValue, deviceTemperatureScale, cmd.precision))
```
greens, line 182: any reason you have to do a type conversion here, but we didn't have to when just passing directly?

PKacprowiczS: type conversion is only really needed for _state.cachedSetpoint_, cause later it's used in Z-Wave commands and theirs constructors require numeric type values. 
Although, as _createEvent_ doesn't care whether it's string or float, I've used what I've got

PKacprowiczS:
I cannot set label here, but please do not merge this yet. I found out that my fix caused another bug... I need to figure this out.

PKacprowiczS:
@greens @dkirker @kianooshST please have a look at this
I think I've managed to get it working. However, DTH got really messy and I'm starting to think that separate DTH would be better.

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -174,9 +177,28 @@ def zwaveEvent(physicalgraph.zwave.commands.thermostatmodev2.ThermostatModeRepor
...
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.thermostatsetpointv2.ThermostatSetpointReport cmd, isResponseOfWakeUp) {
+	if (state.isSetpointChangeRequestedByController) {
+		state.isSetpointChangeRequestedByController = false
+		updateSetpoint(cmd)
+	} else {
+		[:]
+	}
```
greens: Could you combine these two into one handler with a default value for `isResponseOfWakeUp`?

greens (approved):

PR #12402, opened 2019-11-20
ICP-11609 send child events as the parent as well

greens:
for OneApp to change the status of the main device on the all devices view, events need to be duplicated to the parent


devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -130,15 +130,15 @@ def zwaveEvent(physicalgraph.zwave.commands.sceneactivationv1.SceneActivationSet
 	def description = "Button no. ${childId} was ${value}"
 	def event = createEvent(name: "button", value: value, descriptionText: description, data: [buttonNumber: childId], isStateChange: true)
 	sendEventToChild(childId, event)
-	return createEvent(descriptionText: description)
+	return createEvent(event)
```
PKacprowiczS: Cannot we just return _event_?

greens: good point

tpmanley (approved):

PR #13513, opened 2019-11-27
[WWST-5898] Add POPP Smoke Detectar with Siren to zwave-basic-smoke-alarm

MGoralczykS:
@greens @ZWozniakS @PKacprowiczS @KKlimczukS 
Could you review this?


greens:
Should we incorporate this into a DTH that uses the "Alarm" capability? Does the device provide manual control of the siren?

Or maybe we add "alarm" to this dth but specify metadata in such a way that it's only displayed for this device? What do you think?

MGoralczykS:
According to manual this "alarm" is smoke alarm. In my opinion we should leave it as it is or change name. What do you think?

greens (approved):

PR #13526, opened 2019-11-28
[WWST-6065]zigbee multi switch: add fingerprint for heiman 3 gang switch

FenMei007S:
@tpmanley , plz review it


tpmanley:
@FenMei007S I think `getChildCount` also needs to be updated

FenMei007S:
Updated

tpmanley (approved):

PR #13980, opened 2019-12-03
[WWST-6032] Zigbee Non-Holdable Button: New DTH for heiman emergency button

FenMei007S:
@tpmanley , plz review it


tpmanley:
@FenMei007S it looks like this device should work with the `Zigbee Button` DTH with some minimal changes which would be preferable over creating a new DTH. One thing to keep in mind is `Zigbee Button` runs locally so you'll need to test in that configuration to be sure it's really working. Feel free to reach out on Slack if you need assistance.

FenMei007S:
Update

FenMei007S:
For run local case, as i know, china hub is unable to run local. so i have no way to test it

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -64,7 +65,20 @@ def parse(String description) {
     log.debug "description is $description"
     def event = zigbee.getEvent(description)
     if (event) {
-        sendEvent(event)
+        if ((device.getDataValue("model") == "SOS-EM")) {
+            if (description?.startsWith('zone status')) {
```
varzac: This code won't be hit as this is in the branch where `getEvent` returned an event, and neither `zone status` messages nor battery are handled in the generic handling.

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -64,7 +65,20 @@ def parse(String description) {
...
+               def result = [:]
+               result = createEvent(name: "button", value: "pushed",descriptionText: "pushed", isStateChange: true)
+               sendEvent(result)
+            } else {
+               def map = batteyHandler(description)
```
varzac: typo: `batteyHandler` -> `batteryHandler`

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -91,6 +105,27 @@ def parse(String description) {
     }
 }
 
+def batteyHandler(String description){
+	def descMap = zigbee.parseDescriptionAsMap(description)
+	def map = [:]
+	if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.commandInt != 0x07 && descMap.value) {
```
varzac: Instead of just checking that this isn't `0x07` you should check that the command is either `0x01` for the read attr response or `0x0A` for a report attribute.  You should also add a check here that the `attrInt` is what you expect.  I'm guessing in this case from the usage below that it should be `0x0021` as it appears to be the battery percentage remaining.

varzac:
If this DTH is working when you test, it could be that you don't have to have custom handling for `zone status` messages.  Did the pre-existing `zone status` handling not work as expected for this device?

FenMei007S:
@varzac I updated as ur comment, it work well , plz review it

devicetypes/smartthings/zigbee-button.src/zigbee-button.groovy
```diff
@@ -77,7 +78,12 @@ def parse(String description) {
             }
         }
         else if (description?.startsWith('zone status')) {
-            event = parseIasButtonMessage(description)
+            if((device.getDataValue("model") == "SOS-EM")){
```
tpmanley: What happens if you don't include this change to `parse`? Does `parseIasButtonMessage` not do the right thing for this device?

FenMei007S: I checked the log, the device only report release event, no press event, so parseIasButtonMessage can't return a pushed event for it. this why i not let it be handled by parseIasButtonMessage

tpmanley: What is the `zone status` message when the button is pressed? The issue is we cannot make a change like this to `parse` because this DTH runs locally on most hubs so this code will not be executed.

FenMei007S: hm... i see description is zone status 0x0001 -- extended status 0x00 when press, but no log occurr when release

tpmanley: Unfortunately that means this DTH won't work for this device. However I think you should be able to use the `SmartSense Button` DTH instead and won't need these changes to `parse` (you will still need the changes to `configure`).

FenMei007S:
Updated

devicetypes/smartthings/smartsense-button.src/smartsense-button.groovy
```diff
@@ -262,18 +267,21 @@ def refresh() {
     } else {
         refreshCmds += zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020)
     }
-    refreshCmds += zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000) +
-        zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS) +
-        zigbee.enrollResponse()
 
+    if(device.getDataValue("manufacturer") != "HEIMAN") {
+        refreshCmds +=  zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000)
```
tpmanley: I missed that this device doesn't support temperature measurement. That unfortunately means we cannot use this DTH either. 

@varzac can you think of any other options besides a new DTH?

varzac: At this point in time no.  Anything else would be hacky and would require local changes anyways.

tpmanley: @FenMei007S I'm sorry this didn't work out but it seems like we have no choice but to create a new DTH for this device. Please try to match the Zigbee Button implementation as close as you can. Perhaps call it `Zigbee Non-Holdable Button`. I don't like that name so open to other ideas.

FenMei007S:
@tpmanley @varzac updated, Hope this one is okay

tpmanley:
@FenMei007S can you remove the changes to the existing files from this PR? I know the changes are minor but it will keep the noise down.

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,234 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations
+ *  under the License.
+ */
+import physicalgraph.zigbee.clusters.iaszone.ZoneStatus
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+    definition(name: "Zigbee Non-Holdable Button", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.022.0000', executeCommandsLocally: false, mnmn: "SmartThings", vid: "generic-button-4", ocfDeviceType: "x.com.st.d.remotecontroller") {
```
tpmanley: `runLocally` should be `false` and remove `minHubCoreVersion` and `executeCommandsLocally`

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,234 @@
...
+        capability "Sensor"
+
+        fingerprint profileId: "0104", inClusters: "0000, 0001, 0003, 0500", outClusters: "0019", manufacturer: "HEIMAN", model: "SOS-EM", deviceJoinName: "HEIMAN Emergency Button"
+    }
+
+    simulator {
+        status "button 1 pushed": "catchall: 0104 0500 01 01 0140 00 6C3F 00 00 0000 01 01 020000190100"
+    }
+
+    preferences {
```
tpmanley: Remove the `preferences` section

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,234 @@
...
+        capability "Battery"
+        capability "Refresh"
+        capability "Button"
+        capability "Health Check"
+        capability "Sensor"
+
+        fingerprint profileId: "0104", inClusters: "0000, 0001, 0003, 0500", outClusters: "0019", manufacturer: "HEIMAN", model: "SOS-EM", deviceJoinName: "HEIMAN Emergency Button"
+    }
+
+    simulator {
```
tpmanley: Remove the `simulator` section

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,234 @@
...
+
+private Map parseIasMessage(String description) {
+    ZoneStatus zs = zigbee.parseZoneStatus(description)
+
+    translateZoneStatus(zs)
+}
+
+private Map translateZoneStatus(ZoneStatus zs) {
+    if (zs.isAlarm1Set() && zs.isAlarm2Set()) {
+        return getButtonResult('held')
```
tpmanley: You can remove this part since this is a non-holdable button

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,234 @@
...
+    } else if (zs.isAlarm1Set()) {
+        return getButtonResult('pushed')
+    } else if (zs.isAlarm2Set()) {
+        return getButtonResult('double')
```
tpmanley: Can remove this too since it only supports `pushed`

FenMei007S:
@tpmanley  updated, plz review it again

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  License for the specific language governing permissions and limitations
+ *  under the License.
+ */
+import physicalgraph.zigbee.clusters.iaszone.ZoneStatus
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
+    definition(name: "Zigbee Non-Holdable Button", namespace: "smartthings", author: "SmartThings", runLocally: false, mnmn: "SmartThings", vid: "generic-button-4", ocfDeviceType: "x.com.st.d.remotecontroller") {
```
tpmanley, line 20: @dkirker `generic-button-4` includes the "Holdable Button" capability. I wonder, does it make any sense for that vid to include that capability or can we remove it? If we do need to keep it there, I think we should create new UI Metadata without that capability and use it here. Thoughts?

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+}
+
+private Map parseIasMessage(String description) {
+    ZoneStatus zs = zigbee.parseZoneStatus(description)
+
+    translateZoneStatus(zs)
+}
+
+private Map translateZoneStatus(ZoneStatus zs) {
+	if (zs.isAlarm1Set()) {
```
tpmanley, line 111: Check the indentation on this function

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+    // Device-Watch allows 2 check-in misses from device + ping (plus 1 min lag time)
+    // enrolls with default periodic reporting until newer 5 min interval is confirmed
+    // Sets up low battery threshold reporting
+    sendEvent(name: "DeviceWatch-Enroll", displayed: false, value: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, scheme: "TRACKED", checkInterval: 6 * 60 * 60 + 1 * 60, lowBatteryThresholds: [15, 7, 3], offlinePingable: "1"])
+
+    log.debug "Configuring Reporting"
+    def configCmds = []
+    configCmds += zigbee.batteryConfig()
+    
+    return refresh() + configCmds + refresh() // send refresh cmds as part of config
```
tpmanley: This will result in extra messages that aren't helpful. I suggest you instead return the following:

```
zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020) +
zigbee.enrollResponse() +
zigbee.batteryConfig()
```

That will read the current battery level first which is useful since that's visible to the user, then do the IAS Zone enrollment since the user will likely next check that the button works, and finally setup the battery configuration. Unless your testing shows otherwise, I don't think there is any need to read the IAS Zone status attribute at all.

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+        result.name = 'battery'
+        result.translatable = true
+        result.descriptionText = "{{ device.displayName }} battery was {{ value }}%"
+        result.value = Math.round(rawValue / 2)
+    }
+
+    return result
+}
+
+private Map getButtonResult(value) {
```
tpmanley, line 155: Since this DTH only supports `pushed` you can simplify it by removing the handling for held and double tap

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+                   zigbee.readAttribute(zigbee.IAS_ZONE_CLUSTER, zigbee.ATTRIBUTE_IAS_ZONE_STATUS) +
+                   zigbee.enrollResponse()
+    return refreshCmds
+}
+
+def configure() {
+    // Device-Watch allows 2 check-in misses from device + ping (plus 1 min lag time)
+    // enrolls with default periodic reporting until newer 5 min interval is confirmed
+    // Sets up low battery threshold reporting
+    sendEvent(name: "DeviceWatch-Enroll", displayed: false, value: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, scheme: "TRACKED", checkInterval: 6 * 60 * 60 + 1 * 60, lowBatteryThresholds: [15, 7, 3], offlinePingable: "1"])
```
tpmanley: Remove the `lowBatteryThresholds: [15, 7, 3]` part. That was only done for 1st party devices

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
```diff
@@ -0,0 +1,216 @@
...
+
+    def result = [:]
+
+    def volts = rawValue / 10
+
+    if (!(rawValue == 0 || rawValue == 255)) {
+        result.name = 'battery'
+        result.translatable = true
+        result.descriptionText = "${ device.displayName } battery was ${ value }%"
+        if (device.getDataValue("manufacturer") == "SmartThings") {
```
tpmanley: Remove this part since we don't have any 1st party devices using this DTH

devicetypes/smartthings/zigbee-non-holdable-button.src/zigbee-non-holdable-button.groovy
dkirker: Hmmmm I'm not sure that `Holdable Button` is a capability that is active anymore. It is marked deprecated. I don't *think* there would be a negative impact with removing it from the metadata. However, I'm not entirely certain. The `Button` capability basically eclipses `Holdable Button`.

tpmanley (approved):

PR #14449, opened 2019-12-06
WWST-6234 Fibaro Flood Sensor checkinterval is too short

greens:
The device's checkinterval was set to be identical to its reporting interval, which means that one missed checkin could mark the device offline.


greens:
A dev in HQ noticed this through inspection. Not crucial enough that I'm going to fix current installs, but will be good to have for new ones.

tpmanley (approved):

PR #15112, opened 2019-12-13
DVCSMP-3664 Clear up common exceptions.

greens:
Fibaro Motion Sensor - code was erroneously added checking the value of a preference that does not exist.

Zigbee Thermostat - reporting configured messages without values would cause exceptions to be thrown during parsing

Lock without codes - adds old-style OEM parsing to this DTH. The method call had been copy-pasted, but the method had not. Removes all references to codes from the original method in Z-Wave Lock.

### DIFF
--- a/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
@@ -446,8 +446,8 @@ private def handleAlarmReportUsingAlarmType(cmd) {
 		case 18: // Locked with keypad
 			codeID = readCodeSlotId(cmd)
 			map = [ name: "lock", value: "locked" ]
-            map.descriptionText = "Locked by keypad"
-            map.data = [ method: "keypad" ]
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -419,6 +419,103 @@ private def handleBatteryAlarmReport(cmd) {
...
+			break
+		case 16: // Note: for levers this means it's unlocked, for non-motorized deadbolt, it's just unsecured and might not get unlocked
+		case 19: // Unlocked with keypad
+			map = [ name: "lock", value: "unlocked" , method: "keypad"]
+			map.descriptionText = "Unlocked by keypad"
+			break
+		case 18: // Locked with keypad
+			codeID = readCodeSlotId(cmd)
+			map = [ name: "lock", value: "locked" ]
+            map.descriptionText = "Locked by keypad"
```
tpmanley: nit: indentation

greens:
ping

dkirker (approved):

PR #15257, opened 2019-12-16
[WWST-4627] Qubino Thermostat setpoint range change

PKacprowiczS:
Qubino has asked us to change heating setpoint range, so I did...


greens:
@dkirker doesn't this need a metadata change too...?

dkirker:
@greens  It appears that the metadata already has that range.

dkirker (approved):
At some point we should have this DTH send the events `heatingSetpointRange` and `coolingSetpointRange`. Otherwise, this looks good.

PR #15258, opened 2019-12-16
[ICP-11874]: zigbee co sensor: heiman sensor battery remain percentage not be shown.

FenMei007S:


FenMei007S:
@tpmanley , plz review it

tpmanley (approved):

PR #14372, opened 2019-12-06
[CHAD-4106] Set Association Group 1 at initialized() and once later if needed

kianooshkarami:
Ok Two Things: 

1. Adding Set Association to Group 1 at initialized() seems like a very reasonable change to help stop the bleeding. I have verified this myself
2. The second part is to fix the affected devices, which is to send Set Association when we hear anything from the devices, given that these devices are running locally, we would have to wait for them to report battery reports since that is not processed locally.

PR #15345, opened 2019-12-16
Revert "[CHAD-4106] Set Association Group 1 at initialized() and once…

greens:
… later if needed (#14372)"

This reverts commit e71cf45f8f974f0e5e395182c3a9f1e5791c5a94.


greens:
Reverting this in master so that it will be reverted on prod in two weeks, which should be enough time for most of the affected devices to have been cleaned up.

kianooshkarami:
This will be missed.

kianooshkarami (approved):
:(

greens:
this codebase ain't big enough for the both of us

PR #15737, opened 2019-12-19
XXX-9999 First stab at added Window Shade Level to Simulated Window Shade

dkirker:
Still a work in progress. I would like to better simulate opening and closing when you call setLevel. But ultimately, this should allow QA to test some new metadata. I'll create a ticket for this as well.

### DIFF
--- a/devicetypes/smartthings/testing/simulated-window-shade.src/simulated-window-shade.groovy
+++ b/devicetypes/smartthings/testing/simulated-window-shade.src/simulated-window-shade.groovy
@@ -16,7 +16,7 @@
 import groovy.json.JsonOutput
 
 metadata {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/testing/simulated-window-shade.src/simulated-window-shade.groovy
```diff
@@ -31,6 +33,18 @@ metadata {
 		command "opened"
 		command "closed"
 		command "unknown"
+
+		command "shadeLevel0"
```
tpmanley, line 37: What's the value in defining these custom commands?

dkirker: They were just shortcuts that one can use from Classic. Not sure that I have added some as tiles yet. I need to look back at this DTH.

PR #16055, opened 2019-12-23
[WWST-6077]zigbee sence keypad: add fingerprint for heiman sence keypad

FenMei007S:


FenMei007S:
@tpmanley  plz review it

tpmanley (approved):

PR #16545, opened 2019-12-30
[WWST-4328] Tint Remote Controller integration

PKacprowiczS:
@tpmanley @varzac @greens @dkirker can you guys have a look at this? 
@MAblewiczS @KKlimczukS @MGoralczykS @ZWozniakS

### DIFF
--- a/devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
+++ b/devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
@@ -25,7 +25,7 @@ metadata {
 		capability "Light"
 
 		// Muller Licht
-        fingerprint inClusters: "0000, 0003, 1000", outClusters: "0003, 0004, 0006, 0008, 0300, 1000, 0019, 0000", manufacturer: "MLI",  model: "ZBT-Remote-ALL-RGBW", deviceJoinName: "Tint Remote Control"
+		fingerprint inClusters: "0000, 0003, 1000", outClusters: "0003, 0004, 0006, 0008, 0300, 1000, 0019, 0000", manufacturer: "MLI",  model: "ZBT-Remote-ALL-RGBW", deviceJoinName: "Tint Remote Control"
 	}
 
 	tiles(scale: 2) {
@@ -61,31 +61,29 @@ private getSTEP_COMMAND() { 0x02 }
 private getWRITE_ATTRIBUTES_COMMAND() { 0x02 }
 
 def parse(String description) {
-    def event = zigbee.getEvent(description)
+	def event = zigbee.getEvent(description)
 	def result = []
-    if (event) {
-        log.debug "Creating event: ${event}"
-       result = createEvent(event)
-    } else {
-        if (isAttrOrCmdMessage(description)) {
-            result = parseAttrCmdMessage(description)
-        } else {
-            log.warn "Unhandled message came in"
-        }
-    }
+	if (event) {
+		log.debug "Creating event: ${event}"
+		result = createEvent(event)
+	} else {
+		if (isAttrOrCmdMessage(description)) {
+			result = parseAttrCmdMessage(description)
+		} else {
+			log.warn "Unhandled message came in"
+		}
+	}
 	log.debug "Parse returned: ${result}"
 	return result
 }
 
 private parseAttrCmdMessage(description) {
-    def descMap = zigbee.parseDescriptionAsMap(description)
-    if (descMap.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == BATTERY_VOLTAGE_ATTRIBUTE) {
-        getBatteryEvent(zigbee.convertHexToInt(descMap.value))
-    } else if (descMap.clusterInt == zigbee.ONOFF_CLUSTER) {
-        getOnOffEvent()
-    } else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_COMMAND) {
-        getColorEvent(descMap)
-    } else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_TEMPERATURE_COMMAND) {
+	def descMap = zigbee.parseDescriptionAsMap(description)
+	if (descMap.clusterInt == zigbee.ONOFF_CLUSTER) {
+		getOnOffEvent()
+	} else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_COMMAND) {
+		getColorEvent(descMap)
+	} else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_TEMPERATURE_COMMAND) {
 		getColorTemperatureEvent(descMap)
 	} else if (descMap.clusterInt == zigbee.LEVEL_CONTROL_CLUSTER && descMap.commandInt == STEP_COMMAND) {
 		getLevelEvent(descMap)
@@ -97,18 +95,18 @@ private parseAttrCmdMessage(description) {
 }
 
 private isAttrOrCmdMessage(description) {
-    (description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))
+	(description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))
 }
 
 private getOnOffEvent() {
-    state.onOffValue = (state.onOffValue == "off" ? "on" : "off")
-    createEvent(name: "switch", value: state.onOffValue)
+	state.onOffValue = (state.onOffValue == "off" ? "on" : "off")
+	createEvent(name: "switch", value: state.onOffValue)
 }
 
 private getColorEvent(descMap) {
-    def CurrentX = Integer.parseInt(descMap.data[1] + descMap.data[0], 16)
-    def CurrentY = Integer.parseInt(descMap.data[3] + descMap.data[2], 16)
-    def HS = color_xyYtoHSV(CurrentX, CurrentY)
+	def CurrentX = Integer.parseInt(descMap.data[1] + descMap.data[0], 16)
+	def CurrentY = Integer.parseInt(descMap.data[3] + descMap.data[2], 16)
+	def HS = color_xyYtoHSV(CurrentX, CurrentY)
 	return [
 		createEvent(name: "hue", value: HS.H * 100),
 		createEvent(name: "saturation", value: 100)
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,310 @@
...
+            log.warn "Unhandled message came in"
+        }
+    }
+	log.debug "Parse returned: ${result}"
+	return result
+}
+
+private parseAttrCmdMessage(description) {
+    def descMap = zigbee.parseDescriptionAsMap(description)
+    if (descMap.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == BATTERY_VOLTAGE_ATTRIBUTE) {
```
varzac: This doesn't appear to be a battery device. So this and the battery parsing can be removed

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,310 @@
...
+        getBatteryEvent(zigbee.convertHexToInt(descMap.value))
+    } else if (descMap.clusterInt == zigbee.ONOFF_CLUSTER) {
+        getOnOffEvent()
+    } else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_COMMAND) {
+        getColorEvent(descMap)
+    } else if (descMap.clusterInt == zigbee.COLOR_CONTROL_CLUSTER && descMap.commandInt == MOVE_TO_COLOR_TEMPERATURE_COMMAND) {
+		getColorTemperatureEvent(descMap)
+	} else if (descMap.clusterInt == zigbee.LEVEL_CONTROL_CLUSTER && descMap.commandInt == STEP_COMMAND) {
```
varzac, line 88: Check your indentation

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,310 @@
...
+		log.debug "Unhandled message came in, its description map: ${descMap}"
+	}
+}
+
+private isAttrOrCmdMessage(description) {
+    (description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))
+}
+
+private getOnOffEvent() {
+    state.onOffValue = (state.onOffValue == "off" ? "on" : "off")
```
varzac: This device is always a toggle?  It's never just a straight on or off command?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,310 @@
...
+    createEvent(name: "switch", value: state.onOffValue)
+}
+
+private getColorEvent(descMap) {
+    def CurrentX = Integer.parseInt(descMap.data[1] + descMap.data[0], 16)
+    def CurrentY = Integer.parseInt(descMap.data[3] + descMap.data[2], 16)
+    def HS = color_xyYtoHSV(CurrentX, CurrentY)
+	return [
+		createEvent(name: "hue", value: HS.H * 100),
+		createEvent(name: "saturation", value: 100)
```
varzac, line 112: The above conversion function also includes the saturation value, shouldn't you use that value instead of always 100?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,310 @@
...
+}
+
+def configure() {
+	def groupsList = [0x4004, 0x4005, 0x4006]
+		groupsList.each {
+			sendHubCommand(addHubToGroup(it))
+		}
+	sendEvent(name: "checkInterval", value: (2 * 60 + 10) * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+
+	return zigbee.onOffConfig(5, 3600)
```
varzac, line 164: Do you need to configure color temperature or level?

PKacprowiczS: Nah, I've configured on/off just for health check purposes. Which seems to doesn't work anyway. Any idea what configuration I've could use?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
PKacprowiczS: This conversion is returning 100 saturation _almost_ always, anyway. It has some issues with mismatched colors around blue/purple, so to keep consistence between colors on remote and reports (I'm assuming some device issue here, cause same mismatch occurs when I pair remote to Tint's bulb), I've decided to hard-code saturation. However, I've also thought it won't hurt to keep whole conversion for future reference.

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
PKacprowiczS: From device point of view it's straight on/off command, but keeping in mind that user can switch it from mobile app, I've decided to assume that's toggle. Otherwise, any user action on device following switching in app wouldn't make any change.

PKacprowiczS:
I was thinking if it even make sense to implement color control, when we cannot prepare any automation with it, and as its deviceType is defined as "Remote Controller", we cannot add it to a lighting group, while on the other hand with deviceType set to "Light" MCD Button tiles don't show.

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,308 @@
...
+	R = R / max
+	G = G / max
+	B = B / max
+	R = R < 0 ? 0 : R
+	G = G < 0 ? 0 : G
+	B = B < 0 ? 0 : B
+	[R: R, G: G, B: B]
+}
+
+private colorRGBtoHSV(RGB) {
```
greens, line 229: We do have a `colorUtil.hexToHsv` already written that takes a 6 digit hex string and outputs HSV. See `rgbw-light`

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,308 @@
...
+	addChildButtons(numberOfButtons)
+	state.currentLevel = 100
+	sendEvent(name: "level", value: 100, displayed: false)
+	state.onOffValue = "off"
+	createEvent(name: "switch", value: "off", displayed: false)
+	createEvent(name: "colorTemperature", value: 2700, displayed: false)
+	sendEvent(name: "saturation", value: 100, descriptionText: "Color has changed", displayed: false)
+}
+
+private color_xyYtoHSV(CurrentX, CurrentY) {
```
greens, line 193: just checking: xxY is a different color space than the xy one @dkirker did for ikea in zll-rgb-bulb, right?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,308 @@
...
+	} else {
+		log.debug "Unhandled message came in, its description map: ${descMap}"
+	}
+}
+
+private isAttrOrCmdMessage(description) {
+	(description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))
+}
+
+private getOnOffEvent() {
```
greens, line 101: What if the device was turned on/off while the hub was offline? Is the device itself not sending us the actual current state info?

I'm generally a bit confused about how much we can actually interact with this device. It seems like the answer is "almost not at all". Can you explain a bit?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
```diff
@@ -0,0 +1,308 @@
...
+	def CurrentX = Integer.parseInt(descMap.data[1] + descMap.data[0], 16)
+	def CurrentY = Integer.parseInt(descMap.data[3] + descMap.data[2], 16)
+	def HS = color_xyYtoHSV(CurrentX, CurrentY)
+	return [
+		createEvent(name: "hue", value: HS.H * 100),
+		createEvent(name: "saturation", value: 100)
+	]
+}
+
+private getColorTemperatureEvent(descMap) {
```
greens, line 116: could you just use the built-in `zigbee.getEvent` return value here and then clamp it?

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
dkirker: xyY is xyz but in only 2D (http://www.colorwiki.com/wiki/XyY). I believe the function that I wrote in zll-rgb-bulb is xyY.

Keep in mind, the table M in the XyToRgb and RgbToXy functions may actually differ per device (so this might be what you would need to tweak so that you can use the saturation value instead of hard-coding to 100).

Another thing to keep in mind from a comment that I put in zll-rgb-bulb: "Not all XY color values are representable in rgb", and the conversion is not lossless.

@PKacprowiczS We should keep the two device handlers consistent with what they use. There is also an implementation of these conversions in HubCore. If the implementation has a bug, then it should be fixed in this PR, noted in the comments, and then a PR created to fix it in the other DTH as well as a ticket created so that we can fix it in HubCore.



Let's use the logic from zll-rgb-bulb for consistency. Though, we should use colorUtil.hslToRgb(h, s, l) and colorUtil.rgbToHsl(r, g, b) (we can probably even update zll-rgb-bulb and rgbw-light to use those, and that might help with a couple of bugs reported in zll-rgb-bulb's color conversion).

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
dkirker: I left a big comment below in regard to the xyY stuff and it also pertains to this thread.

dkirker:
I left a comment above for the xyY stuff, but the gist of it is:

Let's use the logic from zll-rgb-bulb for consistency. Though, we should use colorUtil.hslToRgb(h, s, l) and colorUtil.rgbToHsl(r, g, b) (we can probably even update zll-rgb-bulb and rgbw-light to use those, and that might help with a couple of bugs reported in zll-rgb-bulb's color conversion).

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
PKacprowiczS: It is sending its current state with no issues. Problem occurs when user uses on/off switch on mobile app, and while we cannot trigger communication with device from hub side, it would cause "missed event issue" from folk's perspective like:
1) Click on/off button on remote - it has sent "off command", _off_ event has been created
2) Current state is _off_, select _on_ command on app in device page. As we cannot interact with device, remote's internal state is still _off_, while in app it's _on_
3) Once again click on/off button on remote - it will send "on command" and _on_ event will be returned, but as device's state in app is already _on_, it would seem that it had done nothing

And yep, there isn't a lot of interaction possible with this device. Basically speaking, it's almost like Z-Wave battery powered sleepy device.

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
PKacprowiczS: Based on my work with the device, I wouldn't go into much trouble with tweaking M matrix, as sending hardcoded 100 saturation might not look pretty, I'd say it's still doing it's job in creating accurate events. Although, I'll elaborate on this matter more on JIRA, cause I believe it would be easier to express myself with a help of a video.

devicetypes/smartthings/tint-remote-controller.src/tint-remote-controller.groovy
PKacprowiczS: Zigbee class covers only reading _ColorTemperature_, but not command which is setting it.

dkirker: `zigbee.getEvent` only handles attribute 0x0007 in the Color Control cluster, it seems, and as @PKacprowiczS not a command. :/

greens:
I almost feel like, with this minimal layer of interactivity, it would be better for us to just represent this as a multi-button remote and forget about trying to accurately reflect the device's on/off and color values. Those capabilities are expected to be two-way interactive and it doesn't seem like this model is going to quite cut it.

greens:
@PKacprowiczS you mentioned it's sleepy. Is it sleepy in the sense that it sends periodic wakeup? If so, is that wakeup frequent (or configurable) enough that we might try a mailboxing-type solution?

dkirker:
I know there is supposed to be work to add a system to allow an integration to indicate which parts of a capability it supports and how it supports them (i.e. allowing to have "stateless", write-only, capabilities without creating new capabilities; or allowing for capabilities that are read-only).

If the device itself is actually tracking the state internally, unlike the Zigbee (Battery) Accessory Dimmer, then we can just have any of the commands return the current state (so, `on()` and `off()` would just return `device.currentValue("switch")`). The mobile app won't like that, but oh well (until it has gotten away from OCF or we get the above capability feature in place). It's a little dirty, but it avoids having the device and the integration get out of sync. I imagine that this device is more meant to be used to control, rather than be controlled. We can certainly remove the device details page entries and the ruleAction entries for switch and color control (or at least make the device details entry non-controllable, if that actually works), that way it is only able to be a rule event (or even just show state).

This at least allows you to use the device in a SmartApp that takes the events from one device and sends them to other devices (that's how I use my Zigbee (Battery) Accessory Dimmer devices).

PKacprowiczS:
@greens unfortunately no, what I meant by comparing this remote to Z-Wave sleepy device, it's just the fact that it doesn't give a damn about what we're trying to send to it. Attempts to configure it to send any periodic events have burned on nothing.

@dkirker I'll work with metadata to see whether what you've said is possible and tell how it went.

PR #17851, opened 2020-01-08
ICP-11631 Lock does not respond to battery polls after battery replac…

greens:
…ement report

Trying to add a delay to test if perhaps there is a short period where the lock will not respond to requests.

### DIFF
--- a/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
@@ -405,6 +405,10 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch (cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
+			result << "delay 1200"
+			result << response(secure(zwave.batteryV1.batteryGet()))
+			break;
 		case 0x0A:
 			map = [name: "battery", value: 1, descriptionText: "Battery level critical", displayed: true, data: [lockName: deviceName]]
 			break
@@ -415,7 +419,7 @@ private def handleBatteryAlarmReport(cmd) {
 			map = [displayed: false, descriptionText: "Alarm event ${cmd.alarmType} level ${cmd.alarmLevel}"]
 			break
 	}
-	result << createEvent(map)
+	if (map != null) result << createEvent(map)
 	result
 }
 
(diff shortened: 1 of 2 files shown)


dkirker:
@greens Feel free to tag me on this for a review when it's done! (Not sure if you're done with it yet.)

greens:
This is based on feedback from Yale that basically confirmed there is a period after new batteries are inserted that the device fails to respond to requests. I had QA here and in SR-Pol verify this fix.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -405,6 +408,11 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch (cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
+			state.queryBattery = true
```
tpmanley: Seems like this DTH needs something that sets `queryBattery` to `false` similar to the other DTH

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -405,6 +408,11 @@ private def handleBatteryAlarmReport(cmd) {
...
+			result << "delay 1200"
```
kianooshkarami, line 413: What is significance of 1200 milliseconds?

Also, `handleBatteryAlarmReport` is handled where the message is received, here we are setting `queryBattery` to true and then we are generating another battery get because the flag is set in line 213. So essentially sending two battery gets.

devicetypes/smartthings/zwave-lock.src/zwave-lock.groovy
```diff
@@ -631,6 +634,8 @@ private def handleBatteryAlarmReport(cmd) {
 	def map = null
 	switch(cmd.zwaveAlarmEvent) {
 		case 0x01: //power has been applied, check if the battery level updated
+			state.queryBattery = true
```
kianooshkarami: See my comment about setting `queryBattery` to true here. and generating double Battery gets.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
greens: yep. thanks

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
greens: Yale told us that the device just doesn't respond after batteries are changed. The ideal solution is that if we just wait a second to query the battery we won't have to do it again later.

I see your point about the flag. I've tried to fix it by delaying 1 second before setting the flag. Thoughts?

kianooshkarami (approved):

kianooshkarami:
@greens I see you did not squash, not a big deal.

PR #17872, opened 2020-01-10
[WWST-5161] Added and integrated Everspring Wall Switch

ZWozniakS:
@greens Could you please review?


greens (changes requested):
the child buttons also have a line where we set their supported button values. You'll need to change that as well.

ZWozniakS:
Of course, I've done it in IDE but forgot to add here, sorry. Done!

dkirker (approved):

greens (approved):

PR #17908, opened 2020-01-13
[ICP-12162]HEIMAN Scene Button: remove un-used button UI of scene keypad plugin

ZouLianghuiS:
@tpmanley , @greens , please have a review of my PR, thanks

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -184,7 +184,7 @@ def initialize() {
 }
 
 private addChildButtons(numberOfButtons) {
-	for(def endpoint : 1..numberOfButtons) {
+	for(def endpoint : 2..numberOfButtons) {
 		try {
 			String childDni = "${device.deviceNetworkId}:$endpoint"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -184,7 +184,7 @@ def initialize() {
 }
 
 private addChildButtons(numberOfButtons) {
-	for(def endpoint : 1..numberOfButtons) {
+	for(def endpoint : 2..numberOfButtons) {
```
tpmanley: I don't think this is the right fix. At the very least this will break the other devices that use this DTH. It appears to me that the reason there is an extra button is because this device is using the `generic-4-button-alt` UI Metadata. The only difference between that and `generic-4-button` is the former adds an extra button for the main component. Is there a reason you're not using `generic-4-button` for this device?

ZouLianghuiS: hi, just tested with generic-4-button, this vid works for the device, I've modified the submit, please check again

tpmanley (approved):

PR #17909, opened 2020-01-13
[WWST-6374, WWST-6379] Zigbee Dimmer: Add fingerprint for Sengled Element Touch

rajendrasamsung:
DTH for WWST-6374 and WWST-6379, Sengled Element Touch

Added FP:
fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0702, 0B05", outClusters: "0019", manufacturer: "sengled", model: "E11-U21U31", deviceJoinName: "Sengled Element Touch"

Please review and approve PR


tpmanley:
@rajendrasamsung is `E11-U21U31` the correct model for both of these tickets? The tickets are for different devices so I'd expect different model names.

tpmanley:
In the future can you name the PRs like this:

`[<list of Jiras>] <DTH Name>: <change summary>`

I'll update this one as an example

rajendrasamsung:
We have on-boarded  the device and verified the in ST IDE. Both the devices are having same model numbers. Only the difference is screw fittings.

tpmanley (approved):

PR #17999, opened 2020-01-13
[ICP-12063] Added "vid:"

ZWozniakS:
@greens could you please check this? Unfortunately, the previous PR(https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/17854/files) was without "vid:" word and wasn't working...


tpmanley:
@ZWozniakS there is a merge conflict on this file. Can you rebase this on the latest `master` branch, squash the commits into a single one and address any merge conflicts?

ZWozniakS:
@tpmanley Yeah, I made a little mess here, sorry about that. I left a comment in ICP-12063 that we should confirm once again whether Innr's color temperature range is correct, as I will probably need to make changes again. It looks like bulb is operating between 2000-6500K not from 1800K like Innr team said. So maybe I should wait for the final decision and then make changes (and resolve conflicts)?

tpmanley:
> So maybe I should wait for the final decision and then make changes (and resolve conflicts)?

I think that makes sense. That way you won't have to resolve conflicts more than once.

k-morawski:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/21324

PR #18064, opened 2020-01-16
[CHAD-4299] SmartSense Motion Sensor: Update battery calculation for SmartThings Motion Sensor

tpmanley:
The manufacturer asked us to calculate the percentage remaining using a
specific formula which effectively changes the minimum operating voltage
from 2.1V to 2.4V. If they ever update the FW to do this on the device we
will need to also update the DTH to only do this calculation for older FW.

https://smartthings.atlassian.net/browse/CHAD-4299


BroderickCarlin (approved):

varzac (approved):
So this will make what was previously 50% battery now 0%, seems like a pretty major change.  The code itself looks fine though.

greens (approved):
🤔

tpmanley:
I'm triple checking the new formula with the manufacturer before merging

tpmanley:
The manufacturer verified the changes are what they want so I'm going to merge this.

PR #18065, opened 2020-01-16
[CHAD-4343] Language translations for the "Temperature Offset" device preference

tpmanley:
These DTHs have a single device preference called "Temperature Offset". There are other DTHs that also have this device preference but they aren't included here because they have other device preferences that we don't yet have translations for. In other words, we're trying to deploy all the device preferences for a given DTH at once.

https://smartthings.atlassian.net/browse/CHAD-4343


greens (dismissed):
woof

tpmanley:
A bunch of the locale codes I used are not known by Java/Groovy so I'm working on an update.

greens (dismissed):

tpmanley:
Sorry @greens I kept finding more issues the more I tested. This should be the last time you need to review :)

greens (approved):
how dare you not get it right the first time

PR #18066, opened 2020-01-17
[ICP-12196] Zigbee Multi Button: The first button isn't creating events correctly

ZouLianghuiS:
@tpmanley , @greens ,please have a review of my PR

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -73,11 +73,9 @@ def parseAttrMessage(description) {
 		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger()) + 1
        
 		log.debug "Number is ${buttonNumber}"
-		def event = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: "pushed", isStateChange: true)
-		if (buttonNumber == 1) {
-			sendEvent(event)
-		}
-		sendEventToChild(buttonNumber, event)
+		def descriptionText = (device.displayName.endsWith(' 1') ? "${device.displayName[0..-2]} button" : "${device.displayName}") + " ${buttonNumber} was pushed"
+		sendEventToChild(buttonNumber, createEvent(name: "button", value: "pushed", descriptionText: descriptionText, isStateChange: true))
+		map = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true)
    	}
 	map
 }
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -76,7 +76,6 @@ def parseAttrMessage(description) {
 		def event = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: "pushed", isStateChange: true)
 		if (buttonNumber == 1) {
 			sendEvent(event)
-			sendEventToChild(buttonNumber, event)
 		}
 		sendEventToChild(buttonNumber, event)
    	}
--- a/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
+++ b/devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
@@ -30,7 +30,7 @@ metadata {
 
 		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L", deviceJoinName: "Iris KeyFob", mnmn: "SmartThings", vid: "generic-4-button"
 		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L2", deviceJoinName: "Iris KeyFob", mnmn: "SmartThings", vid: "generic-4-button"
-		fingerprint profileId: "0104", inClusters: "0004", outClusters: "0000, 0001, 0003, 0004, 0005, 0B05", manufacturer: "HEIMAN", model: "SceneSwitch-EM-3.0", deviceJoinName: "HEIMAN Scene Keypad", vid: "generic-4-button-alt"
+		fingerprint profileId: "0104", inClusters: "0004", outClusters: "0000, 0001, 0003, 0004, 0005, 0B05", manufacturer: "HEIMAN", model: "SceneSwitch-EM-3.0", deviceJoinName: "HEIMAN Scene Keypad", vid: "generic-4-button"
 	}
 
 	tiles {
@@ -70,15 +70,15 @@ def parseAttrMessage(description) {
 		map = getButtonEvent(descMap)
 	} else if(descMap?.clusterInt == 0xFC80) {
 		def buttonNumber
-		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger())
+		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger()) + 1
        
 		log.debug "Number is ${buttonNumber}"
 		def event = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: "pushed", isStateChange: true)
-		if (buttonNumber != 0) {
-			sendEventToChild(buttonNumber, event)
-		} else {
+		if (buttonNumber == 1) {
 			sendEvent(event)
+			sendEventToChild(buttonNumber, event)
 		}
+		sendEventToChild(buttonNumber, event)
    	}
 	map
 }
@@ -233,7 +233,7 @@ private getSupportedButtonValues() {
 private getModelNumberOfButtons() {[
 		"3450-L" : 4,
 		"3450-L2" : 4,
-		"SceneSwitch-EM-3.0" : 3
+		"SceneSwitch-EM-3.0" : 4
 ]}
 
 private getModelBindings(model) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -70,11 +70,11 @@ def parseAttrMessage(description) {
 		map = getButtonEvent(descMap)
 	} else if(descMap?.clusterInt == 0xFC80) {
```
greens: is this cluster only used by the Heiman devices?

ZouLianghuiS: yes, no conflict with other devices

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -30,7 +30,7 @@ metadata {
 
 		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L", deviceJoinName: "Iris KeyFob", mnmn: "SmartThings", vid: "generic-4-button"
 		fingerprint inClusters: "0000, 0001, 0003, 0007, 0020, 0B05", outClusters: "0003, 0006, 0019", manufacturer: "CentraLite", model:"3450-L2", deviceJoinName: "Iris KeyFob", mnmn: "SmartThings", vid: "generic-4-button"
-		fingerprint profileId: "0104", inClusters: "0004", outClusters: "0000, 0001, 0003, 0004, 0005, 0B05", manufacturer: "HEIMAN", model: "SceneSwitch-EM-3.0", deviceJoinName: "HEIMAN Scene Keypad", vid: "generic-4-button"
+		fingerprint profileId: "0104", inClusters: "0004", outClusters: "0000, 0001, 0003, 0004, 0005, 0B05", manufacturer: "HEIMAN", model: "SceneSwitch-EM-3.0", deviceJoinName: "HEIMAN Scene Keypad", vid: "generic-4-button-alt"
```
tpmanley: I think it's better to use `generic-4-button` because the `-alt` version also shows the `main` component in the device details view.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -233,7 +233,7 @@ private getSupportedButtonValues() {
 private getModelNumberOfButtons() {[
 		"3450-L" : 4,
 		"3450-L2" : 4,
-		"SceneSwitch-EM-3.0" : 4
+		"SceneSwitch-EM-3.0" : 3
```
tpmanley: I think this should be left at 4 because it's a 4 button device.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -70,11 +70,11 @@ def parseAttrMessage(description) {
 		map = getButtonEvent(descMap)
 	} else if(descMap?.clusterInt == 0xFC80) {
 		def buttonNumber
-		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger()) + 1
+		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger())
```
tpmanley: What values are sent for each of the 4 button presses? Is it 0, 1, 2, 3 or 1, 2, 3, 4?

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -70,11 +70,11 @@ def parseAttrMessage(description) {
...
        
 		log.debug "Number is ${buttonNumber}"
 		def event = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: "pushed", isStateChange: true)
-		if (buttonNumber != 1) {
+		if (buttonNumber != 0) {
```
tpmanley: I believe the real problem is this check here. On any button press you should create an event on the main device with `buttonNumber` set to the actual button number (1-4) AND you should also generate an event on the appropriate child device. Take a look at how `getButtonResult` does this for an example.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
ZouLianghuiS: when I use generic-4-button, there are 4 child button, no main component, so the status in device card doesn't match any button in plugin

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
ZouLianghuiS: I've checked zigbee scene keypad DTH, for 4 buttons scene keypad, there is 1 main button and 3 child buttons, when create child buttons, index are from 2 to 4. since you requested me to start the index from 1, I have to change it to 3

and from the function
private addChildButtons(numberOfButtons) {
for(def endpoint : 1..numberOfButtons) {
...}

it seems the value is the number of child buttons

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
ZouLianghuiS: 0,1,2,3

0 sent to main button, 1,2,3 sent to child buttons

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
ZouLianghuiS: button number is 0,1,2,3
when number is 0, event is sent to main button,otherwise sent to child buttons

in getButtonResult, event only sent to child buttons, since this function is implemented for Iris KeyFob and vid of KeyFob is generic-4-button, only 4 child buttons are shown in device plugin, so all the event should be sent to child buttons.

I've also tried to use generic-4-button instead of generic-4-button-alt, and sent event to 4 child buttons like KeyFob, but as I said in slack, the status in device card is main component, doesn't match any of child device. so I have to use generic-4-button-alt and create 1 main button and 3 child buttons .

tpmanley: I think this combination of things will work:
* Use `generic-4-button`
* Create 4 components using `addChildDevice`
* On every button event, create an event on the main component with `buttonNumber` set to the appropriate number (1-4) and also create an event on the appropriate child component with `buttonNumber` set to 1

greens (approved):

greens (changes requested):
You need to send an event for to the main device as well as the child device, regardless of the button number. See how this is done in `getButtonEvent`

ZouLianghuiS:
> You need to send an event for to the main device as well as the child device, regardless of the button number. See how this is done in `getButtonEvent`

I've tested after modify my DTH according to your suggestion, but still the status of device card in dashboard doesn't change at all, and here is my modification:

def descriptionText = "pushed"
def event = createEvent(name: "button", value: descriptionText, data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true)
sendEventToChild(buttonNumber, event)
return createEvent(descriptionText: descriptionText)

dkirker:
That definitely looks like a bug in `getButtonResult`. The `return createEvent(descriptionText: descriptionText)` line doesn't actually return an event of anything.

In `parseAttrMessage` I would just do:
```
def descriptionText = (device.displayName.endsWith(' 1') ? "${device.displayName[0..-2]} button" : "${device.displayName}") + " ${buttonNumber} was pushed"

sendEventToChild(buttonNumber, createEvent(name: "button", value: "pushed", descriptionText: descriptionText, isStateChange: true))
map = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true)
```

That will send the button event to the child device as well as create the main device button event (with the correct buttonNumber defined in it) which gets assigned to map and then returned.

dkirker (commented):
That definitely looks like a bug in `getButtonResult`. The `return createEvent(descriptionText: descriptionText)` line doesn't actually return an event of anything. (I think it was from the way the DTH was probably previously implemented.)

In `parseAttrMessage` I would just do:
```
def descriptionText = (device.displayName.endsWith(' 1') ? "${device.displayName[0..-2]} button" : "${device.displayName}") + " ${buttonNumber} was pushed"

sendEventToChild(buttonNumber, createEvent(name: "button", value: "pushed", descriptionText: descriptionText, isStateChange: true))
map = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true)
```

That will send the button event to the child device as well as create the main device button event (with the correct buttonNumber defined in it) which gets assigned to map and then returned.

ZouLianghuiS:
checked again with @dkirker 's suggestion, and the status in device card and plugin all works fine. 
I've made new commit, please review again, thanks

ZouLianghuiS:
checked again based on @dkirker 's suggestion, and the status in device card and plugin are correct now.

I've commit the modification, please have a review again.

dkirker (dismissed):

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -73,12 +73,9 @@ def parseAttrMessage(description) {
 		buttonNumber = Integer.valueOf(descMap?.command[1].toInteger()) + 1
        
 		log.debug "Number is ${buttonNumber}"
-		def event = createEvent(name: "button", value: "pushed", data: [buttonNumber: buttonNumber], descriptionText: "pushed", isStateChange: true)
-		if (buttonNumber != 1) {
-			sendEventToChild(buttonNumber, event)
-		} else {
-			sendEvent(event)
-		}
+		def descriptionText = (device.displayName.endsWith(' 1') ? "${device.displayName[0..-2]} button" : "${device.displayName}") + " ${buttonNumber} was pushed"
```
tpmanley: This code is very similar to the code that calculates the `componentLabel` in `addChildButtons`. Please make a function that both of these usages can share so that the coupling is clear.

devicetypes/smartthings/zigbee-multi-button.src/zigbee-multi-button.groovy
```diff
@@ -73,12 +73,9 @@ def parseAttrMessage(description) {
...
+		sendEventToChild(buttonNumber, createEvent(name: "button", value: "pushed", descriptionText: descriptionText, isStateChange: true))
```
tpmanley: You should include `data: [buttonNumber: 1]` when sending to the child button device

greens (approved):

PR #18083, opened 2020-01-18
Auto arm/disarm alarm

rjhuval:
I would like to be able to automatically change the state of my alarm at a certain time.  Example:  Every night at 11:00 pm, my alarm automatically changes to Arm (Stay).


tpmanley:
@rjhuval you can do this in the new SmartThings app with an automation. Click the `+` button, select Automation and then configure the time you want it to run and the security mode you want to set.

PR #18143, opened 2020-01-22
DVCSMP-4826 Hotfix smartsense multi tile

greens:
my update to change the tile display in classic used "state" instead of "attributeState" resulting in nothing being displayed.


tpmanley (approved):
I tested this in Classic with a sensor in garage door mode and contact sensor mode. It's displaying properly now.

PR #18891, opened 2020-01-28
[ICP-11733] Restrict temperature reports ability to change device status

PKacprowiczS:
@greens @KKlimczukS @MGoralczykS @ZWozniakS 
Yet another fix for Fibaro Heat Controller (to be more accurate, for its Temperature Sensor).


greens (approved):
Looks good but will wait to hear back before merging.

PR #19057, opened 2020-01-29
[ICP-11438] Added support for Cool mode for Qubino Thermostat

PKacprowiczS:
@greens @KKlimczukS @MGoralczykS @ZWozniakS 
A lot of duplicated code, but afaik there's no other option to dynamically change metadata.

### DIFF
--- a/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
+++ b/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
@@ -104,6 +104,7 @@ metadata {
 
 def installed() {
 	state.isThermostatModeSet = false
+    state.supportedModes = ["off"]
 	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 12 * 60 , displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
 	sendEvent(name: "setpointRange", value: [minSetpointTemperature, maxSetpointTemperature], displayed: false)
 	response(refresh())
@@ -187,13 +188,12 @@ def zwaveEvent(physicalgraph.zwave.commands.sensormultilevelv5.SensorMultilevelR
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
-	state.supportedModes = ["off"]
 	//this device doesn't act like normal thermostat, it can support either 'cool' or 'heat' after configuration
 	if (cmd.parameterNumber == 59 && !state.isThermostatModeSet) {
+        state.supportedModes.add(cmd.scaledConfigurationValue ? "cool" : "heat")
         if (cmd.scaledConfigurationValue) {
             setDeviceType("Qubino Flush Thermostat Cool")
         }
-		state.supportedModes.add(cmd.scaledConfigurationValue ? "cool" : "heat")
 		state.isThermostatModeSet = true
 	}
 	createEvent(name: "supportedThermostatModes", value: state.supportedModes.encodeAsJson(), displayed: false)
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -103,16 +93,18 @@ metadata {
...
 				type: "enum",
+                defaultValue: "Heat",
 				options: ["Heat", "Cool"]
 		)
 	}
 }
 
 def installed() {
 	state.isThermostatModeSet = false
+    state.supportedModes = ["off"]
```
greens, line 107: nit: a couple of tabs/spaces problems

greens:
afaik (@dkirker can confirm), we don't have to specify supported modes in the metadata anymore, so it shouldn't be necessary to switch DTHs anymore

PKacprowiczS:
We're switching between thermostat heating and cooling setpoints capabilities, and I don't see how it could be achievable without metadata or misleading an user.

dkirker:
We do have other integrations that have both heating setpoint and cooling setpoint capabilities defined but might not support one or the other with the current device's configuration (ecobee, Zen, Honeywell, LUX KONOz). I believe the plug-in usually does the right thing in terms of those two setpoints. We should also be able to define "visible" conditions in the UI metadata so that one won't be visible when it isn't supported.

Honestly, I would keep one DTH. As long as supportedThermostatModes gets set, it should help guide any integrations. And then just make sure that if you are getting a setHeatingSetpoint or setCoolingSetpoint command called you validate the current mode and ignore it if the mode isn't correct. (The plugin will just have to get over itself if it is stupid and shows the wrong slider and tries to send data and expects a response.)

Some day we'll have a way to be able to turn off parts of a capability (or the capability as a whole).

PKacprowiczS:
@dkirker Let me check if we're on the same page. Are you suggesting that (for now) we should end up with device page displaying both sliders, where either of them would be useless? That's something which will lead to huuuge battle with certification team. Because that's only way to support "cool" mode, except mine, suggested in this PR.

I cannot speak for other devices, cause I've only worked with KONOz, but it has possibility to work with "cool/heat", just "cool" or just "heat" (alongside with auto, but you know what I mean) and I suppose it's also the case for rest thermostats you've mentioned.
This one is different, as it can configured to work as just "heat" (which is default) or just "cool", and to switch between them you need to re-pair device. 
So if we're not willing to choose "DTH change" option, I guess we should leave it as it is, and wait for possibility to exclude capabilities.

greens:
@PKacprowiczS I think Donald is saying that the UI can be configured based on the values of `supportedThermostatModes`. @dkirker can you experiment a bit and see if you can do that?

PKacprowiczS:
If there haven't been any changes recently, `supportedThermostatModes` only affects list of available modes in an app. If metada config has specified e.g. `thermostatCoolingSetpoint` it'll be shown even if "cool" is not an element of a list being sent to `supportedThermostatModes`.

dkirker:
I had thought that maybe if cooling wasn't in supportedThermostatModes then the plugin might not show it, however I was mistaken. I did just verify with an Ecobee, and if the Ebobee is configured for heat support only then the cooling setpoint card still shows in the device details page.

HOWEVER, the `visible` object in the UI metadata can be used. I am currently experimenting with this to verify that it works in the correct places (I have used it in the main action and main state sections of the metadata in the past, but not the device details page or the rule builder).

dkirker:
I have a note to check today how well the plugin reacts to a rule for setting whether coolingSetpoint or heatingSetpoint is visible in UI metadata. There is a function in the metadata to support this, though.

Basically, we can likely show or hide this in the metadata, which should work.

dkirker:
I've been playing around with the metadata. Will leave a comment on the ticket.

PR #19205, opened 2020-01-29
[CHAD-4341] Add a generic Z-Wave Light bulb DTH for Z-Wave Light bulbs

msandstedt:
This commit amends the generic Z-Wave RGBW bulb driver, rgbw-light.groovy, with addition of these fingerprints:

Specific fingerprints:

```
fingerprint deviceId: "0x1101", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
fingerprint mfr: "031E", prod: "0005", model: "0001", deviceJoinName: "ilumin RGBW Bulb"
```

Per agreement with the reviewers, this fingerprint is removed from rgbw-light.groovy, as any bulb that supports command class 0x33 must also support 0x26.  With the inClusters: "0x26,0x33", we don't expect this fingerprint to ever actually be needed:

```
fingerprint inClusters: "0x33"
```

Finally, all non-Aeon-specific fingerprints are removed from aeon-led-bulb.groovy.  This was done to ensure that non-Aeon bulbs match to the rgbw-light.groovy driver, which should better support them.  Note that with this, aeon-led-bulb.groovy no longer has any fingerprints.  The Aeon bulbs for which that driver was written appear to already be matching to rgbw-light.groovy with Aoen-specific fingerprints there.

### DIFF
--- a/devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
+++ b/devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
@@ -27,15 +27,48 @@ metadata {
 		capability "Sensor"
 		capability "Health Check"
 
-		command "reset"
+		/*
+		 * Relevant device types:
+		 *
+		 * * 0x11 GENERIC_TYPE_SWITCH_MULTILEVEL
+		 * * 0x01 SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		 * * 0x02 SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		 *
+		 * Plausible command classes we might see in a color light bulb:
+		 *
+		 * 0x98 COMMAND_CLASS_SECURITY
+		 * 0x5E COMMAND_CLASS_ZWAVEPLUS_INFO_V2
+		 * 0x20 COMMAND_CLASS_BASIC
+		 * 0x26 COMMAND_CLASS_SWITCH_MULTILEVEL
+		 * 0X27 COMMAND_CLASS_SWITCH_ALL
+		 * 0x33 COMMAND_CLASS_SWITCH_COLOR
+		 * 0x70 COMMAND_CLASS_CONFIGURATION
+		 * 0x73 COMMAND_CLASS_POWERLEVEL
+		 *
+		 * Here are the command classes used by this driver that we can fingerprint against:
+		 *
+		 * * 0x26 COMMAND_CLASS_SWITCH_MULTILEVEL -> yes, it is dimmable
+		 * * 0x33 COMMAND_CLASS_SWITCH_COLOR -> yes, it has color control
+		 */
 
-		fingerprint inClusters: "0x26,0x33"
-		fingerprint inClusters: "0x33"
+		// dimmable, color control
+		fingerprint inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1101", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// Manufacturer and model-specific fingerprints.
 		fingerprint mfr: "0086", prod: "0103", model: "0079", deviceJoinName: "Aeotec LED Strip" //US
 		fingerprint mfr: "0086", prod: "0003", model: "0079", deviceJoinName: "Aeotec LED Strip" //EU
 		fingerprint mfr: "0086", prod: "0003", model: "0062", deviceJoinName: "Aeotec LED Bulb" //EU
 		fingerprint mfr: "0086", prod: "0103", model: "0062", deviceJoinName: "Aeotec LED Bulb" //US
 		fingerprint mfr: "0300", prod: "0003", model: "0003", deviceJoinName: "ilumin RGBW Bulb"
+		fingerprint mfr: "031E", prod: "0005", model: "0001", deviceJoinName: "ilumin RGBW Bulb"
 	}
 
 	simulator {
@@ -60,17 +93,27 @@ metadata {
 		}
 	}
 
-	standardTile("reset", "device.reset", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
-		state "default", label:"Reset Color", action:"reset", icon:"st.lights.philips.hue-single"
-	}
-
 	controlTile("colorTempSliderControl", "device.colorTemperature", "slider", width: 4, height: 2, inactiveLabel: false, range:"(2700..6500)") {
 		state "colorTemperature", action:"color temperature.setColorTemperature"
 	}
 
 	main(["switch"])
-	details(["switch", "levelSliderControl", "rgbSelector", "colorTempSliderControl", "reset"])
+	details(["switch", "levelSliderControl", "colorTempSliderControl"])
 }
+
+private getCOLOR_TEMP_MIN() { 2700 }
+private getCOLOR_TEMP_MAX() { 6500 }
+private getWHITE_MIN() { 0 } // min for Z-Wave coldWhite and warmWhite paramaeters
+private getWHITE_MAX() { 255 } // max for Z-Wave coldWhite and warmWhite paramaeters
+private getCOLOR_TEMP_DIFF() { COLOR_TEMP_MAX - COLOR_TEMP_MIN }
+private getRED() { "red" }
+private getGREEN() { "green" }
+private getBLUE() { "blue" }
+private getWARM_WHITE() { "warmWhite" }
+private getCOLD_WHITE() { "coldWhite" }
+private getRGB_NAMES() { [RED, GREEN, BLUE] }
+private getWHITE_NAMES() { [WARM_WHITE, COLD_WHITE] }
+private getCOLOR_NAMES() { RGB_NAMES + WHITE_NAMES }
 
 def updated() {
 	log.debug "updated().."
@@ -79,10 +122,12 @@ def updated() {
 
 def installed() {
 	log.debug "installed()..."
-	state.colorReceived = ["red": null, "green": null, "blue": null, "warmWhite": null, "coldWhite": null]
-	sendEvent(name: "checkInterval", value: 1860, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "0"])
-	sendEvent(name: "level", value: 100, unit: "%")
-	response(refresh())
+	sendEvent(name: "checkInterval", value: 1860, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	sendEvent(name: "level", value: 100, unit: "%", displayed: false)
+	sendEvent(name: "colorTemperature", value: COLOR_TEMP_MIN, displayed: false)
+	sendEvent(name: "color", value: "#000000", displayed: false)
+	sendEvent(name: "hue", value: 0, displayed: false)
+	sendEvent(name: "saturation", value: 0, displayed: false)
 }
 
 def parse(description) {
@@ -115,39 +160,15 @@ def zwaveEvent(physicalgraph.zwave.commands.switchmultilevelv3.SwitchMultilevelR
 
 def zwaveEvent(physicalgraph.zwave.commands.switchcolorv3.SwitchColorReport cmd) {
 	log.debug "got SwitchColorReport: $cmd"
-	if (!state.colorReceived)
-		state.colorReceived = ["red": null, "green": null, "blue": null, "warmWhite": null, "coldWhite": null]
-	state.colorReceived[cmd.colorComponent] = cmd.value
 	def result = []
-	def rgbNames = ["red", "green", "blue"]
-	def tempNames = ["warmWhite", "coldWhite"]
-	// Check if we got all the RGB color components
-	if (rgbNames.every { state.colorReceived[it] != null }) {
-		def colors = rgbNames.collect { state.colorReceived[it] }
-		log.debug "colors: $colors"
-		// Send the color as hex format
-		def hexColor = "#" + colors.collect { Integer.toHexString(it).padLeft(2, "0") }.join("")
-		result << createEvent(name: "color", value: hexColor)
-		// Send the color as hue and saturation
-		def hsv = rgbToHSV(*colors)
-		result << createEvent(name: "hue", value: hsv.hue)
-		result << createEvent(name: "saturation", value: hsv.saturation)
-		// Reset the values
-		rgbNames.collect { state.colorReceived[it] = null}
-	}
-	// Check if we got all the color temperature values
-	if (tempNames.every { state.colorReceived[it] != null}) {
-		def warmWhite = state.colorReceived["warmWhite"]
-		def coldWhite = state.colorReceived["coldWhite"]
-		log.debug "warmWhite: $warmWhite, coldWhite: $coldWhite"
-		// When the device is first installed, warmWhite == coldWhite == 255
-		//  so default to mid-range color temp.
-		def colorTemp = COLOR_TEMP_MIN + (COLOR_TEMP_DIFF / 2)
-		if (warmWhite != coldWhite)
-			colorTemp = (COLOR_TEMP_MAX - (COLOR_TEMP_DIFF * warmWhite) / 255) as Integer
-		result << createEvent(name: "colorTemperature", value: colorTemp)
-		// Reset the values
-		tempNames.collect { state.colorReceived[it] = null }
+	if (state.staged != null && cmd.colorComponent in RGB_NAMES) {
+		// We use this as a callback from our color setter.
+		// Emit our color update event with our staged state.
+		state.staged.subMap("hue", "saturation", "color").each{ k, v -> result << createEvent(name: k, value: v) }
+	} else if (state.staged != null && cmd.colorComponent in WHITE_NAMES) {
+		// We use this as a callback from our temperature setter.
+		// Emit our color temperature update event with our staged state.
+		state.staged.subMap("colorTemperature").each{ k, v -> result << createEvent(name: k, value: v) }
 	}
 	result
 }
@@ -159,10 +180,6 @@ private dimmerEvents(physicalgraph.zwave.Command cmd) {
 		result << createEvent(name: "level", value: cmd.value == 99 ? 100 : cmd.value , unit: "%")
 	}
 	return result
-}
-
-def zwaveEvent(physicalgraph.zwave.commands.hailv1.Hail cmd) {
-	response(command(zwave.switchMultilevelV1.switchMultilevelGet()))
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
@@ -186,16 +203,22 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 	[linkText: linkText, descriptionText: "$linkText: $cmd", displayed: false]
 }
 
-def buildOffOnEvent(cmd){
-	[zwave.basicV1.basicSet(value: cmd), zwave.switchMultilevelV3.switchMultilevelGet()]
-}
-
 def on() {
-	commands(buildOffOnEvent(0xFF), 3500)
+	// Per Z-Wave spec, this multilevel switch set value commands state-transition
+	// to on.  This will restore the most-recent non-zero value cached in the device.
+	def ON = 0xFF
+	commands([zwave.switchMultilevelV3.switchMultilevelSet(value: ON),
+	         zwave.switchMultilevelV3.switchMultilevelGet(),
+	], 3500)
 }
 
 def off() {
-	commands(buildOffOnEvent(0x00), 3500)
+	// Per Z-Wave spec, this multilevel switch set value commands state-transition
+	// to off.  This will not clobber the most-recent non-zero value cached in the device.
+	def OFF = 0
+	commands([zwave.switchMultilevelV3.switchMultilevelSet(value: OFF),
+	         zwave.switchMultilevelV3.switchMultilevelGet(),
+	], 3500)
 }
 
 def refresh() {
@@ -207,17 +230,16 @@ def ping() {
 	refresh()
 }
 
-def setLevel(level) {
-	setLevel(level, 1)
-}
-
-def setLevel(level, duration) {
+def setLevel(level, duration=1) {
 	log.debug "setLevel($level, $duration)"
-	if(level > 99) level = 99
+	level = Math.max(Math.min(level, 99), 1) // See Z-Wave level encoding
+	duration = duration < 128 ? duration : 127 + Math.round(duration / 60) // See Z-Wave duration encodinbg
+	duration = Math.min(duration, 0xFE) // 0xFF is a special code for factory default; bound to 0xFE
+	def tcallback = Math.min(duration * 1000 + 2500, 12000) // how long should we wait to read back?  we can't wait forever
 	commands([
 		zwave.switchMultilevelV3.switchMultilevelSet(value: level, dimmingDuration: duration),
 		zwave.switchMultilevelV3.switchMultilevelGet(),
-	], (duration && duration < 12) ? (duration * 1000) : 3500)
+	], tcallback)
 }
 
 def setSaturation(percent) {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+ *  for the specific language governing permissions and limitations under the License.
+ *
+ *  Z-Wave RGBW Bulb, adapted from Aeon LED Bulb 6 Multi-Color
+ *
+ *  Author: SmartThings
+ *  Date: 2020-01-27
+ */
+
+metadata {
+	definition (name: "Z-Wave RGBW Bulb", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.light", mnmn: "SmartThings", vid: "generic-rgbw-color-bulb") {
```
kianooshkarami: @greens Is there a UI definition for lights with no RGB support?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		// dimmable, color control; supports security command class.
+		//
+		// Note: explicitly delimiting application command classes as secure with 0xF1,0x00
+		fingerprint deviceId: "0x1101", inClusters: "0x98,0xF1,0x00,0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
```
kianooshkarami: 0x98 is not really needed.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control; supports security command class.
+		//
+		// Note: explicitly delimiting application command classes as secure with 0xF1,0x00
+		fingerprint deviceId: "0x1102", inClusters: "0x98,0xF1,0x00,0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// Illumin LZW42 RGBW Bulb.
+		//
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		fingerprint mfr: "031E", prod: "0005", model: "0001", deviceId: "0x1101", deviceJoinName: "Illumin RGBW Bulb"
```
kianooshkarami: Does this device really support RGB?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+}
+
+def updated() {
+	response(refresh())
+}
+
+def installed() {
+	log.debug "installed()..."
+	sendEvent(name: "checkInterval", value: 1860, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "0"])
+	sendEvent(name: "level", value: 100, unit: "%")
```
tpmanley: Set `displayed: false` on all these events so they don't show up in event history

kianooshkarami: Is this a common thing to set the values to 0 to start out with?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+	sendEvent(name: "colorTemperature", value: COLOR_TEMP_MIN)
+	sendEvent(name: "color", value: "#000000")
+	sendEvent(name: "hue", value: 0)
+	sendEvent(name: "saturation", value: 0)
+}
+
+def configure() {
+	commands([
+		// Set the dimming ramp rate
+		zwave.configurationV2.configurationSet(parameterNumber: 0x10, size: 1, scaledConfigurationValue: 5)
```
tpmanley: Is this relevant to all the devices that could fingerprint to this DTH or just the Illumin?

kianooshkarami: Is this parameter support by all devices? In general, the configuration parameters are device specific. If so, I would not send this command to all devices.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.Command cmd) {
+	def linkText = device.label ?: device.name
+	[linkText: linkText, descriptionText: "$linkText: $cmd", displayed: false]
+}
+
+def buildOffOnEvent(cmd){
+	[zwave.basicV1.basicSet(value: cmd), zwave.switchMultilevelV3.switchMultilevelGet()]
```
kianooshkarami: I find it very strange to do a `basic set` and then do `switch multi level get`, I'd rather keep it uniform to one command class for both Set/Get. I would recommend using Switch Multi-Level command class.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+	commands(buildOffOnEvent(0x00), 5000)
+}
+
+def refresh() {
+	commands([zwave.switchMultilevelV3.switchMultilevelGet()] + queryAllColors())
+}
+
+def ping() {
+	log.debug "ping().."
+	unschedule(offlinePing)
```
kianooshkarami: Is this legacy behavior that was copied over here?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,353 @@
...
+def offlinePing() {
+	log.debug "offlinePing()..."
+	sendHubCommand(new physicalgraph.device.HubAction(command(zwave.switchMultilevelV3.switchMultilevelGet())))
+}
+
+def setLevel(level) {
+	setLevel(level, 1)
+}
+
+def setLevel(level, duration) {
```
kianooshkarami: maybe change `duration` to `dimmingDuration`

kianooshkarami (commented):
Looks good.

kianooshkarami:
I have some comments that I'd like you to address specially with regards to RGW support for devices that do not.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: I don't know.  Not sure who to ask?  Probably no.

This seems to provide some naming guidance: https://smartthings.atlassian.net/wiki/spaces/WORKS/pages/660606090/Fingerprint+writing+rules

Do the reviewers have any input on a proper name?  This naming was meant to parallel zigbee-rgbw-bulb.groovy, which has name "ZibBZee RGBW Bulb".

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: This was copied in from the Aeon driver, and I think then may be specific to the Aeon devices.  I should probably remove this.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Yes, indeed it does.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Will do.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Yes.  I take it ping is a function we don't wish to expose?

kianooshkarami (commented):
🤦‍♂ Ignore my comments about UI support for RGB support. All of these lights support RGB

greens:
I left my question on the Jira ticket.

kianooshkarami:
@greens I answered on there. But we can always chat about this as well.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: This was resolved in an offline conversation.  There was a thought that the Illumin bulb may be non-RGB, but it does support color.

Decision here is to keep scope to only RGBW bulbs in this task.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Got it, done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Per offline conversation, legacy offlinePingable functionality is now removed from this DTH.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: My misunderstanding.  The comment was to do with the legacy offlinePingable.  References to this are now removed from the DTH.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+		 * * 0x33 COMMAND_CLASS_SWITCH_COLOR -> yes, it has color control
+		 */
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1101", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
```
tpmanley: The Aeon Led Bulb DTH has very similar fingerprints. Is the thinking that this DTH replaces that DTH? If that's the case I think we should remove the fingerprints from the Aeon Led Bulb DTH.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+			}
+		}
+	}
+
+	controlTile("colorTempSliderControl", "device.colorTemperature", "slider", width: 4, height: 2, inactiveLabel: false, range:"(2700..6500)") {
+		state "colorTemperature", action:"color temperature.setColorTemperature"
+	}
+
+	main(["switch"])
+	details(["switch", "levelSliderControl", "rgbSelector", "colorTempSliderControl"])
```
tpmanley: `rgbSelector` can be removed - there is no tile with that name

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+	response(refresh())
+}
+
+def installed() {
+	log.debug "installed()..."
+	sendEvent(name: "checkInterval", value: 1860, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "0"])
+	sendEvent(name: "level", value: 100, unit: "%", displayed: false)
+	sendEvent(name: "colorTemperature", value: COLOR_TEMP_MIN, displayed: false)
+	sendEvent(name: "color", value: "#000000", displayed: false)
+	sendEvent(name: "hue", value: 0, dispalyed: false)
```
tpmanley: sp: dispalyed

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
kianooshkarami: Yeah, @msandstedt If we can migrate over the Aeon LED Bulb fingerprint over to this DTH and remove that DTH. If there are any manufacturer specific changes such as setting configuration parameters, we can implement those in the DTH by checking the manufacturer/model of the device(Example: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/ed426552b4ffece81cb476a0d5614e7a300da888/devicetypes/smartthings/zwave-siren.src/zwave-siren.groovy#L466)

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd) {
+	dimmerEvents(cmd)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
+	dimmerEvents(cmd)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.switchmultilevelv3.SwitchMultilevelReport cmd) {
+	unschedule(offlinePing)
```
tpmanley: Seems like this should be in a different place where it is hit on receipt of any Z-Wave message, not just SwitchMultilevelReport

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+	runEvery30Minutes(offlinePing)
+	command(zwave.switchMultilevelV3.switchMultilevelGet())
+}
+
+def offlinePing() {
+	log.debug "offlinePing()..."
+	sendHubCommand(new physicalgraph.device.HubAction(command(zwave.switchMultilevelV3.switchMultilevelGet())))
+}
+
+def setLevel(level) {
```
tpmanley: You can remove this function and set a default value on the second parameter in the other `setLevel`

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+	setLevel(level, 1)
+}
+
+def setLevel(level, duration) {
+	log.debug "setLevel($level, $duration)"
+	if(level > 99) level = 99
```
tpmanley: I'd instead do `level = Math.max(Math.min(level, 99), 0)` like many of the other DTHs to do ensure it's in range

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: @greens had a similar question in the Jira issue.  My thinking:

* yes, for sure we should migrate generic RGBW bulbs from the Aeon driver to this one; if they are like the Illumin, they likely won't work quite right in the Aeon DTH
* BUT, I am scared to unify in a single DTH because the Aeon-specific configuration commands are slightly complicated and cryptic to me
* I'm worried about introducing regressions for Aeon bulbs

Thoughts?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Gotcha.  Done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,324 @@
...
+}
+
+def setColorTemperature(temp) {
+	log.debug "setColorTemperature($temp)"
+	def warmValue = tempToZwaveWarmWhite(temp)
+	def coldValue = tempToZwaveColdWhite(temp)
+	if (state.staged == null) {
+		state.staged = [:]
+	}
+	state.staged << [colorTemperature: temp] // stage ST colorTemperature attribute
```
tpmanley: Shouldn't this be an `=` instead of `<<`? I think we want to overwrite the staged value

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: No, you're right.  This is not supported by all devices.  I've now removed it.

msandstedt: Done, removed.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
```diff
@@ -0,0 +1,317 @@
...
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		//
+		// Note: explicitly delimiting application command classes as secure with 0xF1,0x00
+		fingerprint deviceId: "0x1102", inClusters: "0xF1,0x00,0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// Illumin LZW42 RGBW Bulb.
+		//
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		fingerprint mfr: "031E", prod: "0005", model: "0001", deviceId: "0x1101", deviceJoinName: "Illumin RGBW Bulb"
```
tpmanley: Ilumin has one `l`

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: I stripped the offlinePing stuff from the code completely.  Is that OK?

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: - will add that sweet, syntactic sugar - good call

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: I think it also needs to be in the range [1,99].  If we dim to zero, power off, and then power on, we will restore to whatever was set before all of those special state-transition values.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: With the consistency guarantees around read and write of state from the DB and how this interacts with DTH instantiation, I found that it was not safe to have the temp and color handlers clobbering each others' cached state.  This is definitely intentional.

The code attempts to target creation of emitted events by keying these on receipt of zwave RGB color information vs. receipt of zwave coldWhite/warmWhite information.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Got it.  Done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Fixed.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Done.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Fixed.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: OK, now I see that while aeon-led-bulb-6.groovy is Aeon-specific, aeon-led-bulb.groovy is attempting to very generically match to many dimmable bulbs with device type and inClusters.

I see aeon-led-bulb.groovy as somewhat broken and probably less good than the current driver.  So how should I handle this?

msandstedt: Approach agreed upon offline:

* migrate all generic fingerprinting into existing rgbw-light.groovy
* replace body of rgbw-light.groovy with new code (new code works better)
* remove generic fingerprinting from aeon-led-bulb.groovy
* add mfr-specific fingerprint to aeon-led-bulb.groovy

greens (dismissed):
nice

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -27,15 +27,60 @@ metadata {
...
+		// Note: explicitly delimiting application command classes as secure with 0xF1,0x00
+		fingerprint deviceId: "0x1102", inClusters: "0xF1,0x00,0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// Manufacturer and model-specific fingerprints.
 		fingerprint mfr: "0086", prod: "0103", model: "0079", deviceJoinName: "Aeotec LED Strip" //US
 		fingerprint mfr: "0086", prod: "0003", model: "0079", deviceJoinName: "Aeotec LED Strip" //EU
 		fingerprint mfr: "0086", prod: "0003", model: "0062", deviceJoinName: "Aeotec LED Bulb" //EU
 		fingerprint mfr: "0086", prod: "0103", model: "0062", deviceJoinName: "Aeotec LED Bulb" //US
 		fingerprint mfr: "0300", prod: "0003", model: "0003", deviceJoinName: "ilumin RGBW Bulb"
+		fingerprint mfr: "031E", prod: "0005", model: "0001", deviceId: "0x1101", deviceJoinName: "ilumin RGBW Bulb"
```
tpmanley: I'd remove the `deviceId` for consistency with the other fingerprints

msandstedt: Sure thing.  Done.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -27,15 +27,60 @@ metadata {
...
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		// dimmable, color control
+		//
+		// Note: explicitly delimiting application command classes as secure with 0xF1,0x00
+		fingerprint deviceId: "0x1101", inClusters: "0xF1,0x00,0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
```
kianooshkarami: @msandstedt, @gausnes mentioned that the fix for HCS-1050 will be available in Prod environment by two weeks. The changes to these DTHs will more than likely be available in prod in two weeks as well, so I was thinking we can leave these workarounds out since the issues exist in the prod as of now. Thoughts?

kianooshkarami (dismissed):
I had one minor comment about leaving out the workaround for HCS-1050 since that fix has been merged and is making its way to prod. But other than that, everything looks good. Great job.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
msandstedt: Makes sense.  I will strip these out and then rebase.  I believe this will invalidate your approval - sorry for the churn!

Almost done here...

msandstedt: OK, I pushed up what I consider final (apologies - this invalidated the current approvals).

I believe this addresses everything discussed with the reviewers.  Please note that aeon-led-bulb.groovy now actually contains no fingerprints.  I believe this is appropriate because in fact, Aeon bulbs are already matching to the rgbw-light.groovy with these existing fingerprints:

* fingerprint mfr: "0086", prod: "0003", model: "0062", deviceJoinName: "Aeotec LED Bulb" //EU
* fingerprint mfr: "0086", prod: "0103", model: "0062", deviceJoinName: "Aeotec LED Bulb" //US

I verified that no Aeon bulbs have matched to aeon-led-bulb.groovy in the last 30 days.  The only bulbs that have matched there will be better controlled with rgbw-light.groovy.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Note I did all of the above except this:

* add mfr-specific fingerprint to aeon-led-bulb.groovy

The mfr-specific Aeon fingerprint is already in rgbw-light.groovy; Aeon bulbs aren't actually matching to aeon-led-bulb.groovy in the first place, and Sumo confirms this.

Considering this resolved.

devicetypes/smartthings/zwave-rgbw-bulb.src/zwave-rgbw-bulb.groovy
msandstedt: Decided to leave this to reduce the delta in rgbw-light.groovy.

msandstedt:
Hey all.  I invalidated the approvals with my rebase/squash.  But I believe this now addresses everything requested.  Hence, ready for final review.

greens (approved):

PR #20462, opened 2020-02-05
ICP-11631 Refresh Yale lock battery value after battery replacement

greens:
Queues up a battery get after we see the battery changed notification from the lock. Previous implementation relied on sending gets whenever the lock received a new message in parse(), but locks run locally so this happend very rarely. New approach is to query on a regular timer until we get an update.

### DIFF
--- a/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
+++ b/devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
@@ -405,6 +405,10 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch (cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
+			runIn(1, setQueryBattery, [overwrite: true, forceForLocallyExecuting: true])
+			result << response(secure(zwave.batteryV1.batteryGet()))
+			break;
 		case 0x0A:
 			map = [name: "battery", value: 1, descriptionText: "Battery level critical", displayed: true, data: [lockName: deviceName]]
 			break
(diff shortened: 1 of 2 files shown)


greens:
Reason this didn't get caught is largely because we can't reproduce the problems Poland has here.

kianooshkarami (dismissed):

greens:
Going to rebase this and get it to staging after deploy tomorrow.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -686,3 +691,12 @@ private Boolean secondsPast(timestamp, seconds) {
 	return (now() - timestamp) > (seconds * 1000)
 }
 
+private queryBattery() {
+	log.debug "Running queryBattery"
+	if (!state.lastbatt || now() - state.lastbatt > 10*1000) {
+		log.debug "It's been more than 10s since battery was updated after a replacment. Querying battery."
```
cbaumler:
```suggestion
		log.debug "It's been more than 10s since battery was updated after a replacement. Querying battery."
```

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -405,6 +405,10 @@ private def handleBatteryAlarmReport(cmd) {
 	def deviceName = device.displayName
 	def map = null
 	switch (cmd.zwaveAlarmEvent) {
+		case 0x01: //power has been applied, check if the battery level updated
```
cbaumler, line 408: nit: Could add `log.debug "Batteries replaced. Queueing a battery get."` as was done for `zwave-lock.groovy`.

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
```diff
@@ -405,6 +405,10 @@ private def handleBatteryAlarmReport(cmd) {
...
+			runIn(10, "queryBattery", [overwrite: true, forceForLocallyExecuting: true])
+			result << response(secure(zwave.batteryV1.batteryGet()))
+			break;
 		case 0x0A:
```
cbaumler, line 413: Just curious, does case 130 not apply for this DTH as it does for `zwave-lock.groovy`?

cbaumler (dismissed):
LGTM

devicetypes/smartthings/zwave-lock-without-codes.src/zwave-lock-without-codes.groovy
greens: I don't believe any of the locks on this DTH use the OEM/alarmv1 codes.

greens:
@cbaumler  fixed the typos you caught and added that logging, can I get a re-review?

cbaumler (approved):
LGTM

PR #20554, opened 2020-02-06
DevWs for RoomsBeautiful containing containing Curtain

devropes:


devropes:
The only change I made is vid: "generic-shade-2" to display the shade level control.

devicetypes/rooms-beautiful/rooms-beautiful-curtain.src/rooms-beautiful-curtain.groovy
```diff
@@ -136,7 +136,7 @@ def parse(String description) {
                 }
                 break;
             case 0xFC00:
-                if (description?.startsWith('read attr -'))
+                if (description ? .startsWith('read attr -'))
```
greens: I just checked this in the interpreter to be sure, but this won't run.

Interestingly, `description ?. startsWith...` does, but we'd prefer `description?.startsWith`

devropes: I have now fixed that. Must've happened when I was using a code beautifier to fix the indentation.

greens (approved):

greens:
should be live on the 25th in production

PR #21067, opened 2020-02-12
[WWST-4872, WWST-4876, WWST-4892] DawonDNS child in-wall switch Z-wave device. 

juhaki:
…ce for child switch item.

DawonDNS in-wall switch Z-wave device for Korea.
This file for child switch items.

### DIFF
--- a/devicetypes/smartthings/dawon-zwave-smart-switch.src/child-dawon-zwave-wall-smart-switch.groovy
+++ b/devicetypes/smartthings/dawon-zwave-smart-switch.src/child-dawon-zwave-wall-smart-switch.groovy
@@ -1,5 +1,5 @@
 /**
- *  Copyright 2018 SmartThings
+ *  Copyright 2020 SmartThings
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  *  in compliance with the License. You may obtain a copy of the License at:
@@ -37,7 +37,7 @@ metadata {
 
 def installed() {
 	// This is set to a default value, but it is the responsibility of the parent to set it to a more appropriate number
-	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	child.sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 	refresh()
 }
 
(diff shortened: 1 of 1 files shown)


greens:
I would recommend using https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/child-switch-health.src/child-switch-health.groovy instead of creating a new DTH. 

You could then just use `child.sendEvent` in the parent DTH after the child is created to overwrite the default `checkInterval` configuration. (But also do not mark the device as `offlinePingable` please, we try to reserve that for life safety devices.)

Also we would prefer that these be one PR instead of two separate ones, so that it's clear they are related.

juhaki:
> I would recommend using https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/child-switch-health.src/child-switch-health.groovy instead of creating a new DTH.
> 
> You could then just use `child.sendEvent` in the parent DTH after the child is created to overwrite the default `checkInterval` configuration. (But also do not mark the device as `offlinePingable` please, we try to reserve that for life safety devices.)
> 
> Also we would prefer that these be one PR instead of two separate ones, so that it's clear they are related.



Could the Zwave child device use the Zigbee "Child-switch-health.groovy file" ?

SmartThingsPublic\devicetypes\smartthings\child-switch-health.src\child-switch-health.groovy
====================================
...
def installed() {
	// This is set to a default value, but it is the responsibility of the parent to set it to a more appropriate number
	sendEvent(name: "checkInterval", value: 30 * 60, displayed: false, data: [protocol: "zigbee"])
}
...
====================================


SmartThingsPublic\devicetypes\smartthings\dawon-zwave-smart-switch.src\child-dawon-zwave-wall-smart-switch.groovy
====================================
...
def installed() {
	// This is set to a default value, but it is the responsibility of the parent to set it to a more appropriate number
	sendEvent(name: "checkInterval", value: 30 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
	refresh()
}
...
====================================

juhaki (commented):
[WWST-4872, WWST-4876, WWST-4892] DawonDNS child in-wall switch Z-wave device.
DawonDNS child in-wall switch Z-wave device for Korea.
This file for Z-wave child switch items.

Please review this patch.

greens:
Yes. The Z-Wave device could use it. That `checkInterval` value can be overwritten.

Alternatively, you could ignore having health check on your child devices and instead just use `child-switch.groovy`. There are ways you could control the online/offline state of the device from the parent instead of requiring DH to monitor 4 different "devices" that will all go offline at the same time.

juhaki:
Ok. I will use the "child-switch.groovy". 
So. This file does not needed.

PR #21408, opened 2020-02-15
DevWs-acceptance for Roger Role Test User containing containing DevWS Test: RogerAPISubmissionNA04_123

SmartThings:


Lois2020p:
I do not want any more mail from you. I have unsubscribed but get hundreds of this stuff and I am simply not interested.

PR #21813, opened 2020-02-20
ONEAPP-22347 LUX KONOz Celsius logic was being incorrectly applied to Fahrenheit

dkirker:
Toss this one in the embarrassing bin...

1. Typo in the comment.
2. The logic applied specifically to Celsius in setHeatingSetpoint, however this wasn't checked and heatingSetpointRange was being used assuming that it was in Celsius.


greens:
I'm not quite sure I understand this one. You're going to have to talk me through it.

greens (dismissed):

dkirker:
Got a chance to do a bit more heavy testing to make sure that the previous issue was also fixed by this change. It appears so. Had to fix an exception, so that dismissed your review @greens . :(

greens (approved):

PR #21818, opened 2020-02-20
[ICP-12283] Removed deprecated capability

ZWozniakS:
@PKacprowiczS @KKlimczukS @MGoralczykS


greens (approved):
@ZWozniakS have you tested that this doesn't affect the device's operation? I know holdable button was added a long time ago before we could configure the various button actions so I want to make sure nothing was relying on this.

ZWozniakS:
@greens yes, I have checked that before creating PR and it looks like the device is still working as expected :)

PR #22345, opened 2020-02-25
CHAD-3849 Update SmartWeather Station Tile to support the new mobile client

dkirker:


dkirker:
This is currently a work in progress. The goal is to use new existing capabilities, and then create separate custom weather capabilities which will then get promoted to regular weather capabilities.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -22,6 +22,9 @@ metadata {
         capability "Temperature Measurement"
         capability "Relative Humidity Measurement"
         capability "Ultraviolet Index"
+		capability "Wind Speed"
```
tpmanley: nit: indentation

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -22,6 +22,9 @@ metadata {
...
+        capability "stclassic.Wind"
```
tpmanley: Why did you choose `stclassic` as the namespace? Seems a little odd since the point is to get these working in the new app.

tpmanley:
@dkirker is your plan to tackle just these first three capabilities in this PR? I do think it makes sense to break it up all the work into smaller PRs if it can't all be done at once.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
dkirker: I picked it because I really couldn't think of something better (other than maybe stlegacy) that also wouldn't risk conflicting with another namespace. The thinking was "these are being used by device handlers from the classic/legacy platform." Though I am open to suggestions/ideas.

dkirker: I included the custom capability in a JSON file as I have defined it based on what is used (and what makes sense to transfer over). I'll note elsewhere the capabilities that I plan to create and add in another PR.

tpmanley: How about something more specific to this DTH like `st-weatherstation`?

@greens @varzac @BroderickCarlin might have opinions too

devicetypes/smartthings/smartweather-station-tile.src/capability.smartWeather.json
```diff
@@ -0,0 +1,244 @@
...
+          "value": {
+            "type": "string"
+          }
+        },
+        "additionalProperties": false
+      },
+      "required": [
+        "value"
+      ]
+    },
```
BroderickCarlin: Very noobish when it comes to these files, but why do we list `localSunrise` twice in the attributes list?

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
varzac: I don't really like `stclassic`, but I don't have a better idea.

devicetypes/smartthings/smartweather-station-tile.src/capability.smartWeather.json
dkirker: Uhhhhhhh $20 sits on a copy and paste error. :P

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
dkirker: Since custom namespaces are closing, I figured I'd snag something more general that can be added to over time.

dkirker: @tpmanley I liked the idea about a DTH-specific namespace. Originally I was against it, but it at least compartmentalized the capabilities. I opted to break out in to multiple capabilities instead of just a DTH-specific capability. I also created the namespace `stcustom` for future use.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -22,9 +22,21 @@ metadata {
         capability "Temperature Measurement"
         capability "Relative Humidity Measurement"
         capability "Ultraviolet Index"
+        capability "Wind Speed"
+        capability "stsmartweather.Wind Speed" // "Wind Speed" only supports m/s unit, however we want to create both events
+        capability "stsmartweather.Wind Direction"
+        capability "stsmartweather.Apparent Temperature"
+        capability "stsmartweather.Astronomical Data"
+        capability "stsmartweather.Percipitation"
```
tpmanley: sp: Precipitation here and other places

dkirker: Yup, thanks. That was causing an exception that I was facing. A conversion handler deep in the bowels of one of our services was trying to look up a misspelled version of the attribute from a Map using the attr name as the key and value was returned as `null`. Without checking it tries to execute commands on `null` and throw an NPE instead of either gracefully exiting or throwing a more meaningful exception (`CapabilityAttributeNotValid` maybe? heh). That was painful to track down...

dkirker:
Let me verify that the code I was testing in the IDE is the same as this code, but this should be done.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -242,8 +254,8 @@ def pollUsingZipCode(String zipCode) {
 
         send(name: "humidity", value: obs.relativeHumidity, unit: "%")
         send(name: "weather", value: obs.wxPhraseShort)
-        send(name: "weatherIcon", value: obs.iconCode as String, displayed: false)
-        send(name: "wind", value: obs.windSpeed as String, unit: windUnits) // as String because of bug in determining state change of 0 numbers
```
tpmanley, line 246: Is the issue described in the comment still relevant?

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -418,9 +431,41 @@ private localDate(timeZone) {
...
+            "sunriseDate": "stsmartweather.astronomicalData.sunriseDate",
+            "sunsetDate": "stsmartweather.astronomicalData.sunsetDate",
+            "lastUpdate": "stsmartweather.smartWeather.lastUpdate",
+            "uvDescription": "stsmartweather.ultravioletDescription.uvDescription",
+            "forecastToday": "stsmartweather.weatherForecast.forecastToday",
+            "forecastTonight": "stsmartweather.weatherForecast.forecastTonight",
+            "forecastTomorrow": "stsmartweather.weatherForecast.forecastTomorrow"
+        ]
+
+    //log.trace "WUSTATION: event: $map"
```
tpmanley, line 460: Could remove the commented out code

tpmanley (approved):

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
dkirker: I haven't experienced that.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
dkirker: 🤷‍♂ I suppose. I figured that I would leave it in in case someone wants to verify any issues that might come up in the future, but commented out so as not to have production code logging for every event. This log statement was the one original to the function.

gausnes:
I think this commit might have broken the DM integration tests. It looks like it is failing to compile in that environment.

Is this working in dev?


```
16:45:40.695 ERROR {}  <> [ ] (main) CompilerService - key=sandbox_error, owner=null, id=null,
java.lang.RuntimeException: Metadata Error: Capability 'stsmartweather.Wind Speed' not found.
	at physicalgraph.device.api.DeviceTypeMetadataApi.error(DeviceTypeMetadataApi.groovy:41)
	at physicalgraph.device.api.DeviceTypeMetadataApi.capability(DeviceTypeMetadataApi.groovy:69)
	at physicalgraph.device.DeviceTypeExecutorBase.capability(DeviceTypeExecutorBase.groovy:46)
	at org.kohsuke.groovy.sandbox.impl.Checker$1.call(Checker.java:66)
```

PR #22413, opened 2020-02-26
[WWST-6544] Fingerprint for Aurora AOne 13A Smart Plug

MWierzbinskaS:
Added fingerprint for Aurora AOne 13A Smart Plug
@greens  @tpmanley @KKlimczukS  @ZWozniakS  @MGoralczykS @PKacprowiczS 

DTH works this way that usage of 60W light bulb is seen as 6W in both Smartthings apps, thats because in parse method for this DTH in line 92 there is: def value = (event.value as Integer) / 10 . I do not know why there is this division by ten (it works the same way in local execution mode).

I think that this might be a bug. Could you check it so that we can be sure that everything works as intended? Or please tell me if there is anything that could be done about this issue?


tpmanley:
@MWierzbinskaS take a look at the usage of `"divisor"` in Zigbee Switch Power. You're going to want to do something similar in this DTH where you set the "divisor" to 1 for this device and 10 for all other devices. The hub will use that device data field so it will work when the device is running locally or in the cloud.

tpmanley (approved):

varzac (approved):

greens (approved):

PR #22581, opened 2020-02-27
Child Button: Add ocfDeviceType of x.com.st.d.remotecontroller

rboy1:
Include OIC type so that the new app shows the button actions (held/pushed) for child devices

@tpmanley

### DIFF
--- a/devicetypes/smartthings/child-button.src/child-button.groovy
+++ b/devicetypes/smartthings/child-button.src/child-button.groovy
@@ -14,7 +14,7 @@
  *
  */
 metadata {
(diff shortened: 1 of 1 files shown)


tpmanley (approved):
Seems reasonable to me. Any concerns @greens @dkirker ?

greens (approved):

greens:
Nope. Makes sense to me.

dkirker:
@tpmanley I wonder if they fixed that bug where I was having to override these in metadata? Anyway, I will be deploying metadata with the icons defined to production next Tuesday (for all of the other vids beyond `generic-button-2`, including Child Button).

PR #22989, opened 2020-03-02
[WWST-6586] Added FP for Aurora Smart Door/Window Sensor

ZWozniakS:
@greens could you please review? Should I group Aurora's fingerprints below others?
@KKlimczukS @PKacprowiczS @MGoralczykS @MWierzbinskaS


greens (approved):
You could have, but this is fine as is.

PR #23119, opened 2020-03-03
[WWST-5296] Fingerprint for Inovelli Switch

MWierzbinskaS:
Added fingerprint for Inovelli Switch
@greens  @KKlimczukS  @ZWozniakS  @MGoralczykS  @PKacprowiczS 

### DIFF
--- a/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
+++ b/devicetypes/smartthings/zwave-switch-generic.src/zwave-switch-generic.groovy
@@ -19,7 +19,6 @@ metadata {
 		capability "Polling"
 		capability "Refresh"
 		capability "Sensor"
-		capability "Light"
 
 		fingerprint inClusters: "0x25", deviceJoinName: "Z-Wave Switch"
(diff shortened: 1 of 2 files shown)


greens:
I would just put this in `zwave-switch-generic`. Security is handled by default now and that's a better DTH.

MWierzbinskaS:
I changed used DTH, so now fingerprint is in zwave-switch-generic. I deleted ''Light'' capability from this generic DTH, because it is deprecated and can be confusing and does not affect how device works. Do you think it is ok?

greens:
Once upon a time the "Light" capability was used to determine the icon used by a device so I'm not sure I want to delete it yet.

dkirker (approved):
Good to leave `Light` for now.

We should hold off removing capabilities (even if they are deprecated) from devices unless the side-effects have been investigated. While it is possibly safe to remove things like `Light` and `Holdable Button`, there is a chance that old SmartApps or integrations have relied on those capabilities. And certainly removing capabilities like `Thermostat` is dangerous. We really need to have a public communication and capability deprecation process (aside from just "marking" it deprecated) before we can actually remove the use of the capability.

MWierzbinskaS:
Thanks for clearing that up :)

PR #23142, opened 2020-03-04
ICP-12644 Update metadata for IKEA window shades to show Window Shade Preset

dkirker:
`generic-shade-3` has the Window Shade Preset capability and `generic-shade-2` does not. Now that there is code for Window Shade Preset we need to use that. I forgot to make this change in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/21405


tpmanley (approved):
Can you change the description to capture what the user will see? Something like `Display the Window Shade Preset capability for IKEA window shades`

dkirker:
Will merge this when we get the IKEA shades in MV so I can actually verify this myself.

tpmanley:
@dkirker I already tested this with my shades if you don't want to wait for yours to arrive

PR #23347, opened 2020-03-05
[ICP-12696] Added readAttribute for colorTemp in the setColor method

ZWozniakS:
@greens @dkirker could you please review? I added changes to the setColor method to read the color temp attribute to fix the reported issue, and according to this logic, I thought we should be able to change the position of the color picker when changing the color temp. I am not sure if this is possible, but I tried to apply it in a similar way by reading the hue and saturation attributes in the setColorTemperature method. But ... it didn't work. Hue and saturation values ​​have not changed, they just kept their previous values. There is still a question, shouldn't this be possible?

@KKlimczukS @PKacprowiczS @MGoralczykS @MWierzbinskaS

### DIFF
--- a/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zll-rgbw-bulb.src/zll-rgbw-bulb.groovy
@@ -224,7 +224,9 @@ def setColorTemperature(value) {
 
 	zigbee.command(COLOR_CONTROL_CLUSTER, MOVE_TO_COLOR_TEMPERATURE_COMMAND, "$finalHex 0000") +
 	["delay 1500"] +
-	zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_COLOR_TEMPERATURE)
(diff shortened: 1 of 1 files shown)


varzac:
So I am not 100% certain, but by my reading the ZCL spec indicates that the device would not update hue/saturation when the color temperature changes.

See sections 5.2.2.2.1.8, 5.2.2.2.1.9, and 5.2.2.3.2.  Basically there is an attribute on the device called `ColorMode` that determines which set of attributes is controlling the color of the device (either hue/sat, x/y, or colortemp), and it will only ever be one.  When you change from one mode to another, it starts from a conversion of the old value, but it does not look like it will update the value of the other modes.

tpmanley:
Also `executeCommandsLocally: true` is set on this DTH so you cannot make changes to capability commands since in most cases the command will be executed on the hub and not in the DTH.

greens:
The ticket is an invalid test case.

dkirker:
Just a side note, I left this comment on the ticket: I spent 15 minutes and ran the DTH (with the local execution flags to false and some more logging) and can confirm that at least with an OSRAM bulb setting the color does not affect the color temp values that the bulb reports, and setting the color temperature value does not affect the color values.

ZWozniakS:
I am closing this PR as it was discussed on the ticket. Thank you all for your help and clarification on that subject!

PR #23444, opened 2020-03-06
Commit hooks

greens:


greens:
Might add something to `README` indicating people should run `gradle configure`.

BroderickCarlin (dismissed):

varzac (dismissed):
Looks good, but should probably squash.

.githooks/pre-commit
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+STASH_NAME="pre-commit-$(date +%s)"
+git stash save -q --keep-index $STASH_NAME
+
+CHANGED_DTHS=$(git diff --cached --name-only | grep .*.groovy)
```
dkirker: It looks like you didn't end up using this elsewhere other than the commented out line below.

build.gradle
```diff
@@ -138,3 +138,11 @@ slackSendMessage {
         text: messageText
     )
 }
+
+task configure(type: Exec) {
+    File attributeFile = new File("${projectDir}/.git/info/attributes")
+    attributeFile.write("*.groovy filter=tabspace\n")
+    commandLine "git", "config", "filter.tabspace.clean", "unexpand -t 2"
```
dkirker, line 146: I know we still have code that uses spaces, does this affect code outside of the users changes? (I'm not familiar with this git config.)

dkirker (commented):
I left a couple of comments inline.

cbaumler (dismissed):

.githooks/pre-commit
```diff
@@ -0,0 +1,32 @@
...
+  if [[ ${#IMPORTANT_ERRORS} -eq 0 ]]; then
+    echo "No disqualifying compilation errors found"
+  else
+    echo "$DTH failed to compile, run groovyc on your source file for the full error: $ERRORS"
+    ERROR_COUNT=$((ERROR_COUNT + 1))
+  fi
+  echo "======================================================================="
+done < <(git diff --cached --name-only | grep .*.groovy)
+
+STASHES=$(git stash list)
```
msandstedt: I think the stash logic must be left over and doesn't do anything because:
* == will never be true because $STASHES is a list, and each element is annotated with more than just the stash name
* we shouldn't need to stash pop anyway because stash save --keep-index didn't actually stash our changes

So in the end, can't we get rid of the stash logic altogether?

build.gradle
greens: It only works on files included in the commit.

.githooks/pre-commit
greens: It was working when I tested it, but I can take it out since we're not _really_ compiling anything.

msandstedt: OK, I see what's being done here.  However, I think the substring match with == is going be pretty fragile.  This is notoriously unportable.  It definitely does not work in my environment.

If you can get $STASHES to only contain the thing you're looking for, this will be much more likely to work everywhere.

I got this working in my environment with the following:

```
STASHES=$(git stash list -n 1 --format=%s | cut -d: -f2 | cut -d\  -f2)
if [[ $STASHES == $STASH_NAME" ]]; then
  git stash pop -q
fi
```

dkirker: I assume you meant `if [[ $STASHES == "$STASH_NAME" ]]; then` and not `if [[ $STASHES == $STASH_NAME" ]]; then`?

msandstedt (approved):

PR #23516, opened 2020-03-07
Keep compatibility with existing SmartApps

rboy1:
Add capability "Thermostat" so that devices are compatible with existing SmartApps

### DIFF
--- a/devicetypes/smartthings/zwave-battery-thermostat.src/zwave-battery-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-battery-thermostat.src/zwave-battery-thermostat.groovy
@@ -15,6 +15,7 @@ metadata {
 	definition (name: "Z-Wave Battery Thermostat", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.thermostat", genericHandler: "Z-Wave") {
 		capability "Actuator"
 		capability "Temperature Measurement"
+		capability "Thermostat"
 		capability "Thermostat Heating Setpoint"
 		capability "Thermostat Cooling Setpoint"
 		capability "Thermostat Operating State"


rboy1:
@tpmanley

dkirker:
This is probably fine, however the metadata would need to be updated to ensure that it doesn't break the new mobile client.

One concern that I have is that this would change the OCF hrefs that the mobile client is using because of the way ST<->OCF was implemented and I am not sure what the effects on existing rules and automations will be.

So for now I am going to hold off approving this until I can have a chance to actually understand what will break and if it will break.

tpmanley:
@dkirker would the UI metadata have to be updated? New apps (not existing Groovy SmartApps) should be using the new Thermostat* capabilities and Groovy SmartApps don't use the UI metadata.

greens:
@tpmanley The issue (and we've run into this before) is with capabilities duplicating values. I don't think we can 100% safely say that we're muxing those events properly yet.

Similarly I believe it sets a bad precedent to add _back in_ deprecated capabilities. The goal should be moving to change the existing DTHs that _do_ have `Thermostat`

dkirker:
It isn't so much the duplicating values that is the issue here (though, that is an issue for some capabilities -- and at least into the event history service). The issue in the past was that while the mobile client could send OCF data to either the `Thermostat` capability or any of the "sub-capabilities", the only data that would come back to the mobile client would be through the `Thermostat` capability. This resulted in behavior like the network errors and failing to see the mobile client update.

Now, it is possible that this has changed. If it has, then there would technically be no problem with adding `Thermostat` into the capability list. While I agree with @greens about not adding back deprecated capabilities, I think this is one of the capabilities where no thought was put in to the deprecation and implementation, and as such, it has created a mess. (We also have still yet to actually come up with an official deprecation process, aside from the property in the capability definition. But, this is a whole separate can of worms.)

@rboy1 By any chance have you used a device that uses this DTH with the change and had it working in the new app?

rboy1:
@dkirker I did some tests in the new app with the existing DTH and with the proposed changes. I deleted and created a new device in each case, and it looks exactly the same.
The only difference I've noticed is without the Thermostat capability it doesn't work with existing SmartApps.

Here are the screenshots of what it looks like (same for both existing and modifications):

dkirker:
I played a little with this on Monday and the way that the thermostat capability is mapped seems to have been updated. I will check in with the correct team just to be sure.

PR #23677, opened 2020-03-09
[CHAD-4341] Update generic Z-Wave Light bulb DTH for Z-Wave Light bulbs

msandstedt:
This commit moves all devices currently matching aeon-led-bulb.groovy to
the generic Z-Wave RGBW color bulb DTH, rgbw-light.groovy by removal
of the generic device type / command-class fingerprints in the Aeon
driver and addition of appropriate device type / command class
fingerprints in the generic driver.  This is deemed appropriate because
the Aeon driver actually doesn't have any Aeon-specific manufacturer or
model fingerprinting.  In fact, the Aeon / Aeotec LED bulbs are already
matching to the generic driver with these fingerprints:

* fingerprint mfr: "0086", prod: "0003", model: "0062", deviceJoinName: "Aeotec LED Bulb" //EU
* fingerprint mfr: "0086", prod: "0103", model: "0062", deviceJoinName: "Aeotec LED Bulb" //US

This leaves only non-Aeon bulbs matching to the Aeon driver, and these
can be better supported in the generic driver.

The related aeon-led-bulb-6 driver is left intact though, as these bulbs
use non-standard commands for white temperature control, and thus are
incompatible with the generic driver.  It was verified on the bench that
temperature control for the Aeon / Aeotec LED Bulb 6 device does not work
properly with the generic RGBW DTH, but does work with the existing
model-specific aeon-led-bulb-6 DTH.

Additionally, some refinements are made to the RGBW driver to improve
reliability of event emission and state update after user-initiated color
and color temperatures changes.  Finally, we lower-bound cold and warm
white intensity from the color temperature event handler to 1.  Within
standard Z-wave, temperature is controlled by crossfading cold and warm
white channels.  But for devices that only have single cold or warm white
illumination (true for many RGBW LED strips), we cannot dim either white
channel to 0, as this will then completely turn off illumination.  The
lower bound of 1 gives more intuitive operation for these devices, whereby
temperature effectively becomes an intensity modifier for the single cold
or warm white channel.

### DIFF
--- a/devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
+++ b/devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
@@ -27,15 +27,48 @@ metadata {
 		capability "Sensor"
 		capability "Health Check"
 
-		command "reset"
+		/*
+		 * Relevant device types:
+		 *
+		 * * 0x11 GENERIC_TYPE_SWITCH_MULTILEVEL
+		 * * 0x01 SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		 * * 0x02 SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		 *
+		 * Plausible command classes we might see in a color light bulb:
+		 *
+		 * 0x98 COMMAND_CLASS_SECURITY
+		 * 0x5E COMMAND_CLASS_ZWAVEPLUS_INFO_V2
+		 * 0x20 COMMAND_CLASS_BASIC
+		 * 0x26 COMMAND_CLASS_SWITCH_MULTILEVEL
+		 * 0X27 COMMAND_CLASS_SWITCH_ALL
+		 * 0x33 COMMAND_CLASS_SWITCH_COLOR
+		 * 0x70 COMMAND_CLASS_CONFIGURATION
+		 * 0x73 COMMAND_CLASS_POWERLEVEL
+		 *
+		 * Here are the command classes used by this driver that we can fingerprint against:
+		 *
+		 * * 0x26 COMMAND_CLASS_SWITCH_MULTILEVEL -> yes, it is dimmable
+		 * * 0x33 COMMAND_CLASS_SWITCH_COLOR -> yes, it has color control
+		 */
 
-		fingerprint inClusters: "0x26,0x33"
-		fingerprint inClusters: "0x33"
+		// dimmable, color control
+		fingerprint inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_POWER_SWITCH_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1101", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// GENERIC_TYPE_SWITCH_MULTILEVEL:SPECIFIC_TYPE_COLOR_TUNABLE_MULTILEVEL
+		// dimmable, color control
+		fingerprint deviceId: "0x1102", inClusters: "0x26,0x33", deviceJoinName: "Z-Wave RGBW Bulb"
+
+		// Manufacturer and model-specific fingerprints.
 		fingerprint mfr: "0086", prod: "0103", model: "0079", deviceJoinName: "Aeotec LED Strip" //US
 		fingerprint mfr: "0086", prod: "0003", model: "0079", deviceJoinName: "Aeotec LED Strip" //EU
 		fingerprint mfr: "0086", prod: "0003", model: "0062", deviceJoinName: "Aeotec LED Bulb" //EU
 		fingerprint mfr: "0086", prod: "0103", model: "0062", deviceJoinName: "Aeotec LED Bulb" //US
 		fingerprint mfr: "0300", prod: "0003", model: "0003", deviceJoinName: "ilumin RGBW Bulb"
+		fingerprint mfr: "031E", prod: "0005", model: "0001", deviceJoinName: "ilumin RGBW Bulb"
 	}
 
 	simulator {
@@ -60,17 +93,38 @@ metadata {
 		}
 	}
 
-	standardTile("reset", "device.reset", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
-		state "default", label:"Reset Color", action:"reset", icon:"st.lights.philips.hue-single"
-	}
-
 	controlTile("colorTempSliderControl", "device.colorTemperature", "slider", width: 4, height: 2, inactiveLabel: false, range:"(2700..6500)") {
 		state "colorTemperature", action:"color temperature.setColorTemperature"
 	}
 
 	main(["switch"])
-	details(["switch", "levelSliderControl", "rgbSelector", "colorTempSliderControl", "reset"])
+	details(["switch", "levelSliderControl", "colorTempSliderControl"])
 }
+
+private getCOLOR_TEMP_MIN() { 2700 }
+private getCOLOR_TEMP_MAX() { 6500 }
+// For Z-Wave devices, we control illumination by crossfading the cold and warm
+// white channels.  But for devices that only have single cold or warm white
+// illumination (as with many RGBW LED strips), we cannot dim either white
+// channel to 0, as this will then completely turn off illumination.
+// Therefore, we lower-bound both white channels to 1.  This will have no
+// perceptible impact for devices that actually support white temperature
+// cross-fading, but will keep devices that do not doing something sane from
+// the user perspective, which is to modify intensity for the single white
+private getWHITE_MIN() { 1 } // min for Z-Wave coldWhite and warmWhite paramaeters
+private getWHITE_MAX() { 255 } // max for Z-Wave coldWhite and warmWhite paramaeters
+private getCOLOR_TEMP_DIFF() { COLOR_TEMP_MAX - COLOR_TEMP_MIN }
+private getRED() { "red" }
+private getGREEN() { "green" }
+private getBLUE() { "blue" }
+private getWARM_WHITE() { "warmWhite" }
+private getCOLD_WHITE() { "coldWhite" }
+private getRGB_NAMES() { [RED, GREEN, BLUE] }
+private getWHITE_NAMES() { [WARM_WHITE, COLD_WHITE] }
+private getCOLOR_NAMES() { RGB_NAMES + WHITE_NAMES }
+private MIN(a, b) { a < b ? a : b }
+private MAX(a, b) { a > b ? a : b }
+private BOUND(x, floor, ceiling) { x < floor ? floor : x > ceiling ? ceiling : x }
 
 def updated() {
 	log.debug "updated().."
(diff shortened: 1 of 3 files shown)


greens:
The one thing I'd say is that this might be a good time to go ahead and start doing the best practice of querying the device itself for its supported colors and then perhaps based on that we could do more intelligent handling of `colorTemperature`?

msandstedt:
> The one thing I'd say is that this might be a good time to go ahead and start doing the best practice of querying the device itself for its supported colors and then perhaps based on that we could do more intelligent handling of `colorTemperature`?

@greens,

Good point.  There's no reason this can't be collapsed into the single DTH.  And since @kianooshST had the foresight to make sure we have an Aeon LED bulb 6 on hand, I actually have a device to test this with.

I will put together some amendments and push these up.

dkirker:
It is also worth noting that if we do need to hardcode logic for a specific model we have instances where we have done this. https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-battery-accessory-dimmer.src/zigbee-battery-accessory-dimmer.groovy is a good example (see the use of the `is*()` functions). Of course being mindful of whether that logic is executed locally or not when appropriate.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -186,16 +214,22 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 	[linkText: linkText, descriptionText: "$linkText: $cmd", displayed: false]
 }
 
-def buildOffOnEvent(cmd){
-	[zwave.basicV1.basicSet(value: cmd), zwave.switchMultilevelV3.switchMultilevelGet()]
-}
```
dkirker, line 191: I feel like it would be great to reduce the redundant code below and continue to use this function, but change the `basicSet` to the `switchMultilevelSet`, and bring the use of `commands` into this function. And maybe call it `buildSwitchCommand` or something a little more relevant. Then you only have to call `buildSwitchCommand(0xFF)` or `buildSwitchCommand(0x00)`. And hell, I guess `ON` and `OFF` could just be made in to `getSWITCH_ON_VALUE() { 0xFF }` and `getSWITCH_OFF_VALUE() { 0x00 }` or something like that.

It could even be possible to re-use that function elsewhere that `switchMultilevelSet` is used?

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -232,53 +265,54 @@ def setHue(value) {
...
+		rgb = colorUtil.hexToRgb(hex) // separate RGB elements for zwave setter
 	}
-	if (device.currentValue("switch") != "on") {
-		result << zwave.basicV1.basicSet(value: 0xFF)
-		result << zwave.switchMultilevelV3.switchMultilevelGet()
-	}
-	result += queryAllColors()
+	commands([zwave.switchColorV3.switchColorSet(red: rgb[0], green: rgb[1], blue: rgb[2], warmWhite: 0, coldWhite: 0),
+	          zwave.switchColorV3.switchColorGet(colorComponent: RGB_NAMES[0]), // event-publish callback is on any of the RGB responses, so only need to GET one of these
+	], 3500)
```
dkirker, line 286: I think this too could be made into a helper function (like with the `on()` and `off()` calls) since the code is somewhat re-used below.

dkirker (commented):
I added a couple of small suggestions inline.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
msandstedt: Resolved in the update.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
msandstedt: I decided to leave this since the opportunity for reuse was less than the multiLevelSet.  Reusing / wrapping the multiLevelSet was a lot more straight forward.

msandstedt:
> The one thing I'd say is that this might be a good time to go ahead and start doing the best practice of querying the device itself for its supported colors and then perhaps based on that we could do more intelligent handling of `colorTemperature`?

@greens, @dkirker,

I pulled in support of the Aeon  LED Bulb 6.  This is accomplished by mfr/model match against zwaveInfo.  This was a valuable exercise, as I suppose DTH consolidation and code reuse usually is.    The generic RGBW DTH now works better and more reliably than the dedicated DTH did, so that's great too.  Amongst other things, the white temperature control (which is quite finicky with the Aeon/Aeotec LED bulb 6) works a lot better.

I decided not to do add any further dynamic behavior to the DTH at this time, as the ColorSwitchSupportedGet() command does not appear to be supported on all devices I'm working with.  And I believe the rationale and behavior introduced with the initial commit for the LED strips is still very good without this.

Regardless, per the spirit of your comments, the treatment of white temperature is improved and we do have more consolidation into the generic RGBW DTH.

greens (dismissed):

kianooshkarami (dismissed):

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -232,53 +268,109 @@ def setHue(value) {
...
+	// Stage ST colorTemperature attribute.
+	state.staged << [colorTemperature: temp]
+	// Emit our command.  Follow up with callback-trigger getter.  Our
+	// event-publish callback is on any of the while-level responses,
+	// so we only need to GET one of these these.  Make sure that we
+	// only have delay immediately before the callback trigger getter.
+	if (cmds.size() > 1) {
+		def prologue = cmds.init()
+		cmds = cmds.tail()
+		commands(prologue, 0) // emit in quick succession
```
dkirker: I'm not sure that I follow exactly what is going on here. Maybe I don't fully understand Groovy, but wouldn't the `commands` call here ultimately not get returned? `commands` ultimately calls `delayBetween` which returns an array or command strings that you can either return from a Capability command or pass to `sendHubEvent`. I think in this case the `commands(epilogue, 3500)` below will be the only thing returned. Unless Groovy has a feature where it takes both results from the calls to `commands` and adds them and returns that... (I don't admit to having a deep knowledge of Groovy's quirks.)

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -60,29 +100,54 @@ metadata {
...
+private getWARM_WHITE() { "warmWhite" }
+private getCOLD_WHITE() { "coldWhite" }
+private getRGB_NAMES() { [RED, GREEN, BLUE] }
+private getWHITE_NAMES() { [WARM_WHITE, COLD_WHITE] }
+private getCOLOR_NAMES() { RGB_NAMES + WHITE_NAMES }
+private getSWITCH_VALUE_ON() { 0xFF } // Per Z-Wave, this multilevel switch value commands state-transition to on.  This will restore the most-recent non-zero value cached in the device.
+private getSWITCH_VALUE_OFF() { 0 } // Per Z-Wave, this multilevel switch value commands state-transition to off.  This will not clobber the most-recent non-zero value cached in the device.
+private MIN(a, b) { a < b ? a : b }
+private MAX(a, b) { a > b ? a : b }
+private BOUND(x, floor, ceiling) { x < floor ? floor : x > ceiling ? ceiling : x }
```
dkirker: Isn't this basically the same as `Math.max(Math.min(x, ceiling), floor)`? Would it make sense to use one or the other? (You have `Math.max(Math.min(level, 99), 1)` below, which I am thinking could just be `BOUND(level, 1, 99)` and `BOUND` either being what it is or the `Math.min`+`Math.max`.)

msandstedt: Ah, fair enough.  Yes, this is redundant.  Will fixup.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
msandstedt: I assumed that delayBetween was calling sendHubEvent, or something to this effect.  Certainly, that's the only way this _can_ work and I was sure it did.

I'll investigate tomorrow.

Does anybody on the review know what repo provides the DTH API?  I would like to see the delayBetween and sendHubEvent methods and am coming up short.

msandstedt: @dkirker,

You are absolutely correct on this.  My code works, but in a backward unintended way resulting from my misunderstanding of the API.  Indeed, just calling commands and not ultimately returning it from the method doesn't do anything.

Fix will be forthcoming.

msandstedt: Yup, you were right.  Fixed.  See update.

msandstedt:
OK, @dkirker's comments are addressed.  Thanks @dkirker for the detailed review.

This should be good to go.  I have re-tested with the Aeon/Aeotec LED strip, the Aeotec LED Bulb 6 and the Illumin RGBW bulb.  Monitoring both the Graph live logging and the hub-core log, and based on behavior in the app and observation of the bulbs and LED strip, everything now works as I expect.

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
```diff
@@ -232,53 +266,105 @@ def setHue(value) {
...
+	// Emit our command.  Follow up with callback-trigger getter.  Our
+	// event-publish callback is on any of the while-level responses,
+	// so we only need to GET one of these these.  Make sure that we
+	// only have delay immediately before the callback trigger getter.
+	def prologue = cmds.init() // grab all but last
+	def epilogue = [] << cmds.last() // grab last
+	epilogue << zwave.switchColorV3.switchColorGet(colorComponent: WHITE_NAMES[0]) // append callback get
+	def rv = prologue.size() > 0 ? commands(prologue, 0) : [] // collect formatted prologue; check for empty; empty is OK, but can't be passed to commands()
+	rv << commands(epilogue, 3500) // collect formatted epilogue; only delay immediately before callback getter
+	return rv // return formatted command array to execute
```
dkirker, line 307: Just so that I understand this function, the goal is to take all of the temperature set commands, add a delay of 3500, and then add the get for the white color component? If so, `delayBetween` basically just puts `delay $delayValue` in between each item in the array. You can do this manually if you wanted. An example is here: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/1a19a83ce401baf2e6ed74975bbe3ed3d257a3bf/devicetypes/smartthings/zwave-motion-temp-light-sensor.src/zwave-motion-temp-light-sensor.groovy#L174

So you probably could have:
```
def result = [];

result += commands(cmds)
result += ["delay 3500", command(zwave.switchColorV3.switchColorGet(colorComponent: WHITE_NAMES[0]))]

return result
```

That said... It definitely feels like a bad idea to have that hardcoded around (which I am guilty of doing). Too bad we didn't just create a `delay(delayValue) { "delay $delayValue" }` API method for device types to use.

I'll leave it to your discretion if that is the case and if you want to change it. I should have paid a little more attention to it earlier when I left the last comments. My apologies.

dkirker (approved):

devicetypes/smartthings/rgbw-light.src/rgbw-light.groovy
msandstedt: This is fair.  Your solution is simpler.  As you allude, only real advantage to the approach in the PR is that the delay command string is abstracted / obfuscated through the delayBetween API call.

Had I known before that this simple command string was available, I probably would have used it.  For now, I am going to leave this since it does work and I'd like to close out this work.

Thank you for the careful review.

PR #23774, opened 2020-03-09
Changing isComponent to false for new app

InovelliUSA:
Simple change to make this device work when adding it with the new SmartThings app.

### DIFF
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug.src/inovelli-2-channel-smart-plug.groovy
@@ -259,7 +259,7 @@ private void createChildDevices() {
 				device.hubId,
 				[completedSetup: true,
 				 label: "${device.displayName} (CH${i})",
-				 isComponent: true,
+				 isComponent: false,
 				 componentName: "ep$i",
 				 componentLabel: "Channel $i"
 		])


tpmanley:
I might need a second look at how it works today in the new app. The UI Metadata looks correct already but maybe I'm missing something.

InovelliUSA:
Yeah, totally. Just want users to be be able to see the independent channels somewhere in the new app. Also to be able to add them in scenes, automations, etc.

tpmanley:
Actually I remember the issue, the DTH doesn't have the `mcdSync: true` flag so it's not being treated as a true MCD device in the new app.

@dkirker can you help out here and see if the UI Metadata will correctly handle the device as a MCD device if `mcdSync` is set to true?

InovelliUSA:
Hey guys, is there anything that I can change to get this updated?

dkirker:
@InovelliUSA We're currently investigating the work needed to make sure that existing installations will properly be able to work with the changes. As you can imagine over the last couple of weeks things have been slowed. I should have a status update on Friday.

InovelliUSA:
Sounds good @dkirker. Please let me know if you need anything from me.

InovelliUSA:
Hi @dkirker, is there any update on the progress of this?

dkirker:
@InovelliUSA We are currently investigating options for transitioning to the new experience in a way that won't break existing devices that have been set up, including automations. I've identified at least one solution, and am working with some other teams on it.

dkirker:
Small update: We are still working out the best (safest) way to make these changes. We have some of the other teams engaged.

InovelliUSA:
Awesome, thanks for the update!

tpmanley:
@InovelliUSA unfortunately it's pretty difficult to update the DTH of existing devices to MCD without risking breaking rules and apps for users. We're instead planning on deprecating the current DTH and adding a new MCD DTH. Could you take a look at the code, give it a spin and see what you think? The UI Metadata is already in prod so you can self publish the DTH and test it out.

https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/28900

InovelliUSA:
Ok, I will test it out. Thanks.

InovelliUSA:
So this seems to work great! The only problem I ran into is when switching from the old handler to this one, I get 4 child devices and the two old ones show up as EP1 & EP2 in the new app. If I delete the old child devices in the IDE they don't disappear from the new app. 

image

dkirker:
@InovelliUSA Thanks! You'll want to pair the switch as a new device. The plan that we have right now is that the new code is being published as a new DTH so new devices will pair with that, and existing devices will continue to use the old DTH. Eventually we will work out a system to bring the old devices to the new DTH, either by notifying users that they should re-pair, or by some sort of migration. Unfortunately, it isn't easy to do a one-to-one migration of the child devices to the format that is needed to support the new multi-component setup. :(

InovelliUSA:
Ok, that works. I was hoping to have a solution for people that currently have the device installed, but I guess I can just tell them to re-include after you push the device handler to production. You don't have a time frame for that do you?

InovelliUSA:
Sorry to bug you again, but do you have a time frame of when you are going to push the new device handler to production?

tpmanley:
@InovelliUSA it went to production this last Tuesday. Please let us know if you hear about any issues with it, and of course PRs are welcome :)

InovelliUSA:
Awesome, thanks!

PR #24289, opened 2020-03-12
[ICP-12761] Add Chime event to installed, delete old metadata name from definition.

MGoralczykS:
@greens @dkirker @PKacprowiczS @ZWozniakS @KKlimczukS @MWierzbinskaS 
Could you review this?

### DIFF
--- a/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
+++ b/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
@@ -14,7 +14,7 @@
  *
  */
 metadata {
-	definition(name: "Aeotec Doorbell Siren 6", namespace: "smartthings", author: "SmartThings", mcdSync: true,  mnmn: "SmartThings", vid: "generic-8-sound") {
+	definition(name: "Aeotec Doorbell Siren 6", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
 		capability "Actuator"
 		capability "Health Check"
 		capability "Tamper Alert"
@@ -61,6 +61,7 @@ private getNumberOfSounds() {
 def installed() {
 	initialize()
 	sendEvent(name: "alarm", value: "off", isStateChange: true, displayed: false)
+	sendEvent(name: "chime", value: "off", isStateChange: true, displayed: false)
 	sendEvent(name: "tamper", value: "clear", isStateChange: true, displayed: false)
 }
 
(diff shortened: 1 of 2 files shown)


greens (approved):

greens:
@dkirker can you merge this after you've reviewed and merged the metadata?

devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
```diff
@@ -14,7 +14,7 @@
  *
  */
 metadata {
-	definition(name: "Aeotec Doorbell Siren 6", namespace: "smartthings", author: "SmartThings", mcdSync: true,  mnmn: "SmartThings", vid: "generic-8-sound") {
+	definition(name: "Aeotec Doorbell Siren 6", namespace: "smartthings", author: "SmartThings", mcdSync: true) {
```
dkirker, line 17: One thing I would recommend checking is installing a device with the DTH as it currently is in production (maybe best to self-publish so that you aren't changing the device type through the edit device) and then self-publishing this change to make sure that the new vid is used. I *think* on the DTH level there aren't problems, but I seem to recall issues with deleting a vid from a DTH where even if we were just relying on the DTH name we still had to declare it so that it would update for devices that were already installed.

dkirker: If that doesn't seem to be an issue then we should be fine to merge this.

MGoralczykS: @dkirker , @greens 
I left a comment under metadata MR. 
Should we use separately metadata for Siren and Doorbell?
I mean for Siren metadata with Alarm (generic-siren-12) capability and for a Doorbell with Chime.

MGoralczykS: @dkirker ping

dkirker: @MGoralczykS I haven't had a chance to look over the implications of either route. I'll have a look by Friday.

MGoralczykS: @dkirker any updates :)?

dkirker: @MGoralczykS I'm impartial to either way. Since we implement both devices the same way, and they are the same actual device but with a different fingerprint, it might make sense to just use the same metadata. If you want another opinion, you can check with product.

greens: @dkirker I think it's fine, then. Can you verify the metadata is out there and merge?

dkirker (approved):

PR #24828, opened 2020-03-14
Create ilumin bulb device handler with preference for bulb power on state

InovelliUSA:
Used the current device handlers for ilumin bulb so changes are minimal.

Updates to 

RGBW Light
Aeon Multi-White Bulb

### DIFF
--- a/devicetypes/smartthings/ilumin-bulb-multi-color.src/ilumin-bulb-multi-color.groovy
+++ b/devicetypes/smartthings/ilumin-bulb-multi-color.src/ilumin-bulb-multi-color.groovy
@@ -17,7 +17,7 @@
  */
 
 metadata {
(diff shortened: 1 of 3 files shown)


tpmanley:
@msandstedt recently did some internal clean up of our Z-Wave RGBW light DTH: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/23677. That PR includes the ilumin fingerprints. Can you give the DTH from that PR a try and see how it works for you? It won't have the device preference that's in this DTH but perhaps you could add that to the RGBW Light as a PR once @msandstedt 's PR gets merged.

InovelliUSA:
I can definitely check that out. One question I have is that we have that single preference that we would like our customers to have. Since some of them don't want to be blinded by their bulbs at 3am when the power flickers.

What is the best way to handle that? I thought I had to create a new device handler since there is no dynamic way to present certain preferences to certain devices.

PR #25329, opened 2020-03-17
fixes several exceptions (mostly NPEs)

greens:
also switches legacy code checking for secure encapsulation to use "contains" instead of "startsWith"

### DIFF
--- a/devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
+++ b/devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
@@ -204,7 +204,7 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
 }
 
 def isFibaro() {
-	(zwaveInfo?.mfr == "010F" && zwaveInfo.prod == "0C02")
+	(zwaveInfo?.mfr? == "010F" && zwaveInfo?.prod? == "0C02")
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityCommandsSupportedReport cmd) {
@@ -519,7 +519,7 @@ private setSecured() {
 
 private isSecured() {
     if (zwaveInfo && zwaveInfo.zw) {
-        return zwaveInfo.zw.endsWith("s")
+        return zwaveInfo.zw.contains("s")
     } else {
         return getDataValue("secured") == "true"
     }
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -133,7 +133,7 @@ def parse(String description) {
 						map = getBatteryResult(Integer.parseInt(battMap.value, 16))
 					}
 				}
-			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002) {
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07)) {
 				def zs = new ZoneStatus(zigbee.convertToInt(descMap.value, 16))
 				map = translateZoneStatus(zs)
 			} else if (descMap?.clusterInt == zigbee.TEMPERATURE_MEASUREMENT_CLUSTER && descMap.commandInt == 0x07) {
--- a/devicetypes/smartthings/zigbee-motion-temp-humidity-sensor.src/zigbee-motion-temp-humidity-sensor.groovy
+++ b/devicetypes/smartthings/zigbee-motion-temp-humidity-sensor.src/zigbee-motion-temp-humidity-sensor.groovy
@@ -113,7 +113,7 @@ def parse(String description) {
 				if (battMap) {
 					map = getBatteryResult(Integer.parseInt(battMap.value, 16))
 				}
-			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002) {
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07)) {
 				def zs = new ZoneStatus(zigbee.convertToInt(descMap.value, 16))
 				map = translateZoneStatus(zs)
 			} else if (descMap?.clusterInt == zigbee.TEMPERATURE_MEASUREMENT_CLUSTER && descMap.commandInt == 0x07) {
--- a/devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
+++ b/devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
@@ -145,7 +145,7 @@ def parse(String description) {
 			result = zwaveEvent(cmd)
 		}
 	}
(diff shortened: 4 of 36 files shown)


devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -133,7 +133,7 @@ def parse(String description) {
 						map = getBatteryResult(Integer.parseInt(battMap.value, 16))
 					}
 				}
-			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002) {
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07)) {
```
cbaumler, line 136: Is the NPE check not necessary on `attrInt` and `commandInt` because they're guaranteed to exist?

cbaumler (approved):
LGTM

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
greens: I originally had a null check, but I'm like 90% sure these are config reporting responses so I went back to this.

You're probably right.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
```diff
@@ -145,7 +145,7 @@ def parse(String description) {
 			result = zwaveEvent(cmd)
 		}
 	}
-	if (result?.name == 'hail' && hubFirmwareLessThan("000.011.00602")) {
+	if (result?.name? == 'hail' && hubFirmwareLessThan("000.011.00602")) {
```
tpmanley, line 148: I don't believe this change will have any effect. DId you see an error from this?

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
```diff
@@ -92,7 +92,7 @@ private getCommandClassVersions() {
 def parse(String description) {
 	def result = null
 	if (description.startsWith("Err 106")) {
-		if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo?.zw?.endsWith("s")) {
+		if ((zwaveInfo.zw == null && state.sec != 0) || zwaveInfo?.zw?.contains("s")) {
```
tpmanley, line 95: The null checks on `zwaveInfo` are not consistent across all these changes

tpmanley: It's not clear to me that `zwaveInfo` can ever be null. Do you have examples where it is?

greens: I think these are all coming from very old installed devices that predate it.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
greens: `timestamp=2020-03-17 05:44:56.215, level=WARN, ... deviceTypeName=Z-Wave Dimmer Switch Generic, namespace=smartthings, state=LATEST_APPROVED, type=parse, runTime=25,
groovy.lang.MissingPropertyException: Exception evaluating property 'name' for java.util.ArrayList, Reason: groovy.lang.MissingPropertyException: No such property: name for class: physicalgraph.device.HubAction
	at org.kohsuke.groovy.sandbox.impl.Checker$4.call(Checker.java:115)
	at physicalgraph.sandbox.SmartAppSandbox.onGetProperty(SmartAppSandbox.groovy:362)
	at org.kohsuke.groovy.sandbox.impl.Checker$4.call(Checker.java:113)
	at org.kohsuke.groovy.sandbox.impl.Checker.checkedGetProperty(Checker.java:110)
	at script_dth_1bfe4865c2c2c3ee50a5122025aaba2b769733032b8ad675beb7d2799f96a989.parse(script_dth_1bfe4865c2c2c3ee50a5122025aaba2b769733032b8ad675beb7d2799f96a989:148)`

so in retrospect most likely not the right fix, but I did see it

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
greens: I've decided to stick with this because it seems to be a pattern employed elsewhere.

devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
greens: in double-retrospect, this looks like the right fix

this error is thrown in response to the `stopMultilevelChange` command, since it generates two events.

tpmanley: I also see this is specific to very old hub firmware so can probably be completely removed at some point but looks fine for the time being.

devicetypes/smartthings/zwave-door-window-sensor.src/zwave-door-window-sensor.groovy
tpmanley: Based on the `getZwaveInfo` implementation I don't see how it can be null regardless of when the device was installed. I could be misreading the code but that's why I'm asking if you've seen examples where it is null?

greens: So `zwaveInfo.zw` can be null. `zwaveInfo` might never be

devicetypes/smartthings/zigbee-motion-temp-humidity-sensor.src/zigbee-motion-temp-humidity-sensor.groovy
```diff
@@ -113,7 +113,7 @@ def parse(String description) {
 				if (battMap) {
 					map = getBatteryResult(Integer.parseInt(battMap.value, 16))
 				}
-			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002) {
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07)) {
```
dkirker, line 116: Looks like there is an extra `)` here, too.

devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -204,7 +204,7 @@ def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulat
 }
 
 def isFibaro() {
-	(zwaveInfo?.mfr == "010F" && zwaveInfo.prod == "0C02")
+	(zwaveInfo?.mfr? == "010F" && zwaveInfo?.prod? == "0C02")
```
dkirker, line 207: I suspect this *might* fail in the same way that the other DTH did? Though I am not sure why just having the original line would cause a NPE? It seems that `if (null != "some string")` works in a Groovy console. Unless there is something else going on with the way `getZwaveInfo` in DM works?

PR #25414, opened 2020-03-18
ICP-12430 Sinope Technologies:TH1123ZB-TH1124ZB Sinope Thermostat update to revision 1.2.0

SmartThings:
Fix : Offline device don't send event anymore Adaptation for new SmartThings App Capability Thermostat added even if it's deprecated to support thermostat SmartApp General fixes


Sinopetech:
@shrakuma
We have updated our driver to fix multiple issues : 
- Offline devices were coming back online after a simple refresh 
- Zigbee reporting was not configured in the initialization of the device
- Added Capability Thermostat even if it's deprecated to support thermostat SmartApp.

greens (approved):

PR #25500, opened 2020-03-19
[CHAD-4704] Zigbee RGBW Bulb: Fix the color temperature range for innr bulbs

tpmanley:
These bulbs were using the default UI Metadata for the DTH which supports a color temperature range of 2700 to 6500 but these bulbs actually support a range of 1800 to 6500.

Note: this fix will only apply to newly joined devices that match one of the modified fingerprints. Existing devices will not automatically update to the new UI Metadata due to the way fingerprint-specific UI metadata is handled.

https://smartthings.atlassian.net/browse/CHAD-4704


dkirker (dismissed):

greens (dismissed):

tpmanley:
Sorry guys, can you review again? I retargeted this at `staging` so it can go to PROD next week since it's a simple change.

greens (approved):

dkirker (approved):
#

PR #26135, opened 2020-03-25
Revert "CHAD-3849 Update SmartWeather Station Tile to support the new mobile client"

gausnes:
Reverts SmartThingsCommunity/SmartThingsPublic#22345

This appears to be preventing DataMgmt integrationTests from passing. We should resolve those issues before merging this change.

```
16:45:40.695 ERROR {}  <> [ ] (main) CompilerService - key=sandbox_error, owner=null, id=null,
java.lang.RuntimeException: Metadata Error: Capability 'stsmartweather.Wind Speed' not found.
	at physicalgraph.device.api.DeviceTypeMetadataApi.error(DeviceTypeMetadataApi.groovy:41)
```


greens:
Probably just means the custom capability isn't published in dev

dkirker:
@greens Yup. Just fixed that. Worked with the right team to get that in since I can no longer do it. Pinged @gausnes to see if the issue he saw is cleared up.

PR #26397, opened 2020-03-27
fixes build error

greens:


devicetypes/smartthings/zwave-dimmer-switch-generic.src/zwave-dimmer-switch-generic.groovy
```diff
@@ -145,7 +145,7 @@ def parse(String description) {
 			result = zwaveEvent(cmd)
 		}
 	}
-	if (result?.name? == 'hail' && hubFirmwareLessThan("000.011.00602")) {
+	if (result?.name?.equals('hail') && hubFirmwareLessThan("000.011.00602")) {
```
dkirker, line 148: Does this work? A few of the `zwaveEvent` functions return a list.

greens: I believe that the `?` after result will mean it gets marked as false?

greens: yeah, just verified

dkirker: And `result?.name == 'hail'` is what is throwing the original exception? I would expect that to just result in `null == 'hail'` which would be false.

dkirker (approved):

PR #26654, opened 2020-03-31
[ICP-12702] Added clearTamper() after 10s

MGoralczykS:
@greens @KKlimczukS @PKacprowiczS @ZWozniakS @MWierzbinskaS @BJanuszS
Could you review this?

### DIFF
--- a/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
+++ b/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
@@ -198,7 +198,8 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 	if (cmd.notificationStatus == 0xFF) {
 		switch (cmd.event) {
 			case 0x09: //TAMPER
-				createEvent([name: "tamper", value: "detected"])
+				sendEvent(name: "tamper", value: "detected")
+				runIn(10, "clearTamper")
 				break
 			case 0x01: //ON
 				if (state.lastTriggeredSound == 1) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
```diff
@@ -198,7 +198,8 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 	if (cmd.notificationStatus == 0xFF) {
 		switch (cmd.event) {
 			case 0x09: //TAMPER
-				createEvent([name: "tamper", value: "detected"])
+				sendEvent(name: "tamper", value: "detected")
+				runIn(10, "clearTamper")
```
greens, line 202: check the `zwave-motion-sensor` dth. You'll need to put an `unschedule` of any old `clearTamper` calls here just in case tamper got triggered again.

MGoralczykS: @greens I read about `unschedule` and this turns off my scheduling. In this case, I want to call `runIn` every time tamper is detected. 
Maybe to avoid `unschelude` I should use `[overwrite: true]`?

dkirker: I believe `[overwrite: true]` is the default behavior.

cbaumler (commented):
LGTM

devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
greens: Donald is correct. I guess that most places spell it out explicitly for clarity's sake.

@MGoralczykS the way it's done on other DTHs is to unschedule the old call and add a new one, so the delay is always 10s from the last tamper, but since overwriting happens by default anyway this doesn't seem to be necessary.

greens (approved):

PR #27400, opened 2020-04-07
[ICP-12926, ICP-12927] Everspring Wall Switch: handled down_hold event and hid the buttons value for the Hisotry tab

ZWozniakS:
@greens @dkirker @tpmanley 
could you please review this? 

Before I just did not handle the "release" event because I thought we are not supporting it, but then I realized that we made a similar workaround for the Aeotec NanoOne (release is handled as held and the held as held down). So here I tried to do this in the same way.

### DIFF
--- a/devicetypes/smartthings/aeotec-wallmote.src/aeotec-wallmote.groovy
+++ b/devicetypes/smartthings/aeotec-wallmote.src/aeotec-wallmote.groovy
@@ -95,7 +95,7 @@ def zwaveEvent(physicalgraph.zwave.commands.centralscenev1.CentralSceneNotificat
 
     def child = getChildDevice(button)
     child?.sendEvent(name: "button", value: value, data: [buttonNumber: 1], descriptionText: "$child.displayName was $value", isStateChange: true)
-    createEvent(name: "button", value: value, data: [buttonNumber: button], descriptionText: "$device.displayName button $button was $value", isStateChange: true)
+    createEvent(name: "button", value: value, data: [buttonNumber: button], descriptionText: "$device.displayName button $button was $value", isStateChange: true, displayed: false)
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.securityv1.SecurityMessageEncapsulation cmd) {
@@ -169,7 +169,7 @@ def getChildDevice(button) {
 
 private getSupportedButtonValues() {
     if (isEverspring()) {
-        return ["pushed", "held", "double"]
+        return ["pushed", "held", "down_hold", "double"]
     } else {
         return ["pushed", "held"]
     }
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeotec-wallmote.src/aeotec-wallmote.groovy
```diff
@@ -178,7 +178,8 @@ private getSupportedButtonValues() {
 private getButtonAttributesMap() {
     if (isEverspring()) {[
             0: "pushed",
-            2: "held",
+            1: "held",
+            2: "down_hold",
```
greens: why are you changing what event 2 maps to?

ZWozniakS: Before I handled only 02 scene number and mapped that as "held" and skipped the 01 scene number which in fact is for the "release" action. When holding the button the first event which is coming is with 02 scene number, then the 01 is fired. So, the 02 scene number is executing at first while holding the button and then the 01 - because of that I treated the 02 as "held down" and 01 as "held" - the same as it is applied for the Aeotec Nano One button.

dkirker: "down_hold" is more like an up-down toggle switch that is being held down.

So, on this device, if I press down and release the button immediately, I get a scene 0? And then if I press down and hold, I get a scene 2 and a scene 1? Do I get a scene 1 after the scene 0 if I just push and release? Is there ever an instance where I get a scene 1 but never any of the others? If I push and hold, do I get a scene 0, then after some time a scene 2, and then eventually a scene 1?

What I would do to fix ICP-12927 is to have a check in the code that uses `getButtonAttributesMap` to not generate an event if `value == null`. It seems like that is the appropriate solution.

ZWozniakS: Thank you, Donald. Your suggestion sounds much better. 

And yes, if you press once (short press) you get the scene 0, press down+hold you get scene 2 and 1. You don't get scene 1 after scene 0. You never get scene 1 alone, it is always correlated with scene 2 and appears only after scene 2. So the only case when you get scene 1 is while holding the button. If you short press and then press down and hold you'll get in sequence: scene 0 (for the short press), then scene 2 and 1 right after (for the holding). 

And I verified that the ICP-12926 is resolved with this fix, so half of the PR was still good ;) Now I can say the same about ICP-12927.

dkirker (approved):

greens (approved):

ghost (commented):
Duplicate of #

PR #28828, opened 2020-04-21
[ICP-13017] Settings change for Stelpro Ki Zigbee Thermostat

PKacprowiczS:
This is also for ICP-13019

### DIFF
--- a/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
@@ -53,11 +53,11 @@ metadata {
 	preferences {
 		section {
 			input("lock", "enum", title: "Do you want to lock your thermostat's physical keypad?", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: false)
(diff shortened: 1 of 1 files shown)


greens (approved):

dkirker:
I left a comment on the ticket. Another thing to note: The Stelpro integrations all generally behave the same way. I would highly recommend that whatever changes are made to one are made to all for consistency.

While the updating of the outdoor preference looks good, I recommend leaving the "display detailed operating state updates" preference if it behaves the same in the new app as it does in Classic and updating the text for that preference to something like "Do you want to see detailed operating state events in the event history? There may be many.".

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -309,8 +309,7 @@ def parse(String description) {
 
 				// If the user does not want to see the Idle and Heating events in the event history,
 				// don't show them. Otherwise, don't show them more frequently than 30 seconds.
```
dkirker, line 311: If we chose to remove this preference, this comment will need to be updated to reflect that. If we only change the preference text, then this comment can stay as-is.

dkirker (changes requested):

PR #28900, opened 2020-04-21
CHAD-4651 Move Inovelli 2-Channel Smart Plug to new MCD experience

dkirker:
This removes the fingerprints from the old integration and moves them to a new DTH that has `mcdSync: true` added and UI metadata proper for a modern MCD experience.

### DIFF
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
@@ -1,6 +1,6 @@
 /**
  *
- *  Inovelli Smart Plug 2-Channel
+ *  Inovelli 2-Channel Smart Plug MCD
  *
  *  Copyright 2020 SmartThings
  *
@@ -22,7 +22,7 @@
  *
  */
 metadata {
-	definition(name: "Inovelli Smart Plug 2-Channel", namespace: "erocm123", author: "Eric Maycock", ocfDeviceType: "oic.d.smartplug", mcdSync: true) {
+	definition(name: "Inovelli 2-Channel Smart Plug MCD", namespace: "erocm123", author: "Eric Maycock", ocfDeviceType: "oic.d.smartplug", mcdSync: true) {
 		capability "Actuator"
 		capability "Sensor"
 		capability "Switch"
--- a/devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
+++ b/devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
@@ -30,14 +30,14 @@ metadata {
 		capability "Refresh"
 		capability "Health Check"
 
-		fingerprint manufacturer: "015D", prod: "0221", model: "251C", deviceJoinName: "Show Home Outlet" //Show Home 2-Channel Smart Plug
-		fingerprint manufacturer: "0312", prod: "0221", model: "251C", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Smart Plug
-		fingerprint manufacturer: "0312", prod: "B221", model: "251C", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Smart Plug
-		fingerprint manufacturer: "0312", prod: "0221", model: "611C", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Outdoor Smart Plug
-		fingerprint manufacturer: "015D", prod: "0221", model: "611C", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Outdoor Smart Plug
-		fingerprint manufacturer: "015D", prod: "6100", model: "6100", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Outdoor Smart Plug
-		fingerprint manufacturer: "0312", prod: "6100", model: "6100", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Outdoor Smart Plug
-		fingerprint manufacturer: "015D", prod: "2500", model: "2500", deviceJoinName: "Inovelli Outlet" //Inovelli 2-Channel Smart Plug w/Scene
+		fingerprint manufacturer: "015D", prod: "0221", model: "251C", deviceJoinName: "Show Home Outlet" // Show Home 2-Channel Smart Plug
+		fingerprint manufacturer: "0312", prod: "0221", model: "251C", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Smart Plug
+		fingerprint manufacturer: "0312", prod: "B221", model: "251C", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Smart Plug
+		fingerprint manufacturer: "0312", prod: "0221", model: "611C", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Outdoor Smart Plug
+		fingerprint manufacturer: "015D", prod: "0221", model: "611C", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Outdoor Smart Plug
+		fingerprint manufacturer: "015D", prod: "6100", model: "6100", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Outdoor Smart Plug
+		fingerprint manufacturer: "0312", prod: "6100", model: "6100", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Outdoor Smart Plug
+		fingerprint manufacturer: "015D", prod: "2500", model: "2500", deviceJoinName: "Inovelli Outlet" // Inovelli 2-Channel Smart Plug w/Scene
 	}
 	simulator {}
 	preferences {}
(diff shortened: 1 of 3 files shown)


tpmanley:
I think the DTH name will be confusing since it's just a different order than the old name. How about adding `-mcd` to the current name?

dkirker:
> I think the DTH name will be confusing since it's just a different order than the old name. How about adding `-mcd` to the current name?

Done!

dkirker:
I should note, there are a couple of bugs in the old DTH that I did *not* fix. I can add that in follow up commits (basically, the `logging` command references a variable that does not exist, and there is a manufacturerGet command in `installed()` that does not actually do anything because it isn't the last statement).

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,291 @@
...
+	}, delay)
+}
+private channelNumber(String dni) {
+	dni.split("-ep")[-1] as Integer
+}
+private void createChildDevices() {
+	state.oldLabel = device.label
+	for (i in 1..2) {
+		addChildDevice("Switch Child Device",
+				"${device.deviceNetworkId}-ep${i}",
```
tpmanley: Let's switch to our newer style of naming that is `<DNI>:<channel/endpoint>`

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,291 @@
...
+		command(it)
+	}, delay)
+}
+private channelNumber(String dni) {
+	dni.split("-ep")[-1] as Integer
+}
+private void createChildDevices() {
+	state.oldLabel = device.label
+	for (i in 1..2) {
+		addChildDevice("Switch Child Device",
```
tpmanley: Thoughts on switching to `Child Switch`?

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,291 @@
...
+				"${device.deviceNetworkId}-ep${i}",
+				device.hubId,
+				[completedSetup: true,
+				 label: "${device.displayName} (CH${i})",
+				 isComponent: true,
+				 componentName: "ep$i",
```
tpmanley: These component names show up in the device details view in the app. I suggest changing the name to `outlet$i` which is more user friendly and matches the convention we use in other MCD DTHs (ex: button1, switch1, etc..)

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+}
+
+private channelNumber(String dni) {
+	dni.split(":outlet")[-1] as Integer
+}
+
+private void createChildDevices() {
+	state.oldLabel = device.label
+	for (i in 1..2) {
+		def newDevice = addChildDevice("smartthings", "Child Switch Health",
```
tpmanley, line 291: Is it necessary to use the "Health" device? The component is not a standalone device so it should not need the Health Check capability.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+				"${device.deviceNetworkId}:outlet${i}",
+				device.hubId,
+				[completedSetup: true,
+				 label: "${device.displayName} (CH${i})",
+				 isComponent: true,
+				 componentName: "outlet$i",
+				 componentLabel: "Outlet $i"
+		])
+
+		newDevice.sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
tpmanley: Should use `checkInterval` instead of hardcoding the value

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+}
+
+def updated() {
+	log.debug "updated()"
+
+	if (!childDevices) {
+		createChildDevices()
+	} else if (device.label != state.oldLabel) {
+		childDevices.each {
+			if (it.label == "${state.oldLabel} (CH${channelNumber(it.deviceNetworkId)})") {
```
tpmanley: Need to update the label handling here

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+				def newLabel = "${device.displayName} (CH${channelNumber(it.deviceNetworkId)})"
+				it.setLabel(newLabel)
+			}
+		}
+		state.oldLabel = device.label
+	}
+
+	sendEvent(name: "checkInterval", value: checkInterval, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+	sendEvent(name: "needUpdate", value: device.currentValue("needUpdate"), displayed: false, isStateChange: true)
```
tpmanley: Doesn't look like this is used so I think it can be removed

tpmanley (dismissed):
Looks good once the last couple of things are cleaned up

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+
+private channelNumber(String dni) {
+	dni.split(":outlet")[-1] as Integer
+}
+
+private void createChildDevices() {
+	state.oldLabel = device.label
+	for (i in 1..2) {
+		def newDevice = addChildDevice("smartthings", "Child Switch Health",
+				"${device.deviceNetworkId}:outlet${i}",
```
tpmanley: I missed this before but it should be just `"${device.deviceNetworkId}:${i}"`

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
tpmanley: Turns out this is necessary so I'l resolve this comment

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
greens: I think we should just stick with what has been our convention of `parentID:endpointNumber`.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+				event = [createEvent([name: "switch", value: "on"])]
+			}
+		}
+		return event
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicSet cmd) {
+	log.debug "BasicSet ${cmd}"
+	def result = createEvent(name: "switch", value: cmd.value ? "on" : "off", type: "digital")
```
greens: I don't think we're doing the `digital`/`analog` event types anymore

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+		zwaveEvent(encapsulatedCommand, cmd.sourceEndPoint as Integer)
+	}
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.manufacturerspecificv2.ManufacturerSpecificReport cmd) {
+	log.debug "ManufacturerSpecificReport ${cmd}"
+	def msr = String.format("%04X-%04X-%04X", cmd.manufacturerId, cmd.productTypeId, cmd.productId)
+	log.debug "msr: $msr"
+	updateDataValue("MSR", msr)
+}
```
greens: we don't do this/need this anymore

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,309 @@
...
+
+def getCheckInterval() {
+	2 * 15 * 60 + 2 * 60
+}
+
+def installed() {
+	log.debug "installed()"
+
+	sendEvent(name: "checkInterval", value: checkInterval, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
greens: this shouldn't be offlinepingable

dkirker: I removed offlinePingable from the children, but given the experience that @tpmanley had with the device not necessarily reporting physical events, wouldn't we want it as offlinePingable (since the device doesn't report metering).

That said, it existed in the original DTH (and in a few of the Z-Wave switches and dimmers), so I assume it was there for a reason.

dkirker:
The original request was to get this DTH to work as a modern MCD device. While there are a few changes with code formatting and re-usability that I'd like to address I'm hesitant to without scheduling time to try to go to the office. Either way, that would push this back a deploy cycle (unless we got approval and merged to staging).

@tpmanley Could you verify the latest updates to make sure that I didn't miss anything? I tested to make sure that it compiles, which it does.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,332 @@
...
+		command(it)
+	}, delay)
+}
+
+private channelNumber(String dni) {
+	dni.split(":")[-1] as Integer
+}
+
+private void createChildDevices() {
+	state.oldLabel = device.label
```
tpmanley, line 288: I think this needs to be saving `device.displayName` instead of `device.label`. The `label` isn't set when the device is created so `displayName` returns `device.name` instead. As written right now updating the label on the device doesn't update the label on the children.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,332 @@
...
+
+	for (i in 1..2) {
+		def newDevice = addChildDevice("smartthings", "Child Switch Health",
+				"${device.deviceNetworkId}:${i}",
+				device.hubId,
+				[completedSetup: true,
+				 label: "${device.displayName} (CH${i})",
```
tpmanley: I'd go with `${device.displayName} Outlet ${i}` instead which is more user friendly

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,332 @@
...
+		childDevices.each {
+			if (it.label == "${state.oldLabel} (CH${channelNumber(it.deviceNetworkId)})") {
+				def newLabel = "${device.displayName} (CH${channelNumber(it.deviceNetworkId)})"
+				it.setLabel(newLabel)
+			}
+		}
+		state.oldLabel = device.label
+	}
+
+	sendEvent(name: "checkInterval", value: checkInterval, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
tpmanley: This event is already created in `installed`. Do we need it here too?

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
dkirker: Alright, so I was confusing how `offlinePingable` is used today with how it was originally used... Since this and other switch devices were using the old definition, I've removed it.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
dkirker: Turns out that it was the if statement in the closure for updating each child device. I removed it and the code appears to work now.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -272,10 +272,7 @@ def updated() {
 		createChildDevices()
 	} else if (device.label != state.oldLabel) {
 		childDevices.each {
-			if (it.label == "${state.oldLabel} (CH${channelNumber(it.deviceNetworkId)})") {
```
tpmanley: I don't think we want to remove this check. The reason it's there is to update the label if we set it but leave it alone if the user has changed it.

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,305 @@
...
+		}
+	childDevice?.sendEvent(name: "switch", value: cmd.value ? "on" : "off")
+
+	if (cmd.value) {
+		event = [createEvent([name: "switch", value: "on"])]
+	} else {
+		def allOff = true
+
+		childDevices.each { n ->
+				if (n.currentState("switch")?.value != "off")
```
tpmanley: nit: indentation

tpmanley (dismissed):
Just one final nit and squash the commits but otherwise is working well for me. 

@kianooshST or @cbaumler do you want to take a look?

devicetypes/erocm123/inovelli-2-channel-smart-plug-mcd.src/inovelli-2-channel-smart-plug-mcd.groovy
```diff
@@ -0,0 +1,305 @@
...
+
+	return event
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd, ep = null) {
+	log.debug "BasicReport ${cmd} - outlet ${ep}"
+
+	if (ep) {
+		return handleSwitchEndpointEvent(cmd, ep)
+	}
```
kianooshkarami, line 108: Assuming this device never sends Basic Report from the root device?

kianooshkarami: I ask because I generally would recommend to have the same handler for incoming Basic Set and Report

dkirker: @kianooshST That's my assumption. Since I don't have a device on hand and this is the logic from the original code I left it as part of my consolidation. We can always update it later.

kianooshkarami (commented):
This looks good. I do wonder if we can move everything in this DTH to a generic Z-Wave Multi Switch... there is nothing special about this device that wouldn’t be able to leverage a generic DTH

tpmanley:
@kianooshST the reason we decided not to do that right now is not all multi-switch/outlet devices have the native ability to control the endpoints individually or both at once. Thus we felt it was best to keep this a manufacturer specific DTH for now. It's relatively easy to migrate devices from a manufacturer specific DTH to a generic DTH in the future, but a bit harder to go the other direction.

tpmanley (approved):

PR #28936, opened 2020-04-22
DevWs for Neonlite Distribution Ltd containing containing Zigbee Multi Button

Duncankn:


shrakuma:
@Duncankn i see that there is another PR that is closed and linked to here. Seems that this PR is ading some FPs and some other lises are adjusted. 
@greens @KKlimczukS @PKacprowiczS  can you please take a look and provide your feedback please

greens:
Since we've done a renaming pass on some of our device join names this will need to be rebased. Also this DTH is a bit atypical. Most devices don't implement buttons in the same way, so there's a lot of device-specific code here that would probably need to be expanded to add another remote.

Duncankn:
Hi @greens 

Adurosmart is our partner company, she have already launched this button. As our 4-key button is same as their button, except the fingerprint, so I can summit like this.

greens:
@Duncankn ok, but you still need to rebase as your PR reverts our recent join name changes to all the other devices on this DTH.

shrakuma:
@Duncankn can you please rebase and let us know . thanks

KKlimczukS:
@greens
Shouldn't this PR be closed since there were no changes since May?
cc: @Duncankn

PR #29024, opened 2020-04-23
[ICP-12852] Added capability 'Temperature Alarm'

ZWozniakS:
@greens @dkirker @tpmanley 
Could you please review? I've already changed the metadata as well.

### DIFF
--- a/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
+++ b/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
@@ -134,6 +134,13 @@ def updated() {
 	if (settings.temperatureHigh != null || settings.temperatureLow != null) {
         sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false) 
 	}
+
+	def tempAlarmMap = ["clear": "cleared", "overheat": "heat", "underheat": "freeze"]
+	// Convert old device specific temperatureAlarm event to standard Temperature Alarm capability event
+	def temperatureAlarmCurrentValue = device.currentValue("temperatureAlarm")
+	if (tempAlarmMap.containsKey(temperatureAlarmCurrentValue)) {
+		sendEvent(name: "temperatureAlarm", value: tempAlarmMap[temperatureAlarmCurrentValue], displayed: true)
+	}
 	
 	syncStart()
 	state.lastUpdated = now()
--- a/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
+++ b/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
@@ -16,12 +16,12 @@ metadata {
 		capability "Contact Sensor"
 		capability "Tamper Alert"
 		capability "Temperature Measurement"
+		capability "Temperature Alarm"
 		capability "Configuration"
 		capability "Battery"
 		capability "Sensor"
 		capability "Health Check"
-	
-		attribute "temperatureAlarm", "string"
+
 		attribute "multiStatus", "string"
 
 		fingerprint mfr: "010F", prod: "0702", deviceJoinName: "Fibaro Open/Closed Sensor"
@@ -62,15 +62,16 @@ metadata {
 		
 		standardTile("temperatureAlarm", "device.temperatureAlarm", inactiveLabel: false, width: 2, height: 2, decoration: "flat") {
 			state "default", label: "No temp. alarm", backgroundColor:"#ffffff"
-			state "clear", label:'', backgroundColor:"#ffffff", icon: "st.alarm.temperature.normal"
-			state "underheat", label:'underheat', backgroundColor:"#1e9cbb", icon: "st.alarm.temperature.freeze"
-			state "overheat", label:'overheat', backgroundColor:"#d04e00", icon: "st.alarm.temperature.overheat"
+			state "cleared", label:'', backgroundColor:"#ffffff", icon: "st.alarm.temperature.normal"
+			state "freeze", label:'freeze', backgroundColor:"#1e9cbb", icon: "st.alarm.temperature.freeze"
+			state "heat", label:'heat', backgroundColor:"#d04e00", icon: "st.alarm.temperature.overheat"
 		}
 
 		main "FGDW"
 		details(["FGDW","tamper","temperature","battery","temperatureAlarm"])
 	}
-		
+
+
 	preferences {
 		input (
 			title: "Wake up interval",
@@ -117,6 +118,7 @@ def installed() {
 	// Initial states for OCF compatibility
 	sendEvent(name: "tamper", value: "clear", displayed: false)
 	sendEvent(name: "contact", value: "open", displayed: false)
+	sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
 	sendEvent(name: "checkInterval", value: 21600 * 4 + 120, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
 }
 
@@ -132,7 +134,7 @@ def updated() {
 	if ( settings.temperatureHigh as Integer == 0 && settings.temperatureLow as Integer == 0 ) { 
 		sendEvent(name: "temperatureAlarm", value: null, displayed: false) 
 	} else if ( settings.temperatureHigh != null || settings.temperatureHigh != null ) {
-		sendEvent(name: "temperatureAlarm", value: "clear", displayed: false) 
+		sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
 	}
 	
 	syncStart()
@@ -308,15 +310,15 @@ def zwaveEvent(physicalgraph.zwave.commands.alarmv2.AlarmReport cmd) {
 				map.name = "temperatureAlarm"
 				switch (cmd.zwaveAlarmEvent) {
 					case 0:
-						map.value = "clear"
+						map.value = "cleared"
 						map.descriptionText = "Temperature alert cleared"
 						break
 					case 2:
-						map.value = "overheat"
+						map.value = "heat"
 						map.descriptionText = "Temperature alert: overheating detected"
 						break
 					case 6:
-						map.value = "underheat"
+						map.value = "freeze"
 						map.descriptionText = "Temperature alert: underheating detected"
 						break
 				}
--- a/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
+++ b/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
@@ -131,10 +131,8 @@ def updated() {
 	}
 	logging("${device.displayName} - Executing updated()","debug")
 
-	if ( settings.temperatureHigh as Integer == 0 && settings.temperatureLow as Integer == 0 ) { 
-		sendEvent(name: "temperatureAlarm", value: null, displayed: false) 
-	} else if ( settings.temperatureHigh != null || settings.temperatureHigh != null ) {
-		sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
+	if (settings.temperatureHigh != null || settings.temperatureLow != null) {
+        sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false) 
 	}
 	
 	syncStart()
--- a/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
+++ b/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
@@ -139,10 +139,16 @@ def updated() {
 	}
 
 	def currentTemperature = device.currentValue("temperature")
-	def temperatureHigh = settings.temperatureHigh ? new BigDecimal(settings.temperatureHigh) : null
-	def temperatureLow = settings.temperatureLow ? new BigDecimal(settings.temperatureLow) : null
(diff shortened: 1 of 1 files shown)


greens:
The allowed values for the temperatureAlarm attribute are cleared, freeze, heat, and rateOfRise. This DTH uses non-standard values that will need to be adjusted.

Additionally you'd want to remove the custom attribute.

ZWozniakS:
@greens I already adjusted values for the temperatureAlarm, sorry that it took so long but I was waiting for the device. Here is MR for the metadata: [https://codesamsung.com/iot/ocf/SmartThingsResourceTypes/merge_requests/1109]. I left three values (cleared, heat, freeze) as this device is not supporting rateOfRise.

greens:
Please remove the custom attribute `"temperatureAlarm"` and also update the values in the tiles.

dkirker (commented):
We'll need to update the `temperatureAlarm` event for existing devices. Normally, I would suggest adding some code to `parse()` that does this, but that is usually when migrating from an attribute of one name to another.

@tpmanley Do you think it would just be easier to run a script that does this once instead of adding code to `parse()` that checks (if temperatureAlarm is clear or temperatureAlarm is underheat or temperatureAlarm is overheat) every time `parse()` gets called? I think this will be a better option, so I guess my comment here is that we should do that.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -132,7 +134,7 @@ def updated() {
 	if ( settings.temperatureHigh as Integer == 0 && settings.temperatureLow as Integer == 0 ) { 
 		sendEvent(name: "temperatureAlarm", value: null, displayed: false)
```
greens: I'm concerned about this event. I wouldn't expect that we can just send a null value under any circumstances.

greens:
@dkirker I think maybe adding some code to `updated` that checks for the `"temperatureAlarm"` value?

Then we can just run `updated` after deploy.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
dkirker: If the goal is to disable temperature alarm reporting it might be best to just set it to "cleared". I don't think `null` is valid (though, it might work).

dkirker:
@greens That's probably a good option, too.

ZWozniakS:
This is a battery and a sleepy device so I am afraid that checking of an alarm status (response([zwave.alarmv2.AlarmGet(alarmType: 4).format()])) may be missed.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -129,10 +131,8 @@ def updated() {
 	}
 	logging("${device.displayName} - Executing updated()","debug")
 
-	if ( settings.temperatureHigh as Integer == 0 && settings.temperatureLow as Integer == 0 ) { 
-		sendEvent(name: "temperatureAlarm", value: null, displayed: false) 
-	} else if ( settings.temperatureHigh != null || settings.temperatureHigh != null ) {
-		sendEvent(name: "temperatureAlarm", value: "clear", displayed: false) 
+	if (settings.temperatureHigh != null || settings.temperatureLow != null) {
+        sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
```
greens: I think the other case should probably result in us sending a "cleared" as well, although I can't say I understand what the original intent was

greens:
@ZWozniakS I didn't mean actually polling the device. I meant checking for `currentState` of `"TemperatureAlarm"` and synthesizing the capability-correct event from that.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
ZWozniakS: It will anyway send "cleared" event now so I wanted to optimize that with only one line.

greens: I don't understand. How are we still sending a cleared event if `temperatureHigh` and `temperatureLow` are both `0` now?

KKlimczukS: @greens if the intention is to sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false) in both cases then this condition should be enough
`if (settings.temperatureHigh != null || settings.temperatureLow != null) {
        sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
}` 

When temperatureHigh && temperatureLow will be both set to 0 the cleared event will be sent.

greens: oh, you're totally right.

duh

greens:
we still need to add something in `updated` so that previous installs can be brought in line

ZWozniakS:
I am not sure if we need to check what is the tempAlarm value here. The purpose of sending `cleared` in the `updated ` is to clear the tempAlarm value in order to ongoing changes of the tempHigh and tempLow settings. The whole method is in fact dedicated to the settings/parameters changes. The reason of that is that the range of the normal/overheat/underheat (cleared, heat, freeze values) states may differ after the change of the boundary ranges (tempHigh/tempLow) and we will need an actualization of the tempAlarm value but right after.

dkirker:
How does the device behave when you change either the max or the min? (The setting actually goes up to the device and changes its internal logic. So the question is, does it respond to this immediately?) For instance, say it is 10C out and the current min is set to 5C (so the current state is that there is no alarm, "cleared"), if I set the new min to 15C does the device report back a freeze alarm? Does it report immediately, or at the next temperature reading? Since the actual setting results in a config update to the device, I would expect that the device itself should handle reporting to us based on the new parameters. We shouldn't need to do anything to force the event.

What will be needed in `update()` is logic to convert the old "clear", "underheat", and "overheat" to "cleared", "freeze", and "heat" respectively. This way when this is deployed we can run a script to call `update()` on all devices so that they display the correct values. You could do something like:

```
if (["clear", "overheat", "underheat"].contains(device.currentValue("temperatureAlarm"))) {
// ask the device for its alarm state
}
```
(I'm not sure that will actually work.)

or

```
def tempAlarmMap = ["clear": "cleared", "overheat": "heat", "underheat": "freeze"]

// Convert old device specific temperatureAlarm event to standard Temperature Alarm capability event
if (tempAlarmMap.containsKey(device.currentValue("temperatureAlarm"))) {
sendEvent(name: "temperatureAlarm", value: tempAlarmMap[device.currentValue("temperatureAlarm")], displayed: false)
}
```
(Or you can set `displayed: true`, but that probably doesn't matter.)

ZWozniakS:
@dkirker thank you for the explanation, I'll check that and be back with answers soon - but in the meanwhile - what about updating the metadata? How does fetch work once the user has already installed the device?

ZWozniakS:
@dkirker I checked suggested solution and the conversion worked so I pushed the changes. 
When changing the max or min temp settings the update of the current state is usually working immediately and if not - it is happening at the next temperature reading as you assumed.

greens:
@dkirker can you take another look?

dkirker:
@ZWozniakS I think we'd want to move the event conversion before the logic that clobbers the alarm state if either the min or max are set. Otherwise you risk setting the alarm to cleared and then updating it to a different value.

If the device responds to the user updating the min or the max I still don't think we should manually clear the alarm. If you have the alarm threshold set to one value and have the alarm currently triggered (let's say you set the min threshold to 0C and it is currently -2C and temperature alarm is freeze), if you update the threshold to a value which would clear the alarm, does the alarm clear without you having to force it? (So let's say in this example you set the new min threshold to -4C such that the -2C would no longer be a freeze alarm.)

KKlimczukS:
> @ZWozniakS I think we'd want to move the event conversion before the logic that clobbers the alarm state if either the min or max are set. Otherwise you risk setting the alarm to cleared and then updating it to a different value.
> 
> If the device responds to the user updating the min or the max I still don't think we should manually clear the alarm. If you have the alarm threshold set to one value and have the alarm currently triggered (let's say you set the min threshold to 0C and it is currently -2C and temperature alarm is freeze), if you update the threshold to a value which would clear the alarm, does the alarm clear without you having to force it? (So let's say in this example you set the new min threshold to -4C such that the -2C would no longer be a freeze alarm.)

@dkirker @greens 
Would it make sense if clearing temperatureAlarm was left, but with additional conditions ?
For example:
```
def currentTemperature = device.currentValue("temperature")

if ((settings.temperatureHigh != null && (currentTemperature < settings.temperatureHigh))   || 
(settings.temperatureLow != null && (currentTemperature > settings.temperatureLow ) )) {
        sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false) 
}
```

greens:
I agree with @KKlimczukS on this one. I don't think we can rely on the device being clever enough to clear its alarms on its own if the user changes the min/max setting. So I think the order in `updated` should be:

1. If the user has updated the min/max settings, send a clear if necessary.
2. Get `currentState` again and convert if necessary.

ZWozniakS:
@dkirker to answer your question whether the alarm will clear without forcing - yes. But after long work with this device, I see that sometimes it has problems updating alarms, so it's worth leaving forcing.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -129,12 +131,20 @@ def updated() {
...
+	if (tempAlarmMap.containsKey(temperatureAlarmCurrentValue)) {
+		sendEvent(name: "temperatureAlarm", value: tempAlarmMap[temperatureAlarmCurrentValue], displayed: true)
 	}
-	
+
+	def currentTemperature = device.currentValue("temperature")
+	def temperatureHigh = settings.temperatureHigh ? new BigDecimal(settings.temperatureHigh) : null
+	def temperatureLow = settings.temperatureLow ? new BigDecimal(settings.temperatureLow) : null
+	if ((temperatureHigh != null && (currentTemperature < temperatureHigh)) || (temperatureLow != null && (currentTemperature > temperatureLow) )) {
+		sendEvent(name: "temperatureAlarm", value: "cleared", displayed: false)
```
greens: ideally this would only be sent if the temperature alarm isn't already clear

ZWozniakS: right, fix is ready

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -129,12 +131,26 @@ def updated() {
...
+		sendEvent(name: "temperatureAlarm", value: tempAlarmMap[temperatureAlarmCurrentValue], displayed: true)
 	}
-	
+
+	def currentTemperature = device.currentValue("temperature")
+	def alarmCleared = device.currentValue("temperatureAlarm") == "cleared"
+	def alarmFreeze = device.currentValue("temperatureAlarm") == "freeze"
+	def alarmHeat = device.currentValue("temperatureAlarm") == "heat"
+	def temperatureHigh = (settings.temperatureHigh ? new BigDecimal(settings.temperatureHigh) : null) * 0.1
+	def temperatureLow = (settings.temperatureLow ? new BigDecimal(settings.temperatureLow) : null) * 0.1
```
greens: won't you get an error here multiplying null by 0.1?

I think you should be using a default value here instead.

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -129,12 +131,26 @@ def updated() {
...
+	if (!alarmCleared) {
+		if ((temperatureHigh != null && (currentTemperature < temperatureHigh) && !alarmFreeze) ||
+			(temperatureLow != null && (currentTemperature > temperatureLow) && !alarmHeat)) {
```
greens, line 149: another reason to be using a default value is that this should be an AND condition, right?

If temp is below max temp & temp is above min temp, send clear

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
ZWozniakS: We need to multiply these values due to the correct comparison _currentTemperature < temperatureHigh  /  currentTemperature > temperatureLow_. This is needed because values ​​multiplied by 10 are sent to the device (because settings are defined in this way and the device accepts and reports this way as well)

devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
ZWozniakS: These thresholds can be changed independently so it is better to check them separately as well.

greens: right, but if the temperature is below the max temperature this or statement will short-circuit and send a clear, even if it's also below the minimum temperature, right?

ZWozniakS: no, it won't, because if the temperature is below max temperature and at the same time below a minimum temperature, forcing cleared event will be blocked by the AND !alarmFreeze

greens (approved):

PR #29660, opened 2020-04-28
ICP-12952 - removed the cause of duplicated events in the parent dth

KKlimczukS:
ICP-12952 - removed the cause of duplicated events in the parent dth.

Right before sending a button event to a child device -> a redundant event in the parent Dth was created. That was a cause of duplicated events in the activity history.

@greens @dkirker 

@MWierzbinskaS @ZWozniakS @PKacprowiczS @BJanuszS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
+++ b/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
@@ -296,6 +296,10 @@ private Map getButtonEvent(Map descMap) {
 	}
 
 	if (buttonNumber != 0) {
+		// Create old style
+		def descriptionText = "${getButtonName(buttonNumber)} was $buttonState"
+		result = [name: "button", value: buttonState, data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true, displayed: false]
+
 		// Create and send component event
 		sendButtonEvent(buttonNumber, buttonState)
 	}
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -296,10 +296,6 @@ private Map getButtonEvent(Map descMap) {
 	}
 
 	if (buttonNumber != 0) {
-		// Create old style
-		def descriptionText = "${getButtonName(buttonNumber)} was $buttonState"
-		result = [name: "button", value: buttonState, data: [buttonNumber: buttonNumber], descriptionText: descriptionText, isStateChange: true]
```
greens: @tpmanley and @dkirker is this the right direction to go in for mcd now?

I'd favor keeping the "extra" event and simply marking the main event as `displayed: false`, but I can't really think of a compelling reason to keep it around. I guess doing it this way does make the main device on the device list change when you press any button, which seems good, right?

tpmanley: I think we still need the event to ensure that the device works correctly in SmartApps like Smart Lighting. If the user selects the parent device followed by the button number it expects to be created on the parent device and will not see the events creates on the child components.

greens:
@KKlimczukS can you keep both events but change the event for the main component to `displayed: false`?

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
KKlimczukS: @greens @tpmanley i reverted previous changes and added `displayed: false` to a parent event according to Your suggestions.

greens (approved):

PR #29913, opened 2020-04-30
New DTH for our SP2600ZB and SP2610ZB update

SmartThings:
Replacing the pull request 25498 with this one as asked by @PKacprowiczS


PKacprowiczS:
@Sinopetech thank you for this change. One more thing - do you think you could change indentations from spaces to tabs?

Sinopetech:
> @Sinopetech thank you for this change. One more thing - do you think you could change indentations from spaces to tabs?

@PKacprowiczS you may find the update in the following request: #30345

PR #29999, opened 2020-05-02
PR - Z-Wave Window Shade

habhomegit:
PR for Z-Wave Window Shade.  Adding reverse function, Switch and Switchlevel capabilities.


dkirker:
I'm pushing up some of the code that I have. It isn't quite usable yet, but best to have it backed up. :)

PR #30151, opened 2020-05-04
CHAD-4340 Update Window Shade devices to use Window Shade Level

dkirker:
`Window Shade Level` is the preferred capability for Window Shade devices expressing a level value.

The first step is to add this capability to the devices which use it. The next step is to look in to considering removing the `Switch Level` capability. For now we will leave it in the DTH, but remove it from the metadata.

Locally executing integrations will need to be considered, too.

PR #30258, opened 2020-05-06
[WWST-6516] New DTHes to cover Fibaro Walli Roller Shutter functionality

PKacprowiczS:
@greens @dkirker @KKlimczukS @MGoralczykS @ZWozniakS @MWierzbinskaS @BJanuszS 
This device needs 3 separate DTHes to maintain as much of its functionality as possible.
Looking forward to your insights about the code
Metadata which those use were firstly created for Fibaro Roller Shutter 3, so I've decided to just keep those vids unchanged, as they will be used soon anyway. You can look up them in MR 1114

### DIFF
--- a/devicetypes/fibargroup/fibaro-walli-roller-shutter-driver.src/fibaro-walli-roller-shutter-driver.groovy
+++ b/devicetypes/fibargroup/fibaro-walli-roller-shutter-driver.src/fibaro-walli-roller-shutter-driver.groovy
@@ -18,12 +18,14 @@ import java.lang.Math
 metadata {
 	definition (name: "Fibaro Walli Roller Shutter Driver", namespace: "fibargroup", author: "SmartThings", ocfDeviceType: "oic.d.blind", mnmn: "SmartThings", vid: "SmartThings-smartthings-Fibaro_Roller_Shutter") {
 		capability "Window Shade"
-		capability "Switch Level"
+		capability "Window Shade Level"
 		capability "Power Meter"
 		capability "Energy Meter"
 		capability "Refresh"
 		capability "Health Check"
 		capability "Configuration"
+
+		capability "Switch Level"
 	}
 
 	tiles(scale: 2) {
@@ -54,70 +56,47 @@ metadata {
 	}
 
 	preferences {
+		// Preferences template begin
 		parameterMap.each {
-			input (
-					title: it.name,
-					description: it.description,
-					type: "paragraph",
-					element: "paragraph"
-			)
+			input (title: it.name, description: it.description, type: "paragraph", element: "paragraph")
 
 			switch(it.type) {
 				case "boolRange":
 					input(
-							name: it.key + "Boolean",
-							type: "bool",
-							title: "Enable",
+							name: it.key + "Boolean", type: "bool", title: "Enable",
 							description: "If you disable this option, it will overwrite setting below.",
-							defaultValue: it.defaultValue != it.disableValue,
-							required: false
+							defaultValue: it.defaultValue != it.disableValue, required: false
 					)
 					input(
-							name: it.key,
-							type: "number",
-							title: "Set value (range ${it.range})",
-							defaultValue: it.defaultValue,
-							range: it.range,
-							required: false
+							name: it.key, type: "number", title: "Set value (range ${it.range})",
+							defaultValue: it.defaultValue, range: it.range, required: false
 					)
 					break
 				case "boolean":
 					input(
-							type: "paragraph",
-							element: "paragraph",
-							description: "Option enabled: ${it.activeDescription}\n" +
-									"Option disablePending: ${it.inactiveDescription}"
+							type: "paragraph", element: "paragraph",
+							description: "Option enabled: ${it.activeDescription}\n Option disabled: ${it.inactiveDescription}"
 					)
 					input(
-							name: it.key,
-							type: "boolean",
-							title: "Enable",
-							defaultValue: it.defaultValue == it.activeOption,
-							required: false
+							name: it.key, type: "boolean",
+							title: "Enable", defaultValue: it.defaultValue == it.activeOption, required: false
 					)
 					break
 				case "enum":
 					input(
-							name: it.key,
-							title: "Select",
-							type: "enum",
-							options: it.values,
-							defaultValue: it.defaultValue,
-							required: false
+							name: it.key, title: "Select", type: "enum",
+							options: it.values, defaultValue: it.defaultValue, required: false
 					)
 					break
 				case "range":
 					input(
-							name: it.key,
-							type: "number",
-							title: "Set value (range ${it.range})",
-							defaultValue: it.defaultValue,
-							range: it.range,
-							required: false
+							name: it.key, type: "number", title: "Set value (range ${it.range})",
+							defaultValue: it.defaultValue, range: it.range, required: false
 					)
 					break
 			}
 		}
+		// Preferences template end
 	}
 }
 
@@ -272,7 +251,7 @@ def handleConfigurationChange(confgurationReport) {
 					log.info "Device is already configured as Fibaro Walli Roller Shutter Driver"
 					break
 			}
-			break-
+			break
 		default:
 			log.info "Parameter no. ${confgurationReport.parameterNumber} has no specific handler"
 			break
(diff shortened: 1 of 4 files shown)


devicetypes/fibargroup/fibaro-walli-roller-shutter-venetian.src/fibaro-walli-roller-shutter-venetian.groovy
```diff
@@ -0,0 +1,634 @@
...
+		}
+	}
+}
+
+private addVenetianBlind() {
+	if (!childDevices) {
+		try {
+			String childDni = "${device.deviceNetworkId}:1"
+			def componentLabel = "Fibaro Roller Shutter Venetian Blind"
+			def child = addChildDevice("Child Venetian Blind", childDni, device.getHub().getId(), [
```
greens, line 476: I don't see this DTH?

greens:
Can you briefly explain the differences here? I get that the venetian blind has the child device to control the slats, but what about the other two?

PKacprowiczS:
@greens only difference between Roller Shutter and Roller Shutter with Driver, is ability of the first one to automatically calibrate blinds position - parameters 156 and 157 are being set without user's attention, therefore they're not exposed in settings in app. In case in using handler for a built-in driver those parameters have to be set manually.

PKacprowiczS:
In general, I'm not entirely happy with having separate DTHes for non-driver / built-in driver, but due to vexatious lack of flexibility of preferences, I've decided to go this way.

cbaumler (dismissed):

devicetypes/fibargroup/fibaro-walli-roller-shutter-driver.src/fibaro-walli-roller-shutter-driver.groovy
```diff
@@ -0,0 +1,549 @@
...
+				name: "Time of up movement", key: "timeOfUpMovement", type: "range",
+				parameterNumber: 156, size: 4, defaultValue: 6000,
+				range: "1..90000", 
+				description: "This setting determines the time needed for roller blinds to reach the top. [100 = 1s]"
+		],
+		[
+				name: "Time of down movement", key: "timeOfDownMovement", type: "range",
+				parameterNumber: 157, size: 4, defaultValue: 6000,
+				range: "1..90000", 
+				description: "This setting determines time needed for roller blinds to reach the bottom. [100 = 1s]"
```
greens, line 484: so these are necessary to set if you want the device to run properly?

devicetypes/fibargroup/fibaro-walli-roller-shutter-venetian.src/fibaro-walli-roller-shutter-venetian.groovy
```diff
@@ -0,0 +1,636 @@
...
+					setDeviceType("Fibaro Walli Roller Shutter")
+					break
+				case 2: // "Venetian blind (with positioning)"
+					log.info "Device is already configured as Fibaro Roller Shutter Venetian"
+					break
+				case 5: // "Roller blind with built-in driver"
+				case 6: // "Roller blind with built-in driver (impulse)"
+					log.info "Changing device type to Fibaro Walli Roller Shutter Driver"
+					setDeviceType("Fibaro Walli Roller Shutter Driver")
+					break
```
greens, line 283: wouldn't you need to delete the child device in these cases?

greens:
I don't know if the power and energy reporting is worth including. This device doesn't have anything else plugged into it, right?

devicetypes/fibargroup/fibaro-walli-roller-shutter-venetian.src/fibaro-walli-roller-shutter-venetian.groovy
PKacprowiczS: You're right, but I've never saw it been done in any DTH, so I don't know how (even thought that such possibility wasn't provided)

devicetypes/fibargroup/fibaro-walli-roller-shutter-driver.src/fibaro-walli-roller-shutter-driver.groovy
PKacprowiczS: Yes. Without those, user is left with default 1 min up and 1 min down times with no possibility to change.

PKacprowiczS:
@greens right, it has connected only blinds motor. Although, I'd left it like it is. Maintaining those reports does not require much attention.
On top of that, I'm already able to imagine getting pushed back by cert team with the fact that energy/power reports are clearly stated in the manual, so we should include them.

PKacprowiczS:
One more thing - I've just found out that child's DTH must has same namespace, so there's a question - should I switch venetian's namespace to _fibargroup_ or bring back those 3 DTHes to _smartthings_?

tpmanley:
@PKacprowiczS `I've just found out that child's DTH must has same namespace` - this shouldn't be the case. You can specify the namespace when creating the child device. Here is the method definition:

`DeviceWrapper addChildDevice(String namespace, String typeName, String deviceNetworkId, hubId = null, Map properties=[:])`

devicetypes/fibargroup/fibaro-walli-roller-shutter-venetian.src/fibaro-walli-roller-shutter-venetian.groovy
greens: it exists: `deleteChildDevice(String deviceNetworkId, Boolean force = false) `

greens:
@PKacprowiczS if you think we should leave the power/energy in I'll defer to you, but I think the default reporting should be fine for our purposes.

What does the calibration preference do? Is it equivalent to setting manually the open/closing time?

PKacprowiczS:
@greens it triggers automatic calibration process. Blinds are going up -> down -> up, measuring duration of those moves and then set those times as 156 and 157 parameters

dkirker:
I'm taking a look at these today to make sure that things are in sync with where Window Shade support is and will be.

One note, we are starting to add `Window Shade Level` support to the device handlers. I am working on a PR for this. However, if you want to help out you can add that support here (we can work on the metadata aspect a little later). There are some examples in this PR for the steps one would take to start sending the `shadeLevel` events: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/30151

PKacprowiczS:
@dkirker so current approach should be try to support both ways of including Level cap?

dkirker:
@PKacprowiczS Yes.

PKacprowiczS:
Since there's no way that ICP-13556 is gonna be resolved any time soon (if ever), I've been asked to come back to this approach.
@greens @dkirker @cbaumler can you review this again? Should we even consider multiple DTHes approach or we should wait until Malort is introduced?

greens:
just to refresh my memory, all of these devices share a fingerprint, and your approach was that based on certain reports you would switch the DTH because certain preferences were required for operation. And one of them had a child device to control the orientation of the blinds, right?

And the issue is that switching DTHs like that in OneApp doesn't refresh the metadata?

Other approaches we tried were having a main device with no capabilities and having conditional child devices with a metadata that could handle all combinations, but because the preferences are so disparate it wasn't a great solution?

PKacprowiczS:
@greens 
`all of these devices share a fingerprint`
Yes.

`your approach was that based on certain reports you would switch the DTH because certain preferences were required for operation`
Well, yes. In this case "certain reports" are configuration reports of device's mode

`And one of them had a child device to control the orientation of the blinds, right?`
Yes, in venetian blinds mode.

`Other approaches we tried were having a main device with no capabilities`
Actually, main device had common capabilities for all of modes (power/energy meters).

`and having conditional child devices with a metadata that could handle all combinations, but because the preferences are so disparate it wasn't a great solution?`
Main problem is that UI in device page doesn't refresh after changes in child devices list. It stays the same way it was at an inclusion, even if UI elements point to components that don't exist.
Having one main device, and therefore all preferences packed up in one place, was actually a neat solution if you ask me. It solved issues of unpredictable NPEs after DTH change (due no-existing preferences). And we could directly refer to device's manual, where manufacturer states details like "this preference means A in mode X or B in mode Z" etc. 

`And the issue is that switching DTHs like that in OneApp doesn't refresh the metadata?`
In fact, that works. That's why I even consider using this, as this the only one pro among cons (code duplication, possibility of much more issues). 
Sadly, solution with adding/removing child devices, which is my favorite, lacks "in the end, it actually works" kind of thing. It was discussed in ICP-13556 and WWST-6441 if you'd like to know more about it.

greens (approved):
Agree with you that this is enormous and would also prefer we were able to do it the other way, but you did a great job making this work.

PR #30339, opened 2020-05-06
Centralite smart outlet

SmartThings:
It doesn't dim lights


tpmanley:
If you're trying to report a bug please contact our support at https://support.smartthings.com/hc/en-us

PR #30421, opened 2020-05-07
ICP-10928 - Zigbee RGBW Bulb: Fix color temperature value on initial join, code refactoring.

KKlimczukS:
ICP-10928 - add colorTemperatureConfig to configure() method, code refactoring
@tpmanley Please, could You take a look at the code?

@MWierzbinskaS @ZWozniakS @PKacprowiczS @BJanuszS @MGoralczykS

### DIFF
--- a/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
+++ b/devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
@@ -123,7 +123,6 @@ private getHUE_COMMAND() { 0x00 }
 private getSATURATION_COMMAND() { 0x03 }
 private getMOVE_TO_HUE_AND_SATURATION_COMMAND() { 0x06 }
 private getCOLOR_CONTROL_CLUSTER() { 0x0300 }
-private getATTRIBUTE_COLOR_TEMPERATURE() { 0x0007 }
 
 // Parse incoming device messages to generate events
 def parse(String description) {
@@ -191,12 +190,9 @@ def ping() {
 
 def refresh() {
 	zigbee.onOffRefresh() +
-			zigbee.levelRefresh() +
-			zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_COLOR_TEMPERATURE) +
-			zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_HUE) +
-			zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_SATURATION) +
-			zigbee.onOffConfig(0, 300) +
-			zigbee.levelConfig()
+	zigbee.levelRefresh() +
+	zigbee.colorTemperatureRefresh() +
+	zigbee.hueSaturationRefresh()
 }
 
 def configure() {
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	// OnOff, level, colorTemperature minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	zigbee.onOffConfig(0, 300) +
+	zigbee.levelConfig(0, 300) +
+	zigbee.colorTemperatureConfig(0, 300) +
 	refresh()
 }
 
 def setColorTemperature(value) {
 	value = value as Integer
-	def tempInMired = Math.round(1000000 / value)
-	def finalHex = zigbee.swapEndianHex(zigbee.convertToHexString(tempInMired, 4))
 
-	zigbee.command(COLOR_CONTROL_CLUSTER, 0x0A, "$finalHex 0000") +
-			zigbee.readAttribute(COLOR_CONTROL_CLUSTER, ATTRIBUTE_COLOR_TEMPERATURE)
+	zigbee.setColorTemperature(value) +
+	zigbee.colorTemperatureRefresh()
 }
 
 //Naming based on the wiki article here: http://en.wikipedia.org/wiki/Color_temperature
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	// OnOff, level, colorTemperature minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	zigbee.onOffConfig(0, 300) +
+	zigbee.levelConfig(0, 300) +
+	zigbee.colorTemperatureConfig(0, 300) +
```
tpmanley: If I remember correctly we're not using reporting for color temperature and color on purpose. @varzac do you recall why?

varzac: I don't remember any context.

tpmanley: I think it was because the device would report when it started to transition and sometimes wouldn't report again for a while so it would get "stuck". We put in polling and removed the reporting to avoid that.

@KKlimczukS I also don't understand how this fixes the issue. The actual value is 6500 and when converted to mireds and back to color temperature again it becomes 6536.

KKlimczukS: @tpmanley I added setting the color temperature to 6500K in configuration().
To double-check -> will those devices who support color temp up to < 6500K ignore that command or they should consume it and set max possible value ?

tpmanley: The spec doesn't seem to say from my quick check. Personally I think this is fine for now and we can improve it later if we find it doesn't work when joining a device that doesn't support up to 6500.

tpmanley:
@KKlimczukS Can you update the PR title to summarize the important change? Personally I like to use `"<DTH Name>: <Summary>"` so I'd go with `"Zigbee RGBW Bulb: Fix color temperature value on initial join"` or something like that. We should be able to look at just the PR title and have a pretty good idea of what it's changing.

tpmanley (dismissed):

greens (dismissed):
It looks good, but this won't "fix" the issue until 0.31, so not sure the value of putting it in now.

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	// OnOff, level minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
+	zigbee.onOffConfig(0, 300) +
```
tpmanley, line 212: @KKlimczukS on re-review, I suggest doing the configs last like they were done in the original code. That way the current values are read first which is what the user cares about the most. So it should be:
* set color temperature
* refresh
* configure reporting

KKlimczukS: > @KKlimczukS on re-review, I suggest doing the configs last like they were done in the original code. That way the current values are read first which is what the user cares about the most. So it should be:
> 
> * set color temperature
> * refresh
> * configure reporting

@tpmanley ok, done.

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	refresh()
+	zigbee.setColorTemperature(7000) +
```
tpmanley: I intended the 7000 just as a test but thought it would be best to leave it at 6500 for the release. Unless you think 7000 is better?

KKlimczukS: @tpmanley You are right, let's leave 6500K here (and we will test setting 7000K on a side).

tpmanley (dismissed):

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	refresh()
+	zigbee.setColorTemperature(6500) +
```
tpmanley: The reason this isn't working like I thought it should is the `setColorTemperature` command is being generated in the DTH rather than being sent to the hub. 

I think an easy way to fix this is to change 6500 to 5000 (4000 works too). These are nice even numbers and the original value is retained when converting to and from mireds.

KKlimczukS: @tpmanley Did You mean that zigbee.setColorTemperature(6500) should work like You thought if it was wrapped in sendHubCommand() ?

Should we try sendHubCommand(zigbee.setColorTemperature(6500)) separately or use zigbee.setColorTemperature(5000) instead ?

tpmanley: `sendHubCommand` might work but there is also a timing issue in that the device record isn't synced to the hub when `configure` starts executing. So you would need to make sure the command isn't sent to the hub until a few seconds after `configure` starts executing. This is why I suggested just changing the code to `zigbee.setColorTemperature(5000)` because then it doesn't rely on the hub generating the command and avoids the timing issue.

devicetypes/smartthings/zigbee-rgbw-bulb.src/zigbee-rgbw-bulb.groovy
```diff
@@ -205,17 +201,18 @@ def configure() {
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
-	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	refresh()
+	zigbee.setColorTemperature(5000) +
```
tpmanley: @KKlimczukS I thought of one more issue last night. `configure` can be run after initial inboarding so it's possible the color temp is currently at the desired value and this changes it to a different value. What do you think about only setting the color temp to 5000 if the attribute doesn't have a value?

KKlimczukS: @tpmanley The issue was reported because color temperature was set to 6.536 K to right after pairing. If we added the condition You mentioned it would work for general cases where the set values are in range but i'm afraid it wouldn't fix the reported problem.

tpmanley: @KKlimczukS my suggestion is to set the value only if there isn't a current value meaning it will only be run when the device is initially onboarded so I think it will fix the reported problem.

KKlimczukS: @tpmanley ok, fixed, please let me know if that looks better now.

tpmanley (approved):

PR #30501, opened 2020-05-07
CHAD-4921 Add fingerprint for Wink Motion Sensor

greens:


tpmanley (dismissed):
May as well target `staging` to get this out sooner

tpmanley (approved):

PR #31381, opened 2020-05-14
Handler

SmartThings:
Smartthings for hub


tpmanley:
Looks like this is just a test so I'm going to close it. Please let us know if there are changes you'd like to see made to one of these files.

PR #31688, opened 2020-05-19
[WWST-6260] New DTH for Fibaro Roller Shutter 3

PKacprowiczS:
@tpmanley @greens @dkirker @KKlimczukS @MGoralczykS @ZWozniakS @MWierzbinskaS @BJanuszS 

Here's a new DTH for Fibaro Roller Shutter 3 which implements my idea we've talked about on call. I've moved from multiple DTHes and instead I've created "wider" metadata, which define all possible child devices I'm creating in the DTH.

Generally, it works fine, but... That dynamic approach of showing different child devices works (for now) only on device inclusion. After changing current mode, device page keeps showing non existing devices, while missing those which are freshly created. I'll reach to HQ mobile app team to try to fix it.


devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,728 @@
...
+		}
+	}
+}
+
+def installed() {
+	state.calibrationStatus = "notStarted"
+	state.childDevices = [:]
+	state.childDevices.main = null
+	state.childDevices.assistant = null
+	state.currentMode = "windowShade"
```
PKacprowiczS: If I hardcode different mode here, device's page is shown correctly.

devicetypes/smartthings/child-window-shade.src/child-window-shade.groovy
```diff
@@ -0,0 +1,78 @@
...
+ *	on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License
+ *	for the specific language governing permissions and limitations under the License.
+ *
+ */
+
+metadata {
+	definition (name: "Child Window Shade", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.blind") {
+		capability "Window Shade"
+        capability "Window Shade Level"
+        capability "Actuator"
```
greens, line 20: nit: spacing

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,728 @@
...
+		} else {
+			def preferenceName = it.key + "Boolean"
+			settings."$preferenceName" = true
+			state.currentPreferencesState."$it.key".status = "synced"
+		}
+	}
+	// Preferences template end
+}
+
+def updated() {
```
greens, line 78: For all of these preferences, I wonder if it's worth putting in code such that we don't set/query every parameter on install (or at least not default values)? I think it's safe to assume that devices have their factory default values on install.

Maybe another state when we initialize the preference map that denotes that it's still on default? I just worry about the storm of communication this requires.

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,728 @@
...
+		}
+		standardTile("refresh", "device.power", inactiveLabel: false, decoration: "flat", width: 2, height: 2) {
+			state "default", label:'', action:"refresh.refresh", icon:"st.secondary.refresh"
+		}
+		
+		details(["power", "energy", "refresh"])
+	}
+
+	preferences {
+		parameterMap.each {
```
greens, line 44: I wouldn't mind if we inlined a lot of this since it's mostly boilerplate.

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,728 @@
...
+	],
+	[
+		name: "Outputs orientation", key: "outputsOrientation", type: "boolean",
+		parameterNumber: 25, size: 1, defaultValue: 0,
+		optionInactive: 0, inactiveDescription: "default (Q1 - 1st channel, Q2 - 2nd channel)",
+		optionActive: 1, activeDescription: "reversed (Q1 - 2nd channel, Q2 - 1st channel)",
+		description: "This parameter allows reversing the operation of Q1 and Q2 without changing the wiring (in case of invalid motor connection) to ensure proper operation."
+	],
+	[
+		name: "Measuring power consumed by the device itself", key: "measuringPowerConsumedByTheDeviceItself", type: "boolean",
```
greens, line 714: Did we decide to leave the metering in?

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,728 @@
...
+}
+
+private getCommandValue(preference) {
+	def parameterKey = preference.key
+	switch (preference.type) {
+		case "boolean":
+			return settings."$parameterKey" ? preference.optionActive : preference.optionInactive
+		case "boolRange":
+			def parameterKeyBoolean = parameterKey + "Boolean"
+			return !notNullCheck(settings."$parameterKeyBoolean") || settings."$parameterKeyBoolean" ? settings."$parameterKey" : preference.disableValue
```
dkirker: This code seems a bit strange... Was there a reason for not just doing something like  `(settings."$parameterKeyBoolean" == null)`?

If you do need to keep it, I'd recommend using something like `isNull` and `isNotNull`.

dkirker:
@PKacprowiczS I'm still poking around at this, especially in terms of the window shade capability usage. I should have some more feedback over the next couple of days.

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
PKacprowiczS: I'm not sure if I understood you correctly, but right now ` it's safe to assume that devices have their factory default values on install` it's exactly what happens here. 
In `installed()` I'm setting all values of settings according to theirs defaults. There's no any Z-Wave communication at this point.
Then, if any parameter is updated by user, we're comparing `state.currentPreferencesState` and `settings` in search of differences. If there's any, DTH is sending configuration commands related just to those particular parameters. 

I doubt that I'll able to achieve lower radio traffic than that.

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
PKacprowiczS: I'd like to believe there was a reason, in other case this seems to be utterly nonsense. Anyway, I've changed it to something normal.

PKacprowiczS:
I'm bringing this one back from coma. We've been forced to use many DTHes with duplicated code instead of one which utilizes MCD approach, as then device page doesn't refresh, but I've (probably) found a way to tweak this.
As device page does refresh after DTH change, I've decided to use this behaviour, but instead of changing it to another, robust DTH, I'm changing it to small handler, whose only goal is to switch back to the original DTH. 
It's proven to be working, but I haven't yet confirmed whether it can be really reliable. Therefore, treat this as a WIP.

@greens @dkirker @tpmanley could you guys have a look, and tell if you'd merge a solution like this, or I should rather stick to duplicated DTHes?

devicetypes/fibargroup/fibaro-roller-shutter.src/fibaro-roller-shutter.groovy
```diff
@@ -0,0 +1,744 @@
...
+	zwaveEvent(encapsulatedCommand, cmd.sourceEndPoint as Integer)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.switchmultilevelv3.SwitchMultilevelReport  cmd, ep = null) {
+	if (cmd.value != 0xFE) {
+		if (ep != 2) {
+			shadeEvent(cmd.value)
+		} else {
+			def event = [name: "level", value: cmd.value != 0x63 ? cmd.value : 100]
+            sendEventsToAssistantChildDevice([event])
```
greens, line 358: nit: spacing

greens (commented):
This is an incredible hack @PKacprowiczS . A bit hard to follow, though. Can you explain what the preference and state collection children are doing?

PKacprowiczS:
@greens they're both used as a substitute to `state` map, as it is cleared every DTH change. I've decided to separate them and named each one for its own purpose for, idk, "tidiness" (and maybe for any future use, where only state collection would be needed).

greens:
So changing the parent DTH doesn't clear the children? Interesting. Looks good, just the one whitespace error.

PKacprowiczS:
That's correct, it doesn't clear child devices. I think I won't be updating this PR for a while as I'm blocked by ICP-13556 right now. You can have look if you're interested what's going on.

PR #32696, opened 2020-05-29
[WWST-5069] Added Everspring SP815 into Zooz 4 in 1 Motion Sensor and removed from Zwave Motion Sensor

ZWozniakS:
@greens @dkirker @tpmanley could you please review?

### DIFF
--- a/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
+++ b/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
@@ -23,12 +23,9 @@ metadata {
 		capability "Health Check"
 		capability "Tamper Alert"
 
-		fingerprint mfr: "027A", prod: "2021", model: "2101", deviceJoinName: "Zooz Multipurpose Sensor"
-		//Zooz 4-in-1 sensor
-		fingerprint mfr: "0109", prod: "2021", model: "2101", deviceJoinName: "Vision Multipurpose Sensor"
-		//ZP3111US 4-in-1 Motion
-		fingerprint mfr: "0060", prod: "0001", model: "0004", deviceJoinName: "Everspring Motion Sensor", mnmn: "SmartThings", vid: "SmartThings-smartthings-Everspring_Multisensor"
-		//Everspring Immune Pet PIR Sensor SP815
+		fingerprint mfr: "027A", prod: "2021", model: "2101", deviceJoinName: "Zooz Multipurpose Sensor" // Zooz 4-in-1 sensor
+		fingerprint mfr: "0109", prod: "2021", model: "2101", deviceJoinName: "Vision Multipurpose Sensor" // ZP3111US 4-in-1 Motion
+		fingerprint mfr: "0060", prod: "0001", model: "0004", deviceJoinName: "Everspring Motion Sensor", mnmn: "SmartThings", vid: "SmartThings-smartthings-Everspring_Multisensor" // Everspring Immune Pet PIR Sensor SP815
 	}
 
 	tiles(scale: 2) {
@@ -72,7 +69,7 @@ metadata {
 		section {
 			input(
 					title: "Settings Available For Everspring SP815 only",
-					description: "",
+					description: "To activate device's settings changes press the learn key on the device three times or check the device manual.",
 					type: "paragraph",
 					element: "paragraph"
 			)
@@ -129,8 +126,9 @@ def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
 	def results = []
 	results += createEvent(descriptionText: "$device.displayName woke up", isStateChange: false)
 
-	if (isEverspringSP815 && !isConfigured()) {
-		result += lateConfigure(true)
+	log.debug "isConfigured: ${isConfigured()}"
+	if (isEverspringSP815() && !isConfigured()) {
+		result += lateConfigure()
 	}
 
 	results += response([
@@ -220,10 +218,7 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 	} else {
 		result = createEvent(descriptionText: cmd.toString(), isStateChange: false)
 	}
-	log.debug "isConfigured: ${isConfigured()}"
-	if (isEverspringSP815() && !isConfigured()) {
-		result = lateConfigure()
-	}
+
 	return result
 }
 
@@ -312,17 +307,21 @@ def getConfigurationCommands() {
 	}
 	return result
 }
+
 def getEverspringDefaults() {
 	[1: 600,
 	 2: 180]
 }
+
 def lateConfigure() {
 	log.debug "lateConfigure"
 	sendHubCommand(getConfigurationCommands(),200)
 }
+
 private isConfigured() {
 	return state.configured == true
 }
+
 def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
 	if (isEverspringSP815()) {
 		if (cmd.parameterNumber == 1) {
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -300,6 +355,47 @@ private command(physicalgraph.zwave.Command cmd) {
...
+	sendHubCommand(getConfigurationCommands(),200)
+}
+private isConfigured() {
+	return state.configured == true
+}
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	if (isEverspringSP817()) {
+		if (cmd.parameterNumber == 4) {
+			state.retriggerIntervalSettings = scaledConfigurationValue
+			state.intervalConfigured = true
```
greens: since there's only one parameter you probably don't need two booleans to track whether the device is configured

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -300,6 +355,47 @@ private command(physicalgraph.zwave.Command cmd) {
...
+	}
+	return result
+}
+def lateConfigure() {
+	log.debug "lateConfigure"
+	sendHubCommand(getConfigurationCommands(),200)
+}
+private isConfigured() {
+	return state.configured == true
+}
```
greens: nit: put an extra empty line between these functions

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -153,6 +195,11 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 		def value = cmd.v1AlarmLevel == 255 ? "active" : cmd.v1AlarmLevel ?: "inactive"
 		result << createEvent(name: "alarm $cmd.v1AlarmType", value: value, isStateChange: true, displayed: false)
 	}
+
+	log.debug "isConfigured: ${isConfigured()}"
+	if (isEverspringSP817() && !isConfigured()) {
+		result += lateConfigure()
+	}
```
greens: sending a notificiationReport doesn't imply that the device is awake to receive commands, so this should only be in the wakeup handler

ZWozniakS: Yes, but the wakeup events are infrequent and we needed to force quick configuration when the parameter is changed. And it worked, a configuration is really fast when after changing parameter user is provoking for example tamper event. So thanks to that we can be sure that the configuration will be completed after the very next motion event without user's awareness. And this change was merged previously, I just removed checking for Everpring SP815.

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
ZWozniakS: Sure, fixed

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
ZWozniakS: Ok, done

devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
```diff
@@ -23,8 +23,12 @@ metadata {
 		capability "Health Check"
 		capability "Tamper Alert"
 
-		fingerprint mfr: "027A", prod: "2021", model: "2101", deviceJoinName: "Zooz Multipurpose Sensor" //Zooz 4-in-1 sensor
-		fingerprint mfr: "0109", prod: "2021", model: "2101", deviceJoinName: "Vision Multipurpose Sensor" //ZP3111US 4-in-1 Motion
+		fingerprint mfr: "027A", prod: "2021", model: "2101", deviceJoinName: "Zooz Multipurpose Sensor"
+		//Zooz 4-in-1 sensor
```
dkirker: Let's keep the comments indicating the device on the same line as the fingerprint.

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
```diff
@@ -17,25 +17,38 @@
...
-		fingerprint mfr: "0060", prod: "0001", model: "0006", deviceJoinName: "Everspring Motion Sensor" //Everspring Motion Detector
-		fingerprint mfr: "011A", prod: "0601", model: "0901", deviceJoinName: "Enerwave Motion Sensor"  // Enerwave ZWN-BPC //Enerwave Motion Sensor
-		fingerprint mfr: "0063", prod: "4953", model: "3133", deviceJoinName: "GE Motion Sensor" //GE Portable Smart Motion Sensor
-		fingerprint mfr: "0214", prod: "0003", model: "0002", deviceJoinName: "BeSense Motion Sensor" //BeSense Motion Detector
-		fingerprint mfr: "027A", prod: "0001", model: "0005", deviceJoinName: "Zooz Motion Sensor" //Zooz Outdoor Motion Sensor
-		fingerprint mfr: "027A", prod: "0301", model: "0012", deviceJoinName: "Zooz Motion Sensor", mnmn: "SmartThings", vid: "generic-motion-2" //Zooz Motion Sensor
+		capability "Configuration"
+
+		fingerprint mfr: "011F", prod: "0001", model: "0001", deviceJoinName: "Schlage Motion Sensor"
+		// Schlage motion //Schlage Motion Sensor
```
dkirker: Same comment as above. What it might be worth doing is categories by manufacturer sorted alphabetically like with the fingerprints in the Lock and light/switch/dimmer/outlet device handlers. i.e.

```
// Ecolink
fingerprint ... // Ecolink motion

// Eversping
fingerprint ... // Everspring motion
```

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
dkirker: Generally when a user updates a parameter they should press the wakeup button on the device. That typically produces the WakeUpNotification.

What does the device's user manual say about setting the configurations? Usually before the section listing the configuration options and valid values it will indicate that you need to perform some action on the device to tell the controller to send the updated values. (I believe for Fibaro devices this was usually three quick clicks of the button or pressing and holding until a certain color, and Aeotec devices were either one click of the button or three clicks of the button. So there should be something similar.)

I definitely wouldn't rely on squeezing in messages after receiving a notification report. I definitely see that you have it in the WUN handler, which is good. It might be best to remove the calls elsewhere and make sure that the device prefs page has a note indicating what the user should do after saving the device prefs changes. ("Consult the device manual for activating changes on the device." Or "Press the button on the device three times." if the manual states that. Or note both, since sometimes firmware changes on later devices...)

ZWozniakS: There is no such information in the user manual but I checked if it works for pressing the tamper button and yes, there is a need to press it three times in order to complete the requested configuration. I will update the description in preferences to guide the user and remove messages after notification report for both devices.

devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy
ZWozniakS: Fixed

devicetypes/smartthings/zwave-motion-sensor.src/zwave-motion-sensor.groovy
ZWozniakS: Fixed

greens (commented):
looks good but we have conflicts

ZWozniakS:
Here is new PR with the same changes: [#32967](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/32967).

PR #32782, opened 2020-06-01
ICP-13231- removed cause of duplicated events

MWierzbinskaS:
ICP-13231- removed cause of duplicated events
@greens @KKlimczukS @PKacprowiczS @ZWozniakS @MGoralczykS @BJanuszS 

### DIFF
--- a/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
+++ b/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
@@ -95,8 +95,7 @@ def setupChildDevices(){
 	sendEvent(name: "numberOfButtons", value: numberOfButtons, displayed: false)
 	if(!childDevices) {
 		addChildButtons(numberOfButtons)
-	}
-	if(childDevices) {
+
 		for(def endpoint : 1..prodNumberOfButtons[zwaveInfo.prod]) {
 			def event = createEvent(name: "button", value: "pushed", isStateChange: true)
 			sendEventToChild(endpoint, event)
--- a/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
+++ b/devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
@@ -74,7 +74,7 @@ def initialize() {
 	if(childDevices) {
 		def event
 		for(def endpoint : 1..prodNumberOfButtons[zwaveInfo.prod]) {
-			event = createEvent(name: "button", value: "pushed", isStateChange: true)
+			event = createEvent(name: "button", value: "pushed", isStateChange: true, displayed: false)
 			sendEventToChild(endpoint, event)
 		}
 	}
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -74,7 +74,7 @@ def initialize() {
 	if(childDevices) {
 		def event
 		for(def endpoint : 1..prodNumberOfButtons[zwaveInfo.prod]) {
-			event = createEvent(name: "button", value: "pushed", isStateChange: true)
+			event = createEvent(name: "button", value: "pushed", isStateChange: true, displayed: false)
```
greens: I think we still want the event to be displayed in at least the child device's history.

MWierzbinskaS: This event was also duplicated and this is the reason why I added "displayed: false". 
Should I remove this change?

greens: So the reason you're seeing two events is because the parent displays all the children's events (that are marked as displayed) as well as any of its own events (marked as displayed). So what we want is for the child event to be marked as displayed but the parent to be marked as not displayed.

What you've done here is mark both events (since `event` is re-used) as not displayed.

In the other handler you've done it differently.

MWierzbinskaS: Ok, but as I understand, in this function an event is created and then sent to child. I don't see where I use this event as the parent's event. 
I think this is a different situation than in CentralSceneNotification handling, where the event is both sent to child and returned as the parent's event. 
'initialize' is also called in the 'updated' and I think it may be the cause of the duplication. 
After moving child devices setup to "configure" function, those events were visible and not duplicated. 
What do you think about this solution?

devicetypes/smartthings/zwave-multi-button.src/zwave-multi-button.groovy
```diff
@@ -97,9 +86,24 @@ def configure() {
...
+	def numberOfButtons = prodNumberOfButtons[zwaveInfo.prod]
+	sendEvent(name: "numberOfButtons", value: numberOfButtons, displayed: false)
+	if(!childDevices) {
+		addChildButtons(numberOfButtons)
+	}
+	if(childDevices) {
+		for(def endpoint : 1..prodNumberOfButtons[zwaveInfo.prod]) {
+			def event = createEvent(name: "button", value: "pushed", isStateChange: true)
+			sendEventToChild(endpoint, event)
+		}
```
greens, line 102: this code should only run if the device didn't initially have the child devices. This will send a "pushed" event to every child every time configure is called

MWierzbinskaS: Ok, done

greens (approved):

PR #32881, opened 2020-06-02
[ICP-13246] Stelpro Thermostats preferences changes

PKacprowiczS:
@dkirker @greens @KKlimczukS @MGoralczykS @ZWozniakS @MWierzbinskaS @BJanuszS

### DIFF
--- a/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
@@ -53,8 +53,8 @@ metadata {
 			input("heatdetails", "enum", title: "Do you want to see detailed operating state events in the activity history? There may be many.", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		}
 		section {
-			input(title: "Outdoor Temperature", description: 	"To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing)." +
-					"Do not use space. If you don't want a forecast, leave it blank.", displayDuringSetup: false, type: "paragraph", element: "paragraph")
+			input(title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing). Do not use space. If you don't want a forecast, leave it blank.",
+					displayDuringSetup: false, type: "paragraph", element: "paragraph")
 			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "")
 		}
 	}
@@ -559,4 +559,26 @@ def fanCirculate() {
 
 def setThermostatFanMode() {
 	log.trace "${device.displayName} does not support fan mode"
+}
+
+/**
+ * Checks if the time elapsed from the provided timestamp is greater than the number of senconds provided
+ *
+ * @param timestamp: The timestamp
+ *
+ * @param seconds: The number of seconds
+ *
+ * @returns true if elapsed time is greater than number of seconds provided, else false
+ */
+private Boolean secondsPast(timestamp, seconds) {
+	if (!(timestamp instanceof Number)) {
+		if (timestamp instanceof Date) {
+			timestamp = timestamp.time
+		} else if ((timestamp instanceof String) && timestamp.isNumber()) {
+			timestamp = timestamp.toLong()
+		} else {
+			return true
+		}
+	}
+	return (now() - timestamp) > (seconds * 1000)
 }
--- a/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
+++ b/devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
@@ -50,11 +50,12 @@ metadata {
 
 	preferences {
 		section {
-			input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
+			input("heatdetails", "enum", title: "Do you want to see detailed operating state events in the activity history? There may be many.", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		}
 		section {
-			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing).", displayDuringSetup: false, type: "paragraph", element: "paragraph"
-			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+			input(title: "Outdoor Temperature", description: 	"To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing)." +
+					"Do not use space. If you don't want a forecast, leave it blank.", displayDuringSetup: false, type: "paragraph", element: "paragraph")
+			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "")
 		}
 	}
 
@@ -101,7 +102,7 @@ metadata {
 					[value: 84, color: "#f1d801"],
 					[value: 95, color: "#d04e00"],
 					[value: 96, color: "#bc2323"]
-				]
+			]
 		}
 		standardTile("temperatureAlarm", "device.temperatureAlarm", decoration: "flat", width: 2, height: 2) {
 			state "default", label: 'No Alarm', icon: "st.alarm.temperature.normal", backgroundColor: "#ffffff"
@@ -201,7 +202,7 @@ def parse(String description) {
 				value: map.value,
 				unit: map.unit,
 				data: [thermostatSetpointRange: thermostatSetpointRange]
-			])
+		])
 	}
 
 	log.debug "Parse returned $result"
@@ -279,7 +280,7 @@ def poll() {
 			zwave.thermostatModeV2.thermostatModeGet().format(),
 			zwave.thermostatSetpointV2.thermostatSetpointGet(setpointType: 1).format(),
 			zwave.sensorMultilevelV3.sensorMultilevelGet().format() // current temperature
-		], 100)
+	], 100)
 }
 
 // Event Generation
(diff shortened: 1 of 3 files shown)


devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -365,9 +366,12 @@ def zwaveEvent(thermostatoperatingstatev1.ThermostatOperatingStateReport cmd) {
...
 		map.name = "thermostatOperatingState"
 		map.value = operatingState
-
-		if (settings.heatdetails == "No") {
-			map.displayed = false
+		map.displayed = false
+		// If the user want to see each of the Idle and Heating events in the event history,
+		// Otherwise don't show them more frequently than 5 minutes.
+		if (settings.heatdetails == "Yes" ||
+				!secondsPast(device.currentState("thermostatOperatingState")?.getLastUpdated(), 60 * 5)) {
```
greens, line 373: I don't think `secondsPast` is defined.

dkirker: Yeah, it isn't in this DTH. It is in the Zigbee and Maestro ones down at the bottom of the file. @PKacprowiczS You'll need to copy `secondsPast` over.

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -50,11 +50,12 @@ metadata {
 
 	preferences {
 		section {
-			input("heatdetails", "enum", title: "Do you want a detailed operating state notification?", options: ["No", "Yes"], defaultValue: "No", required: true, displayDuringSetup: true)
+			input("heatdetails", "enum", title: "Do you want to see detailed operating state events in the activity history? There may be many.", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		}
 		section {
-			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing).", displayDuringSetup: false, type: "paragraph", element: "paragraph"
-			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+			input(title: "Outdoor Temperature", description: 	"To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing)." +
```
dkirker: Any reason for the odd formatting? I'd take out the tab after `description:` and just combine the string constant instead of using `+` -- but don't forget to add a space in the string before the "Do not" part.

PKacprowiczS: Nah, generally I don't like too long lines, so that's the reason for this

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -362,7 +364,7 @@ def handleTemperature(descMap) {
 			// just mask it.
 			if (!cleared &&
 					((lastAlarm == "freeze" && map.value > FREEZE_ALARM_TEMP) ||
-					 (lastAlarm == "heat" && map.value < HEAT_ALARM_TEMP))) {
+					(lastAlarm == "heat" && map.value < HEAT_ALARM_TEMP))) {
```
dkirker: I'd leave the space there so it is more clear that `(lastAlarm == "heat" && map.value < HEAT_ALARM_TEMP)` is still part of the overall group.

dkirker:
I'm still hesitant about this logic. As it is right now, everything gets forced to not visible, but then an event 5 minutes later is forced to visible. That will mean if you are heating a house for 6 hours, you will see events every 5 minutes, regardless of state change.

I would probably just leave the old logic in place, but update the 30 seconds to something else. That means that the `displayed` logic is left to the DTH API, unless it has been less than the update interval, then we force it to not displayed.

greens:
makes sense

PKacprowiczS:
Personally speaking, I'm also not a fan of any changes in logic here, but cert team has insisted. If you guys could swoop in the ticket it'd be great, cause any resistance from my side is usually opposed by "this is not our decision to make" catchphrase.

greens:
@PKacprowiczS it looks like the ticket is solely about the preference description? So if you want to avoid logic changes I think you'd be fine. 

But if you are going to do them I do agree with Donald that the logic should be:

If I have opted to see detailed events and it has been more than 5 minutes, don't force the displayed flag to true, just let it be set automatically by whether or not this is a state change.

PKacprowiczS:
@greens @dkirker Updated. I've followed Donald's advice and brought back old logic, with updated description and longer time interval.

greens:
I'm still not sure about automatically not showing events. I think we should only hide them if it's been less than 5 minutes AND they're duplicates. I would still want state changes to show no matter what.

PKacprowiczS:
```
if (operatingState) {
	map.name = "thermostatOperatingState"
	map.value = operatingState

	// If the user want to see each of the Idle and Heating events in the event history,
	// Otherwise don't show them more frequently than 5 minutes.
	if ((settings.heatdetails == "No" ||
			!secondsPast(device.currentState("thermostatOperatingState")?.getLastUpdated(), 60 * 5)) &&
			device.currentState("thermostatOperatingState").value == map.value) {
		map.displayed = false
	}
	
} else {
	log.trace "${device.displayName} sent invalid operating state $value"
}
```
Something like that?

greens:
Yeah. I think so. 

If the user has opted not to see details OR it's been less than 5 minutes since the last update, AND this is not a state change, then hide it.

Think you've got it.

PKacprowiczS:
On second thought, won't such conditions make this setting irrelevant and only events that will be shown in the history are those which change device's state? Let's say that user selected not to see that many events and 5 minutes had passed since the last update. Thermostat mode didn't change - event won't be shown in history due Events API. Same situation for less than 5 minutes since the last event - if events values are the same, we manually set _displayed_ flag to false. Only events that are being shown, are those which change thermostat's operating mode, in both possible settings.

greens:
That _should_ be true, actually, yeah. But if that was the case, why did this preference get created in the first place?

PKacprowiczS:
Judging from @dkirker's comment in JIRA ticket, thermostat operating mode in this case may change every several seconds, so he added this setting to avoid history being swamped with mode reports.

greens:
@dkirker can you chime in here. I seem to have lost the plot a little bit

dkirker:
The original logic as submitted by the partner is as you see in the Z-Wave DTH: if the setting is set to "don't show detailed reports", then displayed is set to false. For the Zigbee devices I also added some logic such that if the setting was set to "show detailed reports" but there was an event change in the last 30 seconds (the Zigbee devices at least have a duty cycle of 15 seconds -- the device wakes up, checks the temperature and decides if it needs to turn on or off) then don't show the event.

I would say just change that to 5 minutes, or maybe a little less. You shouldn't need the `device.currentState("thermostatOperatingState").value == map.value` bit as `createEvent` and `sendEvent` will both check the state with the last state and set the flags accordingly (the logic is: if `displayed` is null/undefined then set `displayed` to the value of `isStateChange`).

The alternate is that a function is written that emulates the thermostat operating state logic by checking the value the thermostat sent us, the last state, the current temperature, and the setpoint and doing some comparisons (with some margins of .5C/1F so that you aren't basically just parroting the state from the device). This, however, would require deeper testing and potentially introduce more bugs/edge cases. If certification is dead set on this being an issue, I would push back and state that these devices had already been certified and the devices themselves (nor the integrations) have changed since the last round/region of certification.

PKacprowiczS:
So, I'm reverting back to "original logic, with 5 minutes instead of 30 seconds", and I'll ask cert team to check whether they're happy with it.

devicetypes/stelpro/stelpro-ki-zigbee-thermostat.src/stelpro-ki-zigbee-thermostat.groovy
```diff
@@ -53,11 +53,11 @@ metadata {
...
+			input("heatdetails", "enum", title: "Do you want to see detailed operating state events in the activity history? There may be many.", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		}
-		section {
-			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing).", displayDuringSetup: false, type: "paragraph", element: "paragraph"
-			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+		section {input(title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing). Do not use space. If you don't want a forecast, leave it blank.",
```
dkirker: Not sure you meant to have this on the same line as the `section {`

devicetypes/stelpro/stelpro-maestro-thermostat.src/stelpro-maestro-thermostat.groovy
```diff
@@ -54,11 +54,11 @@ metadata {
...
+			input("heatdetails", "enum", title: "Do you want to see detailed operating state events in the activity history? There may be many.", options: ["No", "Yes"], defaultValue: "No", required: false, displayDuringSetup: true)
 		}
-		section {
-			input title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing).", displayDuringSetup: false, type: "paragraph", element: "paragraph"
-			input("zipcode", "text", title: "ZipCode (Outdoor Temperature)", description: "[Do not use space](Blank = No Forecast)")
+		section {input(title: "Outdoor Temperature", description: "To get the current outdoor temperature to display on your thermostat enter your zip code or postal code below and make sure that your SmartThings location has a Geolocation configured (typically used for geofencing). Do not use space. If you don't want a forecast, leave it blank.",
```
dkirker: Same as above: Not sure you meant to have this on the same line as the `section {`

devicetypes/stelpro/stelpro-ki-thermostat.src/stelpro-ki-thermostat.groovy
```diff
@@ -366,7 +367,10 @@ def zwaveEvent(thermostatoperatingstatev1.ThermostatOperatingStateReport cmd) {
 		map.name = "thermostatOperatingState"
 		map.value = operatingState
 
-		if (settings.heatdetails == "No") {
+		// If the user want to see each of the Idle and Heating events in the event history,
+		// Otherwise don't show them more frequently than 5 minutes.
```
dkirker: Just a little bit of nit-picking for this comment and the other two below:
`If the user wants` and `history, otherwise`.

dkirker:
@PKacprowiczS Sounds good. I left a couple of other comments, but if cert doesn't have a problem I'll approve and merge.

dkirker (approved):
@PKacprowiczS Looks good! Let us know what cert says and we can merge, or push back and then merge. :)

PR #33213, opened 2020-06-05
[ONEAPP-27569] SmartSense Motion Sensor: Use same icon as catalog for ST sensors

tpmanley:
The catalog has two entries for SmartThings motion sensors. One entry for the newest sensor and second entry for all the other sensors. The second entry uses a different motion sensor icon so we need to specify UI Metadata that also uses that same icon.

https://smartthings.atlassian.net/browse/ONEAPP-27569

### DIFF
--- a/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
+++ b/devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
@@ -21,24 +21,24 @@ metadata {
 		capability "Motion Sensor"
 		capability "Configuration"
 		capability "Battery"
-        capability "Temperature Measurement"
+		capability "Temperature Measurement"
 		capability "Refresh"
 		capability "Health Check"
 		capability "Sensor"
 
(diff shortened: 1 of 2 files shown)


tpmanley:
This wasn't working before due a plugin issue which has now been addressed in STG. I do need to do a little more testing before this is merged but I think it's ready for review.

tpmanley:
I finished testing so this is ready to merge once it's approved

dkirker (dismissed):

dkirker:
It isn't part of the PR, but line 24 in SmartSense Motion Sensor has some spacing issue with the temp measurement capability declaration. Dunno if you want to grab that, too.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -21,24 +21,24 @@ metadata {
 		capability "Motion Sensor"
 		capability "Configuration"
 		capability "Battery"
-        capability "Temperature Measurement"
+		capability "Temperature Measurement"
```
tpmanley, line 24: @dkirker fixed

dkirker (approved):

PR #33933, opened 2020-06-12
[WWST-6790] Fingerprints for DAWON devices

PKacprowiczS:
This is also for WWST-6795, WWST-6800, WWST-6805, WWST-6810
@tpmanley @greens @KKlimczukS @MGoralczykS @ZWozniakS @MWierzbinskaS @BJanuszS 
Switches despite being labeled as "PM-SX40-ZB", report model "PM-SX40R-ZB", (and according to quick search on net those are two separate models), so I've included both versions.

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
+++ b/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
@@ -235,8 +235,6 @@ private getChildCount() {
 		return 5
 	} else if (device.getDataValue("model") == "E220-KR6N0Z0-HA") {
 		return 6 
-	} else if (device.getDataValue("model") == "PM-S340-ZB") {
-		return 3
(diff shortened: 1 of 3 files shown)


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -229,6 +235,10 @@ private getChildCount() {
 		return 5
 	} else if (device.getDataValue("model") == "E220-KR6N0Z0-HA") {
 		return 6 
+	} else if (device.getDataValue("model") == "PM-S340-ZB") {
```
tpmanley: This is handled by the `else if` below

PKacprowiczS: Thanks, fixed

tpmanley (approved):

greens (approved):

PR #34180, opened 2020-06-15
[ICP-13020] EZMultipli - fix for timeouts after color change

PKacprowiczS:

### DIFF
--- a/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
+++ b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
@@ -135,6 +135,7 @@ def setupHealthCheck() {
 def installed() {
 	sendEvent(name: "motion", value: "inactive", displayed: false)
 	state.colorReceived = [red: null, green: null, blue: null]
+	state.setColor = [red: null, green: null, blue: null]
 	setupHealthCheck()
 }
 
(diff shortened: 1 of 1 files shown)


devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -247,15 +248,22 @@ def zwaveEvent(switchcolorv3.SwitchColorReport cmd) {
...
-		result << createEvent(name: "hue", value: hsv.hue)
-		result << createEvent(name: "saturation", value: hsv.saturation)
-		// Reset the values
-		RGB_NAMES.collect { state.colorReceived[it] = null}
+		if (state.setColor.red == state.colorReceived.red && state.setColor.green == state.colorReceived.green && state.setColor.blue == state.colorReceived.blue) {
+			unschedule()
+			result << createEvent(name: "hue", value: hsv.hue)
+			result << createEvent(name: "saturation", value: hsv.saturation)
+		} else {
+			runIn(2, "sendColorQueryCommands", [overwrite: true])
```
greens: I can see this easily ending up in an endless loop. For instance if the device never actually received our setColor command.

It *might* be better to attempt to send our set command again instead of querying the colors?

Thinking back about the original issue. Perhaps the problem is with the logic behind the nulling out of `state.colorReceived` in `parse()`?

PKacprowiczS: You mean here https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/2d1c56701c1aebb70c2e04df30f633cf6ac01f3b/devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy#L159 ? I doubt that's the case here, it looks like it's only a backup map initializer if one in _installed()_ failed for some reason.

devicetypes/drzwave/ezmultipli.src/ezmultipli.groovy
```diff
@@ -247,15 +249,30 @@ def zwaveEvent(switchcolorv3.SwitchColorReport cmd) {
 		result << createEvent(name: "color", value: hexColor)
 		// Send the color as hue and saturation
 		def hsv = rgbToHSV(*colors)
-		result << createEvent(name: "hue", value: hsv.hue)
-		result << createEvent(name: "saturation", value: hsv.saturation)
-		// Reset the values
-		RGB_NAMES.collect { state.colorReceived[it] = null}
+		if (state.setColor.red == state.colorReceived.red && state.setColor.green == state.colorReceived.green && state.setColor.blue == state.colorReceived.blue) {
```
greens, line 252: Now I'm confused. Wasn't the original issue that you couldn't count on `state.colorReceived` to be correct?

PKacprowiczS: I think the key part of it, was setting it back to _null_ here https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/34180/files#diff-923fb700963e1771dfa88fd274738f37L253 
Now, even if for a first report there is an invalid value in `state` map, querying it once again, generally solves problem

greens: so do we still want to restrict this to only the values we're setting? If we can avoid it I'd like to in order to preserve inter-operation with other z-wave controllers

dkirker:
I left a comment on the ticket.

greens (approved):

PR #34601, opened 2020-06-18
[ICP-11333] Everspring Radiator Thermostat - fix for timeout after setting mode

PKacprowiczS:
@kianooshST @greens @KKlimczukS @MGoralczykS @ZWozniakS @MWierzbinskaS @BJanuszS

### DIFF
--- a/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
+++ b/devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
@@ -225,7 +225,7 @@ def setThermostatMode(String mode) {
 
 	[
 			secure(zwave.thermostatModeV2.thermostatModeSet(mode: modeValue)),
-			"delay 2000",
+			"delay 5000",
 			secure(zwave.thermostatModeV2.thermostatModeGet())
 	]
 }


kianooshkarami (approved):

devicetypes/smartthings/zwave-radiator-thermostat.src/zwave-radiator-thermostat.groovy
```diff
@@ -225,7 +225,7 @@ def setThermostatMode(String mode) {
 
 	[
 			secure(zwave.thermostatModeV2.thermostatModeSet(mode: modeValue)),
-			"delay 2000",
+			"delay 5000",
```
kianooshkarami, line 228: We may want to make this to a device specific, I'd let @greens weigh in on this

PKacprowiczS: I've thought about that, but other device doesn't even support thermostat mode (it's hidden with different metadata).

greens: If the other device is hiding it anyway I don't mind. I'll wait to merge until Cert confirms they had the same success you did. We've had issues with timing-sensitive changes like this in the past.

PR #34615, opened 2020-06-18
ICP-12967 Configure Aeotec Water Sensor to send non-MC reports to pre…

greens:
…serve local execution


greens:
Confirmed working by cert team.

dkirker (approved):
Looks good.

Looks like part of the PR title was truncated.

PR #34989, opened 2020-06-22
even devices in different namespaces cannot share names

greens:


greens:
quick fixup.

even i didn't realize this was an issue

KKlimczukS (approved):

PR #35253, opened 2020-06-24
CHAD-3909 Change garageSensor preference to a boolean

greens:
DM automatically injects "Tap to set" as the description for preferences that do not set this value themselves. Changing to a boolean removes the issue of a separate call-to-action string.


tpmanley:
I don't think this will work because the preference is used by hub firmware already in the field

greens:
yeah, you're right. forgot the preference was also handled locally

PR #35590, opened 2020-06-29
WWST-6827 - Leviton 4 Speed Fan Controller

BJanuszS:
**WWST-6827**
The zwave-fan-controller.groovy has been adjusted to support the fourth ("Max") fan speed level. The current metadata do not support this "Max" level so some additional metadata for the Leviton 4 Speed Controller should be provided. 

@greens @dkirker @tpmanley could you please take a look at the code before the new metadata is ready?

@KKlimczukS @PKacprowiczS @MWierzbinskaS @ZWozniakS @MWierzbinskaS

### DIFF
--- a/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
+++ b/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
@@ -124,11 +124,10 @@ def fanEvents(physicalgraph.zwave.Command cmd) {
 
 		def fanLevel = 0
 
-		// The GE, Honeywell, and Leviton 3-Speed Fan Controller treat 33 as medium, so account for that
 		if (has4Speeds()) {
-			fanLevel = getValueFor4SpeedDevice(rawLevel)
+			fanLevel = getFanSpeedFor4SpeedDevice(rawLevel)
 		} else {
-			fanLevel = getValueFor3SpeedDevice(rawLevel)
+			fanLevel = getFanSpeedFor3SpeedDevice(rawLevel)
 		}
 		result << createEvent(name: "fanSpeed", value: fanLevel)
 	}
@@ -224,7 +223,8 @@ def ping() {
 	refresh()
 }
 
-def getValueFor3SpeedDevice(rawLevel) {
+def getFanSpeedFor3SpeedDevice(rawLevel) {
+	// The GE, Honeywell, and Leviton 3-Speed Fan Controller treat 33 as medium, so account for that
 	if (rawLevel == 0) {
 		return 0
 	} else if (1 <= rawLevel && rawLevel <= 32) {
@@ -236,7 +236,7 @@ def getValueFor3SpeedDevice(rawLevel) {
 	}
 }
 
-def getValueFor4SpeedDevice(rawLevel) {
+def getFanSpeedFor4SpeedDevice(rawLevel) {
 	if (rawLevel == 0) {
 		return 0
 	} else if (1 <= rawLevel && rawLevel <= 25) {
@@ -251,11 +251,7 @@ def getValueFor4SpeedDevice(rawLevel) {
 }
 
 def has4Speeds() {
-	if (isLeviton4Speed()) {
-		return true
-	} else {
-		return false
-	}
+	isLeviton4Speed()
 }
 
 def isLeviton4Speed() {
--- a/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
+++ b/devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
@@ -27,7 +27,7 @@ metadata {
 		command "raiseFanSpeed"
 		command "lowerFanSpeed"
 
-		fingerprint mfr: "001D", prod: "0038", model: "0002", deviceJoinName: "Leviton Fan" //Leviton 4-Speed Fan Controller
+		fingerprint mfr: "001D", prod: "0038", model: "0002", deviceJoinName: "Leviton Fan", mnmn: "SmartThings", vid: "SmartThings-smartthings-Z-Wave_Fan_Controller_4_Speed" //Leviton 4-Speed Fan Controller
 		fingerprint mfr: "001D", prod: "1001", model: "0334", deviceJoinName: "Leviton Fan" //Leviton 3-Speed Fan Controller
 		fingerprint mfr: "0063", prod: "4944", model: "3034", deviceJoinName: "GE Fan" //GE In-Wall Smart Fan Control
 		fingerprint mfr: "0063", prod: "4944", model: "3131", deviceJoinName: "GE Fan" //GE In-Wall Smart Fan Control
@@ -125,7 +125,7 @@ def fanEvents(physicalgraph.zwave.Command cmd) {
 		def fanLevel = 0
 
 		// The GE, Honeywell, and Leviton 3-Speed Fan Controller treat 33 as medium, so account for that
-		if (maxSupportedSpeeds == 4) {
+		if (has4Speeds()) {
 			fanLevel = getValueFor4SpeedDevice(rawLevel)
 		} else {
 			fanLevel = getValueFor3SpeedDevice(rawLevel)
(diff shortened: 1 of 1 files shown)


joshualyon:
This is a perfect example of needing some way to know _which speeds_ are supported in a Fan Speed device. While it's not directly related to the this PR, I figured I would comment here for visibility. It would be great if there was some sort of attribute to identify which speeds are supported for a Fan Speed device. See [Request: Attribute to Identify Supported Fan Speeds](https://community.smartthings.com/t/request-attribute-to-identify-supported-fan-speeds/197906?u=joshua_lyon) for more details.

kianooshkarami (changes requested):
Instead of `isLeviton4Speed()`, can you determine the low/medium/high levels based on `100/"Max Supported Speeds"? The Max supported Speeds would return the value of 4 or 3 depending on the device?

dkirker:
@joshualyon It is definitely something that I have been pushing to get supported for some time. We've "hacked" it in the past with attributes such as `supportedThermostatModes` and `supportedWindowShadeCommands`, as you have noted (I can provide a response to the thread you mention a little later). Right now this is something that we've had to add manually and I am actually trying to make this an actual pattern of the capability model instead of something that has to be done ad-hoc (and usually post facto) and every time it is needed (with the chance that naming conventions might not be followed, and logic might not be properly followed -- we've already had that problem with `supportedWindowShadeCommands` being implemented in the mobile client).

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -200,14 +201,18 @@ def lowerFanSpeed() {
 }
 
 def low() {
```
greens, line 203: @dkirker are low/medium/high/max a part of any actual fan capability, or are they hacked in?

KKlimczukS:
@dkirker 
We've created and we are testing new UI metadata for this device (they are in ACCEPTANCE) and it seems it works fine for this 4 speed fan controller.
vid: "SmartThings-smartthings-Z-Wave_Fan_Controller_4_Speed"
Should we leave those UI metadata as generic (vid as above) or should they be device specific ?

BJanuszS:
> Instead of `isLeviton4Speed()`, can you determine the low/medium/high levels based on `100/"Max Supported Speeds"? The Max supported Speeds would return the value of 4 or 3 depending on the device?

@kianooshST 
Yes, I have added a getMaxSupportedSpeeds() method for that. Please take a look at the code.

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -217,4 +222,44 @@ def refresh() {
 
 def ping() {
 	refresh()
+}
+
+def getValueFor3SpeedDevice(rawLevel) {
+	if (rawLevel == 0) {
+		return 0
+	}
+	if (1 <= rawLevel && rawLevel <= 32) {
```
dkirker: This should just be an `else if`

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -217,4 +222,44 @@ def refresh() {
...
+		return 2
+	} else if (67 <= rawLevel && rawLevel <= 100) {
+		return 3
+	}
+}
+
+def getValueFor4SpeedDevice(rawLevel) {
+	if (rawLevel == 0) {
+		return 0
+	}
```
dkirker: This should also be an `else if`

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -123,13 +124,11 @@ def fanEvents(physicalgraph.zwave.Command cmd) {
...
 		def fanLevel = 0
 
-		// The GE, Honeywell, and Leviton treat 33 as medium, so account for that
-		if (1 <= rawLevel && rawLevel <= 32) {
-			fanLevel = 1
-		} else if (33 <= rawLevel && rawLevel <= 66) {
-			fanLevel = 2
-		} else if (67 <= rawLevel && rawLevel <= 100) {
-			fanLevel = 3
+		// The GE, Honeywell, and Leviton 3-Speed Fan Controller treat 33 as medium, so account for that
```
dkirker: If you are breaking the logic out in to helper functions, please move this comment in to the 3-speed function. Just put it right above the first `if` as it originally was here.

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -200,15 +201,19 @@ def lowerFanSpeed() {
 }
 
 def low() {
-	setLevel(32)
+	setLevel(100 / maxSupportedSpeeds)
```
dkirker: Given the comment that I had left in this file some revisions ago, `// The GE, Honeywell, and Leviton treat 33 as medium, so account for that`, this logic is going to be troublesome.

dkirker (changes requested):
You'll run in to some issues with @kianooshST 's comment given the way that the 3-speed devices seem to work. See the comment that I had left some time ago: `// The GE, Honeywell, and Leviton treat 33 as medium, so account for that`.

dkirker:
@KKlimczukS I don't see that metadata as being one that actually exists in the repository. Who created it? As far as the vid to use, I would use that vid. No real reason to create one specific for a device in this instance.

KKlimczukS:
> @KKlimczukS I don't see that metadata as being one that actually exists in the repository. Who created it? As far as the vid to use, I would use that vid. No real reason to create one specific for a device in this instance.

@dkirker 
I've created it for test purposes yesterday, but it's on ACCEPTANCE only. 
A snippet from device config JSON file:

`"custom": [
        {
            "component": "",
            "alternatives": [
                {
                    "key": "0"
                },
                {
                    "key": "1"
                },
                {
                    "key": "2"
                },
                {
                    "key": "3"
                },
                {
                    "key": "4"
                }
            ],
            "capability": "fanSpeed",
            "range": [
                0,
                4
            ]
        }
    ],`

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
BJanuszS: @dkirker Right, I see the problem. Maybe I will go for something similar  like in the 1st commit so -> setLevel(has4Speeds() ? 25 : 32)

KKlimczukS:
> > @KKlimczukS I don't see that metadata as being one that actually exists in the repository. Who created it? As far as the vid to use, I would use that vid. No real reason to create one specific for a device in this instance.
> 
> @dkirker
> I've created it for test purposes yesterday, but it's on ACCEPTANCE only.
> A snippet from device config JSON file:
> 
> `"custom": [ { "component": "", "alternatives": [ { "key": "0" }, { "key": "1" }, { "key": "2" }, { "key": "3" }, { "key": "4" } ], "capability": "fanSpeed", "range": [ 0, 4 ] } ],`

@dkirker 
Shall I submit that UI metadata as it is (generic) or rename it and use as device specific?

dkirker:
@KKlimczukS Please submit it with the vid you gave it. That should be good. I imagine that this DTH will accommodate other Z-Wave fan controllers. Until we have a system/attribute in place to indicate supported ranges we'll just use metadata.

BJanuszS:
@dkirker 
I've fixed the issues. Could you please take a look at the code?

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -217,4 +222,42 @@ def refresh() {
...
+		return 2
+	} else if (51 <= rawLevel && rawLevel <= 75) {
+		return 3
+	} else if (76 <= rawLevel && rawLevel <= 100) {
+		return 4
+	}
+}
+
+def has4Speeds() {
+	if (isLeviton4Speed()) {
```
dkirker: This can just be simplified to:
```
def has4Speeds() {
	isLeviton4Speed()
}
```

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
dkirker: Please move this to right above line 228.

devicetypes/smartthings/zwave-fan-controller.src/zwave-fan-controller.groovy
```diff
@@ -217,4 +222,42 @@ def refresh() {
 
 def ping() {
 	refresh()
+}
+
+def getValueFor3SpeedDevice(rawLevel) {
```
dkirker: One more recommendation... Let's use `getFanSpeedFor3SpeedDevice` and `getFanSpeedFor4SpeedDevice` so that it is more clear what `Value` is.

dkirker (changes requested):
Just a couple more.

BJanuszS:
@dkirker I've fixed what you've asked for, please take a look.

kianooshkarami (approved):

dkirker (approved):

PR #35151, opened 2020-06-24
CWD Smart Colour Tuneable Lamp

SmartThings:
Brand name：CollingWood 
App：zhangfangwen's smart app 
device type：bulb 
relevant information： 
manufacturer: "CWD" 
deviceJoinName: "CWD light"

PR #35150, opened 2020-06-24
CWD Smart Dimmer Lamp

SmartThings:
Brand name：CollingWood 
App：zhangfangwen's smart app 
device type：bulb 
relevant information： 
manufacturer: "CWD" 
deviceJoinName: "CWD light"

PR #35658, opened 2020-06-29
Commit cleanup for #35150, #35151, and #35152

dkirker:
This should fix some un-squashed commits from #35150, #35151, and #35152 and a little bit of spacing.


dkirker:
@tpmanley This PR squashes some commits, so it *should* re-write the commit history to remove the "Update <file name>" commits from those PRs. (Unless we just don't care about them.) I did tweak a couple of lines that just had tabs or spaces on them, as well.

tpmanley:
@dkirker since the PRs have already been merged the only way to rewrite history is to force push changes to `master` and we're not going to do that so unfortunately we'll just have to live with it.

dkirker:
@tpmanley Ah, yes... duh... 🤦

PR #36193, opened 2020-07-06
Devws for most target limited co 144

adurosmart:
Fix the bug that there will be multiple events when pressing a button, and resolve code conflicts.


greens (changes requested):
I'm not sure what you're trying to do here. This doesn't add any fingerprints and you're breaking the functionality of several other devices that share these DTHs.

Also you seem to have (mistakenly?) added two empty DTHs and included whitespace changes to otherwise untouched files.

PR #36278, opened 2020-07-06
ICP-13040 Change zigbee devices to report decimal temperature

greens:
This accompanies hubcore and appengine-zigbee changes to allow the parsing and reporting of temperature values as floats.


tpmanley (approved):
Can you confirm that appengine-zigbee returns the value with 1 significant digit? That's what it looks like the code is doing to me but want to double check.

dkirker (approved):

PR #36279, opened 2020-07-07
CHAD-5172 Send follow-ups to verify that window shade level changes o…

greens:
…ccurred

We noticed what seems to be a device bug when window shades send us unsolicited battery reports at the same time they are told to open/close. This seems to cause them to send us immediate position updates rather than waiting until the action is completed. This change should send periodic gets until we see the value we want, up to 5 times, or about 25s after the intial command.

### DIFF
--- a/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
+++ b/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
@@ -272,7 +272,7 @@ def levelChangeFollowUp(expectedLevel) {
 def checkLevelReport(value) {
     if (state.expectedValue != null) {
         if ((state.expectedValue == 99 && value >= 99) ||
-                (state.expectedValue >= value - 2 && state.expectedValue <= value + 2 )) {
+                (value >= state.expectedValue - 2 && value <= state.expectedValue + 2)) {
             unschedule("checkLevel")
         }
     }
--- a/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
+++ b/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
@@ -266,7 +266,7 @@ def refresh() {
 def levelChangeFollowUp(expectedLevel) {
     state.expectedValue = expectedLevel
     state.levelChecks = 0
-    runIn(5, "checkLevel")
+    runIn(5, "checkLevel", [overwrite: true])
 }
 
 def checkLevelReport(value) {
@@ -281,7 +281,7 @@ def checkLevelReport(value) {
 def checkLevel() {
     if (state.levelChecks != null && state.levelChecks < 5) {
         state.levelChecks += 1
-        runIn(5, "checkLevel")
+        runIn(5, "checkLevel", [overwrite: true])
         sendHubCommand(zwave.switchMultilevelV1.switchMultilevelGet())
     } else {
         unschedule("checkLevel")
--- a/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
+++ b/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
@@ -270,8 +270,8 @@ def levelChangeFollowUp(expectedLevel) {
 }
 
 def checkLevelReport(value) {
-    if (state.expectedValue) {
-        if ((state.expectedValue == 99 && value > 99) ||
+    if (state.expectedValue != null) {
+        if ((state.expectedValue == 99 && value >= 99) ||
                 state.expectedValue == value) {
             unschedule("checkLevel")
         }
--- a/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
+++ b/devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
@@ -164,6 +164,7 @@ private handleLevelReport(physicalgraph.zwave.Command cmd) {
         shadeValue = "partially open"
         descriptionText = "${device.displayName} shade is ${level}% open"
     }
+    checkLevelReport(level)
     def levelEvent = createEvent(name: "level", value: level, unit: "%", displayed: false)
     def stateEvent = createEvent(name: "windowShade", value: shadeValue, descriptionText: descriptionText, isStateChange: levelEvent.isStateChange)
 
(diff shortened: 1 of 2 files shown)


cbaumler (dismissed):
LGTM

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
```diff
@@ -266,4 +261,29 @@ def refresh() {
...
+            unschedule("checkLevel")
+        }
+    }
+}
+
+def checkLevel() {
+    if (state.levelChecks != null && state.levelChecks < 5) {
+        state.levelChecks += 1
+        runIn(5, "checkLevel")
+        response(zwave.switchMultilevelV1.switchMultilevelGet())
```
dkirker: I believe that you are actually going to need to pass this to `sendHubCommand`, otherwise nothing will actually get done with it (`response` just wraps the argument in a `HubAction` object).

greens: I think you might be right

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
```diff
@@ -266,4 +261,29 @@ def refresh() {
...
+def levelChangeFollowUp(expectedLevel) {
+    state.expectedValue = expectedLevel
+    state.levelChecks = 0
+    runIn(5, "checkLevel")
+}
+
+def checkLevelReport(value) {
+    if (state.expectedValue) {
+        if ((state.expectedValue == 99 && value > 99) ||
+                state.expectedValue == value) {
```
dkirker: I'm guessing that the `state.levelChecks` that is checked in `checkLevel` is what is protecting us in the event that we tell the shades to go to 60% and they stop at 61%?

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
```diff
@@ -266,4 +261,29 @@ def refresh() {
             zwave.switchMultilevelV1.switchMultilevelGet().format(),
             zwave.batteryV1.batteryGet().format()
     ], 1500)
+}
+
+def levelChangeFollowUp(expectedLevel) {
+    state.expectedValue = expectedLevel
```
dkirker, line 267: This is more of a nit-picking, but `state.expectedLevel` would be more clear than `state.expectedValue`.

dkirker (commented):
Just left a couple of comments only in the Springs file, but they apply to both since it is the same code.

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
greens: yeah, capped at 5 retries.

Do you think I should do some "close enough" if it's within like +/- 5?

Automated-House:
Are these changes also needed in the iBlinds DTH?

greens:
@dkirker ping

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
dkirker: @greens I was wondering that. If you wanted to, I'd do +/- 2. I think I've seen the shades get pretty close. I know the iBlinds folks were saying that shades could take some time (I think up to 2 minutes?). You could also use `state.levelChecks` as a multiplier in the `runIn` timer. Something like `runIn(5 * (state.levelChecks + 1), ...)`, or some other computation to get you a few close values and then a few far out values. I'll leave it up to you.

dkirker (dismissed):
If you don't think any other tweaking is necessary in `checkLevelReport` it looks good. It probably wouldn't hurt to have a bit of a staggered check and maybe a value check with tolerance instead of one that is absolute, but I am not sure how necessary it is.

dkirker:
> Are these changes also needed in the iBlinds DTH?

@prjct92eh2 No, not at the moment. The iBlinds DTH just sends the level values immediately upon setting them. Though, I had been working on some other changes that might need to account for this. But those changes were to take place after some capability additions for Venetian blinds occurred.

greens:
So the user we asked to test this was having his devices pinged almost constantly, so there's almost certainly a bug in my logic here.

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
```diff
@@ -266,4 +261,29 @@ def refresh() {
...
+        if ((state.expectedValue == 99 && value >= 99) ||
+                state.expectedValue == value) {
+            unschedule("checkLevel")
+        }
+    }
+}
+
+def checkLevel() {
+    if (state.levelChecks != null && state.levelChecks < 5) {
+        state.levelChecks += 1
```
dkirker: This will need to be `state.levelChecks = state.levelChecks + 1`

The bit that handles writing to values in `state` can't handle `+=`, so you have to expand it. That is why this was running forever.

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
dkirker: I'd probably do a check that allows +/- 1 or +/- 2. That will likely kill off a couple of the checks instead of running unneeded checks (further reducing cloud cycles as well as network traffic).

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
dkirker: I still think `state.expectedValue` should be `state.expectedLevel`.

dkirker (changes requested):
Figured out the infinite running. The `state` object can't handle `+=` (or any of the `+=`, `-=`, `*=`, `/=`, `++`, or `--` I believe), so you have to manually expand that out in the code.

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
greens: that is very frustrating. Thanks for finding that.

devicetypes/smartthings/springs-window-fashions-shade.src/springs-window-fashions-shade.groovy
```diff
@@ -266,4 +261,29 @@ def refresh() {
...
+def levelChangeFollowUp(expectedLevel) {
+    state.expectedValue = expectedLevel
+    state.levelChecks = 0
+    runIn(5, "checkLevel", [overwrite: true])
+}
+
+def checkLevelReport(value) {
+    if (state.expectedValue != null) {
+        if ((state.expectedValue == 99 && value >= 99) ||
+                (state.expectedValue >= value - 2 && state.expectedValue <= value + 2 )) {
```
dkirker: Looks like ya got an extra space before the )) { (and below). Otherwise it looks good! I have the test program running this change. I'll copy over all of the code to make sure that I didn't leave anything out and let it run.

dkirker (approved):
The tests I've had running look like they are working pretty well.

PR #36283, opened 2020-07-07
Most target limited co 188

adurosmart:
Fixed  'parseAduroSmartButtonMessage' function in zigbee-mulit-button.groovy file, the bug that there will be multiple events when pressing a button. other file not change, only want merge zigbee-mulit-button.groovy .


tpmanley:
@adurosmart Can you rebase your changes on the latest `master` branch so there is only a single commit containing just the changes you want to make to `zigbee-multi-button.groovy`?

PR #36353, opened 2020-07-07
[ICP-8763] Zigbee Metering Plug: Prevent duplicate switch events

BJanuszS:
**<a href="https://smartthings.atlassian.net/projects/ICP/issues/?filter=myopenissues">[ICP-8763]</a> Zigbee Metering Plug: Prevent duplicate switch events**
Fixed an issue with duplicated "On/Off" events according to Tom Manley's comment on <a href="https://smartthings.atlassian.net/projects/ICP/issues/?filter=myopenissues">[ICP-8763]</a>

@KKlimczukS @MWierzbinskaS @PKacprowiczS @MGoralczykS @ZWozniakS , could you please take a look at the code?

### DIFF
--- a/devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
+++ b/devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
@@ -66,15 +66,16 @@ def getATTRIBUTE_HISTORICAL_CONSUMPTION() { 0x0400 }
 def parse(String description) {
     log.debug "description is $description"
     def event = zigbee.getEvent(description)
+    def descMap = zigbee.parseDescriptionAsMap(description)
     def powerDiv = 1
     def energyDiv = 100
 
-    if (event?.name == "switch" && description?.startsWith("catchall: ")) {
-        log.info "Ignoring default response with description $description"
-        return [:]
-    } else if (event) {
+    if (event) {
         log.info "event enter:$event"
-        if (event.name== "power") {
+        if (event.name == "switch" && !descMap.isClusterSpecific && descMap.commandInt == 0x0B) {
+            log.info "Ignoring default response with desc map: $descMap"
+            return [:]
+        } else if (event.name== "power") {
             event.value = event.value/powerDiv
             event.unit = "W"
         } else if (event.name== "energy") {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
```diff
@@ -154,3 +164,14 @@ def configure() {
            zigbee.configureReporting(zigbee.SIMPLE_METERING_CLUSTER, ATTRIBUTE_READING_INFO_SET, DataType.UINT48, 1, 600, 1) +
            zigbee.electricMeasurementPowerConfig(1, 600, 1) 
 }
+
+def isOnOffStateChanged(Map event) {
```
tpmanley: This is different than the change I proposed in the ticket. Since we're trying to avoid duplicate events, I don't think this implementation is going to achieve the desired effect state is not atomic nor does it provide read-after-write guarantees. Instead the change should be to ignore switch events that are created on receipt of a Default Response message. They you won't need to rely on any saved state.

BJanuszS: @tpmanley Sorry for late response. Due to remote work situtation I had to wait for a device. 

I have changed the implementation so that it does not relay on state. Instead, it just ignores the second event that was earlier created from "catchall" description. I've checked the event lists and "switch" is not doubled anymore. 

devicetypes/smartthings/zigbee-metering-plug.src/zigbee-metering-plug.groovy
```diff
@@ -69,7 +69,10 @@ def parse(String description) {
     def powerDiv = 1
     def energyDiv = 100
 
-    if (event) {
+    if (event?.name == "switch" && description?.startsWith("catchall: ")) {
```
tpmanley: I think the code would be a little simpler if you left the `if (event)` as is and put this additional check within the `if` block. 

Also I'd recommend doing this instead of looking for any `catchall` message:

```
if (event.name == "switch" && !descMap.isClusterSpecific && descMap.commandInt == 0x0B)
```

That explicitly looks for the Default Response message.

tpmanley:
We do need to include the Jira ticket in the PR title. My preferred format is this:

`[JIRA-XXXX] <DTH Name>: <Change summary>`

So in this case something like `[ICP-8763] Zigbee Metering Plug: Prevent duplicate switch events`

BJanuszS:
@tpmanley I've changed the code according to your suggestions. Please take a look.

tpmanley (approved):
Looks good, thanks!

PR #36732, opened 2020-07-10
[ICP-13365] Added Tamper Alert for Aeotec Door/Window Sensor 7 Pro

PKacprowiczS:


greens:
I think you need to add the code to the notification event handler to support tamper.

(If the device actually sends them)

dkirker:
@PKacprowiczS Do any of the other devices in this DTH support tamper?

PKacprowiczS:
@dkirker for sure Aeotec Door/Window Sensor 7 has it, I don't really know about the rest

greens (approved):

PR #36908, opened 2020-07-13
CHAD-5256 Update child DNIs when the parent rejoins

greens:


greens:
While I'm doing this I might go ahead and add similar code to all of our multi-endpoint zigbee devices.

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -78,15 +78,20 @@ metadata {
 }
 
 def installed() {
-        createChildDevices()
-        updateDataValue("onOff", "catchall")
-        refresh()
+    createChildDevices()
```
tpmanley: Looks like this is switch to spaces indentation but we should stick with tabs

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -120,7 +125,7 @@ def parse(String description) {
 private void createChildDevices() {
 	def x = getChildCount()
 	for (i in 2..x) {
-		addChildDevice("Child Switch Health", "${device.deviceNetworkId}:0${i}", device.hubId,
+		addChildDevice("Child Switch Health", "${device.deviceNetworkId}:${i}", device.hubId,
```
tpmanley, line 128: I think the leading 0 was here because the endpoint is hex and so we wanted it to be fixed length, similar to how the DNI is hex and includes leading 0s. I'd leave it as is unless there is a strong reason to change.

greens: if it's hex, the function that was parsing it back was treating it as decimal so there wasn't any consistency there so I did this for consistency's sake

tpmanley (dismissed):

tpmanley (approved):

mantiger:
Can you provide how to reproduce this issue to verify code?

tpmanley:
@mantiger I'll PM you

PR #36909, opened 2020-07-13
ICP-12929 Standardize how Fibaro preference default text is created

greens:
There was an existing pattern to to print the enum value of a preference rather than the raw enum, so I've just adopted it across all the fibaro DTHs that were using the incorrect method.

### DIFF
--- a/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
+++ b/devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
@@ -101,7 +101,7 @@ metadata {
 			input (
 				name: it.key,
 				title: null,
-				description: "$DescrDefVal",
+				description: "$descrDefVal",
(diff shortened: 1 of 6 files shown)


devicetypes/fibargroup/fibaro-door-window-sensor-2.src/fibaro-door-window-sensor-2.groovy
```diff
@@ -89,18 +89,19 @@ metadata {
...
 				element: "paragraph"
 			)
-			
+			def defVal = it.def as Integer
+			def descrDefVal = it.options ? it.options.get(defVal) : defVal
 			input (
 				name: it.key,
 				title: null,
-				description: "Default: $it.def" ,
+				description: "$DescrDefVal",
```
tpmanley: Should be lowercase `d`

tpmanley (dismissed):
Looks fine besides that one issue

tpmanley (approved):

dkirker (approved):

PR #37228, opened 2020-07-15
[ICP-13322, ICP-13326, ICP-13336, ICP-13345, ICP-13346, ICP-13355] - fixes for Qubino Dimmers

KKlimczukS:
**ICP-13322, ICP-13355:**

1% is default Minimum dimming value for dimmers,
When device is set to 1% - it turns off and device does not send any level reports
Minimum dimming value has to be set to 2%, so the device's internal range would be 2-100%
Still, for users it will relatively be 1-100% on the UI and device will report it.

**ICP-13326:**
It turned out that response(zwave.switchMultilevelV3.switchMultilevelGet()) didn't work properly in that case. 
Switched to sendHubCommand(encap(zwave.switchMultilevelV3.switchMultilevelGet())).

**ICP-13336**
Removed option 3, because it does not apply to Qubino Temperature Sensor ZMNHEA1.

**ICP-13345, ICP-13346**
Added meterGet calls after switchMultilevel commands.

@greens @dkirker Please, could You take a look at the code ?

cc: @MWierzbinskaS @ZWozniakS @PKacprowiczS @BJanuszS @MGoralczykS

### DIFF
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -167,13 +167,13 @@ def excludeParameterFromSync(preference){
 	return exclude
 }
 
-private readConfigurationFromTheDevice() {
+private getReadConfigurationFromTheDeviceCommands() {
 	def commands = []
 	parameterMap.each {
 		state.currentPreferencesState."$it.key".status = "reverseSyncPending"
 		commands += zwave.configurationV2.configurationGet(parameterNumber: it.parameterNumber)
 	}
-	sendHubCommand(encapCommands(commands))
+	commands
 }
 
 private syncConfiguration() {
@@ -240,6 +240,7 @@ def configure() {
 	// Still, for users it will relatively be 1-100% on the UI and device will report it.
 	// Parameter no. 60 – Minimum dimming value
 	commands << zwave.configurationV2.configurationSet(scaledConfigurationValue: 2, parameterNumber: 60, size: 1)
+	commands + getReadConfigurationFromTheDeviceCommands()
 
 	encapCommands(commands)
 }
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -582,11 +582,9 @@ private getParameterMap() {[
 		values: [
 			0: "Default value - Mono-stable switch type (push button) – button quick press turns between previous set dimmer value and zero)",
 			1: "Bi-stable switch type (on/off toggle switch)",
-			2: "Potentiometer (applies to Flush Dimmer 0-10V only, dimmer is using set value the last received from potentiometer or from z-wave controller)",
-			3: "0-10V Temperature sensor (regulated output, applies to Flush Dimmer 0-10V only)"
+			2: "Potentiometer (applies to Flush Dimmer 0-10V only, dimmer is using set value the last received from potentiometer or from z-wave controller)"
 		],
-		description: "Set input based on device type (switch, potentiometer, temperature sensor,..)." +
-			"After parameter change to value 3 first exclude module (without setting parameters to default value) then wait at least 30s and then re include the module! "
+		description: "Set input based on device type (mono-stable switch, bi-stable switch, potentiometer)."
 	],
 	[
 			name: "Input 2 switch type (applies to Qubino Flush Dimmer only)", key: "inputsSwitchTypes", type: "enum",
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -490,6 +490,11 @@ def setLevel(value, duration = null) {
 	commands << zwave.switchMultilevelV3.switchMultilevelSet(value: level, dimmingDuration: dimmingDuration)
 	commands << zwave.switchMultilevelV3.switchMultilevelGet()
 
+	if(supportsPowerMeter()){
+		commands << zwave.meterV2.meterGet(scale: 0)
+		commands << zwave.meterV2.meterGet(scale: 2)
+	}
+
 	encapCommands(commands, getStatusDelay)
 }
 
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -122,7 +122,6 @@ def installed() {
 		state.currentPreferencesState."$it.key".value = getPreferenceValue(it)
 		state.currentPreferencesState."$it.key".status = "synced"
 	}
-	readConfigurationFromTheDevice()
 	// Preferences template end
 }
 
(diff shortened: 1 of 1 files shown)


devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -114,7 +114,7 @@ private getINPUT_TYPE_TEMPERATURE_SENSOR() {3}
 def installed() {
 	// Device-Watch simply pings if no device events received for 32min(checkInterval)
 	sendEvent(name: "checkInterval", value: 2 * 15 * 60 + 2 * 60, displayed: false, data: [protocol: "zwave", hubHardwareId: device.hub.hardwareID])
-
+	setInitialConfiguration()
```
dkirker: Why not just have `setInitialConfiguration()` and `readConfigurationFromTheDevice()` return the commands to send (instead of calling `sendHubCommand`) and return those commands from `installed`?

dkirker: Actually, these seem like things that should be part of `configure()`.

KKlimczukS: Your are totally right. I moved that part to `configure()`

KKlimczukS: I'd prefer to leave readConfigurationFromTheDevice() as a separate method, because it's a part of our template code for handling device preferences.

greens (commented):
@KKlimczukS so this minimum value basically has the device treat `setlevel(1)` as "go to 2%"? What does it report back?

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
dkirker: @KKlimczukS I understand that. You can leave it as a separate method but change the way that it works. This will keep all of the commands to the device together and they will get sent together. Then you'd have:
```
def configure() {
...
    cmds += readConfigurationFromTheDevice()
...
    return cmds
}
```

And any time outside of a capability command function you would just do:
```
sendHubCommand(readConfigurationFromTheDevice())
```

dkirker: In the past we've had issues with Z-Wave devices where if we started sending them data from the DTH too quickly (usually from within `installed()`) they wouldn't be in a state where they would act on it properly. Having those commands in `configure()` or in `zwaveEvent(WakeUpNotification)` helped. (This is more of a heads up since you use the template elsewhere.)

dkirker: I would also do similar with `syncConfiguration()` and how it is used in `updated()`. I see that is the only thing in `updated()` here, though that isn't the case here: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/37229/files#diff-c4d3d4cb621c024a3943efd31d82d407R176

KKlimczukS: @dkirker I see Your point. 
readConfigurationFromTheDevice() was moved (and renamed) to configure().

KKlimczukS:
> @KKlimczukS so this minimum value basically has the device treat `setlevel(1)` as "go to 2%"? What does it report back?

@greens You can still set the level 1-100%, but in that case the lower bound is 2% - but it's for the device internally. 

If the device is configured with min = 1% -> setting level to 1% will make the device turn OFF report SwitchMultilevelReport(value: 0)
If the device is configured with min = 2% -> setting level to 1% will  make the device setLevel = 1% (internally 2%) and report SwitchMultilevelReport(value: 1).

On other words - we can set min and max level on a device internally but device will still hande setting level 1-100% but for a narrow range.

greens:
So in some cases `setLevel(1)` puts the device to 1% and in other cases it turns it off? That does not seem like ideal behavior. We'd want `setLevel(1)` to behave consistently from the user's perspective (at least with respect to the device being on or off) regardless of the preferences.

KKlimczukS:
> So in some cases `setLevel(1)` puts the device to 1% and in other cases it turns it off? That does not seem like ideal behavior. We'd want `setLevel(1)` to behave consistently from the user's perspective (at least with respect to the device being on or off) regardless of the preferences.

Device turns off only when minimum dimming value is set to 1% **and** user sets 1% on the UI.
Changing device's internal Minimum dimming value settings to >= 2% will surely will help with that (for ST users it will be transparent and they will still use range 1-100%).
I know, it's not ideal, but it might be the only way to configure that device so it would work consistently with ST.
Otherwise we might need to generate a new custom UI metadata with dimming settings that starts from 2% - and that might not look good on the UI.

greens:
I'm just confused about how the mapping works. 

If I set my device's internal range to 1-100, setting the level to 1 turns it off.

If I set my device's internal range to 2-100, setting the level to 1 returns a report of 1%?

What happens for higher minimums? If I set the range to 4-100, do I get different results for setting the device to 1,2,3,4?

It's just confusing to me that in the first scenario, setting the level to the minimum value results in the device being off, but in the second scenario, setting the level **below** the minimum value results in the device still being on?

KKlimczukS:
> I'm just confused about how the mapping works.
> 
> If I set my device's internal range to 1-100, setting the level to 1 turns it off.
> 
> If I set my device's internal range to 2-100, setting the level to 1 returns a report of 1%?
> 
> What happens for higher minimums? If I set the range to 4-100, do I get different results for setting the device to 1,2,3,4?
> 
> It's just confusing to me that in the first scenario, setting the level to the minimum value results in the device being off, but in the second scenario, setting the level **below** the minimum value results in the device still being on?

@greens 
If You set Your device's internal range to 1-100, setLevel(1) turns the device off.
If You set Your device's internal range to **2**-100, setLevel(1) sets **2** on the device internally and device reports 1%
If You set Your device's internal range to **3**-100, setLevel(1) sets **3** on the device internally and device reports 1%
If You set Your device's internal range to **4**-100, setLevel(1) sets **4** on the device internally and device reports 1%
If You set Your device's internal range to **5**-100, setLevel(1) sets **5** on the device internally and device reports 1%
... and so on.

That way reports that come from the device are consistent with ST.

greens (dismissed):

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -234,6 +234,14 @@ def configure() {
 	commands << zwave.multiChannelV3.multiChannelEndPointGet()
 	commands + getRefreshCommands()
 
+	// 1% is default Minimum dimming value for dimmers,
+	// when device is set to 1% - it turns off and device does not send any level reports
+	// Minimum dimming value has to be set to 2%, so the device's internal range would be 2-100%
+	// Still, for users it will relatively be 1-100% on the UI and device will report it.
+	// Parameter no. 60 – Minimum dimming value
+	commands << zwave.configurationV2.configurationSet(scaledConfigurationValue: 2, parameterNumber: 60, size: 1)
+	commands + getReadConfigurationFromTheDeviceCommands()
```
greens: did you mean to do `+=` here?

KKlimczukS: > did you mean to do `+=` here?

@greens 
Yes, thanks for pointing that out!

greens (approved):

PR #37638, opened 2020-07-17
GPD-433 Fibaro Smoke Sensor - Update preference range

greens:


devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -69,7 +69,7 @@ metadata {
 		input "temperatureReportInterval", "enum", title: "Temperature Report Interval",
 				options: ["reports inactive", "5 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours", "18 hours", "24 hours"], defaultValue: "${temperatureReportInterval}", displayDuringSetup: true
 		input "temperatureReportHysteresis", "number", title: "Temperature Report Hysteresis", description: "Available settings: 1-100 C", range: "1..100", displayDuringSetup: true
-		input "temperatureThreshold", "number", title: "Overheat Temperature Threshold", description: "Available settings: 0 or 2-100 C", range: "0..100", displayDuringSetup: true
```
tpmanley: Is 1 valid for this preference?

greens: All of the code for the preference leads me to believe that it is.

tpmanley: Saying the range is 0 -100 seems more natural. But then we'll need to get updated translations so maybe we should just remove the description since we already display the valid range.

greens: I think the intention of the text is to make it clear that a setting of 0 behaves in a way that is not linearly consistent with 1-100.

tpmanley: What does 0 do? It's not clear from the description how it's different

greens: I believe 0 means that the device will never report overheat.

tpmanley: Screen Shot 2020-07-18 at 2 52 32 PM

Only 1-100 is valid. The current code treats a value less than 1 as 1. I'd say we use the same description as `temperatureReportHysteresis` so we can use the same translated string.

devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -69,7 +69,7 @@ metadata {
...
+		input "temperatureThreshold", "number", title: "Overheat Temperature Threshold", description: "Available settings: 1 or 1-100 C", range: "1..100", displayDuringSetup: true
```
tpmanley: Should be `Available settings: 1-100 C`

greens:
@tpmanley I got pinged about this one, didn't realize it had been sitting here for so long.

devicetypes/smartthings/fibaro-smoke-sensor.src/fibaro-smoke-sensor.groovy
```diff
@@ -69,7 +69,7 @@ metadata {
 		input "temperatureReportInterval", "enum", title: "Temperature report interval",
 				options: ["Reports inactive", "5 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours", "18 hours", "24 hours"], defaultValue: "30 minutes", displayDuringSetup: true
 		input "temperatureReportHysteresis", "number", title: "Temperature report hysteresis", description: "Available settings: 1-100 C", range: "1..100", displayDuringSetup: true
-		input "temperatureThreshold", "number", title: "Overheat temperature threshold", description: "Available settings: 0 or 2-100 C", range: "0..100", displayDuringSetup: true
+		input "temperatureThreshold", "number", title: "Overheat temperature threshold", description: "Available settings: 1 or 1-100 C", range: "1..100", displayDuringSetup: true
```
tpmanley: Remove the "1 or" part because its not necessary anymore

greens: done

tpmanley (approved):

PR #37869, opened 2020-07-20
CHAD-5280 Cap number of battery queries

greens:
Locks will query for battery state after a battery replace event, but in some cases, the lock will go offline before responding resulting in an infinite loop of querying the battery, which can clog the z-wave message queue.


dkirker:
Do we want to account for the chance that when this DTH gets deployed there may still be devices in this query loop? Maybe in `queryBattery` we want to have a check `if (state.batteryQueries == null) state.batteryQueries = 0` that way we'll get 5 more runs? (I guess maybe if a device is in that state when this is deployed it won't make a difference if we do 5 more attempts or just end it because state.batteryQueries is null because it probably won't respond anyway?)

greens:
@dkirker I see your point. This doesn't really "fix" anything for devices already in this loop, but if they ever are brought back online and have new batteries installed, the first battery report should break them out of it.

I guess it doesn't hurt to just have that line.

greens:
fixed

greens:
fixed again

kianooshkarami (approved):

PR #38018, opened 2020-07-21
ICP-13140 For IKEA blinds return the current state of the blinds when pause is called if the blinds are not moving

dkirker:
The mobile client expects a response so if `pause()` is called but the blinds aren't moving (and likely to not reply to the command) then send the current state to prevent a network error dialog.

### DIFF
--- a/devicetypes/smartthings/zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy
+++ b/devicetypes/smartthings/zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy
@@ -189,7 +189,12 @@ def setLevel(data, rate = null) {
 
 def pause() {
 	log.info "pause()"
-	zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_PAUSE)
+	// If the window shade isn't moving when we receive a pause() command then just echo back the current state for the mobile client.
+	if (device.currentValue("windowShade") != "opening" && device.currentValue("windowShade") != "closing") {
+		sendEvent(name: "windowShade", value: device.currentValue("windowShade"), isStateChange: true, display: false)
+	} else {
+		zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_PAUSE)
+	}
 }
 
 def presetPosition() {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy
```diff
@@ -189,7 +189,12 @@ def setLevel(data, rate = null) {
 
 def pause() {
 	log.info "pause()"
-	zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_PAUSE)
+	// If the window shade isn't moving when we receive a pause() command then just echo back the current state for the mobile client.
+	if (device.currentValue("windowShade") != "opening" && device.currentValue("windowShade") != "closing") {
+		sendEvent(name: "windowShade", value: device.currentValue("windowShade"), isStateChange: true, display: false)
+	} else {
+		zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_PAUSE)
```
tpmanley: I'd suggest always sending the command because it doesn't hurt anything and it's possible the shade is moving but our current state isn't accurate for some reason.

dkirker: That's actually not a bad idea!

tpmanley (dismissed):

devicetypes/smartthings/zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy
```diff
@@ -189,6 +189,10 @@ def setLevel(data, rate = null) {
 
 def pause() {
 	log.info "pause()"
+	// If the window shade isn't moving when we receive a pause() command then just echo back the current state for the mobile client.
+	if (device.currentValue("windowShade") != "opening" && device.currentValue("windowShade") != "closing") {
+		sendEvent(name: "windowShade", value: device.currentValue("windowShade"), isStateChange: true, display: false)
```
dkirker: Oooops Should have been `displayed: false`

tpmanley (approved):

PR #38146, opened 2020-07-22
CHAD-5256 Hotfix to update child DNIs of zigbee-multi-switch

greens:
There was more code using the leading 0 of child devices than I first noticed, so the leading zero in the child DNI has simply been re-added.


greens:
@tpmanley you're allowed to say "I told you so"

tpmanley (approved):
I've tested these changes as best I can and it's looks good. Please change the PR title to be more descriptive of the issue though.

PR #38295, opened 2020-07-23
[WWST-6159] Qubino Flush Shutter DC - fingerprint and improvements

PKacprowiczS:
@greens @SmartThingsCommunity/srpol-pe-team @BJanuszS 
This PR also includes fix for ICP-13405, as only difference between those two devices is power input.


devicetypes/qubino/qubino-flush-shutter.src/qubino-flush-shutter.groovy
```diff
@@ -355,7 +361,7 @@ def zwaveEvent(physicalgraph.zwave.commands.meterv3.MeterReport cmd, ep = null)
 			eventMap.value = Math.round(cmd.scaledMeterValue)
 			eventMap.unit = "W"
 			events += createEvent(eventMap)
-			if (cmd.scaledMeterValue) {
+			if (Math.round(cmd.scaledMeterValue)) {
```
greens, line 367: is this because you were getting like .01W reports when the device was actually stopped?

PKacprowiczS: exactly

greens (approved):

PR #38387, opened 2020-07-24
CHAD-5315 Update SmartWeather Station Tile DTH

dkirker:
Update SmartWeather Station Tile to ensure backwards compatibility with SmartApps and compatibility with the new client.

### DIFF
--- a/devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
+++ b/devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
@@ -230,7 +230,7 @@ def pollUsingZipCode(String zipCode) {
         send(name: "feelsLike", value: obs.temperatureFeelsLike, unit: tempUnits)
 
         send(name: "humidity", value: obs.relativeHumidity, unit: "%")
-        send(name: "weather", value: obs.wxPhraseShort)
+        send(name: "weather", value: obs.wxPhraseLong)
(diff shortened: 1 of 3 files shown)


dkirker:
I've tested this and this works with existing smartapps (backwards compatibility) and with the mobile client and plugin. I also discovered that when using with a PWS ID (I didn't previously have one to test with, but got one from the community -- thanks!) you can take the geo-coordinates from the PWS and pass them to the APIs and get the rest of the forecast. So, that eliminates the need for "-" for the measurements.

I'd like to get this in production next Tuesday, so if it doesn't seem like we can merge it this morning I'll rebase it to staging.

devicetypes/smartthings/smartweather-station-tile.src/smartweather-station-tile.groovy
```diff
@@ -275,7 +261,7 @@ def pollUsingZipCode(String zipCode) {
         def sunsetDate = dtf.parse(obs.sunsetTimeLocal)
 
         def tf = new java.text.SimpleDateFormat("h:mm a")
-        tf.setTimeZone(TimeZone.getTimeZone(loc.ianaTimeZone))
+        tf.setTimeZone(TimeZone.getTimeZone(loc?.ianaTimeZone))
```
tpmanley, line 265: If loc is `null` what will this code do? You added the null check but I'm wondering if we need to skip the whole line if loc is null.

tpmanley (approved):

greens (approved):
will go to staging today

PR #38946, opened 2020-07-30
HOTFIX: CHAD-5336 Make sure that temperature offset calculations for Zigbee devices are correct and rounded to one decimal

dkirker:
Fixes an issue where a NumberFormatException was being thrown when `zigbee.parseHATempatureValue` or `zigbee.parseSmartThingsTemperatureValue` were being used and a new temperature value with temperature offset was being calculated, as well as enforce single decimal precision for other Zigbee devices that were calculating a new temperature value with temperature offset.

### DIFF
--- a/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
+++ b/devicetypes/plaidsystems/spruce-sensor.src/spruce-sensor.groovy
@@ -258,9 +258,7 @@ private Map getTemperatureResult(value) {
 	def linkText = getLinkText(device)
         
 	if (tempOffset) {
-		def offset = tempOffset as int
-		def v = value as int
-		value = v + offset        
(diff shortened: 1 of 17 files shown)


dkirker:
I have tested this on a SmartSense Moisture Sensor using a self-published DTH. Looking to try with a SmartSense Multi (I believe that this is the first gen).

tpmanley:
Changes look good but let's get an extra review or two and I'm also doing some testing

dkirker:
It seems our platform is fine with `(float)String` and `(String as float)`, but a Groovy console I was also verifying with seemed to scream at `(float)String`, so I went with `(String as float)`.

tpmanley:
I verified the original bug and the fix on the SmartSense Multi

tpmanley (approved):

varzac (approved):

dkirker:
Awesome! I don't have merge access to prod, and I think that deploying hot fixes there require manual steps, so I think that will be on @greens .

greens (approved):

PR #39099, opened 2020-07-31
ONEAPP-31916 Change order of configure commands

greens:
Ensure that the enroll response is sent before any other configuration commands.


tpmanley:
We realized this change wasn't needed so going to close the PR

PR #39101, opened 2020-07-31
CHAD-5355 Zigbee lock event de-duplication

greens:
Uses logic similar to what we do in z-wave locks to delay events created from lock attribute reports in case we receive a much more detailed operation event later.

If the operation event is sent first, the lock attribute event will be marked as not displayed.

### DIFF
--- a/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
+++ b/devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
@@ -493,13 +493,10 @@ private def parseAttributeResponse(String description) {
 			responseMap.value = "unknown"
 			responseMap.descriptionText = "Unknown state"
 		}
-		if (responseMap.value &&
-				device.currentValue("lock") &&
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zigbee-lock.src/zigbee-lock.groovy
```diff
@@ -493,6 +493,14 @@ private def parseAttributeResponse(String description) {
 			responseMap.value = "unknown"
 			responseMap.descriptionText = "Unknown state"
 		}
+		if (responseMap.value) {
+			/*  delay this event for a second in the hopes that we get the operation event (which has more info).
+				If we don't get one, then it's okay to send. If we send the event with more info first, the event
+				with less info will be marked as not displayed
+			 */
+			runIn(1, "delayLockEvent", [data : [map : responseMap]])
```
tpmanley, line 502: Add a log message here so it's clear what's happening

tpmanley (dismissed):
I approve once the log message is added. Let's get this merged before the roll up today and I can help test it with my locks.

PR #39255, opened 2020-08-03
[ICP-13440] Qubino Flush Shutter - missing attribute in child method definition

PKacprowiczS:
@greens @SmartThingsCommunity/srpol-pe-team @BJanuszS


greens (dismissed):

greens (approved):
looks good but will still wait for feedback from testers

PR #40488, opened 2020-08-12
update parseAduroSmartButtonMessage function

adurosmart:
Some customers reported that the buttons 1 and 4 are not sensitive, so restore the zigbee.ONOFF_CLUSTER event trigger of button 1 and button 4, which will increase the success rate of event reporting after the button is pressed


greens (approved):

greens:
@adurosmart this will be live in production on the 25th

PR #40990, opened 2020-08-15
Aeon hem gen5 association fix and Configuration update

Aeotec-ccheng:
- Group Association fix present under "def Refresh()"

- def configuration() parameter settings updated for device setting optimization (kWh and Watt report every 300 seconds)

### DIFF
--- a/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
+++ b/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
@@ -17,7 +17,7 @@
  *  Date: 2013-05-30
  */
 metadata {
-	definition (name: "Aeon Home Energy Meter - code check", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.017.0012', executeCommandsLocally: false, ocfDeviceType: "x.com.st.d.energymeter") {
+	definition (name: "Aeon Home Energy Meter", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.017.0012', executeCommandsLocally: false, ocfDeviceType: "x.com.st.d.energymeter") {
 		capability "Energy Meter"
 		capability "Power Meter"
 		capability "Configuration"
--- a/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
+++ b/devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
@@ -17,7 +17,7 @@
  *  Date: 2013-05-30
  */
 metadata {
-	definition (name: "Aeon Home Energy Meter", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.017.0012', executeCommandsLocally: false, ocfDeviceType: "x.com.st.d.energymeter") {
+	definition (name: "Aeon Home Energy Meter - code check", namespace: "smartthings", author: "SmartThings", runLocally: true, minHubCoreVersion: '000.017.0012', executeCommandsLocally: false, ocfDeviceType: "x.com.st.d.energymeter") {
 		capability "Energy Meter"
 		capability "Power Meter"
 		capability "Configuration"
@@ -139,6 +139,8 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 def refresh() {
 	log.debug "refresh()..."
 	delayBetween([
+			encap(zwave.associationV2.associationRemove(groupingIdentifier: 1, nodeId:[])), // Refresh Node ID in Group 1
+   			encap(zwave.associationV2.associationSet(groupingIdentifier: 1, nodeId:zwaveHubNodeId)), //Assign Node ID of SmartThings to Group 1
 			encap(zwave.meterV2.meterGet(scale: 0)),
 			encap(zwave.meterV2.meterGet(scale: 2))
 	])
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -157,13 +159,12 @@ def configure() {
 	log.debug "configure()..."
 	if (isAeotecHomeEnergyMeter())
 		delayBetween([
-				encap(zwave.configurationV1.configurationSet(parameterNumber: 255, size: 4, scaledConfigurationValue: 1)), // Reset the device to the default settings
```
greens, line 160: I think we'd want to keep this, since it guarantees we get consistent settings on configure.

Unless you can confirm whether or not settings are preserved on removing and re-adding the device to a network.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -157,13 +159,12 @@ def configure() {
...
-				encap(zwave.configurationV1.configurationSet(parameterNumber: 101, size: 4, scaledConfigurationValue: 1)), // report power in Watts...
+				encap(zwave.configurationV1.configurationSet(parameterNumber: 101, size: 4, scaledConfigurationValue: 3)), // report total power in Watts and total energy in kWh...
```
greens, line 162: so this reports power and energy on one association group instead of two, right?

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
Aeotec-ccheng: All default parameter settings will revert back to their default values upon removing and re-adding the device to a network, there are no persistent parameter settings.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
Aeotec-ccheng: That is correct, but this isn't really based on Association Groups, both Parameter 101, 102, and 103 all send to Group Association Group #1 specifically.

Each group just allows you to send various sensors of the HEM Gen5 at different time intervals. (as Group1 interval, Group2 Interval, Group3 Interval)_

Since originally 101 was set to watt, and 102 was set to kWh, but both were adjusted to report at 300, this is essentially the same as setting both kWh + Watt to report on 101 only.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -139,6 +139,8 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 def refresh() {
 	log.debug "refresh()..."
 	delayBetween([
+			encap(zwave.associationV2.associationRemove(groupingIdentifier: 1, nodeId:[])), // Refresh Node ID in Group 1
+   			encap(zwave.associationV2.associationSet(groupingIdentifier: 1, nodeId:zwaveHubNodeId)), //Assign Node ID of SmartThings to Group 1
```
kianooshkarami, line 143: I am kind of curious as to reasons that prompted this change? The hub automatically does this upon join for the past 4-5 years that I can remember. Anything that you'd like the hub team to look into?

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
```diff
@@ -157,13 +159,12 @@ def configure() {
...
+				encap(zwave.configurationV1.configurationSet(parameterNumber: 103, size: 4, scaledConfigurationValue: 0)), // disable group 3...
 				encap(zwave.configurationV1.configurationSet(parameterNumber: 111, size: 4, scaledConfigurationValue: 300)), // ...every 5 min
-				encap(zwave.configurationV1.configurationSet(parameterNumber: 102, size: 4, scaledConfigurationValue: 2)), // report energy in kWh...
-				encap(zwave.configurationV1.configurationSet(parameterNumber: 112, size: 4, scaledConfigurationValue: 300)), // ...every 5 min
-				zwave.configurationV1.configurationSet(parameterNumber: 90, size: 1, scaledConfigurationValue: 1).format(), // enabling automatic reports...
-				zwave.configurationV1.configurationSet(parameterNumber: 91, size: 2, scaledConfigurationValue: 10).format() // ...every 10W change
+				encap(zwave.configurationV1.configurationSet(parameterNumber: 90, size: 1, scaledConfigurationValue: 0)), // enabling automatic reports, disabled selective reporting...
+				encap(zwave.configurationV1.configurationSet(parameterNumber: 13, size: 1, scaledConfigurationValue: 0)) //disable CRC16 encapsulation
```
kianooshkarami, line 167: Could we potentially leverage `CONFIGURATION_BULK_SET` command to generate one command instead of one per each configuration?

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
cbaumler: I'm also curious about this.

greens: my guess is that it's because it's not explicitly shown anywhere that we do this

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
greens: great, then this is fine.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
greens: I think maybe we'd want to do it ourselves first in another DTH before asking a third party to.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
Aeotec-ccheng: I assume that Association Group 1 is set to Node ID 01 automatically, but i have seen and gotten many reported cases where Group 1 is not successfully associated to Node ID 01 where the HEM Gen5 is not reporting automatically.

I have had to help large amount of supports where adjusting Group 01 to report to NodeID 01 using Z-Wave Tweaker custom device handler has allowed the HEM Gen5 to report normally,  which adding these 2 lines of code ensures that the HEM Gen5 is properly assigned without having to use a separate device handler.

---

I have noticed that Z-Wave controllers (not just SmartThings hub) that are using MULTICHANNEL ASSOCIATION to assign Group 1 to Node ID 1.1 which causes this issue, as Node ID 1.1 is not responded to by HEM Gen5 (HEM Gen5 uses an earlier form of Multichannel Association). For this to work, Node ID 1.0 for the use of this command class to work.

This code line is more of a bandage assuming that the association settings are automated and would reduce the amount of reported issues where the HEM Gen5 is not reporting without refreshing the page manually.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
Aeotec-ccheng: Unfortunately HEM Gen5 does not respond to CONFIGURATION BULK SET and BULK GET, please see PC Controller 5 logs below

Capture

**Notes -** The last 2 reports by HEM Gen5 are issued via CONFIGURATION GET 0x65 and CONFIGURATION GET 0x6F individually

----

Parameter 102 and 103 settings can be removed since it typically wouldn't really have any affect on the HEM Gen5 reports as a possible way to reduce configuration lines.

Aeotec-ccheng: Parameter 102 and 103 settings can be removed since it typically wouldn't really have any affect on the HEM Gen5 reports as a possible way to reduce configuration lines.

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
cbaumler: I don't think having this bandage in the DTH hurts anything, but I verified that the SmartThings hub adds node ID 1 / endpoint ID 0 to group 1 (lifeline) if the device supports Multichannel Association V3. So it should work. I don't have a device to verify with, but I'd be more than willing to investigate if you have customers running into this on our hub.

greens (approved):
lgtm, but I'll wait for one other sign-off before I merge

cbaumler (approved):

cjswedes (approved):

greens:
@Aeotec-ccheng this will be live in production on the 1st

devicetypes/smartthings/aeon-home-energy-meter.src/aeon-home-energy-meter.groovy
Aeotec-ccheng: @cbaumler - It should be working based on that information. If i do notice more of these in the future, i'll start debugging the backend to see if there are any problems via ZSniffer as well during its pair process.

PR #41224, opened 2020-08-17
CHAD-5290 Add DoorLockOperationReport delay for ALL z-wave locks

greens:
The same issue (less-useful DoorLockOperationReport arriving before
more-useful NotificationReport) was observed to occur for a user with a
Schlage lock, not just the Yales we'd seen it in before. As a result,
we're extending this workaround to ALL z-wave locks.


rboy1:
@greens there's an issue with code that we've run into in the past. If the user locks and unlocks the door quickly (or vice versa) due to this delayed runIn pending action it will end up reporting an incorrect state.

PR #41391, opened 2020-08-19
[WWST-6945] Development - Somfy Situo 4 Zigbee Pure

BJanuszS:
Handling the Somfy Situo 4 device has been added to the "ikea-button.groovy".

The device specific metada has been created by @KKlimczukS 

Only "UP" and "DOWN" button are supported.

@tpmanley Please, could You take a look at the code ?

### DIFF
--- a/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
+++ b/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
@@ -91,7 +91,7 @@ private getIkeaOnOffSwitchNames() {
 	]
 }
 
-private getIkeaOpenCloseOrSomfySituoRemoteNames() {
+private getOpenCloseRemoteNames() {
 	[
 		"Up", // Up button
 		"Down" // Down button
@@ -106,7 +106,7 @@ private getButtonLabel(buttonNum) {
 	} else if (isIkeaOnOffSwitch()) {
 		label = ikeaOnOffSwitchNames[buttonNum - 1]
 	} else if (isIkeaOpenCloseRemote() || isSomfySituo()) {
-		label = ikeaOpenCloseOrSomfySituoRemoteNames[buttonNum - 1]
+		label = openCloseRemoteNames[buttonNum - 1]
 	}
 
 	return label
--- a/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
+++ b/devicetypes/smartthings/ikea-button.src/ikea-button.groovy
@@ -31,7 +31,7 @@ metadata {
 		fingerprint inClusters: "0000, 0001, 0003, 0009, 0102, 1000, FC7C", outClusters: "0003, 0004, 0006, 0008, 0019, 0102, 1000", manufacturer:"IKEA of Sweden", model: "TRADFRI on/off switch", deviceJoinName: "IKEA Remote Control", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_On/Off_Switch" //IKEA TRÅDFRI On/Off switch
 		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI open/close remote", deviceJoinName: "IKEA Remote Control", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000 //IKEA TRÅDFRI Open/Close Remote
 		fingerprint manufacturer: "KE", model: "TRADFRI open/close remote", deviceJoinName: "IKEA Remote Control", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000 //IKEA TRÅDFRI Open/Close Remote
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -31,6 +31,7 @@ metadata {
...
 		fingerprint manufacturer: "IKEA of Sweden", model: "TRADFRI open/close remote", deviceJoinName: "IKEA Remote Control", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000 //IKEA TRÅDFRI Open/Close Remote
 		fingerprint manufacturer: "KE", model: "TRADFRI open/close remote", deviceJoinName: "IKEA Remote Control", mnmn: "SmartThings", vid: "SmartThings-smartthings-IKEA_TRADFRI_open/close_remote" // raw description 01 0104 0203 01 07 0000 0001 0003 0009 0020 1000 FC7C 07 0003 0004 0006 0008 0019 0102 1000 //IKEA TRÅDFRI Open/Close Remote
+		fingerprint profileId: "0104", inClusters: "0000, 0003", outClusters: "0003, 0005, 0006, 0102", manufacturer: "SOMFY", model: "Situo 4 Zigbee", deviceJoinName: "Situo 4 Zigbee" // raw description 01 0104 0203 00 02 0000 0003 04 0003 0005 0006 0102
```
KKlimczukS: I would remove profileId, inClusters and outClusters and change deviceJoinName to "SOMFY Remote Control"

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -187,7 +188,7 @@ def parse(String description) {
 	} else {
 		if ((description?.startsWith("catchall:")) || (description?.startsWith("read attr -"))) {
 			def descMap = zigbee.parseDescriptionAsMap(description)
-			if (descMap.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
+			if (!isSomfySituo() && descMap.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
```
KKlimczukS: I'm not sure if this check: !isSomfySituo() is necessary since You might not receive any report from POWER_CONFIGURATION_CLUSTER.

BJanuszS: Although the POWER_CONFIGURATION_CLUSTER is not supported it is read when configure() method is called. A response to that command is then received once but it has no value field specified in the descMap so that the NullPointerException occures in the next line: 

event = getBatteryEvent(zigbee.convertHexToInt(descMap.value))

I have added the !isSomfySituo() condition to avoid reaching that line.

KKlimczukS: What do You think about avoiding those calls in configure() method ? That way Somfy devices wouldn't receive those calls in first place and they wouldn't have to respond with no value.

BJanuszS: Ok, that is a better way to solve it. I will prepare the change.

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -90,7 +91,7 @@ private getIkeaOnOffSwitchNames() {
 	]
 }
 
-private getIkeaOpenCloseRemoteNames() {
+private getIkeaOpenCloseOrSomfySituoRemoteNames() {
```
dkirker: Maybe just rename these to `getOpenCloseRemoteNames`. The function name is getting long.

devicetypes/smartthings/ikea-button.src/ikea-button.groovy
```diff
@@ -171,10 +172,13 @@ def configure() {
 	log.debug "Configuring device ${device.getDataValue("model")}"
 
 	def cmds = zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x21, DataType.UINT8, 30, 21600, 0x01) +
-			zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x21) +
```
dkirker, line 174: I would do something like this:

```
def cmds = []

if (!isSomfySituo()) {
	cmds += zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x21, DataType.UINT8, 30, 21600, 0x01) +
		zigbee.readAttribute(zigbee.POWER_CONFIGURATION_CLUSTER, 0x21)
}

cmds += zigbee.addBinding(zigbee.ONOFF_CLUSTER) +
	readDeviceBindingTable() // Need to read the binding table to see what group it's using
```

No need to configure reporting on a cluster that isn't supported.

BJanuszS: When it comes to Somfy Situo something unclear happens. When the POWER_CONFIGURATION_CLUSTER is not configured then there are no reports coming from the ON_OFF_CLUSTER. 

We've supposed that it might be caused by the fact that the device needs to get any kind of configure command to start reporting and we've tried with adding zigbee.onOffConfig() but it did not help.

dkirker: @BJanuszS I would suggest reaching out to the manufacturer in this case.

KKlimczukS:
@greens @dkirker 
I created a new PR (with some other changes): https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/43628
I think we can close this one.

PR #42064, opened 2020-08-27
[WWST-5667] - Development - Somfy Sonesse 30

BJanuszS:
The latest Somfy motors work perfect after adding the fingerprint. 

In order to set the shade's limits or configure the motor (if it has not been done by the manufacturer) the Somfy Set&Go mobile app is required. 

It must be noted that with the Set&Go app the user can change the default model name ("Roller") so that it will no longer match with the fingerprint.

@KKlimczukS @PKacprowiczS @ZWozniakS @MGoralczykS @MWierzbinskaS  

@tpmanley could please take a look at the code?


greens (approved):

greens (approved):

dkirker:
I would have recommended merging the two `isSomfyGlydea ` and `isSomfySonesse ` commands to just `isSomfy` that looked at `device.getDataValue("manufacturer") == "SOMFY"`, which would cover the case of the user changing the model in the Set&Go app. I'd still at the very least do this. (I'm guessing this is how the Set&Go app is implementing renaming the device.)

So, regarding the issue about being able to change the model in the Set&Go app and possibly having the device not pair in SmartThings, I suspect you could have a fingerprint that had the clusters and just the manufacturer with the model removed? `fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0020, 0102", outClusters: "0003", manufacturer: "SOMFY"` (I am pretty sure that should work, right @tpmanley ?) I don't see any other fingerprints that could conflict with.

greens:
Yeah. I think that is actually a good idea. @BJanuszS can you open another PR to implement Donald's suggestion?

PR #42729, opened 2020-09-04
[ICP-13360] - Qubino DIN Dimmer - fixed missing energy meter reports

MWierzbinskaS:
@KKlimczukS @PKacprowiczS @MGoralczykS @ZWozniakS 

@greens Could you please take a look at the code?

### DIFF
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -332,6 +332,10 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 
 def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd, ep = null) {
 	log.debug "BasicReport: ${cmd}"
+	if(isDINDimmer()) {
+		sendHubCommand(encap(zwave.meterV2.meterGet(scale: 0)))
+		sendHubCommand(encap(zwave.meterV2.meterGet(scale: 2)))
+	}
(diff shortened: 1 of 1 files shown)


greens:
For completeness' sake, I think you should add this query to any `zwaveParse` that might result in the switch changing state.

Also, is there any configuration option to let us change the reporting threshold?

greens:
And the bug is that energy consumption doesn't change. It seems like power is reporting fine?

MWierzbinskaS:
Power meter is reporting fine. There aren't any parameters for energy meter configuration. The problem is that when the device is left unattented for energy meter test, it does not send energy meter report even if the value is changed. We have to ask it about energy meter report. This device sends BasicReport every 15 minutes, so function getRefreshCommands() will not be called in this situation. This is the reason why I decided to add asking for meter reports in BasicReport handling.

greens:
@MWierzbinskaS please see my comments on the Jira ticket. I'm not sure this is what we need to do.

MWierzbinskaS:
@greens I will respond here to have this conversation in one place. When first test was performed the issue was that device was tested with not enough load. Then we contacted Partner, and he responded that it should be at least 0,1 kWh. When retest was performed with correct load issue still existed. Device did not send energy meter report automatically when it was left unattended and without any action performed - we had to ask this device about it and then it responded with new load value. The reason why I added asking about energy meter in BasicReport handling I described in comment above.

greens:
@MWierzbinskaS if power reporting works fine then we don't need to query it.

greens (approved):

PR #42736, opened 2020-09-04
DevWs for HAB Home Intelligence containing containing iblinds v3

habhomegit:


devicetypes/iblinds/iblinds-zwave-v3.src/iblinds-zwave-v3.groovy
```diff
@@ -0,0 +1,358 @@
...
+ *
+ *  2020-08-31: Force Lifeline association. 
+ *				Added Schedule for daily battery check (HealthCheck /Check Inverval not seeming to fire correctly)
+ *  2020-09-04: Fixed default ON value and preset 
+ */
+import groovy.json.JsonOutput
+
+
+metadata {
+	definition (name: "iblinds Z-Wave v3", namespace: "iblinds", author: "HABHomeIntel", ocfDeviceType: "oic.d.blind",  mnmn: "SmartThings", vid: "generic-shade-3") {
```
KKlimczukS: Since that's a v3 version of previous "iblinds Z-wave" DTH and both of them support this device:
fingerprint mfr:"0287", prod:"0004", model:"0071", deviceJoinName: "iBlinds Window Treatment", 

I suggest to consider refactoring the existing DTH (iblinds-zwave.groovy) instead of creating a new one.

habhomegit: Hi, thanks for the reply.  The v3 fingerprint will need to be removed from the current v2 DTH.   v3 supports preferences (Z-Wave Configuration Parameters) that v2 will not support.  So, I believe they will need to be separate DTHs.   I put in a request to have the v3 fingerprint removed from the existing v2 device handler but I have not received a response.  

Please advise.

KKlimczukS: @habhomegit I mentioned this to @dkirker , because a migration of currently working devices (to the new DTH) might have to be done. Right after we have the confirmation we will continue the code review.

devicetypes/iblinds/iblinds-zwave.src/iblinds-zwave.groovy
```diff
@@ -16,14 +16,13 @@ import groovy.json.JsonOutput
 
 metadata {
 	definition (name: "iblinds Z-Wave", namespace: "iblinds", author: "HABHomeIntel", ocfDeviceType: "oic.d.blind",  mnmn: "SmartThings", vid: "generic-shade-3") {
+		capability "Window Shade"
+		capability "Window Shade Preset"
 		capability "Switch Level"
-		capability "Switch"
```
dkirker, line 20: I removed this since this DTH isn't actually using `Switch`, and that isn't recommended for window covering devices any more.

dkirker (dismissed):

dkirker (dismissed):

dkirker (dismissed):

dkirker (dismissed):

dkirker (approved):

devicetypes/iblinds/iblinds-zwave.src/iblinds-zwave.groovy
```diff
@@ -239,3 +279,79 @@ def refresh() {
...
+		4					1			NVM_Target_Value				Default on position
+		5					1			NVM_Device_Reset_Support		Turns off the reset button
+		6					1			Speed_Parameter				Speed
+		*/
+
+		log.debug "Configuration Started"
+
+		// If paramater value has changed then add zwave configration command to cmds
+
+		if (NVM_TightLevel != null && state.param1 != NVM_TightLevel) {
```
dkirker, line 301: @habhomegit I went a little crazy with adding null checking (some attempts to make calls to `toInteger()` on a null object showed up in a search I was doing), but it should be more safe. I guess technically if a user doesn't configure a preference then state.paramNo would be null and the preference would be null (and null != null would be false). But... Weird things happen.

habhomegit:
@dkirker Hi Donald,  we have a new model that we would like to have added to our iblinds device handler.   
Here is the new fingerprint. 

fingerprint mfr:"0287", prod:"0004", model:"0072", deviceJoinName: "iBlinds Window Treatment" 

Please let me know what steps we need to take.  

Thanks,
Eric

PR #43964, opened 2020-09-15
[ICP-13278] Changed minimum dimmer level range in settings for Aeotec Nano Dimmer

ZWozniakS:
@greens @dkirker could You please look at the code? Thanks!

Change is applied only to the parameter 131 - minimum dimmer range possible to choose.


greens (approved):
interesting fix, but I thought this was getting fixed with metadata?

dkirker:
I believe this is for the device configuration (the one where you can tell the device what range to support). Since the local execution code isn't supporting 0 as a valid level to pass in setLevel then it would make sense to not allow that as a valid value in the range to support.

dkirker (approved):

greens:
but you're still able to select 0 in the UI, right?

ZWozniakS:
@greens the case with UI is already taken care of by @dkirker here: [ICP-12327](https://smartthings.atlassian.net/browse/ICP-12327)

PR #44143, opened 2020-09-17
update smartsense-motion-sensor.groovy

adurosmart:
update smartsense-motion-sensor.groovy,Optimize that deviceJoinName，change the vid to generic-motion-2.


greens (approved):

greens:
@adurosmart this will be live in production on the 29th

tz0zk6:
Any chance we could add to configuration of this device to include setting the Reportable Range parameter of the temperature cluster?   We have this for the Multi-Sensor as zigbee.configureReporting(0xFC02, 0x0010, DataType.BITMAP8, 0, 3600, 0x01, [mfgCode: manufacturerCode]) 

Many are not overly interested in Temperature logs on motion sensors and the do flood the history of Smarthings with the default 0.1deg and 30sec/5min reporting cycle.

PR #44306, opened 2020-09-18
[CHAD-5501][zigbee multi switch] Need Exception handling for eZEX multi switch

inasail:
https://smartthings.atlassian.net/browse/CHAD-5501

I don't have the eZEX sensor, so I am not sure this change will fix this issue perfectly.

eZEX sensor send undefined packet(refer to below), so need to add exception for filtering of undefined attributeId.
(ZCL Rx: ep:01 prof:0104 clus:0006 lqi:C4 rssi:-51 fc:08 mc:0000 seq:34 cmd:0A body:**11 00 18 07**)

### DIFF
--- a/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
+++ b/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
@@ -107,6 +107,20 @@ def parse(String description) {
 	Map eventMap = zigbee.getEvent(description)
 	Map eventDescMap = zigbee.parseDescriptionAsMap(description)
 
+	if ( device.getDataValue("deviceJoinName") == "eZEX Switch 1") {
+		if (eventMap && eventMap?.name == "switch" && (eventMap?.value == "on" || eventMap?.value == "off")) {
+			if (eventDescMap && eventDescMap?.attrId != "0000") {
+				return [:]
+			}
+		}
+
+		if (!eventMap && eventDescMap) {
+			if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER && eventDescMap?.attrId != "0000") {
+				return [:]
+			}
+		}
+	}
+
 	if (!eventMap && eventDescMap) {
 		eventMap = [:]
 		if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER) {
--- a/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
+++ b/devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
@@ -107,39 +107,19 @@ def parse(String description) {
 	Map eventMap = zigbee.getEvent(description)
 	Map eventDescMap = zigbee.parseDescriptionAsMap(description)
 
-	if ( device.getDataValue("deviceJoinName") == "eZEX Switch 1") {
-		if (eventMap && eventMap?.name == "switch" && (eventMap?.value == "on" || eventMap?.value == "off")) {
-			if (eventDescMap && eventDescMap?.attrId != "0000") {
-				return [:]
-			}
-		}
-
-		if (!eventMap && eventDescMap) {
-			if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER && eventDescMap?.attrId != "0000") {
-				return [:]
-			}
-		}
-	}
-
-	if (!eventMap && eventDescMap) {
-		eventMap = [:]
-		if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER) {
-			eventMap[name] = "switch"
-			eventMap[value] = eventDescMap?.value
-		}
-	}
-
 	if (eventMap) {
-		if (eventDescMap?.sourceEndpoint == "01" || eventDescMap?.endpoint == "01") {
-			sendEvent(eventMap)
-		} else {
-			def childDevice = childDevices.find {
-				it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.sourceEndpoint}" || it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.endpoint}"
-			}
-			if (childDevice) {
-				childDevice.sendEvent(eventMap)
+		if (eventDescMap && eventDescMap?.attrId == "0000") {//0x0000 : OnOff attributeId
+			if (eventDescMap?.sourceEndpoint == "01" || eventDescMap?.endpoint == "01") {
+				sendEvent(eventMap)
 			} else {
-				log.debug "Child device: $device.deviceNetworkId:${eventDescMap.sourceEndpoint} was not found"
+				def childDevice = childDevices.find {
+					it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.sourceEndpoint}" || it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.endpoint}"
+				}
+				if (childDevice) {
+					childDevice.sendEvent(eventMap)
+				} else {
+					log.debug "Child device: $device.deviceNetworkId:${eventDescMap.sourceEndpoint} was not found"
+				}
 			}
 		}
 	}


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -107,6 +107,20 @@ def parse(String description) {
 	Map eventMap = zigbee.getEvent(description)
 	Map eventDescMap = zigbee.parseDescriptionAsMap(description)
 
+	if ( device.getDataValue("deviceJoinName") == "eZEX Switch 1") {
```
greens: @inasail can you verify that this value does not change if you rename the device?

normally we would do this using a manufacturer check, but eZex did not include that string for whatever reason...

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -107,6 +107,20 @@ def parse(String description) {
...
+		if (eventMap && eventMap?.name == "switch" && (eventMap?.value == "on" || eventMap?.value == "off")) {
```
greens: You shouldn't have to check for `on` or `off`. You won't get a switch event from `zigbee.getEvent` without either of those values.

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -107,6 +107,20 @@ def parse(String description) {
...
+			if (eventDescMap && eventDescMap?.attrId != "0000") {
+				return [:]
+			}
+		}
+
+		if (!eventMap && eventDescMap) {
+			if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER && eventDescMap?.attrId != "0000") {
+				return [:]
+			}
+		}
```
greens: I believe the code below:

```
	if (!eventMap && eventDescMap) {
		eventMap = [:]
		if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER) {
			eventMap[name] = "switch"
			eventMap[value] = eventDescMap?.value
		}
	}
```

is probably the source of the issue in the first place. You may want to refactor that instead of adding these checks here.

greens: @KKlimczukS do you know why you added this code for the Aurora double switch? Do you remember what message was being sent?

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
inasail: @greens, I tested with other sensors like moisture, outlet sensor, but in these both cases, the deviceJoinName is returned with 'null' string(even it is defined in the fingerprint), so I can't trust this value in this DTH. I used this condition not to effect other sensors in this DTH. but I think it is okay to adjust as common logic. So I will delete this checking condition.

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
inasail: Okay, I will delete this unnecessary condition.

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
inasail: Okay, then, How about this? (This is all what I want to change including refactoring)
```suggestion
	if (eventMap) {
		if (eventDescMap) {
			if (eventDescMap?.attrId != "0000") {
				eventMap = [:]
			}
		}
	} else {
		if (eventDescMap) {
			eventMap = [:]
			if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER && eventDescMap?.attrId == "0000") {
				eventMap[name] = "switch"
				eventMap[value] = eventDescMap?.value
			}
		}
	}
```

KKlimczukS: > @KKlimczukS do you know why you added this code for the Aurora double switch? Do you remember what message was being sent?

@greens That must have been quite a long time ago and that's was probably because zigbee.getEvent(description) didn't create the eventMap in a first place.

greens: I think you should just be able to remove this code

```
	if (!eventMap && eventDescMap) {
		eventMap = [:]
		if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER) {
			eventMap[name] = "switch"
			eventMap[value] = eventDescMap?.value
		}
	}
```

since the device it was added for is no longer on this DTH. Try that and see if it solves the issue.

devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -107,25 +107,19 @@ def parse(String description) {
...
 	if (eventMap) {
-		if (eventDescMap?.sourceEndpoint == "01" || eventDescMap?.endpoint == "01") {
-			sendEvent(eventMap)
-		} else {
-			def childDevice = childDevices.find {
-				it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.sourceEndpoint}" || it.deviceNetworkId == "$device.deviceNetworkId:${eventDescMap.endpoint}"
-			}
-			if (childDevice) {
-				childDevice.sendEvent(eventMap)
+		if (eventDescMap && eventDescMap?.attrId == "0000") {//0x0000 : OnOff attributeId
```
greens, line 111: were you having problems if you didn't include this new condition?

inasail: https://smartthings.atlassian.net/browse/CHAD-5501
eZEX sensor send undefined packet(refer to below), so need to add exception for filtering of undefined attributeId.
(ZCL Rx: ep:01 prof:0104 clus:0006 lqi:C4 rssi:-51 fc:08 mc:0000 seq:34 cmd:0A body:11 00 18 07)


@greens, I don't have eZEX switch sensor, so I can't check it doesn't have problem without this condition. 
But logically, As I said in decription, eZEX sensor send packet to hub with undefined attributeId(like 0x0011)
and we don't check the validity of attributeId in appengine-zigbee(only check the clusterId), 
so it is recoginized as and 'off' value (even its real value is '07')
https://github.com/PhysicalGraph/appengine-zigbee/blob/master/src/main/groovy/physicalgraph/zigbee/Zigbee.groovy#L832
So I added this condition to activate with only 0x0000(OnOff) attribute.

greens: Right, but that event was being created by the code I've asked you to remove. `zigbee.parse()` most likely will not create that event, so just removing that extra code should solve that problem.

iquix: @greens @inasail
Hi I'm a Smartthings customer in South Korea, and I would like to add comment about it.

1.
Similar error also seems to occur in not only in Ezex, but also in Dawon multi switches.
Below is a thread in Korean Smartthings Community.
https://cafe.naver.com/stsmarthome/18408 (Sorry, It's in Korean, but @inasail, you could read it.)

2.
If you decide to modify source code in either way, you may need to make sure that the change does not cause problem with zigbee-multi-switches from other manufacturers.

There must be some reason that "if (eventDescMap?.clusterId == zigbee.ONOFF_CLUSTER)" part of the code is there. Maybe, some products from other manufacturer (maybe, one of the products from Orvibo, GDKES, HONTAR, HEIMAN, Dawon and eWeLink) send faulty ZHA packets with only clusterId OnOff without attrId or wrong attrId, and that might be the reason of that part of the quirky code.

You may need to check out the history of the zigbee-multi-switch DTH, and look for previous pull request of that part of the code.

Thank you guys for all your efforts.

varzac: > Right, but that event was being created by the code I've asked you to remove. zigbee.parse() most likely will not create that event, so just removing that extra code should solve that problem.

So I don't believe this is correct @greens . The `getEvent` call will generate an incorrect event for this message because we don't verify the attribute ID there.  So we should also make a change https://github.com/PhysicalGraph/appengine-zigbee/blob/master/src/main/groovy/physicalgraph/zigbee/Zigbee.groovy#L832 to check the attribute ID there.

greens (approved):
thanks for checking

PR #44694, opened 2020-09-20
Aeotec

SmartThings:
4 button switch


tpmanley:
The HKZW-SCN04 is a Z-Wave device so the fingerprint should be added to one of our Z-Wave button DTHs instead of a Zigbee one. I'm not personally familiar with this device but this site links to a custom DTH for it https://www.thesmartesthouse.com/products/hank-z-wave-plus-four-button-scene-controller-hkzw-scn04.

PR #44767, opened 2020-09-21
[ICP-13108] Delay off event, add chime when tamper is triggered

MGoralczykS:
Could you check it?

### DIFF
--- a/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
+++ b/devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
@@ -206,7 +206,8 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 			case 0x09: //TAMPER
 				sendEvent(name: "tamper", value: "detected")
 				sendEvent(name: "alarm", value: "both")
-				runIn(2, "clearTamperAndAlarm")
+				sendEvent(name: "chime", value: "chime")
+				runIn(2, "clearTamperAlarmAndChime")
 				break
 			case 0x01: //ON
 				if (state.lastTriggeredSound == 1) {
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/aeotec-doorbell-siren-6.src/aeotec-doorbell-siren-6.groovy
```diff
@@ -206,7 +206,8 @@ def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cm
 			case 0x09: //TAMPER
 				sendEvent(name: "tamper", value: "detected")
 				sendEvent(name: "alarm", value: "both")
-				runIn(2, "clearTamperAndAlarm")
+				sendEvent(name: "chime", value: "chime")
```
dkirker, line 209: Do we actually want to send `chime` here? It should be safe to still clear it with `clearTamperAlarmAndChime`.

greens: I assume we want to send chime wherever we're sending alarm?

MGoralczykS: @greens Yes, in my opinion, this is correct.

dkirker: But a chime isn't an alarm. That said, the device isn't actually integrated in SmartThings as it was designed, and if I recall one of the endpoints is associated with tamper (I think the first one, actually). So I guess for this integration chime == alarm == tamper, which makes no sense at all.

dkirker:
So does the device send us "off" before it is even finished? Is it likely that this is happening because the device integration is actually misusing the endpoints and the endpoint that is sending is the really quick off is something more short term like the doorbell noise instead of an alarm?

greens:
So, @dkirker this was related to an earlier ticket that was raised where when the device is tampered, the alarms will go off, but to the device this is a different alarm and so it does not signal that the alarms are on. Cert did not like that you could have a situation where the alarms were going off without the state being changed.

So we added this event which automatically clears itself to encapsulate the behavior when the alarms are triggered by tamper. Physically, when tampered, the device will sound its alarms for 2s and then turn them off. No indication is otherwise given to us beyond the tamper notification.

k-morawski:
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/49092

PR #3682, opened 2018-11-15
WWST-1788 Update Keen Home Smart Vent DTH

greens:

PR #45127, opened 2020-09-23
CHAD-5488 Update Keen Home Smart Vent

greens:
includes a workaround for a bug introduced in newer vent FWs

adds the draft atmospheric pressure measurement capability instead of
custom attributes

if the vent detects it is obstructed, it will attempt to clear its
obstruction automatically

### DIFF
--- a/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
+++ b/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
@@ -118,7 +118,7 @@ def getBatteryPercentageResult(rawValue) {
 }
 
 def getPressureResult(rawValue) {
-    def kpa = rawValue / (10 * 1000) // reports are in deciPascals, I think
+    def kpa = rawValue / (10 * 1000) // reports are in deciPascals
     return [name: "atmosphericPressure", value: kpa, unit: "kPa"]
 }
 
@@ -133,7 +133,7 @@ def on() {
     if (currentLevel != null) {
         currentLevel = currentLevel as int
     }
-    def levelToSet = currentLevel ? curentLevel : 100
+    def levelToSet = currentLevel ? currentLevel : 100
     cmds << zigbee.setLevel(levelToSet)
 }
 
@@ -209,8 +209,8 @@ private boolean isObstructed() {
     def currentState = device.currentValue("switch")
 
     if (currentState == "obstructed") {
-        sendEvent(name: "level", value: 0) // convert legacy implementation
-        state.obstructed == true
+        sendEvent(name: "switch", value: "off", displayed: false) // convert legacy implementation
+        state.obstructed = true
     }
     return state.obstructed
 }
--- a/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
+++ b/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
@@ -86,8 +86,9 @@ def parse(String description) {
         } else if (descMap?.clusterInt == PRESSURE_MANAGEMENT_CLUSTER && descMap.attrInt == 0x0020) {
             event = getPressureResult(Integer.parseInt(descMap.value, 16))
         }
-    } else if (event.name == "level" && event.value > 100) {
-        state.obstructed = true
+    } else if (event.name == "level" && event.value == 100) {
+        def descMap = zigbee.parseDescriptionAsMap(description)
+        if (Integer.parseInt(descMap.value.trim(), 16) == 255) state.obstructed = true
     } else if (event.name == "level") {
         state.obstructed = false
         if (event.value > 0 && device.currentValue("switch") == "off") {
--- a/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
+++ b/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
@@ -35,8 +35,6 @@ metadata {
         standardTile("switch", "device.switch", width: 2, height: 2, canChangeIcon: true) {
             state "on", action: "switch.off", icon: "st.vents.vent-open-text", backgroundColor: "#00a0dc"
             state "off", action: "switch.on", icon: "st.vents.vent-closed", backgroundColor: "#ffffff"
-            state "obstructed", action: "clearObstruction", icon: "st.vents.vent-closed", backgroundColor: "#e86d13"
-            state "clearing", action: "", icon: "st.vents.vent-closed", backgroundColor: "#ffffff"
         }
         controlTile("levelSliderControl", "device.level", "slider", height: 1, width: 2, inactiveLabel: false) {
             state "level", action:"switch level.setLevel"
@@ -86,14 +84,8 @@ def parse(String description) {
         } else if (descMap?.clusterInt == PRESSURE_MEASUREMENT_CLUSTER && descMap.attrInt == 0x0020) {
             // manufacturer-specific attribute
             event = getPressureResult(Integer.parseInt(descMap.value, 16))
-        } else if (descMap?.clusterInt == zigbee.LEVEL_CONTROL_CLUSTER && descMap.attrInt == 0x0000) {
-            if (Integer.parseInt(descMap.value, 16) == 255) {
-                log.debug("Obstruction detected")
-                state.obstructed = true
-            }
         }
     } else if (event.name == "level") {
-        state.obstructed = false
         if (event.value > 0 && device.currentValue("switch") == "off") {
             sendEvent([name: "switch", value: "on"])
         }
@@ -125,10 +117,6 @@ def getPressureResult(rawValue) {
 /**** COMMAND METHODS ****/
 def on() {
     def cmds = []
-    if (isObstructed()) {
-        cmds << clearObstruction()
-        cmds << "delay 2000"
-    }
     def currentLevel = device.currentValue("level")
     if (currentLevel != null) {
         currentLevel = currentLevel as int
@@ -138,33 +126,12 @@ def on() {
 }
 
 def off() {
-    def cmds = []
-    if (isObstructed()) {
-        cmds << clearObstruction()
-        cmds << "delay 2000"
-    }
-    cmds << zigbee.off()
-}
-
-def clearObstruction() {
-    log.debug "attempting to clear ${device.displayName} obstruction"
-
-    // send a move command to ensure level attribute gets reset for old, buggy firmware
-    // then send a reset to factory defaults
-    // finally re-configure to ensure reports and binding is still properly set after the rtfd
-    [
-            zigbee.setLevel(device.currentValue("level")), "delay 500",
-            zigbee.command(zigbee.BASIC_CLUSTER, 0x00), "delay 5000"
-    ] + configure()
+    zigbee.off()
 }
 
 def setLevel(value, rate = null) {
     log.debug "setting level: ${value}"
     def cmds = []
-    if (isObstructed()) {
-        cmds << clearObstruction()
-        cmds << "delay 2000"
-    }
     cmds << zigbee.setLevel(value)
     cmds << "delay 1000"
     cmds << zigbee.levelRefresh()
@@ -201,16 +168,5 @@ def configure() {
             zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0021, DataType.UINT8, 600, 21600, 0x01) // battery precentage
     ]
 
-    return delayBetween(cmds) + refresh()
-}
-
-
-private boolean isObstructed() {
-    def currentState = device.currentValue("switch")
-
-    if (currentState == "obstructed") {
-        sendEvent(name: "switch", value: "off", displayed: false) // convert legacy implementation
-        state.obstructed = true
-    }
-    return state.obstructed
+    return refresh() + delayBetween(cmds)
 }
--- a/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
+++ b/devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
@@ -83,12 +83,12 @@ def parse(String description) {
         Map descMap = zigbee.parseDescriptionAsMap(description)
         if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
             event = getBatteryPercentageResult(Integer.parseInt(descMap.value, 16))
-        } else if (descMap?.clusterInt == PRESSURE_MANAGEMENT_CLUSTER && descMap.attrInt == 0x0020) {
+        } else if (descMap?.clusterInt == PRESSURE_MEASUREMENT_CLUSTER && descMap.attrInt == 0x0020) {
+            // manufacturer-specific attribute
             event = getPressureResult(Integer.parseInt(descMap.value, 16))
+        } else if (descMap?.clusterInt == zigbee.LEVEL_CONTROL_CLUSTER && descMap.attrInt == 0x0000) {
+            if (Integer.parseInt(descMap.value, 16) == 255) state.obstructed = true
         }
-    } else if (event.name == "level" && event.value == 100) {
-        def descMap = zigbee.parseDescriptionAsMap(description)
-        if (Integer.parseInt(descMap.value.trim(), 16) == 255) state.obstructed = true
     } else if (event.name == "level") {
         state.obstructed = false
         if (event.value > 0 && device.currentValue("switch") == "off") {
@@ -115,8 +115,8 @@ def getBatteryPercentageResult(rawValue) {
 }
 
 def getPressureResult(rawValue) {
-    def kpa = rawvalue / (10 * 1000) // reports are in deciPascals, I think
-    return [name: "atmosphericPressure", value: kpa, units: "kPa"]
+    def kpa = rawValue / (10 * 1000) // reports are in deciPascals, I think
+    return [name: "atmosphericPressure", value: kpa, unit: "kPa"]
 }
 
 /**** COMMAND METHODS ****/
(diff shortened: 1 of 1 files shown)


greens:
Relocated from here https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/3682 because the commit wrangling was far more onerous than I wanted to deal with.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,136 @@ metadata {
...
-                break
+    def event = zigbee.getEvent(description)
+    if (!event) {
+        Map descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
+            event = getBatteryPercentageResult(Integer.parseInt(descMap.value, 16))
+        } else if (descMap?.clusterInt == PRESSURE_MANAGEMENT_CLUSTER && descMap.attrInt == 0x0020) {
+            event = getPressureResult(Integer.parseInt(descMap.value, 16))
+        }
+    } else if (event.name == "level" && event.value > 100) {
```
varzac: Is this possible? I don't think `getEvent` will return a level event with a value > 100

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -477,47 +185,25 @@ def configure() {
...
-        // "zcl global send-me-a-report 0x0403 0x20 0x22 60 60 {010000}", "delay 200",
-        // "send 0x${device.deviceNetworkId} 1 1", "delay 1500",
 
-        // report with these parameters is preconfigured in firmware, can be overridden here
-        // battery - type: int8u, change: 1
-        // "zcl global send-me-a-report 1 0x21 0x20 60 3600 {01}", "delay 200",
-        // "send 0x${device.deviceNetworkId} 1 1", "delay 1500",
-    ]
+private boolean isObstructed() {
+    def currentState = device.currentValue("switch")
```
varzac: Where does this get set to `obstructed`?

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
greens: Looking at the code in `getEvent` it doesn't look like there's any clamping.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
greens: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/45127/files#diff-bea535a5e9f948f1d19a1a80bb0be167L215

varzac: Ah so this is just a workaround for existing values that are out there, it won't happen anymore?

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
varzac: https://github.com/PhysicalGraph/appengine-zigbee/blob/master/src/main/groovy/physicalgraph/zigbee/Zigbee.groovy#L836 is what I'm looking at.  Since level is one byte 255 is the max it could be and that should be rounded to 100, but maybe I'm missing something.

greens: Ah, yeah, it'll just round to 100. Okay I'll need to work on this.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
greens: Right.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,137 @@ metadata {
...
+    if (!event) {
+        Map descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
+            event = getBatteryPercentageResult(Integer.parseInt(descMap.value, 16))
+        } else if (descMap?.clusterInt == PRESSURE_MANAGEMENT_CLUSTER && descMap.attrInt == 0x0020) {
+            event = getPressureResult(Integer.parseInt(descMap.value, 16))
+        }
+    } else if (event.name == "level" && event.value == 100) {
+        def descMap = zigbee.parseDescriptionAsMap(description)
+        if (Integer.parseInt(descMap.value.trim(), 16) == 255) state.obstructed = true
```
greens: @varzac I think this should work

varzac: So I think you will need to do something like https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy#L170 as `0xFF` is invalid for a uint8 and I think it should be filtered by https://github.com/PhysicalGraph/appengine-zigbee/blob/master/src/main/groovy/physicalgraph/zigbee/Zigbee.groovy#L810 

Sorry I should have realized this earlier.

greens: bothersome

damn our standards-compliant parsing

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,137 @@ metadata {
...
-
-            case 0x0006:
-                return makeOnOffResult(cluster.data[-1])
-                break
+    def event = zigbee.getEvent(description)
+    if (!event) {
+        Map descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
+            event = getBatteryPercentageResult(Integer.parseInt(descMap.value, 16))
+        } else if (descMap?.clusterInt == PRESSURE_MANAGEMENT_CLUSTER && descMap.attrInt == 0x0020) {
```
tpmanley: Attribute ID 0x0020 isn't defined in ZCL8. Is this a manufacturer specific attribute?

tpmanley: Nevermind, I see that it is. Probably worth adding a comment that this is a manufacturer specific cluster.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,137 @@ metadata {
...
-    def rangeMax = 254
-    def scaledLevel = Math.round(level * rangeMax / 100)
-    log.debug "scaled level for ${level}%: ${scaledLevel}"
-
-    // convert to hex string and pad to two digits
-    def hexLevel = new BigInteger(scaledLevel.toString()).toString(16).padLeft(2, '0')
-
-    "st cmd 0x${device.deviceNetworkId} 1 8 4 {${hexLevel} 0000}"
+def getPressureResult(rawValue) {
+    def kpa = rawvalue / (10 * 1000) // reports are in deciPascals, I think
```
tpmanley: deciPascals is correct I believe. I compared the value reported by this attribute to the value reported by the standard MeasuredValue attribute:

0x0020: 979339
0x0000: 979

The definition of attribute 0x0000 is the pressure in kPa times 10

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,137 @@ metadata {
...
+    return [name: "atmosphericPressure", value: kpa, units: "kPa"]
```
tpmanley: `units` should be `unit`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -477,47 +186,25 @@ def configure() {
...
+                    zigbee.addBinding(zigbee.LEVEL_CONTROL_CLUSTER) +
+                    zigbee.configureReporting(PRESSURE_MEASUREMENT_CLUSTER, 0x0020, DataType.UINT32, 600, 21600, 0x010000, [mfgCode: MFG_CODE]) +
+                    zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0021, DataType.UINT8, 600, 21600, 0x01) // battery precentage
+    ]
 
-        // report with these parameters is preconfigured in firmware, can be overridden here
-        // temperature - type: int16s, change: 0xA = 10 = 0.1C
-        // "zcl global send-me-a-report 0x0402 0 0x29 60 60 {0A00}", "delay 200",
-        // "send 0x${device.deviceNetworkId} 1 1", "delay 1500",
+    return delayBetween(cmds) + refresh()
```
tpmanley: `delayBetween` should not be necessary. There are already delays added by the `zigbee` library. Also the `refresh` commands should be sent first so the current state is reflected more quickly.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
tpmanley: Should be `PRESSURE_MEASUREMENT_CLUSTER`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,137 @@ metadata {
...
 def refresh() {
-    getOnOff() +
-    getLevel() +
-    getTemperature() +
-    getPressure() +
-    getBattery()
+    zigbee.onOffRefresh() +
+            zigbee.levelRefresh() +
+            zigbee.readAttribute(zigbee.TEMPERATURE_MEASUREMENT_CLUSTER, 0x0000) +
+            zigbee.readAttribute(PRESSURE_MANAGEMENT_CLUSTER, 0x0020, [mfgCode: MFG_CODE]) +
```
tpmanley: Should be `PRESSURE_MEASUREMENT_CLUSTER`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
tpmanley: Should be `rawValue`

tpmanley:
I made fxes I suggested to my local copy and the normal functionality is all working. The problem I noticed with the copy we have published is the level shown in the app was always 0%. On/Off controls and display worked correctly, and temperature/battery were displayed correct. SetLevel worked but it was just the current status that was wrong.

Do you plan to add Atmospheric Pressure to the UI Metadata?

greens:
@tpmanley thanks for all those catches.

I'm investigating the atmospheric pressure, but it's only a proposed capability right now so I don't know if there are widgets for it.

greens:
Have verified that the new implementation for determining an obstruction works.

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -194,7 +194,7 @@ def configure() {
                     zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0021, DataType.UINT8, 600, 21600, 0x01) // battery precentage
     ]
 
-    return delayBetween(cmds) + refresh()
+    return refresh + cmds
```
tpmanley: Should be `refresh()`

tpmanley:
How is the obstruction detection supposed to work? I tried to actuate the vent while I prevented it from moving and now the red LED is blinking and all the commands I send from the app result in errors.

greens:
@tpmanley Based on what I read in the commented code it factory resets and then reconfigures the device. I may need to add a delay onto the end.

The thing I changed was that if an obstruction is detected, we attempt to clear it the next time the vent is told to move. It could be that the factory reset + configure doesn't have enough breathing room before any commands are sent for this to work

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
tpmanley: I think this should be left as `refresh() + cmds`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -87,7 +87,10 @@ def parse(String description) {
             // manufacturer-specific attribute
             event = getPressureResult(Integer.parseInt(descMap.value, 16))
         } else if (descMap?.clusterInt == zigbee.LEVEL_CONTROL_CLUSTER && descMap.attrInt == 0x0000) {
-            if (Integer.parseInt(descMap.value, 16) == 255) state.obstructed = true
+            if (Integer.parseInt(descMap.value, 16) == 255) {
```
tpmanley: Are you sure that a value of 255 is supposed to mean it is obstructed? Mine is reporting 255 after join even though it's not obstructed.

greens: It might be similar to the garage door opener where it's unsure of its actual position until it has been actuated?

this 255 is cribbed directly from the previous DTH

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,144 @@ metadata {
...
+        cmds << "delay 2000"
     }
-
-    sendEvent(makeOnOffResult(1))
-    "st cmd 0x${device.deviceNetworkId} 1 6 1 {}"
+    def currentLevel = device.currentValue("level")
+    if (currentLevel != null) {
+        currentLevel = currentLevel as int
+    }
+    def levelToSet = currentLevel ? curentLevel : 100
```
tpmanley: curentLevel -> currentLevel

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +46,144 @@ metadata {
...
-    def rangeMax = 254
-    def scaledLevel = Math.round(level * rangeMax / 100)
-    log.debug "scaled level for ${level}%: ${scaledLevel}"
-
-    // convert to hex string and pad to two digits
-    def hexLevel = new BigInteger(scaledLevel.toString()).toString(16).padLeft(2, '0')
-
-    "st cmd 0x${device.deviceNetworkId} 1 8 4 {${hexLevel} 0000}"
+def getPressureResult(rawValue) {
+    def kpa = rawValue / (10 * 1000) // reports are in deciPascals, I think
```
tpmanley: can remove the "I think" part

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -477,47 +193,24 @@ def configure() {
...
-        // battery - type: int8u, change: 1
-        // "zcl global send-me-a-report 1 0x21 0x20 60 3600 {01}", "delay 200",
-        // "send 0x${device.deviceNetworkId} 1 1", "delay 1500",
-    ]
+private boolean isObstructed() {
+    def currentState = device.currentValue("switch")
 
-    return configCmds + zigbee.onOffConfig() + refresh()
+    if (currentState == "obstructed") {
+        sendEvent(name: "level", value: 0) // convert legacy implementation
```
tpmanley: I think we should also set `switch` to a valid value here

tpmanley: The `switch` and `level` events should also not be displayed

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -477,47 +193,24 @@ def configure() {
...
+        state.obstructed == true
```
tpmanley: Should be `=`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
greens: I think this was just an error on my part. This was always meant to be a switch event. The old DTH currently sets "switch" to "obstructed" and this is just meant to give it capability-aligning values. It never sets "level" to anything weird.

tpmanley (approved):

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +44,113 @@ metadata {
...
-        descriptionText: "${linkText} has zigbeeId ${device.zigbeeId}" ])
-    return [
-        name: "zigbeeId",
-        value: device.zigbeeId,
-        descriptionText: "${linkText} has zigbeeId ${device.zigbeeId}" ]
+    def cmds = []
+    cmds << zigbee.setLevel(value)
+    cmds << "delay 1000"
+    cmds << zigbee.levelRefresh()
+    cmds
```
dkirker, line 138: Could probably just make this `[zigbee.setLevel(value), "delay 1000", zigbee.levelRefresh()]`

devicetypes/keen-home/keen-home-smart-vent.src/keen-home-smart-vent.groovy
```diff
@@ -51,423 +44,113 @@ metadata {
...
-                return makeTemperatureResult(value)
-                break
-
-            case 0x0006:
-                return makeOnOffResult(cluster.data[-1])
-                break
+    def event = zigbee.getEvent(description)
+    if (!event) {
+        Map descMap = zigbee.parseDescriptionAsMap(description)
+        if (descMap?.clusterInt == zigbee.POWER_CONFIGURATION_CLUSTER && descMap.attrInt == 0x0021) {
```
dkirker, line 82: Might be worth defining this attrs instead of magic numbers.

dkirker (commented):
Just a couple of comments. Otherwise, 👍 .

dkirker (approved):

PR #45830, opened 2020-09-29
Add three sensor products from ThirdReality, Inc

XQ-Huang:


XQ-Huang:
Add three zigbee sensor products from ThirdReality, Inc

BronzeWareH (approved):

XQ-Huang:
@greens, Please help to review the merge require, Thanks.

greens (changes requested):
Do these devices really not report battery level or temperature?

Do you really not want to treat the motion or water sensors like motion and water sensors?

Why did you copy and paste an entire existing DTH instead of just adding the fingerprints to the DTH you copied from (`open-closed-sensor.groovy`)?

greens (changes requested):
Do these devices really not report battery level or temperature?

Do you really not want to treat the motion or water sensors like motion and water sensors?

Why did you copy and paste an entire existing DTH instead of just adding the fingerprints to the DTH you copied from (`open-closed-sensor.groovy`)?

jiangshanyang0203:
@greens Sorry for the late reply.At present, our three kinds of sensors only realize simple alarm and non alarm functions. So use a simple way to meet the use. Thanks.

greens:
@jiangshanyang0203 if this is what you want to do, then just add the fingerprints for these three devices to the existing DTH you copied instead of making a new one.

PR #45873, opened 2020-09-29
WWST-6892, WWST-6896, WWST-6898, WWST-6904 - New DTH for Viconics/Schneider Room Controllers

KKlimczukS:
**New DTH for Viconics/Schneider Room Controllers:**

WWST-6898 - VT8350 - Viconics Technologies - Low Voltage Fan Coil Controller and Zone Controller
WWST-6892 - VT8650 - Viconics Technologies - Heat Pump and Indoor Air Quality Controller
WWST-6904 - SE8350 - Schneider Electric - Low Voltage Fan Coil Unit (FCU) and Zone Control
WWST-6896 - SE8650 - Schneider Electric - Roof Top Unit Controller

This is the first working version of the DTH for Viconics/Schneider Room Controllers.
Since:
- many custom attributes are used by the manufacturer - a separate DTH had to be done. 
- device supports also humidity, occupancy and fan capabilities
- a separate child device to handle unoccupied setpoints was needed too 

@ZWozniakS @MWierzbinskaS 
Please let me know if it works correctly with your devices.

@tpmanley @greens @dkirker 


devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,551 @@
...
+def configure() {
+	log.debug "Configuration"
+
+	if (!childDevices) {
+		createChildThermostat()
+	}
+
+	def configurationCommands = []
+	// todo: check if following binding is necessary here
+	configurationCommands += "zdo bind 0x${device.deviceNetworkId} 1 0xA 0x201 {${device.zigbeeId}} {}"
```
PKacprowiczS: I think `zigbee.addBinding(_CLUSTER_)` might also work here.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,551 @@
...
+		// Viconics VT8350 humidity reports are parsed as floating point numbers (range 0 - 1%)
+		if (eventMap.name == "humidity") {
+			eventMap.value = eventMap.value * 100
+		}
+		result = createEvent(eventMap)
+	}*/ else {
+		eventMap = [:]
+		def descMap = zigbee.parseDescriptionAsMap(description)
+
+		if ((descMap.clusterInt == THERMOSTAT_CLUSTER && descMap.attrId)) {
```
PKacprowiczS: nit: unnecessary double brackets

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,551 @@
...
+	sendEvent(name: "supportedThermostatFanModes", value: JsonOutput.toJson(state.supportedFanModes), displayed: false)
+	sendEvent(name: "supportedThermostatModes", value: JsonOutput.toJson(state.supportedThermostatModes), displayed: false)
+	sendEvent(name: "coolingSetpointRange", value: coolingSetpointRange, displayed: false)
+	sendEvent(name: "heatingSetpointRange", value: heatingSetpointRange, displayed: false)
+}
+
+private void createChildThermostat() {
+	log.debug "Creating child thermostat to handle unoccupied cooling/heating setpoints"
+	def label = "Unoccupied setpoints"
+	def childName = "${device.displayName} " + label
```
PKacprowiczS: nit: maybe just `"${device.displayName} ${label}"`?

KKlimczukS: good point

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Since binding is added in zigbee.configureReporting() method those binding commands probably  will not be needed here, but i will take another look at this (remove those commands first and check if all works correctly)

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: sure, fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+			"02":"Override",
+			"03":"Standby"
+	]
+}
+
+def installed() {
+	log.debug "installed"
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
+
+	if (isViconicsVT8350()|| isSchneiderSE8350()) {
```
ZWozniakS: nit: space between _**isViconicsVT8350()**_ and _**||**_

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+import physicalgraph.zigbee.zcl.DataType
+metadata {
+	definition(name: "Viconics Schneider Room Controller", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.thermostat", mcdSync: true) {
+
+		capability "Actuator"
+		capability "Sensor"
+		capability "Occupancy Sensor"
+		capability "Temperature Measurement"
+		capability "Relative Humidity Measurement"
+		capability "Thermostat"
```
greens: I think you should just leave this deprecated capability off

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+			"00":"Occupied",
+			"01":"Unoccupied",
+			"02":"Override",
+			"03":"Standby"
+	]
+}
+
+def installed() {
+	log.debug "installed"
+	sendEvent(name: "checkInterval", value: 2 * 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID, offlinePingable: "1"])
```
greens: should not be offlinePingable

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+			"02":"heating"
+	]
+}
+
+private getEFFECTIVE_OCCUPANCY_MAP() {
+	[
+			"00":"Occupied",
+			"01":"Unoccupied",
+			"02":"Override",
+			"03":"Standby"
```
greens: These last two are not legal values for the capability.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+	if (description?.startsWith("humidity: ")) {
+		// Viconics VT8350 humidity reports are parsed as floating point numbers (range 0 - 1%)
+		def humidityVal = (description - "humidity: " - "%").trim()
+		if (humidityVal.isNumber()) {
+			humidityVal = new BigDecimal(humidityVal) * 100
+		}
+		eventMap.name 	= "humidity"
+		eventMap.value 	= humidityVal
+		eventMap.unit 	= "%"
+	}
```
greens: Can you just handle this the way you're handling the other events? i.e. using the results of `parseDescriptionAsMap`?

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+		}
+		result = createEvent(eventMap)
+	}*/ else {
+		eventMap = [:]
+		def descMap = zigbee.parseDescriptionAsMap(description)
+
+		if (descMap.clusterInt == THERMOSTAT_CLUSTER && descMap.attrId) {
+			def attributeInt = zigbee.convertHexToInt(descMap.attrId)
+
+			if (attributeInt == OCCUPANCY || attributeInt == CUSTOM_EFFECTIVE_OCCUPANCY) {
```
greens: I think you should be using a switch statement here. If you're clever about placement of breaks, you should be able to condense this code significantly.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+					setFanSpeed(1)
+					break
+				case "auto":
+					setFanSpeed(4)
+					break
+			}
+		} else if (isViconicsVT8650() || isSchneiderSE8650()) {
+			switch (mode) {
+				case "on":
+					getThermostatFanModeCommands(CUSTOM_FAN_MODE_ON)
```
greens: I think you should avoid a switch statement here and delegate this lookup of the fan mode value to the `getThermostatFaneModeCommands` function you defined.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+		log.debug "Unsupported mode $mode"
+	}
+}
+
+def off() {
+	delayBetween([
+			zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE, DataType.ENUM8, THERMOSTAT_MODE_OFF),
+			zigbee.readAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE),
+			zigbee.readAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_OPERATING_STATE)
+	], 500)
```
greens: seems like you could use the same pattern for these 4 that you did for the fan modes.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+	delayBetween([
+			zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE, DataType.ENUM8, THERMOSTAT_MODE_HEAT),
+			zigbee.readAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE),
+			zigbee.readAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_OPERATING_STATE)
+	], 500)
+}
+
+def ping() {
+	log.debug "ping"
+	refresh()
```
greens: the ping method should not be querying all of these attributes. Just temperature is fine.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, COOLING_SETPOINT, DataType.INT16, 1, 60, 10)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, HEATING_SETPOINT, DataType.INT16, 1, 60, 10)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, COOLING_SETPOINT_UNOCCUPIED, DataType.INT16, 1, 60, 10)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, HEATING_SETPOINT_UNOCCUPIED, DataType.INT16, 1, 60, 10)
+	//configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, 0x0A58, 0x10, 1, 300, 1) //GFan
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_FAN_MODE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_FAN_SPEED, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_OPERATING_STATE, DataType.ENUM8, 1, 60, 1)
+    configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, OCCUPANCY, DataType.ENUM8, 1, 60, null)
```
greens: nit: spacing difference here

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,550 @@
...
+		def celsius = Integer.parseInt(value, 16) / 100
+		if (temperatureScale == "C") {
+			return celsius//Math.round(celsius)
+		} else {
+			return celsiusToFahrenheit(celsius)//Math.round(celsiusToFahrenheit(celsius))
+		}
+	}
+}
+
+def mapFanSpeedSliderValue(rawValue) {
```
greens: you could just make this an actual map

devicetypes/smartthings/child-thermostat-setpoints.src/child-thermostat-setpoints.groovy
```diff
@@ -0,0 +1,57 @@
...
+def installed() {
+	configure()
+}
+
+def updated() {
+	configure()
+}
+
+def configure() {
+	parent.configureChild()
```
greens: undefined

devicetypes/smartthings/child-thermostat-setpoints.src/child-thermostat-setpoints.groovy
```diff
@@ -0,0 +1,57 @@
...
+	refresh()
+}
+
+def ping() {
+	refresh()
+}
+
+def refresh() {
+	parent.refreshChild()
```
greens, line 42: undefined

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: getThermostatFaneModeCommands receives constant hex values, but a map will be in the middle - that way switch statement won't be used here.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed, thank You.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: I'm not sure about it. When I got rid of that line, all  other plugin elements that are related to the Thermostat weren't displayed.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: good point, fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/child-thermostat-setpoints.src/child-thermostat-setpoints.groovy
KKlimczukS: fixed

devicetypes/smartthings/child-thermostat-setpoints.src/child-thermostat-setpoints.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: yes, it will be done as you suggest.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
greens: some metadata is defined by referencing the `Thermostat` attributes, but others, such as Z-Wave Battery Thermostat, specifically reference the discrete capabilities. You might just need to poke at the metadata bit.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+                childEvent.unit = temperatureScale
+                sendEventToChild(UNOCCUPIED_SETPOINT_CHILD_DEVICE_ID, childEvent)
+                break
+            case LOCAL_TEMPERATURE:
+                log.debug "LOCAL TEMPERATURE, descMap.value: ${descMap.value}"
+                eventMap.name = "temperature"
+                eventMap.value = getTemperature(descMap.value)
+                eventMap.unit = temperatureScale
+                break
+            case CUSTOM_HUMIDITY:
```
greens, line 203: just double-checking, they included this under the thermostat cluster, right?

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
greens: I think @PKacprowiczS  is correct.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: sure, done

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: @greens 
It's possible to obtain humidity values either from RELATIVE_HUMIDITY_CLUSTER or THERMOSTAT_CLUSTER (custom attribute)

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: @greens i will give it a try ( merge request with UI metadata wasn't created yet)

dkirker: Yeah, if the metadata references `/capability/thermostat/X` where `X` is some number then it is referring to the `Thermostat` capability. If you drop this and use metadata that references the individual capabilities it should work.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+                eventMap.data = [supportedThermostatModes: state.supportedThermostatModes]
+                break
+            case CUSTOM_THERMOSTAT_OPERATING_STATE:
+                log.debug "CUSTOM THERMOSTAT OPERATING STATE, descMap.value: ${descMap.value}"
+                eventMap.name = "thermostatOperatingState"
+                eventMap.value = THERMOSTAT_OPERATING_STATE_MAP[descMap.value]
+                break
+            default:
+                log.debug "UNHANDLED ATTRIBUTE, descMap.inspect(): ${descMap.inspect()}"
+
```
dkirker: You've got a blank line here

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+def setSetpoint(degrees, setpointAttr) {
+	log.debug "degrees: ${degrees}, setpointAttr: ${setpointAttr}"
+	if (degrees != null && setpointAttr != null) {
+		log.debug "temperatureScale: ${temperatureScale}"
+		def celsius = (temperatureScale == "C") ? degrees : fahrenheitToCelsius(degrees)
+		celsius = (celsius as Double).round(2)
+
+		delayBetween([
+				zigbee.writeAttribute(THERMOSTAT_CLUSTER, setpointAttr, DataType.INT16, zigbee.convertToHexString(celsius * 100)),
```
dkirker: Let's bring this indenting to the left one so there isn't so much wasted space

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+}
+
+def fanCirculate() {
+	getThermostatFanModeCommands(CUSTOM_FAN_MODE_CIRCULATE)
+}
+
+def getThermostatFanModeCommands(mode) {
+	if (mode) {
+		delayBetween([
+				zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_FAN_MODE, DataType.ENUM8, mode),
```
dkirker: Same with the indenting here

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+def setFanSpeed(speed) {
+	log.debug "setFanSpeed: ${speed}"
+
+	if (speed == 0 || speed >= 4) { //if by any chance user selects 0 or a value higher than 3, it fan will be set to AUTO
+		speed = 3
+	} else {
+		speed = speed - 1
+	}
+	delayBetween([
+			zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_FAN_SPEED, DataType.ENUM8, speed),
```
dkirker: And here

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+				zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_FAN_MODE, DataType.ENUM8, mode),
+				zigbee.readAttribute(THERMOSTAT_CLUSTER, CUSTOM_FAN_MODE)
+		], 500)
+	}
+}
+
+def setFanSpeed(speed) {
+	log.debug "setFanSpeed: ${speed}"
+
+	if (speed == 0 || speed >= 4) { //if by any chance user selects 0 or a value higher than 3, it fan will be set to AUTO
```
dkirker, line 348: `it fan will be set to AUTO` -> `the fan will be set to AUTO`

So there is no `off` then?

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+}
+
+def off() {
+	getThermostatModeCommands(THERMOSTAT_MODE_OFF)
+}
+
+def getThermostatModeCommands(mode) {
+	if (mode) {
+		delayBetween([
+				zigbee.writeAttribute(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE, DataType.ENUM8, mode),
```
dkirker: Same with the indenting here

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
dkirker: Yes, `zigbee.configureReporting` has a call to `zigbee.addBinding` as part of it. So in this case, clusters 0x201 and 0x405 will be handled with all of the `zigbee.configureReporting(THERMOSTAT_CLUSTER` and `zigbee.configureReporting(RELATIVE_HUMIDITY_CLUSTER` calls below (in fact, 13 times for the thermostat cluster). So these calls are superfluous.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+	(getTemperatureScale() == "C") ? [4.5, 32] : [40, 90]
+}
+
+def getTemperature(value) {
+	if (value != null) {
+		def celsius = Integer.parseInt(value, 16) / 100
+		if (temperatureScale == "C") {
+			celsius//Math.round(celsius)
+		} else {
+			celsiusToFahrenheit(celsius)//Math.round(celsiusToFahrenheit(celsius))
```
dkirker: I know in some other device handlers there is rounding logic. Are we not wanting to do that here? If not, let's remove the comments.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,532 @@
...
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_FAN_SPEED, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_OPERATING_STATE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_EFFECTIVE_OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_UI_CONFIGURATION_CLUSTER, TEMPERATURE_DISPLAY_MODE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_HUMIDITY, DataType.UINT16, 1, 60, 10)
+	configurationCommands += zigbee.configureReporting(RELATIVE_HUMIDITY_CLUSTER, RELATIVE_HUMIDITY_MEASURED_VALUE, DataType.UINT16, 1, 60, 5)
+
+	delayBetween(getRefreshCommands()+configurationCommands)
```
dkirker: Space around the `+` for readability.

dkirker (commented):
Did a quick first pass.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Yes, I agree. As expected, I removed those zdo bind commands.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Thank You @dkirker, I removed those references from UI metadata and it works now.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,549 @@
...
+	[
+		"00":"on",
+		"01":"auto",
+		"02":"circulate"
+	]
+}
+
+private getTHERMOSTAT_FAN_MODE_ATTRIBUTE_ID_MAP() {
+	[
+		"on":  0x00,		// CUSTOM_FAN_MODE_ON
```
greens: Can you use the enums you created in these maps? You should also be using them for the keys in the other maps, but you'll obviously have to cast to ints when retrieving the values, but I think if you're going to define all these magic numbers you should not have magic numbers in the code.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,549 @@
...
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_FAN_MODE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_FAN_SPEED, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_OPERATING_STATE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_EFFECTIVE_OCCUPANCY, DataType.ENUM8, 1, 60, null)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_UI_CONFIGURATION_CLUSTER, TEMPERATURE_DISPLAY_MODE, DataType.ENUM8, 1, 60, 1)
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_HUMIDITY, DataType.UINT16, 1, 60, 10)
+	configurationCommands += zigbee.configureReporting(RELATIVE_HUMIDITY_CLUSTER, RELATIVE_HUMIDITY_MEASURED_VALUE, DataType.UINT16, 1, 60, 5)
```
greens: I think you can remove the reporting configuration using the humidity cluster, since you don't handle it in the parse method.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,549 @@
...
+
+	def configurationCommands = []
+
+	// set initial values
+	configurationCommands += setSetpoint(initialCoolingSetpoint, COOLING_SETPOINT)
+	configurationCommands += setSetpoint(initialHeatingSetpoint, HEATING_SETPOINT)
+	configurationCommands += setSetpoint(initialCoolingSetpoint, COOLING_SETPOINT_UNOCCUPIED)
+	configurationCommands += setSetpoint(initialHeatingSetpoint, HEATING_SETPOINT_UNOCCUPIED)
+
+	configurationCommands += zigbee.configureReporting(THERMOSTAT_CLUSTER, CUSTOM_THERMOSTAT_MODE, DataType.ENUM8, 1, 60, 1) //formerly THERMOSTAT MODE: 0x001C
```
greens: I just realized that you're setting the max reporting interval for _all_ of these attributes at a minute. That's way too low. That _requires_ them to report every minute, even if there's no change. If your min reporting interval is 1 second and the delta is 1 they'll report every change anyway, so please raise the max reporting interval to at least an hour. 

If the device is sending us like 20 updates every minute it will be an incredibly expensive device to run.

KKlimczukS: @greens Good point (max reporting was set to short period for testing purposes).

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: @dkirker  Sure. 
Actually setpoints can be changed manually with a step = 0.5
Additionally - since the device displays his values on the screen with only one digit after the decimal point and reports its values with 2 digits after decimal point - it' looks really bad if You want to change setpoints on the ST UI (step = 0.5), for example: 18.72 <-19.22 -> 19.72 
so those values will be rounded to the nearest half (for example: 19.22 >> 19.00; 19.32 >> 19.50)
We will also configure  setpoints initially to rounded default values so the device would work on "rounded" numbers from the very beginning.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: No, the only possible options are: **L**ow, **M**ed, **H**igh, **A**uto, **O**n (user can also configure the device so the concrete "fan sequence" can be set manually :  On-Auto, L-M-H, L-H, L-M-H-A and L-H-A).

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Sure, but I had to add some new constants.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Sure, fixed (i kept those configuration commands for testing purposes).

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,554 @@
...
+	]
+}
+
+private getTHERMOSTAT_MODE_MAP() {
+	[
+		(THERMOSTAT_MODE_OFF):"off",
+		(THERMOSTAT_MODE_AUTO):"auto",
+		(THERMOSTAT_MODE_COOL):"cool",
+		(THERMOSTAT_MODE_HEAT):"heat",
+		"04":"heat"
```
greens: is this emergency heat?

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,554 @@
...
+		(OPERATING_STATE_IDLE):"idle",
+		(OPERATING_STATE_COOLING):"cooling",
+		(OPERATING_STATE_HEATING):"heating"
+	]
+}
+
+private getEFFECTIVE_OCCUPANCY_MAP() {
+	[
+		(OCCUPANCY_OCCUPIED):"Occupied",
+		(OCCUPANCY_UNOCCUPIED):"Unoccupied"
```
greens: I believe these are not capitalized in the capability.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,554 @@
...
+	]
+}
+
+private getFAN_SPEED_SLIDER_MAP() {
+	[
+		(FAN_SPEED_LOW): 1,
```
greens: This map is not necessary. Just refactor the code in the one place that you use it.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,554 @@
...
+	log.debug "child.inspect() ${child}"
+}
+
+def parse(String description) {
+    def result = []
+    def eventMap = [:]
+    def descMap = zigbee.parseDescriptionAsMap(description)
+
+    if (descMap.clusterInt == THERMOSTAT_CLUSTER && descMap.attrId) {
+        def attributeInt = zigbee.convertHexToInt(descMap.attrId)
```
greens: just use `descMap.attrInt`

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
```diff
@@ -0,0 +1,554 @@
...
+private getCUSTOM_FAN_MODE_CIRCULATE() { 0x02 }
+private getOPERATING_STATE_IDLE() { 0x00 }
+private getOPERATING_STATE_COOLING() { 0x01 }
+private getOPERATING_STATE_HEATING() { 0x02 }
+private getOCCUPANCY_OCCUPIED() { 0x00 }
+private getOCCUPANCY_UNOCCUPIED() { 0x01 }
+private getFAN_SPEED_LOW() { 0x00 }
+private getFAN_SPEED_MEDIUM() { 0x01 }
+private getFAN_SPEED_HIGH() { 0x02 }
+private getFAN_SPEED_AUTO() { 0x03 }
```
greens: I think it might be worthwhile to look into using an enum for some of these.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: I'm still not sure about that one. Viconics/Schneider uses this value (i saw it in their very old dths), but it's not mentioned in their manual.
I left this value intentionally (for testing purposes), but it will be removed probably.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: fixed

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: In that case, we can get rid of it and just use "value +1" instead (because the device reports values in the range 0-3, where 0 is LOW fan speed).

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: hmm... why didn't I use that in the first place

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: @greens 
wouldn't that look bad in the code if some of those were enums and some not ? 
Old way (declaration and usage) looks more concise.

enums for occupancy might look like this:

```
enum Occupancy{
	OCCUPIED(0), UNOCCUPIED(1)
	Occupancy(int value) {
		this.value = value
	}
	private final int value
	int getValue() {
		value
	}
}
```

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: That's because in most cases there are no **attrInt** in the map returned by zigbee.parseDescriptionAsMap(description)
so it had to be that way:
def attributeInt = zigbee.convertHexToInt(descMap.attrId)

greens: I'm looking at the code for `parseDescriptionAsMap`

```
if (outMap.containsKey("attrId")) {
	outMap["attrInt"] = convertHexToInt(outMap["attrId"])
}
```

If this is what you're seeing, I want to see the case in which you are getting a message that is parsing out to have an `attrId` but not an `attrInt`.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
greens: yeah fair enough

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
greens: right, or just `value < 3`

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: It was noticed mostly when OCCUPIED reports were processed, that's why i reverted that change.
When i checked it today on a side,  it turned out it worked fine:

example code:
`    def description = "read attr - raw: 230A0A02010A0200001800, dni: 230A, endpoint: 0A, cluster: 0201, size: 10, attrId: 0002, result: success, encoding: 18, value: 00"
    log.debug "test descMap.inspect(): ${descMap}"`

That was a strange behavior, because yesterday attInt was null in most cases for a while.
Anyway, i'm reverting those changes for now, but i will keep an eye on it.

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
KKlimczukS: Reported speeds are in range 0-2 (0 is LOW, not OFF), so it has to be incremented +1 to indicate (on the UI/slider) that the fan is working, if it stayed at 0 - it wouldn't be the best UX.

greens: ah yeah, it's passed in the event later

I was just looking at the if statement

either way, this map should go away

devicetypes/smartthings/viconics-schneider-room-controller.src/viconics-schneider-room-controller.groovy
greens: did you still want to remove it? It's still a string and I think all the other indices are not.

KKlimczukS: I performed few tests and the device consumed 0x04 and 0x05 values, but after a while it switched to previous mode. 
Additionally: values >= 0x06 were rejected (device didn't even report setting them).
Anyway,  i removed "04" already.

greens (approved):

PR #46568, opened 2020-10-05
CHAD-5555 Update z-wave device multichannel to work in oneapp

greens:
Because of its variable number of components, the child devices cannot be added as components.

### DIFF
--- a/devicetypes/smartthings/zwave-device-multichannel.src/zwave-device-multichannel.groovy
+++ b/devicetypes/smartthings/zwave-device-multichannel.src/zwave-device-multichannel.groovy
@@ -434,3 +434,9 @@ private encap(cmd, endpoint) {
 private encapWithDelay(commands, endpoint, delay=200) {
 	delayBetween(commands.collect{ encap(it, endpoint) }, delay)
 }
+
+def updated() {
+    childDevices.each {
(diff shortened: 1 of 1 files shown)


tpmanley (dismissed):

devicetypes/smartthings/zwave-device-multichannel.src/zwave-device-multichannel.groovy
```diff
@@ -435,3 +434,9 @@ private encap(cmd, endpoint) {
 private encapWithDelay(commands, endpoint, delay=200) {
 	delayBetween(commands.collect{ encap(it, endpoint) }, delay)
 }
+
+def updated() {
+    childDevices.each {
+        if (it.device.isComponent) { it.save([isComponent: false, componentLabel: null, componentName: null]) }
+    }
+}
```
greens, line 442: @tpmanley happy news is that I have confirmed that this works as we expect.

dkirker: So I can probably use this for the Inovelli device, then. (I snagged a device last time I was in the office, so I can actually try that.)

dkirker (approved):

tpmanley (approved):

PR #46726, opened 2020-10-07
ICP-13361 Added min dimming level preference

MWierzbinskaS:
@KKlimczukS @ZWozniakS @MGoralczykS @PKacprowiczS 

I've added minimum dimming level preference to resolve issue ICP-13361. 
Because after setting this preference, device reports values adjusted to its range, they need to be readjusted in our DTH.

@greens  Could you please take a look at the code?

### DIFF
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -400,8 +400,7 @@ Integer adjustValueToRange(value){
 		return 0
 	}
 	def minDimmingLvlPref = settings.minimumDimmingValue ?: parameterMap.find({it.key == 'minimumDimmingValue'}).defaultValue
-	def range = 100 - minDimmingLvlPref
-	def adjustedValue = (((value-minDimmingLvlPref)*100)/range) as Integer
+	def adjustedValue = (((value - minDimmingLvlPref)/(100 - minDimmingLvlPref)) * 100) as Integer
 	return Math.max(adjustedValue, minDimmingLvlPref)
 }
 
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -401,7 +401,7 @@ Integer adjustValueToRange(value){
 	}
 	def minDimmingLvlPref = settings.minimumDimmingValue ?: parameterMap.find({it.key == 'minimumDimmingValue'}).defaultValue
 	def adjustedValue = (((value - minDimmingLvlPref)/(100 - minDimmingLvlPref)) * 100) as Integer
-	return Math.max(adjustedValue, minDimmingLvlPref)
+	return Math.max(adjustedValue, 1)
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.sensormultilevelv5.SensorMultilevelReport cmd, ep = null) {
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -378,7 +378,6 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
 	def value = (cmdValue ? "on" : "off")
 	def result = [createEvent(name: "switch", value: value)]
 	if (cmdValue && cmdValue <= 100) {
-		//cmdValue = adjustValueToRange(cmd.value)
 		result << createEvent(name: "level", value: cmdValue == 99 ? 100 : cmdValue)
 	}
 
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -383,6 +383,7 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
 	def value = (cmdValue ? "on" : "off")
 	def result = [createEvent(name: "switch", value: value)]
 	if (cmdValue && cmdValue <= 100) {
+		cmdValue = adjustValueToRange(cmd.value)
 		result << createEvent(name: "level", value: cmdValue == 99 ? 100 : cmdValue)
 	}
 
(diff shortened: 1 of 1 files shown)


devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -387,11 +387,24 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
 	def value = (cmdValue ? "on" : "off")
 	def result = [createEvent(name: "switch", value: value)]
 	if (cmdValue && cmdValue <= 100) {
+		if(isDINDimmer()){
```
KKlimczukS: Since the "minimumDimmingValue" parameter is exposed here, should we also consider adjusting that value for all Qubino devices that are supported by this DTH.

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -387,11 +387,24 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
...
 		result << createEvent(name: "level", value: cmdValue == 99 ? 100 : cmdValue)
 	}
 	return result
 }
 
+Integer adjustValueToRange(value){
+	if(value == 0){
+		return 0
+	}
+	def minDimmingLvlPref = getCommandValue(parameterMap.find({it.key == 'minimumDimmingValue'})) ?: parameterMap.find({it.key == 'minimumDimmingValue'}).defaultValue
```
greens: if the value has been set it should just be accessible at `settings.minimumDimmingValue`

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
greens: Agree with Konrad. I expect the same problem will occur for all of these devices.

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -387,11 +387,24 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
...
+	def range = 100 - minDimmingLvlPref
+	def adjustedValue = (((value-minDimmingLvlPref)*100)/range) as Integer
+	return Math.max(adjustedValue, minDimmingLvlPref)
```
greens: This math doesn't seem right to me.

If I want to set the level to 50, but my minimum value is 10:

range = 90
adjustedValue = ((40*100)/90) = 44.44

I would expect, by inputting 50, that I would get a value halfway between (50%) my minimum and maximum, which would be 55, right?

We want to treat [`minLevel`-100] as our range of values we will send to the device, but the value sent by the user should be interpreted as a percentage of that range. The way to get that number is `(100 - min) * level + min`.

greens: but retain the special case of 0 here.

MWierzbinskaS: The issue here is that when I have:
minimum dimming value - 10
1.I chose to set 50% on UI and this value is being sent to the device
2.device responds that it set 55% - which is half of the range

I was recalculating those values to show 50 on UI in our application when creating ST events.

So it is: (55 - 10)*100/90

I see your point, but I think something is wrong 
(100 - min) * level + min = (100 - 10) * 55 + 10 = 4960

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
MWierzbinskaS: Done

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
MWierzbinskaS: Done

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
greens: Oh, I thought this was the code for sending the commands, not for parsing the reports. My mistake.

So if we send 50 to the device it reports 55, but we want to transform that back into 50, so we'd just do the above equation in reverse, solving for `value`: `(reportedValue - min)/(100 - min)`

`(55 - 10) / 90 = .5` so just multiply by 100 and boom, right?

MWierzbinskaS: Ok, done

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -387,11 +387,21 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
...
 	return result
 }
 
+Integer adjustValueToRange(value){
+	if(value == 0){
+		return 0
+	}
+	def minDimmingLvlPref = settings.minimumDimmingValue ?: parameterMap.find({it.key == 'minimumDimmingValue'}).defaultValue
+	def adjustedValue = (((value - minDimmingLvlPref)/(100 - minDimmingLvlPref)) * 100) as Integer
+	return Math.max(adjustedValue, minDimmingLvlPref)
```
greens: I don't think you want to be doing this `math.max` anymore, right? If we're doing 50->55->50, presumably the user can still set the level to 1 even if the min dimming level is 10, right?

MWierzbinskaS: I changed that and now user can set 1% level. I left Math.max with 1 instead of minDimmingLvlPref, because I wanted to be sure that after calculation we will not get any out of range values.

greens (dismissed):

greens:
Looks good, but waiting for verification.

MWierzbinskaS:
@greens  After discussion in SRPOL we decided to try a workaround.

Setting minimum dimming level preference has an influence on how this device levels are being converted inside device logic. We get a rounded response from the device which is usually different from level set by user (even by over 2%). Recounting those values by us can be misleading because different devices with different firmware versions have different behavior. We decided that:

- parameter “Minimum dimming value” shouldn’t be sent to device - it will be set for UI purposes only

- When user changes dimmer value manually and device reports it and this value is lower than parameter Minimum dimming value then Minimum dimming value has to be sent to device.

- parameter “Minimum dimming value” should start from 1 (by default). We decided to change device UI metadata to dim in 2-100% range.

@KKlimczukS

devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -394,6 +389,14 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
 	return result
 }
 
+Integer adjustValueToRange(value){
+	if(value == 0){
+		return 0
+	}
+	def minDimmingLvlPref = settings.minimumDimmingValue ?: parameterMap.find({it.key == 'minimumDimmingValue'}).defaultValue
+	return Math.max(value, minDimmingLvlPref)
```
greens, line 397: won't you get a "network error" for the same reason if you try to set the level lower than the minimum dimming value? i.e. if I set level to 15 but my minimum dimming level is 20, the value I get returned is 20. How is that any better?

MWierzbinskaS: There is no network errors in application. When set level is lower than minimum dimming level, then the level set by the device is minimum dimming level. We do not get network error, we change the level value.

greens (approved):
lgtm, but I'll wait to hear back if this workaround is acceptable

greens (approved):

PR #49013, opened 2020-11-02
[ICP-13795] Honeywell Home T6 Pro - device specific metadata

PKacprowiczS:


greens (approved):

greens (changes requested):
actually I realized v5 get also needs a `scale` field

for temperature it should be `0`, which is Celsius and for humidity it should be `0` as well, for percentage.

Those may be filled it automatically as defaults, though, so it might not even be needed, but I will wait to see what the testing returns.

greens (approved):

PR #49278, opened 2020-11-05
ICP-13656 - Query the device for power meter values in response to dimmer reports. Code refactoring.

KKlimczukS:
**Previously:** device was queried for power meter values only only when its state was changed from the app. 

Since the device doesn't report power meter values when its state is changed manually (turn off) we have to query for power meter values in response to dimmer values reported by the device.

@greens @dkirker 


greens (approved):
looks good to me

PR #49282, opened 2020-11-06
DevWs for frient containing containing Zigbee Metering Plug

frient-design:

PR #49283, opened 2020-11-06

frient-design:


PKacprowiczS:
It seems to be duplicate of #49282 
@frient-design if you're including more than device in one DTH, don't create multiple PRs for it, just commit changes to a single PR

PR #49364, opened 2020-11-06
DevWs for frient containing containing SmartSense Motion Sensor

frient-design:


devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -133,7 +134,7 @@ def parse(String description) {
 						map = getBatteryResult(Integer.parseInt(battMap.value, 16))
 					}
 				}
-			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07) {
+			} else if (descMap?.clusterInt == 0x0500 && descMap.attrInt == 0x0002 && descMap.commandInt != 0x07 && descMap.value != null) {
```
PKacprowiczS, line 137: Why this condition is necessary here?

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -212,7 +213,13 @@ private Map getBatteryResult(rawValue) {
...
 			pct = pct > 0 ? pct : 1
 			result.value = Math.min(100, pct)
-		} else { // Centralite
+		} else if (isFrientSensor()) {
+			def minValue = 23
+			def maxValue = 30
+			def pct = Math.round((rawValue - minValue) * 100 / (maxValue - minValue))
+			pct = pct > 0 ? pct : 1
+			result.value = Math.min(100, pct)
+        } else { // Centralite
```
PKacprowiczS: Please use tabs as intendations

PKacprowiczS (changes requested):
Please pay attention to consistency of indentations.

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -43,6 +43,7 @@ metadata {
         //AduroSmart
         fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001,FFFF", manufacturer: "ADUROLIGHT", model: "VMS_ADUROLIGHT", deviceJoinName: "ERIA Motion Sensor", mnmn: "SmartThings", vid: "generic-motion-2" //ERIA Motion Sensor V2.0
         fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001,FFFF", manufacturer: "AduroSmart Eria", model: "VMS_ADUROLIGHT", deviceJoinName: "ERIA Motion Sensor", mnmn: "SmartThings", vid: "generic-motion-2" //ERIA Motion Sensor V2.1
+        fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,000F,0020,0500", outClusters: "000A,0019", manufacturer: "frient A/S", model :"MOSZB-140", deviceJoinName: "frient Motion Sensor Pro"
```
PKacprowiczS: Please just leave `frient Motion Sensor` as a deviceJoinName

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
rboy1: This is a bug in the current DTH which is causing a `java.lang.IllegalArgumentException` error for devices which don't support reading this attribute. The way it currently coded, it assumes that response messages sent by a the device using the IAS Zone cluster and the IAS ZoneStatus attribute in response to Configure Reporting, the response command should contain a value. That's incorrect. When the DTH tries to read the IAS Zone Status attribute in the configuration, some devices (like the frient motion sensor) can return UNSUPPORTED_ATTRIBUTE, which leads to a `null` value for `desc.Map` and it should be ignored by the DTH otherwise it leads to an exception when `ZoneStatus(zigbee.convertToInt(descMap.value, 16))` tries to parse it in the next line.

TLDR; The message should only be parsed by ZoneStatus if the value is non null (this change won't affect devices which respond to the IAS Zone with a value but it'll skip parsing for those devices which respond with unsupported response).

e..g. `[raw:0104 0500 22 01 0000 00 6F71 00 00 0000 01 01 020086, profileId:0104, clusterId:0500, sourceEndpoint:22, destinationEndpoint:01, options:0000, messageType:00, dni:6F71, isClusterSpecific:false, isManufacturerSpecific:false, manufacturerId:0000, command:01, direction:01, attrId:0002, resultCode:86, data:[02, 00, 86], clusterInt:1280, attrInt:2, commandInt:1]`

frient-design:
We have made all the requested changes, can you please accept the request to continue the certification or let us know if more modification is needed

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
PKacprowiczS: Thanks for elaborating, that's reasonable

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -305,12 +312,19 @@ def configure() {
 	configCmds += zigbee.enrollResponse()
 	// temperature minReportTime 30 seconds, maxReportTime 5 min. Reporting interval if no activity
 	// battery minReport 30 seconds, maxReportTime 6 hrs by default
-	if (device.getDataValue("manufacturer") == "Samjin") {
-		configCmds += zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0021, DataType.UINT8, 30, 21600, 0x10)
+    if (device.getDataValue("manufacturer") == "Samjin") {
```
PKacprowiczS: You've missed wrong indentation here, ...

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -305,12 +312,19 @@ def configure() {
...
+        configCmds += zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0021, DataType.UINT8, 30, 21600, 0x10)
+    } else if (isFrientSensor()) {
+        configCmds += zigbee.configureReporting(zigbee.POWER_CONFIGURATION_CLUSTER, 0x0020, DataType.UINT8, 30, 21600, 0x1, powerEndpoint())
 	} else {
 		configCmds += zigbee.batteryConfig()
 	}
-	configCmds += zigbee.temperatureConfig(30, 300)
+    
+    if (isFrientSensor()) {
```
PKacprowiczS: ...here, ...

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -335,3 +349,23 @@ private shouldUseOldBatteryReporting() {
 
 	return isFwVersionLess // If f/w version is less than 1.15.7 then do NOT smooth battery reports and use the old reporting
 }
+
+private Boolean isFrientSensor() {
```
PKacprowiczS, line 353: ...and here.

PKacprowiczS (changes requested):

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
rboy1: fixed

rboy1 (approved):

PKacprowiczS (approved):
@greens

KKlimczukS (approved):

devicetypes/smartthings/smartsense-motion-sensor.src/smartsense-motion-sensor.groovy
```diff
@@ -335,3 +349,23 @@ private shouldUseOldBatteryReporting() {
...
+	}
+}
+
+private Map powerEndpoint() {
+	if (isFrientSensor()) {
+		[destEndpoint: 0x23]
+	} else {
+		[:]
+	}
+}
```
greens, line 371: too late to point this out now, but these calls only ever occur after we've already checked for `isFrientSensor` so they should have just been inlined I think

rboy1: Steven I thought about that initially but then realized that separating the endpoint mappings would lead to a cleaner way to reuse them across API's in future as more devices are added. So one could just have single call to `zigbee.configureReporting` for multiple devices with the same configuration requirement but with different sensor endpoints without any additional conditions like `isFrientSensor`

greens: The usual way we handle this is essentially to cross that bridge when we get to it (but I understand your rationale). The double-call to the model check really rubbed me the wrong way so I've already changed it in both DTHs.

PR #49513, opened 2020-11-09
WWST-5862, added FP for the ABL light

chw-choi:
. Add ZigBee manufacture name for ABL light (Samsung Electronics and Juno)


KKlimczukS:
Why are there so many other commits from 23rd and 25th of October ? Is it an old branch ? (since task WWST-5862 is already marked as DONE)

Please create a fresh branch for that change.

PR #49675, opened 2020-11-10
[ICP-13816] Qubino Flush 2 Relay - making sure energy usage is polled

PKacprowiczS:
For whatever reason, out of the blue, `response()` is "refusing" to send a command...

### DIFF
--- a/devicetypes/qubino/qubino-flush-2-relay.src/qubino-flush-2-relay.groovy
+++ b/devicetypes/qubino/qubino-flush-2-relay.src/qubino-flush-2-relay.groovy
@@ -291,7 +291,7 @@ def zwaveEvent(physicalgraph.zwave.commands.meterv3.MeterReport cmd, ep = null) 
 		String childDni = "${device.deviceNetworkId}:$ep"
 		def child = childDevices.find { it.deviceNetworkId == childDni }
 		child?.sendEvent(createMeterEventMap(cmd))
-		response(encap(zwave.meterV3.meterGet(scale: 0x00), ep))
+		sendHubCommand(encap(zwave.meterV3.meterGet(scale: 0x00), ep))
 	}
 }
 
--- a/devicetypes/qubino/qubino-flush-2-relay.src/qubino-flush-2-relay.groovy
+++ b/devicetypes/qubino/qubino-flush-2-relay.src/qubino-flush-2-relay.groovy
@@ -281,18 +281,24 @@ private changeSwitch(endpoint, cmd) {
 }
 
 def zwaveEvent(physicalgraph.zwave.commands.meterv3.MeterReport cmd, ep = null) {
+	def result = []
+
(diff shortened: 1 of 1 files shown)


greens:
I can't see why this would be the case, but if it is, there are several other places you'd need to fix it as well.

dkirker:
All of the `zwaveEvent` functions are only ever called from `parse`, unless you are doing some weird stuff. As long as `parse` is returning the correct data (DM takes the return value from `parse` and processes that), you shouldn't need the `sendHubCommand`. Does the hub log show that the command is being sent to the hub to send? Is the data being returned from `parse` valid? Is there a possible timing edge case?

If you are doing everything correct you shouldn't need to force call `sendHubCommand` from the `parse`->`zwaveEvent` pipeline. If you are returning correct data and it isn't being sent to the hub then we should file a ticket to have that looked at instead of forcing the `sendHubCommand` (this would also be affecting other things).

dkirker:
Try without the `response()`. You shouldn't need that when returning from `parse`, so the `zwaveEvent` functions need not call it on protocol commands they are returning.

PKacprowiczS:
Leaving out response() is definitely not a way to do it, but after some time I came up with another idea which solves the issue...

It seems like response() needs to be inside a list to be considered as useful resource. Why? Dunno, but that's the way it is.

dkirker:
`response()` is just a wrapper for `new HubAction()`. I *think* the `parse` execution chain eventually just yanks out the string (`toString()`) later on (for Z-Wave and Zigbee). The passed in `physicalgraph.zwave.Command` to `response` has `format()` called on it.

The result of `parse` is expected to be a `Map` (which would be the result of `createEvent` -- the creation of an event as a result of a state change from the device) or a `List` (a list of commands to send to the device and/or results from `createEvent` to create new events based on state changes from the device).

I'd just do
```
if (cmd.scale == 2) {
	[response(encap(zwave.meterV3.meterGet(scale: 0x00), ep))]
}
```
It's cleaner and easier to read.

Actually, I'd just re-write the whole function as:
```
def zwaveEvent(physicalgraph.zwave.commands.meterv3.MeterReport cmd, ep = null) {
	def result = []

	log.debug "Meter ${cmd}" + (ep ? " from endpoint $ep" : "")

	if (ep == defaultEndpoint()) {
		result << createEvent(createMeterEventMap(cmd))
	} else if (ep) {
		String childDni = "${device.deviceNetworkId}:$ep"
		def child = childDevices.find { it.deviceNetworkId == childDni }

		child?.sendEvent(createMeterEventMap(cmd))
	}
	// Query energy when we receive power reports
	if (cmd.scale == 2) {
		result << response(encap(zwave.meterV3.meterGet(scale: 0x00), ep))
	}

	result
}
```

dkirker:
Poked around a little more and updated my comment. :)

dkirker (approved):

PR #49685, opened 2020-11-11
Fix Smart Switch 7 US fingerprint

Aeotec-ccheng:
Model identifier is changed from "00AF" to "0017", the new model number for Series 700 Smart Switch 7 US is ZWA023-A with Product ID 0x0017.

Certified product ID can be found here: https://products.z-wavealliance.org/products/3844?selectedFrequencyId=-1


Aeotec-ccheng:
@greens - Could you take a look at this one? 

The Smart Switch 7 which will be released soon for US Frequency won't automatically use this device handler without the change in fingerprint for Smart Switch 7 US ZWA023. (It otherwise defaults to the Fortrezz valve DTH which is a bit odd)

greens (commented):
@Aeotec-ccheng Is the EU fingerprint correct?

Aeotec-ccheng:
@greens - EU Fingerprinting is correct, that line doesn't need to change, only the US version needs to change.

greens (approved):

PR #49771, opened 2020-11-11
CHAD-5726 Update Thermostat DTHs to discrete capabilities

greens:
This is the first step towards transitioning away from the deprecated monolithic Thermostat capability.


devicetypes/smartthings/fidure-thermostat.src/fidure-thermostat.groovy
```diff
@@ -12,6 +12,11 @@ metadata {
 		capability "Actuator"
 		capability "Temperature Measurement"
 		capability "Thermostat"
+		capability "Thermostat Heating Setpoint"
+		capability "Thermostat Cooling Setpoint"
+		capability "Thermostat Operating State"
```
tpmanley, line 17: This DTH has a custom attribute called `thermostatOperatingState` that I think should be removed. Also the values it uses in `thermostatOperatingState` events don't match the capability exactly and should be updated.

dkirker (approved):

PR #49935, opened 2020-11-13
[WWST-7147] Fingerprint to Qubino Mini Dimmer

MWierzbinskaS:
@greens 


greens (commented):
Are all of the preferences supported in this new DTH supported by this device?

Does this device need specific metadata like the flush dimmer?

MWierzbinskaS:
This device supports preferences in this DTH except those described as: "applies to Qubino Flush Dimmer only".
This device does not require specific metadata, it works similarly to other Qubino dimmers.

greens (approved):

PR #50407, opened 2020-11-19
[WWST-6946] Somfy Glydea Ultra - fingerprint update and tweaks

PKacprowiczS:
@greens @tpmanley @SmartThingsCommunity/srpol-pe-team

### DIFF
--- a/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
+++ b/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
@@ -32,7 +32,7 @@ metadata {
 		fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0006, 0008, 0102", outClusters: "000A", manufacturer: "Feibit Co.Ltd", model: "FTB56-ZT218AK1.8", deviceJoinName: "Wistar Window Treatment" //Wistar Curtain Motor(CMJ)
 		fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0102", outClusters: "0003", manufacturer: "REXENSE", model: "KG0001", deviceJoinName: "Window Treatment" //Smart Curtain Motor(BCM300D)
 		fingerprint profileId: "0104", inClusters: "0000, 0003, 0004, 0005, 0102", outClusters: "0003", manufacturer: "REXENSE", model: "DY0010", deviceJoinName: "Window Treatment" //Smart Curtain Motor(DT82TV)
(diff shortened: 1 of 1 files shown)


greens:
I think this is the wrong ticket.

Also could you explain what's going on here a little bit?

PKacprowiczS:
Oh yeah, sorry about that.

Tweak is related with ICP-13709. Originally, this DTH ignores level reports which have the same value as the current level. 
Thing is, this Somfy device won't move if change is equal or less than 3 percentage points. If user tries so, device just reports its current level, which, without this tweak, is ignored causing timeout.

Why using states for extra conditions in favor of simply ignoring setLevel command when necessary? 
Well, curtain rail we have is pretty short, so change resolution is pretty bad. I'm just worried that with different equipment an user might be able to actually change in a more accurate way and I don't want to deprive anybody of this privilege.

devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
```diff
@@ -122,8 +122,9 @@ def parse(String description) {
 def levelEventHandler(currentLevel) {
 	def lastLevel = device.currentValue("level")
 	log.debug "levelEventHandle - currentLevel: ${currentLevel} lastLevel: ${lastLevel}"
-	if (lastLevel == "undefined" || currentLevel == lastLevel) { //Ignore invalid reports
+	if ((lastLevel == "undefined" || currentLevel == lastLevel) && state.invalidSameLevelEvent) { //Ignore invalid reports
 		log.debug "Ignore invalid reports"
+		state.invalidSameLevelEvent = true
```
KKlimczukS: Could You tell why do invalidSameLevelEvent is set to true here if the above condition already checks it ?
Did You mean to check first if that state variable is defined ? (if it will be defined and set to false then the whole condition might not be met anyway).

PKacprowiczS: Oh yeah, thanks. It was meant for `else { ... }` block.

devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
```diff
@@ -163,6 +164,12 @@ def open() {
 
 def setLevel(data, rate = null) {
 	log.info "setLevel()"
+	def currentLevel = device.currentValue("level")
+
+	if (isSomfy() && Math.abs(data - currentLevel) <= GLYDEA_MOVE_THRESHOLD) {
+		state.invalidSameLevelEvent = false
```
KKlimczukS, line 170: should state.invalidSameLevelEvent be set to true in this case?

PKacprowiczS: No, by setting it to `false`, we're basically saying "hey, that next report which might has same level as now, should not be ignored"

greens (approved):

PR #51389, opened 2020-11-30

frient-design:


frient-design:
Adding one more frient smart plug min variant (Danish version)

KKlimczukS (approved):

PR #51473, opened 2020-11-30
ICP-5864 Add health check to original SmartSense Multi

greens:


devicetypes/smartthings/smartsense-multi.src/smartsense-multi.groovy
```diff
@@ -80,6 +81,10 @@ metadata {
 	}
 }
 
+def updated() {
+	sendEvent(name: "checkInterval", value: 60 * 60 + 2 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
```
tpmanley: How did you arrive at this number? My multi's are checking in every 2.5 minutes so I think it could be shorter. I'd probably go with 12 minutes like many of our Zigbee sensors.

greens: Didn't see any reporting configuration in the DTH so I just picked a big number. I'll change to 12 minutes.

tpmanley (approved):

PR #51554, opened 2020-12-01
BUG-1315: Remove checkInterval event from aeon key fob

cjswedes:
https://smartthings.atlassian.net/browse/BUG-1315


greens (commented):
Is there a ticket for this? Convention is to put the ticket number in the commit message and PR title.

kianooshkarami:
> Is there a ticket for this? Convention is to put the ticket number in the commit message and PR title.

There is a BUG ticket, Would you like to create a CHAD ticket or will the BUG ticket suffice?

greens:
Any jira ticket is fine.

cjswedes:
The ticket link is already in the commit message, and I just updated the the title to include the ticket number.

greens (approved):
This looks good, but just a heads up that this will not "fix" existing devices. Only new installs.

cjswedes:
We will have to deploy the DTH so that each device is re-initialized. `updated()` will be called when this happens which in turn calls `initialize()` where my change to the tracking is.

PR #52021, opened 2020-12-07
ICP-13878 - skips rounding of Danfoss Ally Thermostat temperature values

KKlimczukS:
ICP-13878 - skips rounding of Danfoss Ally Thermostat temperature values

@greens 


greens (commented):
I think we just want to round to 1 decimal place if the device is using Celsius and to whole numbers if the device is using Fahrenheit.

KKlimczukS:
> I think we just want to round to 1 decimal place if the device is using Celsius and to whole numbers if the device is using Fahrenheit.

@greens 
Done. Now it should look more concise.

greens (approved):

PR #52115, opened 2020-12-08
BUG-1315: Remove checkInterval event from aeon key fob (#51554)

greens:
https://smartthings.atlassian.net/browse/BUG-1315


greens:
cherry-picking this to staging so it's deployed this week instead of next

kianooshkarami (approved):

PR #52121, opened 2020-12-08
Disable threshold settings

Aeotec-ccheng:
Found error with parameter setting used, Change Parameter 90 to Parameter 3 which is used for threshold settings on HEM Gen5.


greens:
@Aeotec-ccheng These config parameters, right? https://products.z-wavealliance.org/products/1289/configs

Setting parameter 3 to 0 looks like it would turn _off_ automatic reporting. Am I reading that wrong?

Aeotec-ccheng:
> @Aeotec-ccheng These config parameters, right? https://products.z-wavealliance.org/products/1289/configs
> 
> Setting parameter 3 to 0 looks like it would turn _off_ automatic reporting. Am I reading that wrong?

Parameter 3 = 0 disables selective reporting, this disables parameters 4 - 11 which sets the selective report settings. 

Automatic/timed reporting relies on Parameter 101/111 based on the current settings. Timed reporting operates separately from Parameter 3.

greens (approved):

PR #52125, opened 2020-12-09
DevWs for Vision-Elec. Technology Co., Ltd. containing containing Vision 4-in-1 sensor

vision-raytseng:


devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,552 @@
...
+  definition (name: "Vision 4-in-1 Motion Sensor", namespace: "vision-raytseng", author: "Ray Tseng", vid: "generic-motion-8", ocfDeviceType: "x.com.st.d.sensor.motion") {
+    capability "Battery"
+    capability "Motion Sensor"
+    capability "Relative Humidity Measurement"
+    capability "Temperature Measurement"
+    capability "Illuminance Measurement"
+    capability "Tamper Alert"
+    capability "Health Check"
+    
+    fingerprint mfr:"0109", prod:"2021", model:"2112", deviceJoinName: "Vision ZP3113-7 4-in-1 Motion"
```
KKlimczukS: Please change deviceJoinName to "Vision Multipurpose Sensor" in this fingerprint and add Raw Description in a comment after the fingerprint so it would look like:
fingerprint mfr:"0109", prod:"2021", model:"2112", deviceJoinName: "Vision Multipurpose Sensor" // Raw Description // Raw Description >>paste the raw description here<<

(Raw description can be found at the Smartthings graphapi in device details)

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,552 @@
+/**
+ *  Vision 4-in-1 Motion Sensor
+ *
+ *  Author: Ray Tseng
+ */
+metadata {
```
KKlimczukS, line 6: Please reformat the whole code and fix indentation (change 4 spaces to 1 tab) - applies to a whole file.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,552 @@
...
+    capability "Relative Humidity Measurement"
+    capability "Temperature Measurement"
+    capability "Illuminance Measurement"
+    capability "Tamper Alert"
+    capability "Health Check"
+    
+    fingerprint mfr:"0109", prod:"2021", model:"2112", deviceJoinName: "Vision ZP3113-7 4-in-1 Motion"
+  }
+
+  tiles(scale: 2) {
```
KKlimczukS: Please remove the whole tiles section since tiles are not supported in the SmartThings app.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,552 @@
...
+  wakeUp.title = "Wake up interval [Hours]"
+  wakeUp.description = ""
+  wakeUp.type = "number"
+  wakeUp.range = "1..4660"
+  wakeUp.default = 24
+  
+  return wakeUp."${text}"
+}
+
+private getCommandClassVersions() {
```
KKlimczukS, line 219: CommandClassVersions are not obligatory. You may use zwave.parse(description) instead.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,552 @@
...
+  [min: 30, max: 39.99, multiplier: 4.981],
+  [min: 40, max: 49.99, multiplier: 5.194],
+  [min: 50, max: 59.99, multiplier: 6.016],
+  [min: 60, max: 69.99, multiplier: 4.852],
+  [min: 70, max: 79.99, multiplier: 4.836],
+  [min: 80, max: 89.99, multiplier: 4.613],
+  [min: 90, max: 100, multiplier: 4.5]
+]}
+
+def logInfo(msg) {
```
KKlimczukS: Please remove this empty method.

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
KKlimczukS: What about the Raw Description in a comment after the fingerprint ?

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+	definition (name: "Vision 4-in-1 Motion Sensor", namespace: "vision-raytseng", author: "Ray Tseng", vid: "generic-motion-8", ocfDeviceType: "x.com.st.d.sensor.motion") {
+		capability "Battery"
+		capability "Motion Sensor"
+		capability "Relative Humidity Measurement"
+		capability "Temperature Measurement"
+		capability "Illuminance Measurement"
+		capability "Tamper Alert"
+		capability "Health Check"
+		
+		fingerprint mfr:"0109", prod:"2021", model:"2112", deviceJoinName: "Vision Multipurpose Sensor" // Vision ZP3113-7 4-in-1 Motion
```
KKlimczukS: What about the Raw Description in a comment after the fingerprint ?

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+			description: getWakeUpInfo("description"),
+			type: getWakeUpInfo("type"), range: getWakeUpInfo("range"),
+			defaultValue: getWakeUpInfo("default"),
+			required: true, displayDuringSetup: true
+	}
+}
+
+def installed() {
+	def cmds = []
+	def lightEnableDefault = [:]
```
KKlimczukS: Could You remove this map since it's not used ?

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+		logDebug "Unable to parse description: ${description}"
+	}
+	
+	return result
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
+	def cmds = []
+	
+	cmds += configure()
```
KKlimczukS, line 245: Could You explain why do You call configure() on every WakeUpNotification ?

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+def zwaveEvent(physicalgraph.zwave.commands.notificationv3.NotificationReport cmd) {
+	def result = []
+	
+	if (cmd.notificationType == 0x07) {
+		if (cmd.eventParametersLength) {
+			cmd.eventParameter.each {
+				if (it == 0x03) {
+					result = createEvent(name: "tamper", value: "clear") 
+				}
+				else if( it == 0x08) {
```
KKlimczukS: Please use this convention of formatting (applies to all occurrences)
`} else if() {
...
}`

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+def zwaveEvent(physicalgraph.zwave.commands.sensormultilevelv5.SensorMultilevelReport cmd) {
+	def map = [:]
+	
+	switch (cmd.sensorType) {
+		case 0x01:
+			map.name = "temperature"
+			map.value = cmd.scaledSensorValue
+			map.unit = cmd.scale == 0 ? "C": "F"
+			break
+
```
KKlimczukS: Please remove all those empty lines (those ones after break statements) in this switch block.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,510 @@
...
+	}
+	cmds << zwave.wakeUpV2.wakeUpNoMoreInformation().format()
+	
+	return cmds ? response(cmds) : []
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpIntervalReport cmd) {
+	if (cmd.nodeid == zwaveHubNodeId) {
+		def interval = [:]
+		interval."${getWakeUpInfo("name")}" = second2Hour(cmd.seconds)
```
KKlimczukS: Since second2Hour() is used only once, can You just inline this code here ?

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: This is battery-powered device that only wakes up and communicates when an event has occurred.
If the smartThings hub would like to communicate with this device, the smartThings hub must wait for this device awake. When this device wakes up, it will send WakeUpNotification to inform the smartThings hub it is ready to communicate, and then the smartThings hub could send the configuration from the user settings to this device. So I call the configure() on every receiving WakeUpNotification to check that command the user configured needs to be sent to this device.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: Okay, I modified it.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: OKay, I have removed.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
KKlimczukS: That way configure() will be called regardless changes in settings and every few hours the device will receive the set of commands even if user doesn't intent to configure the device. You can improve that approach by using another state variables to determine if the configuration has to be done or not (for example: "Z-Wave Motion Sensor")

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,495 @@
...
+	definition (name: "Vision 4-in-1 Motion Sensor", namespace: "vision-raytseng", author: "Ray Tseng", vid: "generic-motion-8", ocfDeviceType: "x.com.st.d.sensor.motion") {
+		capability "Battery"
+		capability "Motion Sensor"
+		capability "Relative Humidity Measurement"
+		capability "Temperature Measurement"
+		capability "Illuminance Measurement"
+		capability "Tamper Alert"
+		capability "Health Check"
+		
+		fingerprint mfr:"0109", prod:"2021", model:"2112", deviceJoinName: "Vision Multipurpose Sensor" // zw:Ss2a type:0701 mfr:0109 prod:2021 model:2112 ver:32.32 zwv:7.13 lib:03 cc:5E,22,98,9F,6C,55 sec:85,59,80,70,5A,7A,87,8E,72,71,73,31,86,84
```
KKlimczukS: Please add "Raw description" (literally), so it would look:
...deviceJoinName: "Vision Multipurpose Sensor" // Raw description: zw:Ss2a type...

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: Actually, the device does not receive the set of commands every few hours if the settings have not changed, since the configure() checks which configuration was changed by the user and then sends the command to the device. You could check the code of updated() and configure(). In updated(), it will set the corresponding flag if user changes in settings, and then in configure() it will check the flag and decide which command needs to be sent to the device. This mechanism is the same as you said.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: Okay, I have added "Raw description" comment in fingerprint.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: I do not modify it, since I would like to specify the command class version for used.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,495 @@
...
+	parameter.parameter6options = ["Highest", "Higher", "High", "Medium", "Low", "Lower", "Lowest"]
+	parameter.parameter6enumMap = ["Highest": 1, "Higher": 2, "High": 3, "Medium": 4, "Low": 5, "Lower": 6, "Lowest": 7]
+	parameter.parameter6default = "Medium"
+	parameter.parameter6size = 1
+	
+	parameter.parameter7name = "LedDispMode"
+	parameter.parameter7title = "LED display mode"
+	parameter.parameter7description = ""
+	parameter.parameter7type = "enum"
+	parameter.parameter7options = ["LED off when Temperature report/Motion active",
```
KKlimczukS: Please, could You fix indentations in parameter7options and parameter7enumMap ?

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: Okay, I will fix the indentations in parameter7options and parameter7enumMap.

KKlimczukS: Looks better now, but now spacing has to be fixed (replace 4 spaces to 1 tab).

KKlimczukS (changes requested):

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: These was used 8 tabs and 3 spaces to make these indentations originally before I fixed today, but it looks not alignment. So I replace the 7 tabs to 28 spaces to fix this look. If I replace 4 spaces to 1 tab, it will be the same as previous committed code and look not good.

KKlimczukS: Since indentation in this file contains tabs - please use tabs in this case too (it doesn't have to be super aligned with the first element)

vision-raytseng: So these indentations are allowed?
```
	parameter.parameter7options = ["LED off when Temperature report/Motion active",
		"LED blink when Temperature report/Motion active",
		"LED blink when Motion active/LED off when Temperature report"]
	parameter.parameter7enumMap = ["LED off when Temperature report/Motion active": 1,
		"LED blink when Temperature report/Motion active": 2,
		"LED blink when Motion active/LED off when Temperature report": 3]
```

KKlimczukS: Yes, it would look better this way.

vision-raytseng: I got it, thank you.

greens:
@vision-raytseng just checking, other than preferences/configuration, what parts of the handler code for this device would not fit in https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zooz-4-in-1-sensor.src/zooz-4-in-1-sensor.groovy ?

vision-raytseng:
@greens other than preferences and configurations, some differences are list as followed:
1. This DTH handles the commands of wakeUpIntervalSet and wakeUpIntervalReport.
2. This DTH does not send batteryGet command on every receiving wakeUpNotification, but it sends batteryGet command if it is more than 8 hours from receiving last battery report.
3. The temperature unit is according to the user's settings in this DTH, not according to location.
4. This DTH does not use basicSet command to handle motion event, just use notification command to handle motion event.
5. This DTH supports two secure encapsulation commands (secure v1 and secure v2), and which secure encapsulation command was used depends on adding to smartThings hub by which secure level. So this DTH does not specify any secure encapsulation command, and to use which secure encapsulation command to communicate with this device is decided by smartThings hub.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,495 @@
...
+}
+
+def getConfigurationNumber() {
+	return [1, 2, 3, 4, 5, 6, 7, 8]
+}
+
+def getConfigurationInfo(num, text) {
+	def parameter = [:]
+	
+	parameter.parameter1name = "TemperatureUnit"
```
greens: wouldn't it make more sense to do all of this as a map of maps? Would certainly make iteration etc a lot simpler for you.

i.e.: 

```
def parameter = [:]
parameter[1] = {name: "TemperatureUnit", "title": "Temperature Unit [C/F]" ..}
```

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,495 @@
...
+		0x8E: 3,
+		0X31: 5,
+		0x71: 3,
+		0x73: 1,
+		0x98: 1,
+		0x9F: 1,
+		0x6C: 1,
+		0x55: 2,
+		0x86: 3,
+		0x84: 2
```
greens: you do not need to include all of these, only the ones whose messages you are parsing/handling

vision-raytseng: Okay, I fixed it. Now I only include these command version that I parse and handle.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: I think not. Using a map of maps does not change the iteration way in my case, since I still use the index to make iteration to get key and value. In my case, I use the parameter name as a part of key name of state (i.e. state."${getConfigurationInfo(idx, "name")}Refresh" and state."${getConfigurationInfo(idx, "name")}") to store which configuration is changed by user and check which configuration command needs to be sent.
I think a map of maps that maybe looks better but here it does not make iteration more simpler for me.
I paste the snippet code of this function using a map of maps.
```
def getConfigurationInfo(num, text) {
	def parameter = [:]
	parameter[1] = [
		name: "TemperatureUnit", 
		title: "Temperature Unit [°C/°F]",
		description: "", 
		type: "enum", 
		options: ["°C", "°F"], 
		enumMap: ["°C": 0, "°F": 1], 
		default: "°C", 
		size: 1
	]
	...
	parameter[8] = [
		name: "RetryTimes", 
		title: "Motion notification retry times", 
		description: "",
		type: "number", 
		options: "0..10", 
		enumMap: [], 
		default: 3, 
		size: 1
	]
	return parameter[num.toInteger()]."${text}"
}
```

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,480 @@
...
+		map.value = cmd.batteryLevel
+	}
+	
+	state.lastBatteryReport = new Date().time
+	createEvent(map)
+}
+
+def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
+	def size = getConfigurationInfo(cmd.parameterNumber, "size")
+	def value = array2Integer(cmd.configurationValue)
```
greens: here and elsewhere we've provided as a convenience the `scaledConfigurationValue` where this conversion has already been done

conversely you can pass integers directly in configuration sets as well

you should be able to do away with your conversion functions

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
```diff
@@ -0,0 +1,480 @@
...
+	getConfigurationNumber().each { idx ->
+		if (config?."${getConfigurationInfo(idx, "name")}" != null) {
+			if (config?."${getConfigurationInfo(idx, "name")}" == state."${getConfigurationInfo(idx, "name")}") {
+				if (state."${getConfigurationInfo(idx, "name")}Refresh" == true) {
+					state."${getConfigurationInfo(idx, "name")}Refresh" = false
+				}
+			}
+		}
+	}
+}
```
greens: if you wrote your configuration report hander function differently and refactored your configuration parameters as a map of maps, I think this method could instead be written:

```
config.each { state[key].Refresh = false}
```

where `state[x]` is a map containing a value with key `value` and a boolean with key `Refresh` and your configuration report handler returns a list of numbers.

Because as far as I can tell you're not actually comparing the value you've received from the configuration report to the value the user has set. Any configuration report indicating a parameter number the user has set will get marked as not needing to be refreshed, regardless of the actual value reported.

This is just one example here. I'm not really satisfied with how you've handled configuration options. For a more acceptable example take a look at any of the Fibaro DTHs that include preferences.

vision-raytseng: > Because as far as I can tell you're not actually comparing the value you've received from the configuration report to the value the user has set. Any configuration report indicating a parameter number the user has set will get marked as not needing to be refreshed, regardless of the actual value reported.

Actually, I will compare the value I am received from the configuration report to the value the user has set. If these values are different, I will send the configuration value the user has set on next time I received the wake up notification until these value are the same. 

Here is my configuration snippet code, and its procedure. Please look this, and give me a suggestion, thanks.

Variable definition:
1. _**state."${getConfigurationInfo(idx, "name")}"**_: The values the user has set.
2. _**state."${getConfigurationInfo(idx,"name")}Refresh"**_:  This configuration is needed to be sent or not: true is needed to be sent while I received the wake up notification, false is not.

Procedure:
1. After the user changed the configuration value, I will check which configuration value is changed. If this configuration value has changed, I will store this configuration value to corresponding _**state."${getConfigurationInfo(idx, "name")}"**_, and set _**state."${getConfigurationInfo(idx, "name")}Refresh"**_ to true to indicate this needs to be updated to this device. Here are **updated()** and **configurationUpdate(settings)**.
2. I will send this configuration set with _**state."${getConfigurationInfo(idx, "name")}"**_ if its _**state."${getConfigurationInfo(idx, "name")}Refresh"**_ is true after I received the wake up notification. And then I will send the configuration get to check the configuration report value is the same or not. Here are z**waveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd)** and **configure()**.
3. I will compare the configuration report I received to _**state."${getConfigurationInfo(idx, "name")}"**_ from the user has set. I will set _**state."${getConfigurationInfo(idx, "name")}Refresh"**_ to false if these are the same. Here are **zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd)** and **configurationCheck(config)**.
4. Loop from step 2 to step 3 until these configuration value are the same.

```
def updated() {
	configurationUpdate(settings)
	...
}

def configurationUpdate(config) {
	getConfigurationNumber().each { idx ->
		if (config?."${getConfigurationInfo(idx, "name")}" != null) {
			if (state."${getConfigurationInfo(idx, "name")}" != config?."${getConfigurationInfo(idx, "name")}") {
				state."${getConfigurationInfo(idx, "name")}" = config?."${getConfigurationInfo(idx, "name")}"
				state."${getConfigurationInfo(idx, "name")}Refresh" = true
			}
		}
	}
}

def zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
	def cmds = []
	
	cmds += configure()
	if (cmds) {
		cmds << "delay 5000"
	}
	cmds << zwave.wakeUpV2.wakeUpNoMoreInformation().format()
	
	return cmds ? response(cmds) : []
}

def configure() {
	def cmds = []
	def value
	def optionValue
	...
	getConfigurationNumber().each { idx ->
		if (state."${getConfigurationInfo(idx, "name")}Refresh" == true) {
			switch (getConfigurationInfo(idx, "type")) {
				case "enum":
					optionValue = getConfigurationInfo(idx, "enumMap").find { it.key == state."${getConfigurationInfo(idx, "name")}" }?.value
					value = integer2Array(optionValue, getConfigurationInfo(idx, "size"))
					break
				case "number":
					value = integer2Array(state."${getConfigurationInfo(idx, "name")}", getConfigurationInfo(idx, "size"))
					break
			}
			if (value != null) {
				cmds << zwave.configurationV2.configurationSet(parameterNumber: idx, defaultValue: false, size: getConfigurationInfo(idx, "size"), configurationValue: value).format()
				cmds << zwave.configurationV2.configurationGet(parameterNumber: idx).format()
				if (idx == 1) {
					cmds << zwave.sensorMultilevelV5.sensorMultilevelGet(sensorType: 0x01, scale: optionValue?:0x00).format()
				}
				value = null
			}
		}
	}
	...
	return cmds ? delayBetween(cmds, 500) : []
}

def zwaveEvent(physicalgraph.zwave.commands.configurationv2.ConfigurationReport cmd) {
	def size = getConfigurationInfo(cmd.parameterNumber, "size")
	def value = array2Integer(cmd.configurationValue)
	def config = [:]
	
	if (size && size == cmd.size) {
		switch (getConfigurationInfo(cmd.parameterNumber, "type")) {
			case "enum":
				def optionName = getConfigurationInfo(cmd.parameterNumber, "enumMap").find { it.value == value}?.key
				if (optionName) {
					config."${getConfigurationInfo(cmd.parameterNumber, "name")}" = optionName
				}
				break
			case "number":
				config."${getConfigurationInfo(cmd.parameterNumber, "name")}" = value
				break
		}
		
		if (config) {
			configurationCheck(config)
		}
	}
	[]
}

def configurationCheck(config) {
	getConfigurationNumber().each { idx ->
		if (config?."${getConfigurationInfo(idx, "name")}" != null) {
			if (config?."${getConfigurationInfo(idx, "name")}" == state."${getConfigurationInfo(idx, "name")}") {
				if (state."${getConfigurationInfo(idx, "name")}Refresh" == true) {
					state."${getConfigurationInfo(idx, "name")}Refresh" = false
				}
			}
		}
	}
}
```

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
vision-raytseng: Okay, I already use **scaledConfigurationValue** to replace my conversion functions. By the way, I can't see the **scaledConfigurationValue** in [class Configuration Report Command](https://graph.api.smartthings.com/ide/doc/zwave-utils.html#configurationV2), but **scaledConfigurationValue** still works fine in class Configuration Report Command.

devicetypes/vision-raytseng/vision-4-in-1-motion-sensor.src/vision-4-in-1-motion-sensor.groovy
greens: Okay so you are verifying they've been set.

My mistake, but this is just reinforcing my point that the code you've written is hard to follow because you're using integers combined with strings as your dictionary keys.

Iterating from 1..8 and checking to see if each param was in your updated configuration report is much more laborious and confusing than just structuring the output of your configuration report such that you could simply iterate over that itself.

There are other existing DTHs that solve this problem more elegantly: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/fibargroup/fibaro-dimmer-2-zw5.src/fibaro-dimmer-2-zw5.groovy for example.

vision-raytseng: I think I understand your meaning. Your approach simplifies this procedure and make it more elegantly.
I refactored configuration parameter using a map of maps, and refactored wake up info parameter using a map. Please review it. Thank you very much.

greens (approved):

greens:
@KKlimczukS this now looks good to me, just waiting on your approval

KKlimczukS (approved):

PR #52199, opened 2020-12-09
ICP-13858 - fix for setting the thermostat mode to OFF

KKlimczukS:
ICP-13858 - fix for setting the thermostat mode to OFF

@greens
Please, could You review it ?


greens (changes requested):
You'll have the same bug for fan mode as well so you might want to fix that as well. Also you control every place these functions are called so I don't think you need the null check at all, frankly.

KKlimczukS:
> You'll have the same bug for fan mode as well so you might want to fix that as well. Also you control every place these functions are called so I don't think you need the null check at all, frankly.

@greens Thanks for pointing that out. I fixed that one too (since methods are called with proper parameters - i removed the null check according to Your suggestion)

greens (approved):

PR #52444, opened 2020-12-11
[ICP-13654] Added specific UI metadata for Qubino dimmers

MWierzbinskaS:
@greens 

### DIFF
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -12,7 +12,7 @@
  *
  */
 metadata {
-	definition(name: "Qubino Dimmer", namespace: "qubino", author: "SmartThings", mnmn: "SmartThings", vid:"generic-dimmer-power-energy", ocfDeviceType: "oic.d.switch", runLocally: false, executeCommandsLocally: false) {
+	definition(name: "Qubino Dimmer", namespace: "qubino", author: "SmartThings", mnmn: "SmartThings", vid:"qubino-dimmer-power-energy", ocfDeviceType: "oic.d.switch", runLocally: false, executeCommandsLocally: false) {
 		capability "Actuator"
 		capability "Configuration"
 		capability "Energy Meter"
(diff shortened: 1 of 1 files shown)


devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -12,7 +12,7 @@
  *
  */
 metadata {
-	definition(name: "Qubino Dimmer", namespace: "qubino", author: "SmartThings", mnmn: "SmartThings", vid:"generic-dimmer-power-energy", ocfDeviceType: "oic.d.switch", runLocally: false, executeCommandsLocally: false) {
+	definition(name: "Qubino Dimmer", namespace: "qubino", author: "SmartThings", mnmn: "SmartThings", vid:"qubino-dimmer-power-energy", ocfDeviceType: "oic.d.switch", runLocally: false, executeCommandsLocally: false) {
```
greens, line 15: What's different about this metadata?

MWierzbinskaS: @greens It changes dimming range in UI metadata from 1-100% to 2-100% (More info why we wanted to set 2% as minimum, please see this PR comments: https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/37228)

But as I checked, those metadata are still not merged. I will change its status as draft

greens: I thought @KKlimczukS 's changes made it so that despite the device's internal range being 2-100, we scaled our commands so that the user still saw 1-100?

MWierzbinskaS: Yes, Konrad set minimum dimming level as 2% in the device configuration, but because device starts recalculating those values when the range is different we had to remove this configuration (ICP-13654, ICP-13361). We decided to set this range by UI metadata.

KKlimczukS: @greens We had to find another approach/workaround and constraint those min dimming values from the application (DTH and UI) side instead of setting this configuration parameter on the device - that's was the only way to make it work and avoid dimmer level jumping (on the UI slider, in a quite non deterministic way).

greens: okay, well then whenever you're ready @MWierzbinskaS

greens (approved):

PR #52598, opened 2020-12-13
Initial support for ZRC-90US first 6 buttons only

dluyer:
The current device handler for this device is not useful in the new SmartThings app.

This sets up a device handler which almost works, for 6 of the 8 buttons on the device.  (Changing the number to 8 makes this fully functional in the classic app, but even more broken in the new app.)

Attempts to save actions from the new app still fail with "A network or server error occurred, try again later", but at least the UI elements exist to attempt to set them (for 6 buttons).


dluyer:
Note as to failure mode in new app with 8 buttons and type generic-8-button:
- app displays all 8 buttons (great!)
- but it also displays the 'main' button (wrong)
- and, for all the child buttons, it displays an insane number of possible actions (vs the correct list, displayed for the non-existent main button)
- trying to set an action on any button crashes the app

dluyer:
And, of course, everything looks correct in the web UI.  supportedButtonValues is displayed and correct for all child buttons.

PR #52774, opened 2020-12-15
Adding the device handler for Zigbee Metering Plug Power Consumption Report

dwd-kwon:


greens (approved):

lukhong:
All Dawon devices supported by this DTH should have their unique vid. The purpose of this DTH with powerConsumptionReport is for SmartThings Energy service. Energy Service wants to allow only pre-contracted devices. For the 3rd party device like this Dawon plug we can only distinguish it by vid. Energy Service will add the Dawon vid to the whitelist to allow it. But this DTH is generic one and has the generic vid. We don't want any other vendor's device is coming to Energy Service yet.
@greens, for this purpose I think the vid should be hidden can this DTH go to private and not opened to public? Or any other suggestion?

greens:
@lukhong you can make a new PR targeted to private, put the correct vid on that one, and then make another PR that removes this file.

lukhong:
@greens Can Dawon get the access to the private repo to do that?

greens:
@lukhong Send me an email or slack messages with the github accounts you want to add.

PR #53142, opened 2020-12-16
[ICP-13896] Fixed lack of energy meter reports

MWierzbinskaS:
@greens 

This device did not send energy meter report itself and because it sends BasicReport every 15 minutes it was not asked about this report via ping and refresh.

### DIFF
--- a/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
+++ b/devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
@@ -327,6 +327,7 @@ def zwaveEvent(physicalgraph.zwave.Command cmd) {
 
 def zwaveEvent(physicalgraph.zwave.commands.basicv1.BasicReport cmd, ep = null) {
 	log.debug "BasicReport: ${cmd}"
+	sendHubCommand(encapCommands(getPowerMeterCommands()))
 	dimmerEvents(cmd)
 }
 


devicetypes/qubino/qubino-dimmer.src/qubino-dimmer.groovy
```diff
@@ -363,6 +363,8 @@ private dimmerEvents(physicalgraph.zwave.Command cmd, ep = null) {
 		result << createEvent(name: "level", value: cmdValue == 99 ? 100 : cmdValue)
 	}
 
+	sendHubCommand(encapCommands(getPowerMeterCommands()))
```
KKlimczukS: @MWierzbinskaS Is it needed for all Qubino dimmers? As far as I know Qubino Flush Dimmer and Qubino Flush Dimmer 0-10V are reporting power meter values [W] itself.

MWierzbinskaS: @KKlimczukS I moved this to BasicReport handling as we spoke

greens:
Are we certain there's not a way to do this automatically via association or configuration?

MWierzbinskaS:
There are no configuration parameters for energy meter reporting, only for power meter reporting.

Those are association groups for this device:
Group 1: Lifeline group (reserved for communication with the hub).
Group 2: Basic on/off (status change report for output), up to 16 nodes.
Group 3: Start level change/stop (status change report for I1 input).
Group 4: Multilevel set (status change report of the output).
Group 5: Multilevel sensor report (external temperature sensor report).

I don't know about any other solution

greens:
@KKlimczukS or @PKacprowiczS didn't you solve a similar problem by changing the association groups on another qubino device?

KKlimczukS:
> @KKlimczukS or @PKacprowiczS didn't you solve a similar problem by changing the association groups on another qubino device?

@greens 
All Qubino dimmers are supported by this DTH, so my previous change was in this DTH.

- In this case health check will not be able to trigger ping() method (where both power meter commands are sent) because device repots its presence by sending BasicReports quite often. 
- Qubino dimmers report power [W] themselves (after every on/off) and they don't have to be polled for that.
- It seems that Qubino DIN dimmer may report energy [kWh] a bit differently than other Qubino dimmers and it has to be polled when device reports power [W] - but I'm not 100% sure, because changes from this PR may also fix [kWh] reporting for other Qubino dimmers.

greens (approved):

PR #53789, opened 2020-12-23
Add Devicetype of RealilySwitch-Gen3 Zigbee Mode from Third Reality, Inc

Jackie-Kilby:
@greens @tpmanley 
This is our new switch product. Please help us to merge into published code.
We have tested it on "For me" option. It works well.
And if there is any other things we need to do, please tell us.


tpmanley (approved):

tpmanley:
@Jackie-Kilby thanks, this will be released on Jan 5

PR #53872, opened 2020-12-23
enable local execution for fibaro single switch 2

greens:


greens:
As far as I can tell, this can be turned on.

The only slight concern I have is that this DTH ignores basic reports, but those would be parsed locally as switch events on the hub.

PR #55184, opened 2021-01-08
DevWs for SHINA SYSTEM containing containing SiHAS OSM-300-Z

shinasys:


devicetypes/shinasys/sihas-osm-300-z.src/sihas-osm-300-z.groovy
```diff
@@ -0,0 +1,226 @@
...
+ *	Unless required by applicable law or agreed to in writing, software
+ *	distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *	WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *	License for the specific language governing permissions and limitations
+ *	under the License.
+ */
+import physicalgraph.zigbee.clusters.iaszone.ZoneStatus
+import physicalgraph.zigbee.zcl.DataType
+
+metadata {
```
KKlimczukS, line 19: I took a look at Your 3 pull requests: 
#55182, #55183, #55184

Since those devices share some functionalities, device settings and configuration - I looks like all of them may be handled by one DTH, so please:
- create one DTH (instead of 3) and name it -> SiHas Multipurpose sensor
- put all 3 fingerprints (for TSM-300Z, USM-300Z and OSM-300Z) - it's important to also add mnmn: "SmartThings", vid: <proper vid for each model> after each fingerprint
- You may need some tweaks in the code to distinguish device model, so helper methods (for example isTSM300Z() etc) , might be needed it use them later in a code, for example  if  isTSM300Z() {}  (in configuration and parse method to handle messages that come from devices)

That way helps to avoid code duplication and Your DTH will be more concise.

KKlimczukS (changes requested):

PR #55624, opened 2021-01-13
Add ThirdReality Zigbee Motion Sensor and Door Sensor.

Jackie-Kilby:
This is Motion Sensor and Door Sensor from ThirdReality, Inc.
We have tested the Motion Sensor and Door Sensor by "Publish for Me" and it works well.
Please help us to merge these code into SmartThingsPublic. Thanks a lot.
2021-01-15 I will close this PR. The two file has no dependence. Split them as two commits.


Jackie-Kilby:
<Close> The two file has no dependence. Split them as two commits then.

PR #55830, opened 2021-01-15
Add ThirdReality Motion Sensor DeviceType.

Jackie-Kilby:
@greens @tpmanley 
This is the device handler of Motion Sensor from ThirdReality, Inc.
We create a new groovy file. Because the battery value formula is different with ours from the existing motion-sensor device handler, and the capabilities list is not the same.
Please help to merge into public code. Thanks a lot.


greens (commented):
Maybe I'm missing something, but couldn't you have just added fingerprints to https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy ?

Jackie-Kilby:
> Maybe I'm missing something, but couldn't you have just added fingerprints to https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy ?

I'm so sorry that I cannot easily finish this little job. I encountered a problem, which is really confusing:
step-1. Click "Update from Repo" under "My Device Handlers" in Integration IDE and choose my Branch(ThirdReality-Motion). My DeviceHandler URL is https://graph-na04-useast2.api.smartthings.com/ide/devices .
step-2. Choose devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy to "Execute Update" with "Publish" checked.
step-3. Add my Device's fingerprint into the file, and add the Device by a SmartThings Hub. It was showed correctly (I can guarantee this by changing deviceJoinName). 
step-4. Open "Live Logging" to debug.
step-5. Click on the Device icon in App, and slide once to call "Refresh"
step-6. **I should see debug level log"Refreshing Values", but I didn't.**
step-7. After that, I tried the new groovy I created. It works well. (This will be a golden sample for this problem)
step-8. I copied all the code of my new goovy into "Zigbee Motion Detector", except definition name in metadata. It still not worked.

**In a word, I cannot call "refresh()" method by sliding on the screen in App.** This phenomenon can be reproduced. Because I retried again when I wrote the repro steps above.
Though I know that the problem may only occurs on me. I still cannot guarantee and verify the Device's performance without debugging method.
So can you help me on this problem first? @greens @tpmanley

greens:
@Jackie-Kilby if you watch the live logs during add, you should see the `refresh()` log message, as it is called as part of `configure`.

Jackie-Kilby:
@greens 
Yeah, I saw the debug log "refresh" at configure.
But, in my opinion, the capability "Refresh" means you can call `refresh()` method when you slide down on your App with your finger. Isn't that right? However, I can only get "refresh" log at configure step. When I slide on App, the "refresh" log is not shown. Is that a design on purpose?

greens:
I'll bring that up. I do not think that is on purpose. Functionality should be what you describe.

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
```diff
@@ -25,10 +25,12 @@ metadata {
 		capability "Refresh"
 		capability "Health Check"
 		capability "Sensor"
+		
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0001,0003,0500", outClusters: "0003", manufacturer: "eWeLink", model: "MS01", deviceJoinName: "eWeLink Motion Sensor" //eWeLink Motion Sensor
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001", manufacturer: "ORVIBO", model: "895a2d80097f4ae2b2d40500d5e03dcc", deviceJoinName: "Orvibo Motion Sensor" //Orvibo Motion Sensor
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000,0003,0500,0001,FFFF", manufacturer: "Megaman", model: "PS601/z1", deviceJoinName: "INGENIUM Motion Sensor" //INGENIUM ZB PIR Sensor
 		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000, 0003, 0500, 0001", outClusters: "0019", manufacturer: "HEIMAN", model: "PIRSensor-N", deviceJoinName: "HEIMAN Motion Sensor" //HEIMAN Motion Sensor
+		fingerprint profileId: "0104", deviceId: "0402", inClusters: "0000, 0001, 0500", outClusters: "0019", manufacturer: "Third Reality, Inc", model: "3RMS16BZ", deviceJoinName: "Motion Sensor" //ThirdReality Motion Sensor
```
tpmanley: @Jackie-Kilby Usually we put the brand name in the `deviceJoinName` like "Third Reality Motion Sensor". Do you want to do that or leave it more generic?

Jackie-Kilby: Thank you. I will use "ThindReality Motion Sensor". However, may I state my opinion? 
I think a user will pair only one or two sensors, or smart devices once a time. As a normal user, he do care the brand of the device when purchasing. However, he do not care what the brand is when he is pairing the device. So I think all deviceJoinName should be generic, only the manufacturer info will tell the difference backstage.

greens:
@Jackie-Kilby looks like your most recent commit accidentally deleted a lot of code.

devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy
tpmanley: @Jackie-Kilby thanks for the feedback. I understand what you mean but this was a thoughtful decision and it's done this way for most devices so if nothing else it's helpful for consistency.

Jackie-Kilby:
> @Jackie-Kilby looks like your most recent commit accidentally deleted a lot of code.
@greens @tpmanley 
Yes, I think I commit the wrong fake file. I copied the original one back again and modified based on it. Now I think the ThirdReality Motion Sensor can work. Please help us to merge the code into Publish code. Thanks.
BTW, the "refresh problem" is still there. :)

greens (approved):

Jackie-Kilby:
@greens @tpmanley 
Sorry, I am really confused. I have added our Motion Sensor Device Handler and it is merged and published. I can see the release in Tags. However, when I check the Motion Sensor support today, the Motion is recognized as a "Thing", which is same with no Device Handler.
I want to know if there is an operation that I can get the latest Device Handler published on my SmartThings Hub. Or if there is an indicator saying which version of Published Device Handler it is now, shown in App.

greens:
@Jackie-Kilby it's only in staging. We have a two week release cycle that progresses on Tuesday. You can always double check by looking at the "production" branch of the file:

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/production/devicetypes/smartthings/zigbee-motion-detector.src/zigbee-motion-detector.groovy

PR #56420, opened 2021-01-21
Adding model type to existing Smartthings device handler

SmartThings:
The standard device handler "Zigbee Lock" works well with the Zigbee Module for ID Lock 150.
Please update in accordance with attached changes in Device Handler.


oyvindbrown (changes requested):
See newer request #61595 with correct model name.

PR #56512, opened 2021-01-22
[ICP-12754] Fix for: "Qubino Flush On Off Thermostat 2 – Cannot set a heating temperature"

MGoralczykS:
When I changed the setpoint value I get this error 
`groovy.lang.GroovyRuntimeException: Ambiguous method overloading for method java.lang.Math#min.
Cannot resolve which method to invoke for [class java.math.BigDecimal, class java.lang.Integer] due to overlapping prototypes between:
[double, double]
[float, float] @line 281 (updateSetpoint)`

I cast those two values to double to fix this. 

### DIFF
--- a/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
+++ b/devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
@@ -278,7 +278,7 @@ def setCoolingSetpoint(setpoint) {
 
 def updateSetpoint(setpoint, setpointType) {
 	setpoint = temperatureScale == 'C' ? setpoint : fahrenheitToCelsius(setpoint)
-	setpoint = Math.max(Math.min(setpoint.doubleValue(), maxSetpointTemperature.doubleValue()), minSetpointTemperature)
+	setpoint = Math.max(Math.min(setpoint.doubleValue(), maxSetpointTemperature.doubleValue()), minSetpointTemperature.doubleValue())
 	[
 			secure(zwave.thermostatSetpointV2.thermostatSetpointSet([precision: 1, scale: 0, scaledValue: setpoint, setpointType: setpointType, size: 2])),
 			"delay 2000",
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
```diff
@@ -278,7 +278,7 @@ def setCoolingSetpoint(setpoint) {
 
 def updateSetpoint(setpoint, setpointType) {
 	setpoint = temperatureScale == 'C' ? setpoint : fahrenheitToCelsius(setpoint)
-	setpoint = Math.max(Math.min(setpoint, maxSetpointTemperature), minSetpointTemperature)
+	setpoint = Math.max(Math.min(setpoint.doubleValue(), maxSetpointTemperature.doubleValue()), minSetpointTemperature)
```
KKlimczukS: Can You also cast minSetpointTemperature to double ?

KKlimczukS (changes requested):

devicetypes/smartthings/qubino-flush-thermostat.src/qubino-flush-thermostat.groovy
MGoralczykS: Changed

KKlimczukS (approved):

greens:
what if `maxSetpointTemperature` and `minSetpointTemperature` just returned floats instead? I think you could just append a `.0` to all those values for a more elegant fix.

KKlimczukS:
> what if `maxSetpointTemperature` and `minSetpointTemperature` just returned floats instead? I think you could just append a `.0` to all those values for a more elegant fix.

@greens I think that use case was checked by @MGoralczykS and it was working fine (with the code from this PR).
@MGoralczykS please confirm it.

MGoralczykS:
@greens @KKlimczukS Everything is working correctly.

PR #56525, opened 2021-01-25
WWST-7201, WWST-7223, WWST-7239 moves Enbrighten, Jasco devices between DTHs

KKlimczukS:
WWST-7201, WWST-7223, WWST-7239 moves Enbrighten, Jasco devices between DTHs
Those all devices support also "Energy meter" capability, so they have to be moved between DTHs.
(in both cases, a manufacturer is: "Jasco Products" )

@greens Please, could You take a look at the code ?

### DIFF
--- a/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
+++ b/devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
@@ -57,8 +57,6 @@ metadata {
 
 		// Enbrighten
 		fingerprint manufacturer: "Jasco Products", model: "43078", deviceJoinName: "Enbrighten Switch" //Enbrighten In-Wall Smart Switch With Energy Monitoring 43078, Raw Description: 01 0104 0100 00 07 0000 0003 0004 0005 0006 0702 0B05 02 000A 0019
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zigbee-switch-power.src/zigbee-switch-power.groovy
```diff
@@ -57,8 +57,6 @@ metadata {
 
 		// Enbrighten
 		fingerprint manufacturer: "Jasco Products", model: "43078", deviceJoinName: "Enbrighten Switch" //Enbrighten In-Wall Smart Switch With Energy Monitoring 43078, Raw Description: 01 0104 0100 00 07 0000 0003 0004 0005 0006 0702 0B05 02 000A 0019
```
greens: this device's name indicates that it _also_ supports energy monitoring. Are you sure you shouldn't be moving it as well?

KKlimczukS: @greens Yes, I was checking that device yesterday too. 
If it's ok to move a switch to zigbee-metering-plug DTH then I'll do it.

greens (approved):

PR #58041, opened 2021-02-12
Add fingerprint for Heimantech HS1SA-M smoke alarm

beiske:
I'm no expert in these device handlers, but this seems to work well with mine that are the HS1SA-M model.

Battery status, autodetection of device on add, alarms, tests etc all appear good.


beiske:
I guess nobody cares about contributions

PR #59780, opened 2021-03-04
ICP-14065 Notification Gets resulting in false positives

greens:
FireAngel users reported getting periodic false positives from notification gets sent in response to wakeups. I believe this is due to some gray areas in the Z-Wave spec, but notification gets are not the same as gets for other command classes.


cbaumler:
Okay, help me understand. So when the hub sends the notification get for notification type `0x01` / event/state `0x00`, the device responds with a notification report indicating unsolicited messages are enabled (`0xFF`) and indicating that for notification type `0x01`, the state is `0x00` (idle). The DTH interprets this as a smoke clear event.

```
2021-03-04_14:27:01.832 | I | zw         | Dst:[15] CC:Notification ID:0x04 Len:3 Payload:0x00 01 00 Encap:Auto Source:GROOVY_DTH 
2021-03-04_14:27:01.868 | I | sc         | Src:HUB Ch:1 TID:8C Cmd:ACK_RESP             Len:15
2021-03-04_14:27:02.109 | I | zw         | Src:[15] CC:Notification ID:0x05 Len:7 Payload:0x00 00 00 FF 01 00 00 Encap:None 
2021-03-04_14:27:02.109 | I | zw         | [15] NotificationReport(v1type:0, v1level:0, status:FF, type:1, event:0)
```

It looks like the device is also sending unsolicited notifications that smoke is detected (type `0x01` and event `0x02`):

```
2021-03-04_14:26:55.789 | I | zw         | Src:[15] CC:Notification ID:0x05 Len:8 Payload:0x00 00 00 FF 01 02 80 33 Encap:None 
2021-03-04_14:26:55.789 | I | zw         | [15] NotificationReport(v1type:0, v1level:0, status:FF, type:1, event:2, p: 33 )
```

Isn't the real issue that the device is sending these unsolicited smoke notifications rather than the responses to the Notification Get commands?

greens:
Yeah, I think you're right, Chris. It looks like the device is constantly in a "smoke detected" state and our notification gets are actually generating false _negatives_.

PR #59781, opened 2021-03-05
Aeotec Doorbell Siren 6 sound/volume configuration fix

Aeotec-ccheng:
Fixes issue where Sound/Volume cannot be configured if paired with No Security (S2_FAILED or No Security level)


Aeotec-ccheng:
@greens - recently been having issues with non-secure or S2_FAILED security level inclusion not being able to change sounds/volumes. Would you be able to take a look? This is more of a bandage fix that was missed for non-secure levels.

greens (approved):
@Aeotec-ccheng if you have a contact with us you could e-mail some associated zniffer logs we could take a look and see if we see anything.

PR #60116, opened 2021-03-09
DevWs for NIE-TECH CO., LTD. containing containing Minoston Min Smart Plug

sky-nie:


sky-nie:
I have add the fingerprints to another DTH file, so cancel this request.

PR #60277, opened 2021-03-10
ICP-14003 - fixes configuration of Leviton devices, query Leviton devices after setting level. Code refactoring.

KKlimczukS:
ICP-14003 - fixes configuration of Leviton devices, query Leviton devices after setting level. Code refactoring.
@greens @varzac 

### DIFF
--- a/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
@@ -257,7 +257,5 @@ def configure() {
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	def cmds = refresh() + zigbee.onOffConfig(0, 300)
-	cmds += isLeviton() ? zigbee.levelConfig(1,3600, 0x00) : zigbee.levelConfig()
-	return cmds
+	refresh() + zigbee.onOffConfig(0, 300) + zigbee.levelConfig()
 }
--- a/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
+++ b/devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
@@ -174,7 +174,7 @@ def parse(String description) {
 			} else {
 				log.warn "ON/OFF REPORTING CONFIG FAILED- error code:${cluster.data[0]}"
 			}
-		} else if (device.getDataValue("manufacturer") == "sengled" && descMap && descMap.clusterInt == 0x0008 && descMap.attrInt == 0x0000) {
+		} else if (isSengled() && descMap && descMap.clusterInt == 0x0008 && descMap.attrInt == 0x0000) {
 			// This is being done because the sengled element touch/classic incorrectly uses the value 0xFF for the max level.
 			// Per the ZCL spec for the UINT8 data type 0xFF is an invalid value, and 0xFE should be the max.  Here we
 			// manually handle the invalid attribute value since it will be ignored by getEvent as an invalid value.
@@ -204,7 +204,7 @@ def setLevel(value, rate = null) {
 	def additionalCmds = []
 	if (device.getDataValue("model") == "iQBR30" && value.toInteger() > 0) { // Handle iQ bulb not following spec
 		additionalCmds = zigbee.on()
-	} else if (device.getDataValue("manufacturer") == "MRVL") { // Handle marvel stack not reporting
+	} else if (isMRVL() || isLeviton()) { // Handle marvel stack not reporting
 		additionalCmds = refresh()
 	}
 	zigbee.setLevel(value) + additionalCmds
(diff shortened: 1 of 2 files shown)


devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
```diff
@@ -221,19 +221,41 @@ def refresh() {
...
 def configure() {
 	log.debug "Configuring Reporting and Bindings."
 	// Device-Watch allows 2 check-in misses from device + ping (plus 1 min lag time)
 	// enrolls with default periodic reporting until newer 5 min interval is confirmed
 	sendEvent(name: "checkInterval", value: 2 * 10 * 60 + 1 * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
 
 	// OnOff minReportTime 0 seconds, maxReportTime 5 min. Reporting interval if no activity
-	refresh() + zigbee.onOffConfig(0, 300) + zigbee.levelConfig()
+	def cmds = refresh() + zigbee.onOffConfig(0, 300)
+	cmds += isLeviton() ? zigbee.levelConfig(1,3600, 0x00) : zigbee.levelConfig()
```
varzac: I'm not sure this different reporting config will do anything.  It already has a reportable change of 1, and a value of 0 is not functionally different than 1.

devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
```diff
@@ -204,7 +204,7 @@ def setLevel(value, rate = null) {
 	def additionalCmds = []
 	if (device.getDataValue("model") == "iQBR30" && value.toInteger() > 0) { // Handle iQ bulb not following spec
 		additionalCmds = zigbee.on()
-	} else if (device.getDataValue("manufacturer") == "MRVL") { // Handle marvel stack not reporting
+	} else if (isMRVL() || isLeviton()) { // Handle marvel stack not reporting
 		additionalCmds = refresh()
 	}
```
varzac: Doing it this way does mean we are also going to send a read for the on/off status every time level is set, which seems unnecessary.

KKlimczukS: Good point. I've changed it already.

devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
KKlimczukS: Well, it has helped with Leviton DL6HD. Use case -> 
**before:** 
step 1:  current is Y,  set to X (where X < Min),  device sets to Min, reports Min.
step 2:  current is Min, set to X (where X < Min),  device stays at Min, **reports nothing.**
**after:** 
step 1:  current is Y,  set to X (where X < Min),  device sets to Min, reports Min.
step 2:  current is Min, set to X (where X < Min),  device stays at Min, **reports Min.**

But this part of code may be redundant if the device will be queried for a level (in local execution code).

KKlimczukS: I removed that part of the code.

devicetypes/smartthings/zigbee-dimmer.src/zigbee-dimmer.groovy
```diff
@@ -204,8 +204,10 @@ def setLevel(value, rate = null) {
 	def additionalCmds = []
 	if (device.getDataValue("model") == "iQBR30" && value.toInteger() > 0) { // Handle iQ bulb not following spec
 		additionalCmds = zigbee.on()
-	} else if (device.getDataValue("manufacturer") == "MRVL") { // Handle marvel stack not reporting
+	} else if (isMRVL()) { // Handle marvel stack not reporting
 		additionalCmds = refresh()
+	} else if (isLeviton()) {
+		additionalCmds = zigbee.levelRefresh()
```
greens, line 210: as far as I know, none of this would ever have been run for a device running locally in the first place

greens, line 208: @varzac would it work if we just treated this device as a zll device by moving the fingerprint over?

varzac: Yes it should, but it would likely be a worse experience as on average it would be slower reporting of the level.  It would have the advantage of not requiring any local changes though.

greens: I think that's probably the right course of action as it also lets us avoid messy special-casing in hubcore

KKlimczukS: @greens @varzac ok then, please let me know if I should revert some changes from current DTH.

greens: Yeah I'd revert these changes and move the relevant fingerprints to `zll-dimmer-bulb` with a correct `ocfDeviceType`

KKlimczukS: I'm having second thoughts about it. 
Since Leviton DG3HL and Leviton DG6HD are not certified yet - moving those fingerprints may be an easy thing (from WWST certification point of view). 
But this issue occurs also in case of WWST certified device -> Leviton DL6HD (and maybe the other ones too). 
It should be reconsidered to leave those fingerprints here and fix this issue in a hub core.
@varzac @greens What do You think ?

greens: @KKlimczukS The migration isn't a big deal. It's just tedious, but I'm willing to take care of it.

greens (approved):

PR #60782, opened 2021-03-16
BUG-1910 Garage Door events being duplicated

greens:
The Garage Door Control and Door Control capabilities are largely redundandant with each other. We should just be able to remove the Door Control capability.


bflorian:
Don't we want to do the opposite? The _doorControl_ capability is currently live and _garageDoorControl_ is deprecated:

https://st-schema.com/capabilities/doorControl/1
https://st-schema.com/capabilities/garageDoorControl/1

dkirker (approved):

aarontimbo (approved):

PR #61244, opened 2021-03-22
[WWST-7366] Aeotec MultiSensor 7 - fingerprints and necessary tweaks

PKacprowiczS:


greens (approved):

Aeotec-ccheng:
@greens - May need a separate device handler for the MS7 or to separate out the configuration settings pushed to MS6 and MS7. The Parameter setting byte size values for the MS7 compared to MS6 differ. 

This interface will work, but the MS7 may not be configurable with the size byte settings for its parameter.

Critical parameter settings changed (Parameter size differs from MultiSensor 7 especially with the below parameters):
Parameter 101 [1 byte] = 240
Parameter 102 [1 byte] = 1
Parameter 111 [2 byte] value range = 30 - 65535
Parameter 112 [2 byte] value range = 30 - 65535
Parameter 3 [2 byte] value range = 30 - 3600
Parameter 4 [1 byte] value range = 0 - 11 (min to max)

Critical new parameter setting added:
Parameter 2 [2 byte] value range = 0 - 3600

Not sure iif this is needed, but hopefully this is helpful information.

greens:
@Aeotec-ccheng you should just be able to add a similar model check in the configuration step when collecting the configuration commands instead of a brand new DTH

PR #63302, opened 2021-04-14
DevWs for SHINA SYSTEM containing containing SiHAS Zigbee Contact Sensor

shinasys:


MWierzbinskaS (changes requested):
This DTH looks very similar to https://github.com/shinasys/SmartThingsPublic/blob/f03a97881d59828ea282dc4ac5480ba264d2e1c5/devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy I think You can add fingerprint with contact sensor implementation to that DTH to avoid code duplication

shinasys:
> This DTH looks very similar to https://github.com/shinasys/SmartThingsPublic/blob/f03a97881d59828ea282dc4ac5480ba264d2e1c5/devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy I think You can add fingerprint with contact sensor implementation to that DTH to avoid code duplication

I want to add in sihas-multipurpose-sensor.groovy. But this have error in the smartthing app as the following image.
So I separated this.
And I want to make the Preference method not appear according to the device model, but there is no way.

smartthings_device_add_error_20210421

MWierzbinskaS:
We try to avoid adding new DTH if there are no functionality (like preferences) which require a separate DTH.
You are right that if this integration would be made in sihas-multipurpose-sensor.groovy the preferences would be visible.
If you would like to avoid exposing preferences for this sensor I'd suggest to check if this device can be integrated with some of the other existing DTH.
Maybe: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/a7c980340992ca37508d42a70f354b5fd7d30048/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy

This checking status error which you send looks if there was a problem with setting correct status in the app after pairing. If you make any action on this device does it change to correct status or there is still this problem?

shinasys:
1. I'd like to know the procedure for adding another existing DTH.
2. Even if I added it to my DTH and EVENT occurred on my device, the result was the same. On the history tab, the processing for the event is displayed as Motion. And there are only pairing-related messages on the Live logging tab of the SmartThings Groovy IDE, and no other events. 
 My modified DTH is https://github.com/shinasys/SmartThingsPublic/blob/49716ceac7f8081aaf7eb4d3c814d3ad19665154/devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy
Anyway, the follows is a related screen capture.
smartthings_device_add_error_20210423_1
I think the smartthings platform doesn't seem to be able to handle it properly.

MWierzbinskaS:
Could you please add some logging to verify which event is created in this case for contact sensor:
```
private Map translateZoneStatus(ZoneStatus zs) {
	if (isDSM300()) {
    	return (zs.isAlarm1Set() || zs.isAlarm2Set()) ? getContactResult('open') : getContactResult('closed')
    } else {    
    	// Some sensor models that use this DTH use alarm1 and some use alarm2 to signify motion
        return (zs.isAlarm1Set() || zs.isAlarm2Set()) ? getMotionResult('active') : getMotionResult('inactive')
    }    
}
```
It seems that events are created but with wrong capability.

shinasys:
In parse(), there is no logging for the close/open event.  
But POWER_CONFIGURATION_CLUSTER event is following logging.

>  83068b7e-64b5-40d5-a960-5de756594e77 오전 9:57:35: debug Parsing message from device: read attr - raw: 9623010001082000201E, dni: 9623, endpoint: 01, cluster: 0001, size: 8, attrId: 0020, result: success, encoding: 20, value: 1e

Please confirm.

```
>def parse(String description) {
>	log.debug "Parsing message from device: $description"
>   
>    Map map = zigbee.getEvent(description)  
```

shinasys:
When will this problem be confirmed?

MWierzbinskaS:
I see, but I meant to add logging in the translateZoneStatus method (which branch of if statement is executed) and publish DTH with changes. There might be some wrong value in isDSM300() or with creating event. It seems that there is proper communication with the device but created event is for motion instead of contact.

shinasys:
As you know, if the parsing method does not have logging, the translateZoneStatus method does not have logging.
So there's no way I can check.
As I said before, the first connection and power configuration messages show logging.

shinasys:
` // capability "Motion Sensor"`

Removing capability "Motion Sensor" works properly. However, equally logging is not output.
So I don't think I can add this to this DTH unless this problem is solved.
How should I proceed ?

MWierzbinskaS:
@greens Could You please share your opinion about integrating this device in a new DTH?

shinasys:
> @greens Could You please share your opinion about integrating this device in a new DTH?

I want to add it to the existing DTH, but it doesn't work, so I want to add it to a new DTH.

greens:
@shinasys can you explain what doesn't work? This looks like it should fit very easily inside an existing DTH.

shinasys:
@greens My test reporting is https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/63302#issuecomment-823709034, https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/63302#issuecomment-825371387, https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/63302#issuecomment-827232130, https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/63302#issuecomment-834022053
It doesn't work to be added to my DTH(sihas-multipurpose-sensor.groovy). 
Then I want to add it to other existing DTH(https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/a7c980340992ca37508d42a70f354b5fd7d30048/devicetypes/smartthings/smartsense-open-closed-sensor.src/smartsense-open-closed-sensor.groovy), but I don't know how.

greens:
@shinasys I imagine that difficulty you were having was because you had `runLocally` and `executeCommandsLocally` set to true.

The way you would add your fingerprint to an existing DTH is to 

`git clone git@github.com:shinasys/SmartThingsPublic.git` 
`git checkout devws-for-shina-system-containin_67`

locally, then make your additions to our DTHs, commit, and push them.

PR #63418, opened 2021-04-14
CHAD-6481 Incorrect temp reports for Zooz multisiren

greens:
Zooz had reported incorrect temperature reports, and this should fix any inconsistencies between how this device reports vs how all other devices report.


cbaumler (approved):
Zooz has verified this change resolves the issue.

dkirker (approved):

PR #63423, opened 2021-04-15
MultiSensor 7 parameter settings update

Aeotec-ccheng:
Separated MS6 and MS7 configuration SET settings (byte size difference and allowable range settings)

### DIFF
--- a/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
+++ b/devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
@@ -74,12 +74,12 @@ metadata {
 
 	preferences {
 		input "motionDelayTime", "enum", title: "Motion Sensor Delay Time",
-			options: ["20 seconds", "40 seconds", "1 minute", "2 minutes", "3 minutes", "4 minutes"]
+			options: ["30 seconds", "40 seconds", "1 minute", "2 minutes", "3 minutes", "4 minutes"]
 
 		input "motionSensitivity", "enum", title: "Motion Sensor Sensitivity", options: ["maximum", "normal", "minimum", "disabled"]
 
 		input "reportInterval", "enum", title: "Report Interval", description: "How often the device should report in minutes",
-			options: ["8 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours", "18 hours", "24 hours"]
+			options: ["1 minute", "2 minutes", "3 minutes", "4 minutes", "8 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours"]
 	}
 
 	tiles(scale: 2) {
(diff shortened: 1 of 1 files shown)


Aeotec-ccheng:
@greens - could you take a look if this is good enough to merge for parameter setting byte size changes? I've tested this one out and it works well with MS6 and MS7 on my side.

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -74,12 +74,12 @@ metadata {
...
 	preferences {
 		input "motionDelayTime", "enum", title: "Motion Sensor Delay Time",
-			options: ["20 seconds", "40 seconds", "1 minute", "2 minutes", "3 minutes", "4 minutes"]
+			options: ["30 seconds", "40 seconds", "1 minute", "2 minutes", "3 minutes", "4 minutes"]
 
 		input "motionSensitivity", "enum", title: "Motion Sensor Sensitivity", options: ["maximum", "normal", "minimum", "disabled"]
 
 		input "reportInterval", "enum", title: "Report Interval", description: "How often the device should report in minutes",
-			options: ["8 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours", "18 hours", "24 hours"]
+			options: ["1 minute", "2 minutes", "3 minutes", "4 minutes", "8 minutes", "15 minutes", "30 minutes", "1 hour", "6 hours", "12 hours"]
```
greens, line 83: we don't want to allow the interval this low

we find that customers get very concerned when their batteries die quickly

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -398,46 +398,96 @@ def ping() {
...
+
+            		//6. report automatically ONLY on threshold change
+            		//From manual:
+            		//Enable/disable the selective reporting only when measurements reach a certain threshold or percentage set in 41-44.
+            		//This is used to reduce network traffic.  (0 = disable, 1 = enable)
+            		//Note: If USB power, the Sensor will check the threshold every 10 seconds. If battery power, the Sensor will check the threshold
+            		//when it is waken up.
+            		request << zwave.configurationV1.configurationSet(parameterNumber: 40, size: 1, scaledConfigurationValue: 1)
+			break
+	}
```
greens, line 460: a lot of these parameters are the same across models, couldn't you build a `request` list that has all the parameters in common and only separate out those parameters that are different inside the switch statement?

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
Aeotec-ccheng: Even when set to 1 minute, the lowest setting is 4 minutes on battery power, Wakeup interval generally sets the minimum most setting so it generally won't change unless wakeup interval is set to 240 seconds which is the lowest setting.

If Wakeup Interval = 3600
If Parameter 111 = 60

Sensors will still report every 60 seconds. 

Only on USB power is this setting unlocked to as low as once per a second (or in this case 1 minute) since wakeup does not play a hand in this anylonger as a repeater device.

greens: ok

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
Aeotec-ccheng: Got it, i moved a bunch of the same lines from the switch statement outside if they are the same between MS6 and MS7. I left Parameter 3 since they have different range supports.

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
```diff
@@ -398,46 +399,74 @@ def ping() {
...
+        request << zwave.associationV1.associationSet(groupingIdentifier: 1, nodeId: zwaveHubNodeId)
+
+        //2. automatic report flags
+        // param 101 -103 [4 bytes] 128: light sensor, 64 humidity, 32 temperature sensor, 15 ultraviolet sensor, 1 battery sensor
+        // set value  241 (default for 101) to get all reports. Set value 0 for no reports (default for 102-103)
+        //association group 1
+        request << zwave.configurationV1.configurationSet(parameterNumber: 101, size: 1, scaledConfigurationValue: 240)
+
+        //association group 2
+        request << zwave.configurationV1.configurationSet(parameterNumber: 102, size: 1, scaledConfigurationValue: 1)
```
greens: nit: looks like you've accidentally done spaces instead of tabs here

Aeotec-ccheng (commented):
accidentally started this review/comment - just needs to be deleted (sorry)

devicetypes/smartthings/aeon-multisensor-6.src/aeon-multisensor-6.groovy
Aeotec-ccheng: I did a quick change to the tabs, let me know if there is anything else that i need to fix.

greens (approved):

PR #63570, opened 2021-04-16
BUG-2032 Send previous setpoint value if we don't intend to update the setpoint

dkirker:
The thermostat won't accept commands to change a setpoint that doesn't match the current mode. To appears the mobile client we need to send an event for the specified setpoint. So, we will just send the current value.


dkirker:
The only thing that might be wonky is if `device.currentValue` returns a `null` value. But that would be an odd edge case anyway as we make an effort to get all of the current state at device creation.

greens (approved):

PR #64360, opened 2021-04-22
[ICP-14142] Aeotec Multisensor 7 - added custom vid 

PKacprowiczS:


dkirker (approved):

greens (approved):

greens:
this works, but I think it needs to be discussed on my ticket. If we change this here we'd have to change it everywhere, and I don't think that's the right call.

PR #65135, opened 2021-04-30
CHAD-4340 Update Window Shade devices to use Window Shade Level

dkirker:
Mostly unchanged from https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/30151 with the exception of:
* A little more formatting in `axis/axis-gear-st.src/axis-gear-st.groovy`
* Bringing changes made to `zigbee-window-shade-battery.src/zigbee-window-shade-battery.groovy` and `zigbee-window-shade.src/zigbee-window-shade.groovy` in
* Re-working of the ternary logic for getting the latest shade level values to account for 0 AND null being treated as false

### DIFF
--- a/devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy
+++ b/devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy
@@ -15,252 +15,262 @@ import groovy.json.JsonOutput
 
 
 metadata {
-    definition (name: "Z-Wave Window Shade", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.blind") {
-        capability "Window Shade"
-        capability "Window Shade Preset"
-        capability "Battery"
-        capability "Refresh"
-        capability "Health Check"
-        capability "Actuator"
-        capability "Sensor"
+	definition (name: "Z-Wave Window Shade", namespace: "smartthings", author: "SmartThings", ocfDeviceType: "oic.d.blind") {
+		capability "Window Shade"
+		capability "Window Shade Level"
+		capability "Window Shade Preset"
+		capability "Switch Level"
+		capability "Battery"
+		capability "Refresh"
+		capability "Health Check"
+		capability "Actuator"
+		capability "Sensor"
 
-        command "stop"
+		command "stop"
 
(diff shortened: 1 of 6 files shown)


devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy
```diff
@@ -15,252 +15,262 @@ import groovy.json.JsonOutput
...
+	} else if (level <= 0) {
+		level = 0  // unlike dimmer switches, the level isn't saved when closed
+		shadeValue = "closed"
+	} else {
+		shadeValue = "partially open"
+		descriptionText = "${device.displayName} shade is ${level}% open"
+	}
+	checkLevelReport(level)
+
+	def levelEvent = createEvent(name: "level", value: level, unit: "%", displayed: false)
```
KKlimczukS, line 159: @dkirker
Since level and shadeLevel keep the same value - I'm wondering why levelEvent is still needed. Is it because of existing automations (so not to break anything) ? Are there any plans for migrating such automations ?
What about the UI (metadata) ? Are we going to have two sliders to set level value?

KKlimczukS:
@greens @tpmanley 
Will the engineers who have ported those DTHs to LUA drivers be notified about @dkirker's changes (right after they are merged) ?

devicetypes/smartthings/zwave-window-shade.src/zwave-window-shade.groovy
dkirker: @KKlimczukS `Switch Level` is going away, but it can't be pulled out wholesale. The UI metadata will be updated accordingly.

dkirker: Step 1 is to add `Window Shade Level` to the DTH.
Step 2 is to use `Window Shade Level` for the device details control and both as rule options, removing `Switch Level` as the device details control.
Step 3 is to remove `Switch Level` from use with rules and migrate all existing to `Window Shade Level`.

It might be easier to just have step 2 and step 3 combined as step 2 to reduce the risk of duplicate rules.

greens (approved):

PKacprowiczS (approved):

PR #66239, opened 2021-05-12
DevWs for TechniSat Digital GmbH containing containing Z-Wave Metering Dimmer

SmartThings:


MWierzbinskaS:
@NArlt This DTH looks similar to: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/technisat/technisat-on-off-switch.src/technisat-on-off-switch.groovy . I would suggest to check if those two integrations can be merged into one DTH. You will need to add correct mnmn and vid for each fingerprint to ensure that UI will be correct e.g. https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/a7c980340992ca37508d42a70f354b5fd7d30048/devicetypes/smartthings/zwave-motion-light-sensor.src/zwave-motion-light-sensor.groovy#L29-L31

NArlt:
Hi @MWierzbinskaS ,
sorry for the late reply and thanks for the advice. I didn't know it is possible to merge 2 devices with different capabilities into 1 DTH. I will create a new PR with the requested changes and link it here.

BR NArlt

NArlt:
Hi @MWierzbinskaS ,
I've tried to merge the 2 DTH but unfortunately I discovered a problem. The 2 DTH have similar but not equal preferences. I haven't found a way to implement the 2 different preferences into 1 DTH. Is it even possible? If not, I would suggest to keep the 2 DTH separate.

BR NArlt

MWierzbinskaS:
@NArlt unfortunately there is no possibility to implement 2 different preferences into 1 DTH

greens:
@NArlt I'd just add to your existing DTH with a note (and some model number-checking code) that indicates that the dimming curve preference does not apply to all devices.

NArlt:
Hi @greens ,
after discussing internally, we think that

a) offering settings that are only applicable to certain devices - even with a note "this parameter will only affect certain devices" - could confuse the end-user and thus degrade the user experience

b) we are always trying to keep code as maintainable as possible - so rather having two files with clear functionality , obvious association and fewer lines, than one larger file which does everything

Considering a) and b), we would prefer to keep the two DTHs separate and hope that you can follow our reasoning in the interest of the SmartThings user experience and better, more maintainable code.

greens (approved):

PR #66394, opened 2021-05-14
DevWs for Sensative AB containing containing Strips-Comfort-700

SmartThings:

PR #2898, opened 2018-03-12
Sensative's Strips Multi-Sensor Product's Device Handler.

dparyani:
As recommended by Steven Green, creating this pull request.

PR #66395, opened 2021-05-14
DevWs for Sensative AB containing containing Strips-Drip-700

SmartThings:


devicetypes/sensative/sensative-strips-drip-700.src/sensative-strips-drip-700.groovy
```diff
@@ -0,0 +1,617 @@
...
+	0x8E: 2,	// Multi Channel Association
+	0x9F: 1 	// Security 2
+]
+
+@Field static int wakeUpIntervalSeconds = 43200
+@Field static int tempSensorType = 1 
+@Field static int waterSensorType = 31 
+@Field static int leakageCalibrationParamNum = 23
+
+metadata {
```
KKlimczukS, line 56: Did You consider supporting Your both devices (this and the one from #66394) by one DTH ?
You would have to move ocfDeviceType, vid, mnmn from definition() to the fingerprints. Some tweaks in configuration code etc might be needed too (and some "if"-s) . 
That way You might avoid a code duplication.

KKlimczukS (changes requested):

devicetypes/sensative/sensative-strips-drip-700.src/sensative-strips-drip-700.groovy
krlaframboise: If SmartThings supported conditionally showing preferences based on which device was joined then using one handler to support both devices would be possible, but it doesn't and it also doesn't make sense to combine them because ones a Contact Sensor and the other is a Water/Temperature Sensor...

krlaframboise: Another reason these handlers can't be merged is that some of the built-in SmartApps don't use the the presentation so both devices would appear in their Water Sensor and Contact Sensor lists.

An updated version of this handler will be posted soon that includes the same types of changes you requested on PR #66394 so you might want to hold off on reviewing this until the new version is posted.

KKlimczukS:
> I am lost, not sure where to find your comment (@KKlimczukS) regarding the merging of all device handlers for our products into one.

@dparyani My it's above in this pull request:
_"Did You consider supporting Your both devices (this and the one from #66394) by one DTH ?"_

> I am wondering about the benefits? Are there benefits of merging all device handlers into one?

To avoid code duplication as much as possible.

> Else we agree with @krlaframboise to keep the device handlers separate for our three products (third one in development) as they have different capabilities.

If those devices need to expose their all parameters (not only those parameters that can be shared by both: Strips-Comfort-700 and Strips-Drip-700) then we will continue reviewing 2 separate DTHs.

dparyani:
@KKlimczukS , For Strips Drip 700, the new commit:

https://github.com/dparyani/SmartThingsPublic/commit/38973cfc0e77b180fd256a2f4daf987c946a1f74

KKlimczukS:
> @KKlimczukS , For Strips Drip 700, the new commit:
> 
> [dparyani@38973cf](https://github.com/dparyani/SmartThingsPublic/commit/38973cfc0e77b180fd256a2f4daf987c946a1f74)

@krlaframboise @dparyani We have to ask You to create new pull request with those changes [dparyani@38973cf](https://github.com/dparyani/SmartThingsPublic/commit/38973cfc0e77b180fd256a2f4daf987c946a1f74), so we could continue reviewing the code. And please add the link here.

dparyani:
> 
> 
> > @KKlimczukS , For Strips Drip 700, the new commit:
> > [dparyani@38973cf](https://github.com/dparyani/SmartThingsPublic/commit/38973cfc0e77b180fd256a2f4daf987c946a1f74)
> 
> @krlaframboise @dparyani We have to ask You to create new pull request with those changes [dparyani@38973cf](https://github.com/dparyani/SmartThingsPublic/commit/38973cfc0e77b180fd256a2f4daf987c946a1f74), so we could continue reviewing the code. And please add the link here.

@KKlimczukS Does the following work?
https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/2898

k-morawski:
I am closing this pull request.

PR #67326, opened 2021-05-27
DevWs for SHINA SYSTEM containing containing ZigBee Multi Switch

shinasys:


devicetypes/smartthings/zigbee-multi-switch.src/zigbee-multi-switch.groovy
```diff
@@ -146,6 +153,8 @@ def parse(String description) {
 }
 
 private void createChildDevices() {
+	if (isSBM300Z1()) return
```
MWierzbinskaS: If that device does not have any childs I would suggest moving it to a different switch DTH.  E.g. https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-switch.src/zigbee-switch.groovy

MWierzbinskaS (changes requested):

greens (approved):

PR #68969, opened 2021-06-23
DevWs for Plaid Systems LLC containing containing Spruce Controller SST

natec007:


PKacprowiczS (changes requested):
Overall looks good, just one little thing. You're using spaces instead of tabs in some places. Could you change that?

natec007:
@PKacprowiczS should be fixed now... somehow I always miss checking those spaces :)

PKacprowiczS (approved):
Sorry for late response, somehow I've treated your latest message as "I'll get to that" kind of thing.

@greens I think we can merge this.

greens (approved):

greens:
@natec007 should be live in prod on the 13th in NA and the 14th elsewhere.

natec007:
Thanks @PKacprowiczS @greens !

PR #69361, opened 2021-06-29
[ICP-14234] Zigbee Window Shade - included lastLevel in methods scope as local variable

PKacprowiczS:
This is needed, cause comparison `currentLevel < (or) > lastLevel` was always false, as lastLevel value was being acquired after shadeLevel was set.

### DIFF
--- a/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
+++ b/devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
@@ -132,10 +132,10 @@ def getLastLevel() {
 }
 
 def levelEventHandler(currentLevel) {
-	def lastLevel = getLastLevel()
-	log.debug "levelEventHandle - currentLevel: ${currentLevel} lastLevel: ${lastLevel}"
+	def priorLevel = lastLevel
+	log.debug "levelEventHandle - currentLevel: ${currentLevel} priorLevel: ${priorLevel}"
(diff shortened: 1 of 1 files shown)


devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
```diff
@@ -132,6 +132,7 @@ def getLastLevel() {
 }
 
 def levelEventHandler(currentLevel) {
+	def lastLevel = getLastLevel()
```
dkirker: I feel like we should name the variable something else just to be safe. But otherwise, I think it is ok.

dkirker (dismissed):

devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
greens: yeah, I'm sure that the local scope blows the groovy syntactic sugar away, but I agree with Donald here just so it's obvious what's going on.

`def priorLevel`? perhaps?

PKacprowiczS: Yeah, I thought of new name so hard, that I've forgotten this PR. Anyway, now it's updated.
@greens @dkirker

greens (approved):

PR #70293, opened 2021-07-08
Add smartplug and modify switch of ThirdReality

jiangshanyang0203:
Add smartplug and modify switch of ThirdReality Plug and Switch
@greens Please review . Thanks.


greens (approved):

greens:
will be in production in two weeks

PR #70757, opened 2021-07-13
[BUG-3052] Updating Ecolink URL

AaronHuus:
Old URL caused a mismatch in SSL ciphers and updating to api.ecobee.com resolved the issue.

### DIFF
--- a/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
+++ b/smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
@@ -201,7 +201,7 @@ def callback() {
 			redirect_uri: callbackUrl
 		]
 
-		def tokenUrl = "https://www.ecobee.com/home/token?${toQueryString(tokenParams)}"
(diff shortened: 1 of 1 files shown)


smartapps/smartthings/ecobee-connect.src/ecobee-connect.groovy
```diff
@@ -201,7 +201,7 @@ def callback() {
 			redirect_uri: callbackUrl
 		]
 
-		def tokenUrl = "https://www.ecobee.com/home/token?${toQueryString(tokenParams)}"
+		def tokenUrl = "https://api.ecobee.com/home/token?${toQueryString(tokenParams)}"
```
marstorp: since this is the same API uri that is used in other places, e.g. L187,  could we use that instead of a hard coded version?
i.e. def tokenUrl = "${apiEndpoint}/home/token?${toQueryString(tokenParams)}"

nathankooistra (approved):

marstorp (approved):
LGTM ⛵IT

PR #70915, opened 2021-07-14
DevWs for Smartenit, Inc containing containing Smartenit Zigbee MLC30

lecontr:


devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,175 @@
...
+	sendEvent(name:"level", value:val)
+
+	int newval = 0
+	if (val != 0) {
+		newval = (255.0 / (100.0 / val))
+	}
+
+	zigbee.command(zigbee.LEVEL_CONTROL_CLUSTER, MoveToLevelWOnOff,
+		DataType.pack(newval, DataType.UINT8, 1) + DataType.pack(0xffff, DataType.UINT16, 1), [destEndpoint: 3])
+}
```
PKacprowiczS: Generally we prefer to not send events in commands body, but rather add additional Zigbee command, which polls affected attribute, and then send an event based on that response.

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,175 @@
...
+def configureHealthCheck() {
+	Integer hcIntervalMinutes = 10
+	sendEvent(name: "checkInterval", value: hcIntervalMinutes * 60, displayed: false, data: [protocol: "zigbee", hubHardwareId: device.hub.hardwareID])
+}
+
+def updated() {
+	log.debug "in updated()"
+	def cmds = configureHealthCheck()
+	cmds.each{ sendHubCommand(new physicalgraph.device.HubAction(it)) }
+}
```
PKacprowiczS: `configureHealthCheck()` doesn't return any Zigbee commands, so that's unnecessary to call `sendHubCommand()` here.

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,175 @@
...
+		capability "Energy Meter"
+		capability "Health Check"
+		capability "Voltage Measurement"
+		capability "monthpublic25501.current"
+		capability "monthpublic25501.load1"
+		capability "monthpublic25501.load2"
+		capability "monthpublic25501.levelControl"
+
+		fingerprint model: "ZBMLC30NC", manufacturer: "Smartenit, Inc", deviceJoinName: "Smartenit Dual Switch Meter"
+		fingerprint model: "ZBMLC30NO", manufacturer: "Smartenit, Inc", deviceJoinName: "Smartenit Dual Switch Meter"
```
PKacprowiczS: Please change deviceJoinNames to `Smartenit Switch`

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,175 @@
...
+			if (mapDescription.clusterInt == zigbee.SIMPLE_METERING_CLUSTER) {
+				if (mapDescription.attrInt == MeteringCurrentSummation) {
+					return sendEvent(name:"energy", value: getFPoint(mapDescription.value)/EnergyDivisor)
+				} else if (mapDescription.attrInt == MeteringInstantDemand) {
+					return sendEvent(name:"power", value: getFPoint(mapDescription.value/EnergyDivisor))
+				} else if (mapDescription.attrInt == Voltage) {
+					return sendEvent(name:"voltage", value: getFPoint(mapDescription.value) / 100)
+				} else if (mapDescription.attrInt == Current) {
+					return sendEvent(name:"current", value: getFPoint(mapDescription.value) / 100, unit: "A")
+				}
```
PKacprowiczS: If you want to return an event, it is more correct to use `createEvent()`. It takes same arguments as `sendEvent()`.
Main difference between them is that `createEvent()` returns an event object which needs to be returned from `parse()` method to be acknowledged by the platform, while `sendEvent()` sends an event immediately and returns no object (as far as I can tell).

PKacprowiczS (changes requested):
Looking good, just some minor things.

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,177 @@
...
+		return cmds.collect { new physicalgraph.device.HubAction(it) }
+	}
+
+	if (event && event.name != "switch") {
+		log.debug "Collecting event: ${event}, ${event.name}, ${event.value}"
+		if ((eventDescMap?.sourceEndpoint == "01") || (eventDescMap?.endpoint == "01")) {
+			if (event.name == "power") {
+				return createEvent(name: "power", value: (event.value/EnergyDivisor))
+			} else {
+				sendEvent(event)
```
PKacprowiczS: Perhaps you haven't known that, but in Groovy an object returned from last call of a method, is actually returned also by this whole method, for example.
```
def someMethod() {
    def object  = createEvent()
    object
}
```
(note that you don't have to use `return` statement, but it's not an error if you do, obviously)
and also
```
def someMethod() {
    createEvent()
}
```
would return created event in `someMethod`. 
Keeping that in mind, what I wrote in https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/70915#discussion_r670229771 also applies here, as call `sendEvent(event)` would be the last call in `parse()` method.
So if we want to keep things consistent, can you rework rest of the code, in a way that it would utilize `createEvent()`?

PKacprowiczS (changes requested):

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
lecontr: Ah I wasn't aware of that! Thank you so much for the explanation, that was very helpful.

devicetypes/smartenit/smartenit-metering-dual-load-controller.src/smartenit-metering-dual-load-controller.groovy
```diff
@@ -0,0 +1,177 @@
...
+				} else if (mapDescription.attrInt == Current) {
+					return createEvent(name:"current", value: getFPoint(mapDescription.value) / 100, unit: "A")
+				}
+			} else if (mapDescription.clusterInt == zigbee.ONOFF_CLUSTER) {
+				if (mapDescription.attrInt == OnOff) {
+					def nameVal = mapDescription.sourceEndpoint == "01" ? "loadone" : "loadtwo"
+					def status = mapDescription.value == "00" ? "off" : "on"
+					return createEvent(name:nameVal, value: status)
+				} else {
+					return createEvent(event)
```
PKacprowiczS, line 99: Are you sure that you don't need a null check for `event`?

PKacprowiczS (approved):
@greens

greens (approved):

PR #73948, opened 2021-08-17
DevWs Plaid Systems LLC sensor update fingerprint to correct join issue

natec007:


natec007:
@dkirker  - corrected PR with only fingerprint changes

dkirker:
@natec007 Don't forget the controller fingerprint!

dkirker (approved):

PR #74454, opened 2021-08-23
Update zigbee-metering-plug-power-consumption-report.groovy

dwd-kwon:
Changed the unit of PowerConsumptionReport


HyoungSangCho:
This modification should be applied as soon as possible because there is a market problem at the moment.

unisist (approved):
LGTM
The value of Wh unit is required for powerConsumptionReport.

greens (approved):

PR #74783, opened 2021-08-27
DevWs for NIE-TECH CO., LTD. containing containing Min Smart Plug

sky-nie:


sky-nie:
Some info will be changed for a N4001, I will resart PR after change.

PR #75070, opened 2021-08-31
DevWs for Zooz (The Smartest House) containing containing Zooz Remote Switch ZEN34

KevinTSH:


KevinTSH:
This requires a custom DTH because it looks like a regular wall dimmer, but it's a battery powered remote that sends several up/down button events.

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
```diff
@@ -0,0 +1,321 @@
...
+		capability "Battery"
+		capability "Button"
+		capability "Refresh"
+		capability "Configuration"
+		capability "Health Check"
+		capability "platemusic11009.firmware"
+		capability "platemusic11009.syncStatus"
+
+		//zw:Ss2a type:1800 mfr:027A prod:7000 model:F001 ver:1.01 zwv:7.13 lib:03 cc:5E,55,9F,6C sec:86,85,8E,59,72,5A,73,80,5B,70,84,7A
+		fingerprint mfr: "027A", prod: "7000", model: "F001", deviceJoinName: "Zooz Remote" //Zooz Remote Switch ZEN34
```
KKlimczukS: @KevinTSH Did You try to integrate this device in the existing DTH ? -> 
https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-button.src/zwave-button.groovy

`ledMode, upperPaddleLedColor, lowerPaddleLedColor` - those configuration parameters don't seem to be crucial and device might be configured with default settings (in the `configure()` function)

KKlimczukS (changes requested):

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
krlaframboise: Multiple sections of that handler would need to be re-written/modified because it only supports 1 scene and the device supports 2.  

The other built-in button DTHs that support multiple scenes use child devices which is a messy approach for this device because it's a paddle switch and the button capability already supports a bunch of "up" and "down" button values so there's no need for child devices.

Since there isn't an existing DTH that already works well with this device and a lot of users will find those LED settings useful, I personally think that this device should qualify for a custom DTH.

KevinTSH:
@KKlimczukS @greens  I'm just checking to see you'll be reviewing this DTH... Thanks

FYI, I accidentally posted the explanation above with my other GItHub account.

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
```diff
@@ -0,0 +1,321 @@
...
+		state[name] = val
+		logDebug "${configParams[name]?.title}(#${configParams[name]?.num}) = ${val}"
+	} else {
+		logDebug "Parameter #${cmd.parameterNumber} = ${cmd.scaledConfigurationValue}"
+	}
+}
+
+void zwaveEvent(physicalgraph.zwave.commands.versionv1.VersionReport cmd) {
+	logDebug "${cmd}"
+	sendEvent(name: "firmwareVersion", value: (cmd.applicationVersion + (cmd.applicationSubVersion / 100)))
```
KKlimczukS, line 232: Could You return `createEvent(...)` instead of `sendEvent(...)` ?

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
```diff
@@ -0,0 +1,321 @@
...
+}
+
+void zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
+	def val = (cmd.batteryLevel == 0xFF ? 1 : cmd.batteryLevel)
+	if (val > 100) {
+		val = 100
+	}
+	logDebug "Battery is ${val}%"
+	sendEvent(name:"battery", value:val, unit:"%", isStateChange: true)
```
KKlimczukS, line 241: Could You return `createEvent(...)` instead of `sendEvent(...)` ?

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
KevinTSH: No...  There's no benefit to doing that and my 4 other DTHs that passed the review process only used sendEvent and no one had an issue with it.

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
KevinTSH: No

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
```diff
@@ -0,0 +1,321 @@
...
+	logDebug "$cmd"
+	runIn(4, refreshSyncStatus)
+	state.wakeUpInterval = cmd.seconds
+}
+
+void zwaveEvent(physicalgraph.zwave.commands.wakeupv2.WakeUpNotification cmd) {
+	logDebug "Device Woke Up..."
+	runIn(4, refreshSyncStatus)
+	configure()
+}
```
greens: I don't see you ever sending a `NoMoreInformation` which technically violates the Z-Wave spec and probably will not do good things for your battery life.

I'd recommend you include it

KevinTSH: I've made the change, but it wasn't really necessary because SmartThings made a platform change about 3-6 months that causes it to suppress that command if it's specified in the handler and automatically send it 4 seconds after the last command is sent.

The first 2 lines in the screenshot below are from the original handler and the last 2 are from the version I just posted which shows that the WakeUpNoMoreInformation command is automatically sent by the hub after 4 seconds and sending that command from the DTH doesn't make a difference.

image

**Update:**  It appears that the V3 hub sends it automatically after 4 seconds, but the V2 hub sends it after 2 seconds so if that behavior is hub specific then maybe I'm wrong and it is still needed for other hub versions/models.

greens (approved):
Thanks.

Hopefully you get back all of your configuration gets before the nomoreinfo command is sent.

greens:
@KKlimczukS

KevinTSH:
@greens  The hub automatically sends the wakeUpNoMoreInfo command 2-4 seconds after communication between the device and hub stops so as long as you don't exceed the platform's z-wave message guardrail/limit then all the config reports will get sent back.

devicetypes/zooz/zooz-remote-switch-zen34.src/zooz-remote-switch-zen34.groovy
KevinTSH: Just to be clear, I have no problem making changes for consistency, readability, optimization, or are necessary to meet ST standards and every time a reviewer requests a change I make sure that I include that change in every DTH I submit going forward.

Using createEvent instead of sendEvent will require me to change the return type of those methods, make the code harder to follow, make it inconsistent with the central scene method, and I've submitted 5 custom DTHs _(3 different manufacturers)_ that only use sendEvent and you approved the other 4 without requesting this change so those are the reasons I said no...

KKlimczukS (approved):

PR #75141, opened 2021-08-31
DevWs for NIE-TECH CO., LTD. containing containing Min Smart Plug Dimmer

sky-nie:
restore for trace history code


sky-nie:
New PR will be started,so ingore this.

PR #75541, opened 2021-09-07
DevWs for SHINA SYSTEM containing containing SiHas Multipurpose Sensor

shinasys:


devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy
```diff
@@ -209,6 +209,12 @@ private Map getAnalogInputResult(value) {
     ]
 }
 
+def setPeopleCounter(peoplecounter) {
+    int pc =  Float.floatToIntBits(peoplecounter);
+    log.debug "SetPeopleCounter = $pc"
+    zigbee.writeAttribute(ANALOG_INPUT_BASIC_CLUSTER, ANALOG_INPUT_BASIC_PRESENT_VALUE_ATTRIBUTE, DataType.FLOAT4, pc)
+}
```
PKacprowiczS: Can you elaborate a bit how does it work and why is it needed? As far as I understand this device tries to estimate number of people in a room based on some measurement, but why is it needed to set a value to the same attribute from which that estimation comes from?

shinasys: This device measures the number of people passing through a certain point (toilet door or front door).
And this device is not a 100% accurate sensor. 
For example, If 2 people pass by in a row, there is a possibility that it will be recognized as 1 person.
So a user wants to make it possible to change the number of people.

PKacprowiczS (dismissed):
Personally, I don't see how overwriting wrong report manually can help, but if you think so (and Zigbee allows writing this attribute), let be it.

devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy
KKlimczukS: @shinasys Don't You also need a command in this custom capability to reset the counter ? (setting 0, without any input values -> it might call -> `setPeopleCounter(0)` )

shinasys: @KKlimczukS Since this custom capability is sufficient, it seems that there is no need to add reset separately.

shinasys:
@KKlimczukS @PKacprowiczS I want to add one more attribute. So some more updates are needed.
The attribute I'm trying to add is the max number of people counter. 
I'll post the edited part again. Please wait.

devicetypes/shinasys/sihas-multipurpose-sensor.src/sihas-multipurpose-sensor.groovy
KKlimczukS: @shinasys I meant -> to extend the custom capability by adding a command that will reset this counter. But it's a suggestion only.

shinasys: @KKlimczukS  @PKacprowiczS  I'll proceed without a command to a reset the counter.

PKacprowiczS (approved):

KKlimczukS (approved):

PR #75852, opened 2021-09-16
DevWs for SHINA SYSTEM containing containing Zigbee Metering Plug

shinasys:


PKacprowiczS (changes requested):
@shinasys can you fix this PR, so diff would show actual change instead of rewriting of whole file? Perhaps something's wrong with line ending encodings.

PR #75857, opened 2021-09-17
DevWs for SHINA SYSTEM containing containing Zigbee Metering Plug

shinasys:


shinasys:
I've imported the original file. 
I don't know why the whole file is marked as changed?
Is there any way?

PKacprowiczS:
I'm not sure what exactly causes problem in your case, but maybe you'll find some insights here: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

k-morawski:
> I'm not sure what exactly causes problem in your case, but maybe you'll find some insights here: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Dear @shinasys , could you please continue the work on this integration? Many thanks

PR #76136, opened 2021-10-14
HEIMAN Outlet WWST-7664

LUZhanchang:


chonghoon:
We need to change DTH code for Heiman outlet. Because outlet do not support power meter and energy meter. 
So Luchange want to remove fingerprint on zigbee-metering-plug.groovy

KKlimczukS (approved):

PR #77213, opened 2021-12-23
DevWs for NIE-TECH CO., LTD. containing containing Min Smart Plug

sky-nie:


sky-nie:
Because the function isButtonAvailable is called invalid in the preference ConfigParams, some fingerprints have to be separated.

devicetypes/sky-nie/min-smart-plug.src/min-smart-plug.groovy
```diff
@@ -79,31 +83,16 @@ metadata {
...
         fingerprint mfr: "0312", prod: "FF00", model: "FF0C", deviceJoinName: "Minoston Outlet", ocfDeviceType: "oic.d.smartplug" //MP21Z Minoston Mini Smart Plug
         fingerprint mfr: "0312", prod: "AC01", model: "4001", deviceJoinName: "New One Outlet",  ocfDeviceType: "oic.d.smartplug" // N4001 New One  Mini Smart Plug
-        fingerprint mfr: "0312", prod: "EE00", model: "EE01", deviceJoinName: "Minoston Switch", ocfDeviceType: "oic.d.switch"    //MS10ZS Minoston Smart Switch
-        fingerprint mfr: "0312", prod: "EE00", model: "EE03", deviceJoinName: "Minoston Switch", ocfDeviceType: "oic.d.switch"    //MS12ZS Minoston Smart on/off Toggle Switch
-        fingerprint mfr: "0312", prod: "A000", model: "A005", deviceJoinName: "Evalogik Switch", ocfDeviceType: "oic.d.switch"    //ZW30
-        fingerprint mfr: "0312", prod: "BB00", model: "BB01", deviceJoinName: "Evalogik Switch", ocfDeviceType: "oic.d.switch"    //ZW30S Evalogik Smart on/off Switch
-        fingerprint mfr: "0312", prod: "BB00", model: "BB03", deviceJoinName: "Evalogik Switch", ocfDeviceType: "oic.d.switch"    //ZW30TS Evalogik Smart on/off Toggle Switch
```
KKlimczukS, line 86: @sky-nie Why do You need to get rid of those fingerprints ? Are You going to add anything new ?

KKlimczukS: All changes from this PR should be a part of #77215

sky-nie: Because the products with those fingerprints can not correctly be supported by this DTH, so I have to remove them. and move them into [#77215](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/77215)

KKlimczukS: @sky-nie Could You explain more ? And why calling the function isButtonAvailable is problematic for those products ? Cannot You just fix it here instead of duplicating the code in the new DTH ?

sky-nie: The function isButtonAvailable returns true on line 426 when called in the preferences procedure.It does not change the return value no matter which product is in use.When users modify parameters through preferences, some products have incorrect feedback results.  Except for the preferences, the return value of this function is normal.Since the preferences are statically initialized (can't be modified dynamically), I can't think of a better way to solve this, or do you have a better solution suggestion?

KKlimczukS (approved):

PR #77214, opened 2021-12-23
DevWs for NIE-TECH CO., LTD. containing containing Min Smart Plug Dimmer

sky-nie:


sky-nie:
Because the function isButtonAvailable is called invalid in the preference ConfigParams, some fingerprints have to be separated.

devicetypes/sky-nie/min-smart-plug-dimmer.src/min-smart-plug-dimmer.groovy
```diff
@@ -102,39 +106,16 @@ metadata {
         fingerprint mfr: "0312", prod: "FF00", model: "FF0D", deviceJoinName: "Minoston Smart Plug Dimmer", ocfDeviceType: "oic.d.smartplug" //MP21ZD
         fingerprint mfr: "0312", prod: "FF07", model: "FF03", deviceJoinName: "Minoston Outdoor Dimmer", ocfDeviceType: "oic.d.smartplug" //MP22ZD
         fingerprint mfr: "0312", prod: "AC01", model: "4002", deviceJoinName: "New One Smart Plug Dimmer",  ocfDeviceType: "oic.d.smartplug" //N4002
-        fingerprint mfr: "0312", prod: "0004", model: "EE02", deviceJoinName: "Minoston Dimmer Switch", ocfDeviceType: "oic.d.switch"    //MS11ZS Minoston Smart Dimmer Switch
-        fingerprint mfr: "0312", prod: "EE00", model: "EE04", deviceJoinName: "Minoston Dimmer Switch", ocfDeviceType: "oic.d.switch"    //MS13ZS Minoston Smart Toggle Dimmer Switch
-        fingerprint mfr: "0312", prod: "BB00", model: "BB02", deviceJoinName: "Evalogik Dimmer Switch", ocfDeviceType: "oic.d.switch"    //ZW31S Evalogik Smart Dimmer Switch
-        fingerprint mfr: "0312", prod: "BB00", model: "BB04", deviceJoinName: "Evalogik Dimmer Switch", ocfDeviceType: "oic.d.switch"    //ZW31TS Evalogik Smart Toggle Dimmer Switch
```
KKlimczukS, line 108: @sky-nie Why do You need to get rid of those fingerprints ? Are You going to add anything new ?

KKlimczukS: All changes from this PR should be a part of #77216

sky-nie: Because the products with those fingerprints can not correctly be supported by this DTH, so I have to remove them. and move them into [#77216](https://github.com/SmartThingsCommunity/SmartThingsPublic/pull/77216)

KKlimczukS: @sky-nie Could You explain more ? And why calling the function `isButtonAvailable `is problematic for those products ? Cannot You just fix it here instead of duplicating the code in new DTH ?

sky-nie: The function isButtonAvailable returns true on line 426 when called in the preferences procedure.It does not change the return value no matter which product is in use.When users modify parameters through preferences, some products have incorrect feedback results.  Except for the preferences, the return value of this function is normal.Since the preferences are statically initialized (can't be modified dynamically), I can't think of a better way to solve this, or do you have a better solution suggestion?

KKlimczukS (approved):

PR #77221, opened 2021-12-24

tklee2020:


unisist (dismissed):
HomeIoT T/F in HQ tested this changes with WWST TCs and new TCs for Energy Service without any issue.

PKacprowiczS (commented):
Adding a capability to already existing capability can be a risky business. We need to discuss this internally within the team.

unisist:
We (HomeIoT T/F in HQ) had discussion with MX division (former IM division) Ryan Kong already. and did same thing before for DAWON zigbee smart plug.

lustre2028:
This case is for supporting the DA business in urgent. It's one time support without considering the proper process. 
I got the Internal Test Result with the approval by VP SeungSoo about this DTH.  It's adding capability at the existing DTH for avoiding the impact to General customer. The capability will support the showing the energy at the Energy service plugin. Not for changing or adding some function of the device. 
DA need to launch the Energy Service with this supporting together. If there's no critical issue on it, let's help to publish this DTH immediately. Thank you in advance.

greens (changes requested):
Why not move these devices' fingerprints over to the DTH that already has this capability? 

https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zigbee-metering-plug-power-consumption-report.src/zigbee-metering-plug-power-consumption-report.groovy

Either way, it's my understanding that you need to use a different UI-metadata for this capability to work correctly, and you don't seem to have done that here. I would have expected you to alter the specific device fingerprints to point to a specific metadata.

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -104,6 +105,36 @@ def parse(String description) {
                         map.name = "energy"
                         map.value = zigbee.convertHexToInt(it.value)/(energyDivisor * 1000)
                         map.unit = "kWh"
+                        
+                        if (isEZEX()) {
+                            def currentEnergy = zigbee.convertHexToInt(it.value) / 1000
+                            def prevPowerConsumption = device.currentState("powerConsumption")?.value
+                            Map previousMap = prevPowerConsumption ? new groovy.json.JsonSlurper().parseText(prevPowerConsumption) : [:]
```
greens, line 114: you need to import `groovy.json.JsonSlurper` at the top of the file or I expect this to throw an error.

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
```diff
@@ -104,6 +105,36 @@ def parse(String description) {
...
+                            def deltaEnergy = calculateDelta(currentEnergy, previousMap)
+                            def currentTimestamp = Calendar.getInstance().timeInMillis
+                            def prevTimestamp = device.currentState("powerConsumption")?.date?.time
+                            if (prevTimestamp == null) prevTimestamp = 0L
+                            def timeDiff = currentTimestamp - prevTimestamp
+                            log.debug "currentTimestamp= $currentTimestamp, prevTimestamp= $prevTimestamp, timeDiff= $timeDiff"
```
greens, line 118: this is all unused

devicetypes/smartthings/zigbee-power-meter.src/zigbee-power-meter.groovy
tklee2020: Actually, it has not caused error yet. Also, in the DTH, zigbee-metering-plug-power-consumption-report, there is no import statement for groovy.json.JsonSlurper either. 

Even though, do you want me to add "import groovy.json.JsonSlurper"?

tklee2020: > you need to import `groovy.json.JsonSlurper` at the top of the file or I expect this to throw an error.

I've added import statement cause you've not answered. please check it asap.

greens (approved):

PR #77720, opened 2022-02-08
[ICP-14430] Qubino Flush Thermostat - allow non-zero power meter values only when device is not idle

PKacprowiczS:


greens (approved):

greens:
waiting for feedback from tester

PR #77897, opened 2022-03-02
DevWs for Zooz (The Smartest House) containing containing Zooz ZSE11 Q Sensor

KevinTSH:


devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,448 @@
...
+	sendHubCommand(cmds)
+}
+
+void zwaveEvent(physicalgraph.zwave.commands.batteryv1.BatteryReport cmd) {
+	int val = (cmd.batteryLevel == 0xFF ? 1 : cmd.batteryLevel)
+	if (val > 100) {
+		val = 100
+	} else if (val < 1) {
+		val = 1
+	}
```
KKlimczukS: what about `val = Math.min(Math.max(1, val), 100)` ?

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,448 @@
...
+}
+
+void sendHumidityEvent(reportedVal, boolean onlyIfNew=false) {
+	reportedVal = safeToInt(reportedVal)
+	state.reportedHumidity = reportedVal
+
+	def adjVal = (safeToInt(settings?.humidityOffset) + reportedVal)
+	if (adjVal < 0) adjVal = 0
+	if (adjVal > 100) adjVal = 100
```
KKlimczukS: what about ` adjVal = Math.min(Math.max(0, adjVal ), 100)` ?

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,448 @@
...
+			break
+		case humiditySensorType:
+			sendHumidityEvent(cmd.scaledSensorValue)
+			break
+		default:
+			log.debug "Unhandled: ${cmd}"
+	}
+}
+
+void sendTempEvent(reportedVal, boolean onlyIfNew=false) {
```
KKlimczukS: After the same event value is sent then the previously created event is updated and that's valuable too.
I wouldn't use flag: `onlyIfNew` (here and in other functions).
@greens any thoughts on this ?

KKlimczukS (changes requested):

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: I've made the requested change

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: I've made the requested change

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: I've removed it.

KKlimczukS (dismissed):
lgtm
@greens

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,439 @@
...
+		} else {
+			sendHubCommand(getConfigCmds())
+		}
+	} else {
+		sendHubCommand(getRefreshCmds())
+		state.firstConfig = false
+	}
+}
+
+void initialize() {
```
greens, line 150: same note as the other DTH

if you put this in `installed` it's guaranteed to only be called once, so you can leave out these if statements.

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,439 @@
...
+		sendEvent(name: "powerSource", value: powerSource)
+
+		if (powerSource == "dc") {
+			sendEvent(name: "battery", value: 100, unit: "%")
+		}
+	}
+
+	if (state.reportedTemp == null) {
+		state.reportedTemp = device.currentValue("temperature")
+	}
```
greens: when would `state.reportedTemp` be `null` when the current temperature value isn't?

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
```diff
@@ -0,0 +1,439 @@
...
+	if (!device.currentValue("checkInterval")) {
+		sendEvent(name: "checkInterval", value: ((60 * 60 * 24) + (60 * 5)), displayed: false, data:[protocol: "zwave", hubHardwareId: device.hub.hardwareID])
+	}
+
+	if (!device.currentValue("tamper")) {
+		sendEvent(name: "tamper", value: "clear")
+	}
+
+	if (device.currentValue("powerSource") == null) {
+		boolean hasBatteryCC = ((zwaveInfo?.cc?.find { it.toString() == batteryCC }) || (zwaveInfo?.sec?.find { it.toString() == batteryCC }))
```
greens, line 160: Does the device change its reported command classes after being added? It looks like all of your fingerprints have the battery command class.

KevinTSH: Yes, the command classes are different based on whether or not the device is plugged in when you join it.
I happened to have it joined on batteries when I copied the description and then just changed the product in it for the other 2 fingerprints since there just for different regions.

Do I need to include the raw description for both ways it can join?   I've never seen any of the built-in handlers do that before...

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: Same reason as the other DTH...

The onboarding process for DTHs has been horrible ever since drivers launched and I have no clue why, but preventing the excess commands seems to make a difference...

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: Technically, I think the raw description would also be different if secure inclusion failed so technically to make that statement match the raw description list I'd have to add 4 comments...

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: It wouldn't.  This is a modified version of my community DTH and I added that offset settings in a later version which is why that code was there.  

It looks like you already approved this PR so am I correct to assume I can leave this as-is?

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
greens: Okay. So the command classes are consistent once the device joins. That's all I wanted to know. The `zwaveInfo` object doesn't get updated after join and I just wanted to make sure you weren't expecting it.

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
greens: That was Konrad. You don't have to change it if you don't want to, but it would make the code cleaner.

KevinTSH: I've removed that code.

devicetypes/zooz/zooz-zse11-q-sensor.src/zooz-zse11-q-sensor.groovy
KevinTSH: correct.  Let me know if this PR is all set now.  Thanks

KKlimczukS (approved):

PR #77900, opened 2022-03-02
DevWs for HELTUN containing containing HE-HT01 Handler X

Sarkis008:


KKlimczukS (commented):
@Sarkis008 It's the empty PR. Why there are no files ?

k-morawski:
Dear @Sarkis008 , I have noticed that your PR is empty. Moreover your request looks like this is an update for something but not for catalog as it's empty as well. May I ask you what's the purpose of it?

Sarkis008:
@k-morawski @KKlimczukS I only changed the spacing of "Pairing Instructions".
The Pairing Instructions text has very strange spacing in the app. I might change the spacing for the remaining devices too.
Is there any easier way to do it ?

k-morawski:
> @k-morawski @KKlimczukS I only changed the spacing of "Pairing Instructions". The Pairing Instructions text has very strange spacing in the app. I might change the spacing for the remaining devices too. Is there any easier way to do it ?

Hi @Sarkis008 , as per my task which I double checked today, there are no planned changes in the catalog and I can see 'null' value. Could you please cancel this request and try to issue a new one? Definitely in DevWS there is a possibility to update catalog content, like pairing instructions. I believe something went wrong. 

Just want to add, that if I approve your current request it can add more the same devices to the catalog. Definitely I would like to avoid that.

PR #77938, opened 2022-03-12
DevWs for NIE-TECH CO., LTD. containing containing ZigBee Switch

sky-nie:


sky-nie:
I added a new product fingerprint on the offical public version - as much as possible to make sure nothing else has changed.

KKlimczukS (approved):

PR #78017, opened 2022-03-29
Sinope fix

greens:


greens:
fyi @KKlimczukS and @aguilbault-Sinope I reverted your original change and did my own that added the fingerprint back because the file rename/move creates some unwanted headaches on our end.

PR #78029, opened 2022-04-01
  SmartThingsPublic WWST-7893, Added a health check code for ABL CPX Lighting(Light with a motion sensor)

RaihaPark:
By the reason for WWST certification failed, I added a health check code in the configure function. After added a health check code the device changed to offline about 20 mins later from when I remove the power of device or reset the device.

TC_Health_Check_3 - Coming back to the online state after performing action on a device(Fail)
problem : This test case fails because the device does not show as Offline in the SmartThings app after remove power from device.

TC_Reset_1 - Correct device behavior after manual reset(Fail)
problem : This test case fails because the device does not show as Offline in the SmartThings app after reset.


chonghoon:
@greens Could you check it also? There is no Healtcheck capability, It makes a connection update error.

greens (approved):

PR #78039, opened 2022-04-02
DevWs for HELTUN containing containing HELTUN TPS04 Switch

SmartThings:


KKlimczukS (changes requested):
@Sarkis008 
Since all of new devices support the same capabilities. Please integrate all devices from PR: #78039, #78040, #78041, #78042 in one DTH. That will allow to avoid code duplication.

KKlimczukS (changes requested):
@Sarkis008 
Since all of new devices support the same capabilities. Please integrate all devices from PR: #78039, #78040, #78041, #78042 in one DTH. That will allow to avoid code duplication.

Sarkis008:
@KKlimczukS we already discussed this.
This issue is that the devices have different parameters, different button/relay numbers, and different inclusion steps.
Please let me know if there is a way to keep all this in one DTH

KKlimczukS:
> @KKlimczukS we already discussed this. This issue is that the devices have different parameters, different button/relay numbers, and different inclusion steps. Please let me know if there is a way to keep all this in one DTH

- the list of parameters (titles) is exactly the same, but the sub-options differ, so here You could add notes (in those options that  it will work with certain model(s) and skip sending it if particular option is selected when user tries to do it with the model that don't support that number of buttons - it could be done by defining some helper function in the DTH.
- different number of relays - this can be determined basing on current device model  in helper function that might be called from `configure()` or when some events are reported, example here: https://github.com/SmartThingsCommunity/SmartThingsPublic/blob/master/devicetypes/smartthings/zwave-multi-metering-switch.src/zwave-multi-metering-switch.groovy

k-morawski:
Dear @Sarkis008 , this is just a friendly reminder. Could you please continue the discussion with Konrad? Thanks

PR #78040, opened 2022-04-02
DevWs for HELTUN containing containing HELTUN TPS03 Switch

SmartThings:


KKlimczukS (changes requested):
@Sarkis008 
Since all of new devices support the same capabilities. Please integrate all devices from PR: #78039, #78040, #78041, #78042 in one DTH. That will allow to avoid code duplication.

PR #78128, opened 2022-04-26
Add fingerprint in ZigBee Switch for Vimar xx592 and 03981

sciaccom-vimar:
Support for Vimar 2-way Smart Switch xx592 and Vimar Actuator Module 03981


sciaccom-vimar:
PR closed in favor of Edge Drivers.

PR #78251, opened 2022-05-09
BUG-4775 Only prevent sending pause for Somfy shades

greens:


devicetypes/smartthings/zigbee-window-shade.src/zigbee-window-shade.groovy
```diff
@@ -222,14 +222,10 @@ def pause() {
 	log.info "pause()"
 	def currentShadeStatus = device.currentValue("windowShade")
 
-        if (device.getDataValue("manufacturer") == "Third Reality, Inc") {
-	        zigbee.command(CLUSTER_WINDOW_COVERING, COMMAND_PAUSE)
+	if (isSomfy() && (currentShadeStatus == "open" || currentShadeStatus == "closed")) {
```
KKlimczukS, line 225: @greens Are You 100% sure that it should work that way only for Somfy devices ?

greens: Przemek originally only added it as a workaround for the Somfy devices.

I don't _think_ sending a pause should cause an error on most devices.

KKlimczukS (dismissed):

PR #78730, opened 2022-07-15
update ikea-button.groovy

weihuan1111:
update ikea-button.groovy


weihuan1111:
@greens please help to review the merge require,thanks

greens (changes requested):
Please handle everything in `getButtonEvent` as has been done for other manufacturers rather than inside `parse`.

PR #78738, opened 2022-07-18
upadte the fingerprint info

sky-nie:
remove the inof about the brand

@KKlimczukS


sky-nie:
Because the DTH will be applied to a different product (only the brands are different), the relevant brand information is removed.

KKlimczukS:
@sky-nie
Even if You change device join name Evalogik Switch 1 -> Outdoor Smart Plug 1
It will affect only the name of the device after You onboard it.

The entry in the catalog (in the ST app) will remain the same: Evalogik. 
Previously it was possible to change it by submitting a request in developer workspace.

cc: @k-morawski

sky-nie:
> @sky-nie Even if You change device join name Evalogik Switch 1 -> Outdoor Smart Plug 1 It will affect only the name of the device after You onboard it.
> 
> The entry in the catalog (in the ST app) will remain the same: Evalogik. Previously it was possible to change it by submitting a request in developer workspace.
> 
> cc: @k-morawski

Yes, you're right! I'll cancel this PR and start a request with the new brand info in developer workspace.

PR #78770, opened 2022-07-29
DevWs for NIE-TECH CO., LTD. containing containing Z-Wave Dual Switch

sky-nie:


sky-nie:
I have added fingerprints for  two new products. @moorage @pennig @charliek

PR #78787, opened 2022-08-03
Devws for shina system containin 114

shinasys:
SiHAS Dual Motion Sensor


shinasys:
@greens please review,thanks

greens (commented):
First, your whitespace is inconsistent in several places and you've left in a fair amount of commented-out lines that should be removed or converted to actual comments.

Second, can I ask why you didn't write this as a multi-component device? You could have avoided using a custom capability.

shinasys:
> Second, can I ask why you didn't write this as a multi-component device? You could have avoided using a custom capability.
1. The names that appear in the History tab of the SmartThings app are not distinguishable.
2. I wanted to make a custom label in Korean language in multi-component.
If this can be resolved, I will do so.

greens (approved):
That makes sense. Thanks for explaining.

PR #78836, opened 2022-08-18
update

weihuan1111:
update


weihuan1111:
@greens please help to review the merge request, thanks

greens (commented):
There is more to enabling this flag than this. This would break some of the functionality of the device. Specifically, the firmware version-based battery handling as well as the automatic motion clear

